Issue #7, opened 2020-02-27
Skeleton for the relayer CLI 


ancazamfir:
Skeleton for the relayer CLI

ancazamfir:
Addressed in e5970e7

PR #8, opened 2020-02-27
Skeleton for the relayer CLI

romac:
This PR adds a skeleton for the relayer CLI based on the [Abscissa crate](https://github.com/iqlusioninc/abscissa/).

For now there is only one command which parses and prints the relayer configuration.

**Note:** This branch should be merged after [`anca/ibc_relayer_spec`](https://github.com/informalsystems/ibc-rs/tree/anca/ibc_relayer_spec).

See #7


romac:
Failure is because we can't set up Codecov yet.

PR #20, opened 2020-03-03
Improve validation of the configuration file at deserialization time

romac:
## Description

Leverage `serde` to improve validation of the configuration file at deserialization time.

The idea is to validate as many static properties of the configuration file at deserialisation time by replacing `String` and `Option<String>` with more specific types such as `std::time::Duration`, `Strategy` or `Direction`.

______

For contributor use:

- [ ] Wrote tests
- [ ] Updated CHANGELOG.md

### DIFF
--- a/relayer/cli/src/application.rs
+++ b/relayer/cli/src/application.rs
@@ -1,6 +1,6 @@
 //! Cli Abscissa Application
 
-use crate::{commands::CliCmd, config::CliConfig};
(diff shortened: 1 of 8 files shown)


romac:
> Looks good. At some point we should identify parameters that are chain type specific and probably rethink the .toml configuration "language".

Agreed! Perhaps even look at something like the Dhall configuration language: https://dhall-lang.org

PR #22, opened 2020-03-06
Run cargo-audit daily or when dependencies have changed

romac:
Closes: N/A

## Description

Change the Audit GitHub Action to run daily, and on pushes but only when dependencies have actually changed. This should speed up CI quite a bit as this action takes a considerable amount of time to complete.

The config comes from https://github.com/interchainio/tendermint-rs/pull/144#issuecomment-595322485

______

For contributor use:

- [ ] Wrote tests

### DIFF
--- a/.github/workflows/audit.yaml
+++ b/.github/workflows/audit.yaml
@@ -13,7 +13,7 @@ jobs:
     name: Security Audit
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - name: Cache cargo bin
         uses: actions/cache@v1
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/audit.yaml
```diff
@@ -1,11 +1,25 @@
...
+  schedule:
+    - cron: '0 0 * * *'
 
 jobs:
   security_audit:
+    name: Security Audit
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v1
```
Liamsi: Is there a particular reason this uses `v1`?

romac: No, that's a mistake on my part, I blindly copy-pasted the config file used above. Will change it back to v2. Thanks for noticing!

Issue #37, opened 2020-03-17
Add relayer client queries


ancazamfir:
## Summary
Add CLI to query:
- client consensus state
- client full state 

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #38, opened 2020-03-17
Add client queries to the relayer

ancazamfir:
Closes: #37

## Description
Add consensus state and full state query clis to the relayer:
- command to query full state of client
- command to query consensus state of client
- validate parameters against loaded configuration file
- move to tm v0.33 branch that fixes the new merkle proof format in the abci response

Note: full deserialization does not yet work, waiting for migration of SDK/ TM to protobuf.


ancazamfir:
More work needs to be done here but at this point trying to figure out why we get error when we ask for proof vs. with no proof.  Will add more soon.

modules/src/ics07_tendermint/client_state.rs
```diff
@@ -0,0 +1,82 @@
...
+use crate::ics07_tendermint::header::Header;
+use crate::ics24_host::client::ClientId;
+use serde_derive::{Deserialize, Serialize};
+use std::time::Duration;
+use tendermint::lite::Header as liteHeader;
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct ClientState {
+    id: String,
+    trusting_period: Duration,
```
Liamsi, line 13: Note (mostly to myself): 

in the go code, time.Duration actually is this type alias:
```go
// A Duration represents the elapsed time between two instants
// as an int64 nanosecond count. The representation limits the
// largest representable duration to approximately 290 years.
type Duration int64
```
while rust's `std::time::Duration` actually is 
```rust
pub struct Duration {
    secs: u64,
    nanos: u32, // Always 0 <= nanos < NANOS_PER_SEC
}
```
Same for the `unbonding_period` field below. 

So for deserialization (independent from amino or not) we need to translate between the two.

modules/src/ics07_tendermint/client_state.rs
```diff
@@ -0,0 +1,82 @@
...
+    unbonding_period: Duration,
+    latest_header: Header,
```
Liamsi, line 15: Note that the Header here contains a `SignedHeader` which in turn contains a `Commit`. Looking at the `Commit` struct in the iqlusion/relayer (on master) shows that they are already using tendermint `v0.33.2`which modified the `Commit` struct:
```go
type Commit struct {
	// NOTE: The signatures are in order of address to preserve the bonded
	// ValidatorSet order.
	// Any peer with a block can gossip signatures by index with a peer without
	// recalculating the active ValidatorSet.
	Height     int64       `json:"height"`
	Round      int         `json:"round"`
	BlockID    BlockID     `json:"block_id"`
	Signatures []CommitSig `json:"signatures"`
        // ...
}
```
while this branch uses the v0.32 version of the commit:
```rust
pub struct Commit {
    /// Block ID of the last commit
    pub block_id: block::Id,

    /// Precommits
    pub precommits: Precommits,
}
```
Or in golang: https://github.com/tendermint/tendermint/blob/64e61365c16f74d56ccaa9eb784e05df5b01764c/types/block.go#L489-L494

v0.33 tendermint-rs is currently worked on in https://github.com/informalsystems/tendermint-rs/pull/196

Does it make sense to assume we want to decode v0.33 commits directly (for the sake of #45)? @ancazamfir

ancazamfir: yes, let's look at 0.33! thx!

modules/src/ics02_client/state.rs
```diff
@@ -18,3 +19,24 @@ pub trait ConsensusState {
...
+    /// Height of consensus state
+    fn get_latest_height(&self) -> Height;
+
+    /// Freeze status of the client
+    fn is_frozen(&self) -> bool;
+
+    fn verify_client_consensus_state(
+        &self,
+        root: &CommitmentRoot,
+    ) -> Result<(), Self::ValidationError>;
```
Liamsi, line 41: I guess the other state verification functions are not included on purpose?
https://github.com/cosmos/cosmos-sdk/blob/b5a658729170e577b0e6e809e73d5e0756b90e07/x/ibc/02-client/exported/exported.go#L26-L92

modules/src/ics07_tendermint/client_state.rs
```diff
@@ -0,0 +1,82 @@
...
+    frozen_height: crate::Height,
```
Liamsi, line 16: The order in the go-code is different (latest header should be last):

```suggestion
    frozen_height: crate::Height,
    latest_header: Header,
```

modules/src/ics07_tendermint/client_state.rs
Liamsi, line 15: Only commit batched together with the suggestion below (frozen_height):
```suggestion
```

modules/src/ics23_commitment/mod.rs
```diff
@@ -17,11 +18,11 @@ impl CommitmentPath {
...
-pub struct CommitmentProof;
-
+pub type CommitmentProof = abci::Proof;
+/*
 impl CommitmentProof {
     pub fn from_bytes(_bytes: &[u8]) -> Self {
         todo!()
     }
 }
+*/
```
Liamsi, line 28: Shouldn't we delete this but but extend the `abci::Proof` type instead?

relayer/cli/src/commands/utils.rs
```diff
@@ -0,0 +1,11 @@
...
+
+/// block on future
+pub fn block_on<F: Future>(future: F) -> F::Output {
+    tokio::runtime::Builder::new()
+        .basic_scheduler()
+        .enable_all()
+        .build()
+        .unwrap()
+        .block_on(future)
+}
```
Liamsi, line 11: You can probably get rid of this `block_on` method (might be relevant: https://github.com/informalsystems/tendermint-rs/pull/171, https://github.com/informalsystems/tendermint-rs/pull/173, https://github.com/informalsystems/tendermint-rs/pull/125#discussion_r388390015).

relayer/relay/src/query/client_consensus_state.rs
```diff
@@ -67,17 +74,15 @@ impl<CS> ConsensusStateResponse<CS> {
     pub fn new(
         client_id: ClientId,
         consensus_state: CS,
-        abci_proof: abci::Proof,
+        abci_proof: Option<abci::Proof>,
```
Liamsi, line 77: Why isn't the type alias (`CommitmentProof`) used consistently?

relayer/relay/src/query/client_state.rs
```diff
@@ -0,0 +1,123 @@
...
+use relayer_modules::ics23_commitment::{CommitmentPath, CommitmentProof};
+use relayer_modules::ics24_host::client::ClientId;
+use relayer_modules::path::{ClientStatePath, Path};
+use relayer_modules::Height;
+
+use super::{ibc_query, IbcQuery, IbcResponse};
+use crate::chain::Chain;
+use crate::error;
+
+pub struct QueryClientFullState<CLS> {
```
Liamsi, line 16: Is that the corresponding request to `ClientStateResponse`?

relayer/relay/src/query/client_state.rs
```diff
@@ -0,0 +1,123 @@
...
+) -> Result<ClientStateResponse<C::ClientState>, error::Error>
+where
+    C: Chain,
+{
+    let query = QueryClientFullState::new(chain_height, client_id, prove);
+    ibc_query(chain, query).await
+}
+
+fn amino_unmarshal_binary_length_prefixed<T>(_bytes: &[u8]) -> Result<T, error::Error> {
+    todo!()
```
Liamsi, line 122: My bad :-/

relayer/cli/src/commands/utils.rs
romac: Unfortunately, I have been unable to get `abscissa_tokio` to work, `abscissa_tokio::run` [just hangs](https://github.com/informalsystems/ibc-rs/blob/88e0c0cc49b8ea773b445b1bb95419a1f8117ab0/relayer/cli/src/commands/start.rs#L25-L26) (on my machine at least).

I haven't really put any time into debugging this, as I'd rather wait until we figure the new light client + relayer architecture, before spending time on that (`block_on` does the job just fine for now imho as it is called at the top level only).

Liamsi: Makes sense!

modules/src/ics02_client/state.rs
ancazamfir: Yes, a bit more work to get there :)

modules/src/ics07_tendermint/client_state.rs
ancazamfir: Oy! good catch! thx, just changed. The problem is that we cannot test properly with SDK since we don't deserialize yet. Once we move to protobuf and we do the decoding we will catch more of these.

modules/src/ics07_tendermint/client_state.rs
ancazamfir: fixed

modules/src/ics23_commitment/mod.rs
ancazamfir: Yes, but will do in a separate issue/PR

relayer/cli/src/commands/utils.rs
ancazamfir: @romac just a note that we use this now for the query also.

relayer/relay/src/query/client_consensus_state.rs
ancazamfir: my bad, changed.

relayer/relay/src/query/client_state.rs
ancazamfir: yes, changed the name of response.

Issue #5, opened 2020-02-26
TLA+ for Relayer Algorithms


ancazamfir:
TLA+ for Relayer Algorithms, see [spec](https://github.com/cosmos/ics/tree/master/spec/ics-018-relayer-algorithms)

PR #39, opened 2020-03-18
[WIP] TLA+ specification of the Relayer algorthm

istoilkovska:
Closes: #5
(not yet, still work in progress)

A TLA+ specification of the Relayer algorithm.

The specification has two modules: 
  - `relayer.tla` (the main module)
  - `environment.tla` (the environment module)

The module `relayer.tla` contains the specification of the relayer algorithm. It instantiates the module `environment.tla`, which takes care of the chain logic. Currently, the relayer and the environment can only handle ICS 002 Client datagrams.

To import the specification in the TLA+ toolbox and run TLC:
  - add the spec `relayer.tla` in TLA+ toolbox
  - add `environment.tla` as a module in the spec
  - create a model
  - assign values to the constants `MaxHeight` and `NrChains` (I used 3 and 2, respectively)
  - choose "Temporal formula" as the behavior spec, and use the formula `Spec`
  - add the invariants `TypeOK`, `CreateClientInv`, `ClientUpdateInv`
  - add the properties `Env!ClientCreated`, `Env!ClientUpdated`, `Env!HeightsDontDecrease`
  - run TLC on the model

______

For contributor use:

- [ ] Wrote tests


verification/spec/relayer/environment.tla
```diff
@@ -0,0 +1,275 @@
...
+AdvanceChain ==
+    \E chainID \in ChainIDs :
+        /\ chains[chainID].height < MaxHeight
+        /\ chains' = [chains EXCEPT 
+                        ![chainID].height = chains[chainID].height + 1                                            
+                     ]
+        /\ UNCHANGED pendingDatagrams
+
+\* Update the height of the light client on chainID for chainID 
+UpdateOwnClient ==
```
ancazamfir: Is this needed if we let the relayer update with src=dest? Or initialize chains not to create its own clients?
I will have to confirm if we support this currently but in principle we allow loopback connections and they would require a client for a chain A to be created on chain A. But this would also be done by the relayer.

verification/spec/relayer/relayer.tla
```diff
@@ -0,0 +1,225 @@
+------------------------------ MODULE relayer ------------------------------
+
+EXTENDS Naturals, FiniteSets
+
+CONSTANTS MaxHeight, \* maximal height of all the chains in the system 
+          NrChains \* number of chains in the system
+
+VARIABLES chains, \* a function that assigns to each chainID a chain record
```
ancazamfir: It seems that the relayer does not (and should not) update `chains`. Can we remove this and provide `Env!<getter>` for all chain queries needed by the relayer? (see below)

verification/spec/relayer/relayer.tla
```diff
@@ -0,0 +1,225 @@
...
+GetClientID(onChainID, forChainID) == Env!GetClientID(onChainID, forChainID)
+    
+ValidClientID(chainID, clientID) == Env!ValidClientID(chainID, clientID) 
+ 
+(***************
+Client datagrams
+****************) 
+ 
+\* get the latest height of the chainID
+GetLatestHeight(chainID) ==
```
ancazamfir: can we define this in environment and import it here?

verification/spec/relayer/relayer.tla
```diff
@@ -0,0 +1,225 @@
...
+              clientID |-> GetClientID(srcChainID, dstChainID)]} 
+        ELSE \* the dst client exists at the src chain
+            IF dstClientHeight < dstChainHeight
+            THEN \* the height of the dst client is smaller than the height of the dst chain
+                {[type |-> "ClientUpdate",
+                  height |-> relayer.clientHeights[dstChainID],
+                  clientID |-> GetClientID(srcChainID, dstChainID)]} 
+            ELSE {} IN
+                    
+    LET srcChainHeight == GetLatestHeight(srcChainID) IN
```
ancazamfir: do we need the rest of this? `ClientDatagrams` should be called for the flipped case, i.e. should see both `ClientDatagrams(A, B)` and `ClientDatagrams(B, A)`

verification/spec/relayer/relayer.tla
```diff
@@ -0,0 +1,225 @@
...
+(***************
+Client datagrams
+****************) 
+ 
+\* get the latest height of the chainID
+GetLatestHeight(chainID) ==
+    chains[chainID].height    
+    
+\* get the client height of the client on chain srcChainID for chain dstChainID    
+GetClientHeight(srcChainID, dstChainID) ==
```
ancazamfir: same as above, define this in Env?

ancazamfir:
Could you add a README.md file with the instructions you have in this PR?

verification/spec/relayer/relayer.tla
istoilkovska: That's a good point, thanks!

verification/spec/relayer/environment.tla
```diff
@@ -0,0 +1,275 @@
...
+\* check if the clientID is a valid client ID w.r.t. chainID    
+ValidClientID(chainID, clientID) ==
+    /\ GetChainIDs(clientID).src = chainID
+    /\ GetChainIDs(clientID).dst \in ChainIDs 
+    
+(***************
+Chain update operators
+***************)    
+
+LightClientUpdate(chainID, chain, datagrams) ==
```
milosevic: This function is pretty long so it's a bit hard to read. Can we refactor it into few smaller functions so it's easier understanding the core logic?

verification/spec/relayer/environment.tla
```diff
@@ -0,0 +1,275 @@
...
+                                      ELSE clientsCreated.clientHeights[cID] 
+                ]
+         ] IN
+    
+    clientsUpdated
+ 
+\* Update chainID with the received datagrams
+\* Currently, only supporting ICS2: Client updates
+\* TODO: Support the remaining datagrams
+UpdateChain(chainID, datagrams) ==
```
milosevic: I realised that TLA+ specs should be read bottom up as high level functions come after helper functions are defined :)

verification/spec/relayer/environment.tla
```diff
@@ -0,0 +1,275 @@
...
+\* Advance the height of the chain until MaxHeight is reached
+AdvanceChain ==
+    \E chainID \in ChainIDs :
+        /\ chains[chainID].height < MaxHeight
+        /\ chains' = [chains EXCEPT 
+                        ![chainID].height = chains[chainID].height + 1                                            
+                     ]
+        /\ UNCHANGED pendingDatagrams
+
+\* Update the height of the light client on chainID for chainID
```
milosevic: This comment is not really clear.

verification/spec/relayer/environment.tla
milosevic: We can probably remove this for sake of simplicity (at least initially).

istoilkovska: I was looking at the ICS, in ICS 018 in the specification of `pendingDatagrams`, the relayer is only creating datagrams when src != dst. But in principle, we could allow relayer updates when src = dst. 

Are loopback connections those where src = dst? I.e., is it allowed that chain A creates a connection with itself? I'm modeling the connection datagrams now and I wasn't sure if this is allowed.

verification/spec/relayer/environment.tla
```diff
@@ -0,0 +1,275 @@
...
+    \E chainID \in ChainIDs :
+        /\ pendingDatagrams[chainID] /= {} 
+        /\ chains' = [chains EXCEPT 
+                        ![chainID] = UpdateChain(chainID, pendingDatagrams[chainID])
+                     ]
+                        
+        /\ pendingDatagrams' = [pendingDatagrams EXCEPT
+                                    ![chainID] = {}
+                               ]
+\* Initial value if the chain:
```
milosevic:
```suggestion
\* Initial value of the chain: 
```

verification/spec/relayer/environment.tla
```diff
@@ -0,0 +1,275 @@
...
+                        
+        /\ pendingDatagrams' = [pendingDatagrams EXCEPT
+                                    ![chainID] = {}
+                               ]
+\* Initial value if the chain: 
+\*      - height is initialized to 1
+\*      - all the light clients are uninitialized
+InitChain ==
+    [height |-> 1,
+     clientHeights |-> [chID \in ChainIDs |-> nullHeight]]
```
milosevic: Would it be simpler if we refer to particular two chains A and B (or Alice and Bob) in the spec?

milosevic: I am wondering if it would be simpler (easier to understand) if we just consider two chains (Alice and Bob) and then chain state is height and lightClient of other chain.

verification/spec/relayer/environment.tla
```diff
@@ -0,0 +1,275 @@
...
+    /\ pendingDatagrams = [chainID \in ChainIDs |-> {}]
+
+\* Next state action
+\*    Some chain either
+\*        - advances its height
+\*        - updates the height of its light client
+\*        - receives datagrams and updates its state
+Next == 
+    \/ AdvanceChain
+    \/ UpdateOwnClient
```
milosevic: Why we update our own client?

verification/spec/relayer/relayer.tla
```diff
@@ -0,0 +1,225 @@
...
+        ELSE \* the dst client exists at the src chain
+            IF dstClientHeight < dstChainHeight
+            THEN \* the height of the dst client is smaller than the height of the dst chain
+                {[type |-> "ClientUpdate",
+                  height |-> relayer.clientHeights[dstChainID],
+                  clientID |-> GetClientID(srcChainID, dstChainID)]} 
+            ELSE {} IN
+                    
+    LET srcChainHeight == GetLatestHeight(srcChainID) IN    
+    LET srcClientHeight == GetClientHeight(dstChainID, srcChainID) IN
```
milosevic: Can we extract this logic into a function? It seems that it is the same logic in these two blocks.

PR #42, opened 2020-03-20
[WIP] Connection handshake English spec

### LINKED ISSUE #34: Connection Handshake: English specification of the problem
## Summary

This issue tracks the deliverable "L1: High level English specification of the problem" of #33.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Contributor assigned/self-assigned

### LINKED ISSUE #35: Connection Handshake: English specification of system model
## Summary

This issue tracks the deliverable "L2: High-level English specification of system model" of #33.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Contributor assigned/self-assigned

### LINKED ISSUE #36: Connection Handshake: English protocol description
## Summary

Issue for tracking "L2: English description of the protocol" of #33.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Contributor assigned/self-assigned

adizere:
Resolves #34.
Resolves #35.

## Description

The following sections are ready for review:

- [L1. Problem statement / outside view](https://github.com/informalsystems/ibc-rs/blob/adi/connection_handshake_spec/docs/spec/connection-handshake/L1_2.md#l1-problem-statement--outside-view) in its entirety. (re: #34)

- [L2 / 2.1 System model specification](https://github.com/informalsystems/ibc-rs/blob/adi/connection_handshake_spec/docs/spec/connection-handshake/L1_2.md#21-system-model-specification) in its entirety (re: #35)

Not yet ready to close #36.

______

For contributor use:

- [ ] Unit tests written


docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,264 @@
...
+
+> Reviews are welcome.
+
+
+## L1. Problem statement / outside view
+
+Suppose there are two chains (e.g., distributed ledgers) that execute independently of each other.
+What should be a basic abstraction for representing a path of communication between the two?
+We use the term _connection_ to represent this path of communication.
+The purpose of the IBC connection handshake protocol, at the highest level of abstraction, is to enable two chains to establish a connection with each other [[ICS 003](#references)].
```
milosevic: I think that it is more precisely to talk about connection establishment between two modules that are executing as deterministic state machines on different chains.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,264 @@
...
+## L2. Protocol specification / protocol view
+
+In this section we attempt to make the problem specification more concrete.
+We define what the processes are, their fault-model, as well as communication and synchrony assumptions.
+
+### 2.1 System model specification
+
+#### System model
+
+A **chain** is a state machine.
```
milosevic:
```suggestion
A **chain** is a deterministic state machine.
```

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,264 @@
...
+Each chain may be replicated, i.e., consisting of a replicated state machine (RSM), or it may be a standalone process.
+In practice, there are certain requirements which each chain must satisfy (e.g., see [[ICS 024](#references)]).
+At this level of specification, however, we can assume each chain is a single **process**.
```
milosevic: It is maybe better talking about module instead of chain as we can have multiple modules from some chain establishing connections with modules from remote chain.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,264 @@
...
+
+The connection handshake protocol is asymmetric, meaning that the two chains that interact play different roles within the protocol.
```
milosevic: We should be careful here as every module (on corresponding chain) can act as both initiator and responder. Although protocol is asymmetric we cannot necessarily define a priori who is playing what role. In addition, there is potentially a problematic case in which both modules at the same time trigger connection handshake (act as initiator) and we want to formalise this case and talk about guarantees we provide in that case.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,264 @@
...
+Neither of these functions are blocking, and they have no return value.
+We also assume that datagrams cannot be tampered with, i.e., they are authenticated by the sender process when calling `Send`.
+We assume that the destination process implicitly verifies authenticity upon calling `Deliver` for any datagram.
+
+The communication layer is _asynchronous_.
+There is no guarantee on how much time it will take for the destination of a datagram to deliver that datagram.
+
+
+###### Fault assumptions
+This specification operates under the assumption that processes `I` and `R` are not faulty, i.e., do not deviate from any protocol we describe below.
```
milosevic: My understanding is that we operate in the model where processes can behave arbitrarily (Byzantine fault), but guarantees are given only for correct case.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,264 @@
...
+     |     I     |   ConnHandshakeAck    |     R     |
+     |           |--------- 3 --------->>|           |
+     |           |                       |           |
+     |           | ConnHandshakeConfirm  |           |
+     |           |<<------- 4 -----------|           |
+     +-----------+                       +-----------+
+```
+
+
+###### Protocol handlers
```
milosevic: It is probably useful taking a look at this high level spec: https://github.com/informalsystems/VDD/pull/5. We should probably try to align with protocol description style used there, i.e., defining process state and then set of functions (handlers) where preconditions and postconditions are defined. This avoids writing pseudocode (as TLA+ specs will have some) but put clear expectations.

milosevic:
Good start! In my view it is maybe too high level. I am really starting to wonder if L1 part is doing more harm that it helps, as it tends to create formalism that is not really helping us. Protocol section should be up to the point, i.e., it should define all important terms and assumptions. In this case we probably want to define consensus state and light client, probably with some abstractions, but we still need to introduce those terms. The same with data store as state of the store is what we send proofs of to counter party (this is done in the current proposal). As section 2 is what engineer should read to start implementation, we need to be careful to provide right insights and introduce all important terms.

docs/spec/connection-handshake/L1_2.md
adizere: I agree we should be precise and use these terms accurately.

I hesitated a lot between module, chain, and process. I ended up using "chain" as a compromise, but primarily motivated by two ideas: (1) connections are a per-chain abstraction (i.e., a chain executes the logic of establishing a connection; multiple modules  can share the same connection), and (2) the term _module_ is generic and overused (to an outsider reader, a "module" means not more than "subsystem" and so is quite vague; to an IBC expert, a [module](https://github.com/cosmos/ics/tree/master/spec/ics-024-host-requirements#module-system) is an implementation detail for separating different IBC functionalities).

It's worth noting that idea (1) is more complicated than what I wrote. It is very abstract to claim that a chain "executes the logic of establishing a connection". Concretely, this logic resides in a module. But this is just an implementation detail, just like the consensus logic of Tendermint Core resides in the "consensus reactor" (we don't usually speak about the reactor in high-level specs). This analogy to "reactors" was useful for me, but not sure how correct it is. What do you think?

docs/spec/connection-handshake/L1_2.md
adizere: Hmm, I wasn't aware we require chains to be deterministic -- I was thinking that maybe randomized-based algorithms, e.g., Bitcoin, are allowed in IBC? But I guess when viewed from the "outside" (from the perspective of other chains and IBC) each chain must appear to behave deterministically -- is that what you mean?

docs/spec/connection-handshake/L1_2.md
adizere: Indeed, there are some tricky cases to consider there. I tried to rule out some of these corner-cases with the property **[CH/Proto/1/Uniqueness]**. As long as connections have different (unique) identifiers, they can proceed concurrently and should not matter who was initiator or responder.

docs/spec/connection-handshake/L1_2.md
adizere: Hmm, I think I was vague here. Maybe it would help to restate this as follows? "Processes `i` and `r` may be subject to arbitrary/Byzantine faults, but the properties we define below only hold as long as both `i` and `r` behave correctly.

adizere:
> Good start! In my view it is maybe too high level. I am really starting to wonder if L1 part is doing more harm that it helps, as it tends to create formalism that is not really helping us. Protocol section should be up to the point, i.e., it should define all important terms and assumptions. In this case we probably want to define consensus state and light client, probably with some abstractions, but we still need to introduce those terms. The same with data store as state of the store is what we send proofs of to counter party (this is done in the current proposal). As section 2 is what engineer should read to start implementation, we need to be careful to provide right insights and introduce all important terms.

Thanks for the feedbacK!

Writing L1 forced me to think about the problem abstractly. Personally, that was a valuable step, it helped my mind to "decouple" the problem of connection handshake from absolutely every dependency and think about it in isolation. I don't know how valuable L1 is, however, to others, e.g., novice readers, outsiders... If it brings confusion, then indeed we should eliminate it.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,264 @@
...
+
+connectionEnd = createConnectionEnd(connectionEnd_A)
+connectionEnd.state = TRYOPEN
+
+consensusStateA = getConsensusState(connectionEnd.clientIdentifier)
+
+send ConnOpenAck( connectionEnd, getProof(connectionEnd), consensusStateA, getProof(consensusStateA))
+}
+```
+
```
ancazamfir: The code is not aligned, missing parenthesis, etc. It is not clear on what side (A vs B) these are run. I think code  and figure above are inconsistent.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,264 @@
...
+Each chain may be replicated, i.e., consisting of a replicated state machine (RSM), or it may be a standalone process.
+In practice, there are certain requirements which each chain must satisfy (e.g., see [[ICS 024](#references)]).
+At this level of specification, however, we can assume each chain is a single **process**.
+
+The connection handshake protocol is asymmetric, meaning that the two chains that interact play different roles within the protocol.
+We take advantage of this asymmetry and explicitly label one of the chains with `I` (the initiating, or source, process), and the other with `R` (the responder, or destination, process).
+
+On top of each chain, one or multiple **applications** may be running (e.g., payment system or token exchanges, with their associated smart contracts).
+In the context of the connection handshake problem, it is an application that invokes the connection handshake protocol, and provides the necessary input for executing this protocol (we give more details on the invocation & input in section [2.2 Protocol](#22-protocol) below).
+
```
ancazamfir: Maybe it is better to use the spec term of `actor` here.
 `An actor, or a user (used interchangeably), is an entity interacting with the IBC protocol. An actor can be a human end-user, a module or smart contract running on a blockchain, or an off-chain relayer process capable of signing transactions.`
It is possible that the connection is initiated by an end-user or relayer, and the handshake completed by the relayer.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,264 @@
...
+At a high-level, the handshake protocol consists of processes `I` and `R` exchanging a few messages called **datagrams**.
+Each datagram has a well-defined syntax which we describe later.
+__[FIX: describe this and put ref to [2.2 Protocol](#22-protocol).]__
+
+We define a **connection** as a bidirectional or unidirectional path of communication that applications on the initiator and responder chains use to exchange **packets**.
+Packets are payloads from the point of view of a connection; they contain arbitrary application-level data, and their content is opaque to a connection; suffice to say that applications generate and consume packets.
```
ancazamfir: I am not sure about the correctness of these statements, what does it mean exactly "content is opaque to a connection"? or "packets are payloads from the point of view of a connection"? Where are the channels for example? The packets carry port and channel identifiers.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,264 @@
...
+
+We now restate the connection handshake problem in a slightly more precise (or lower level) formulation:
+A connection handshake enables two processes to agree on a consistent view of their state, to verify each other's state, and to allocate a new connection.
+
+
+###### Guarantees
+We refine the safety guarantee **[CH/Seq/1/Agreement]**  (defined in the [sequential problem statement](#sequential-problem-statement)) by splitting this into four complementary safety properties:
+
+- **[CH/Proto/1/Uniqueness]**
+If a process finishes a handshake for a `ConnectionEnd` `c`, no handshake has previously finished for `c`.
```
ancazamfir: Would it be enough to say that `connectionEnd c` should be unique? Even intermediate stages of the handshake require there is no other handshake in progress or finished for `c`

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,264 @@
...
+
+- **[CH/Proto/2/Validity]**
+If either of the responder or initiator agree to a `ConnectionEnd` `c`, then the initiator proposed `c`.
```
ancazamfir: I don't understand this.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,264 @@
...
+
+- **[CH/Proto/3/StateConsistency]**
+If process `i` establishes a connection with some process `r`, then process `i` has the `State` `si` representing process `r` and this state is consistent with the actual state of `r`.
+
+- **[CH/Proto/4/StateVerification]**
+If the processes establish a connection  for their respective `State` `si` (state of initiator) and `sr` (state of responder), then the initiator verified `sr` and the responder verified `si`.
+
+The liveness property below is functionally identical to the **[CH/Seq/2/Termination]** property we defined earlier:
+
+- **[CH/Proto/5/Termination]** If the initiator and responder are correct, then the handshake eventually terminates (i.e., both processes allocate a new `ConnectionEnd` object).
```
ancazamfir: should we say `correct termination` and require consistent `ConnectionEnd` objects?

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,264 @@
...
+
+The CH protocol consists of four datagrams that the initiator and responder exchange.
+We enumerate the names of these datagrams in the following code snippet.
+
+```golang
+type ConnHandshakeInit struct {  }
+type ConnHandshakeTry struct {  }
+type ConnHandshakeAck struct {  }
+type ConnHandshakeConfirm struct {  }
+```
```
ancazamfir: Should we use names closer to the IBC spec? Maybe they are not perfect but I think in this case using `ConnOpenInit`, etc. it's not too confusing. 
Also, the connection spec also covers the closing handshake.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,264 @@
...
+     |           |<<------- 2 -----------|           |
+     | Initiator |                       | Responder |
+     |     I     |   ConnHandshakeAck    |     R     |
+     |           |--------- 3 --------->>|           |
+     |           |                       |           |
+     |           | ConnHandshakeConfirm  |           |
+     |           |<<------- 4 -----------|           |
+     +-----------+                       +-----------+
+```
+
```
ancazamfir: This is not entirely correct. The `init` is sent to the initiator by some actor, the `try` should be from `I` to `R`, ...i.e. all arrows should be flipped here.

docs/spec/connection-handshake/L1_2.md
adizere: This means that a process never opens a connection having some details (encapsulated in variable `c`) unless that connection was in fact initiated with those precise details by the initiator. This property rules out bugs or corner-cases such as:

- a process proposes a connection `c` but opens a connection `c'` (i.e., there was tampering with `c` such as a man-in-the-middle attack).
- a process opens a connection `c` but there never existed a proposal to open this connection (the connection was fabricated/initiated by some other process).

None of the other properties cover these two corner-cases.

adizere: I can add these insights in the .spec in the [remarks] section below the properties if they help!

docs/spec/connection-handshake/L1_2.md
adizere: These properties are only intended to cover the case when both initiator and responder are correct, and in such a case, there is no possibility of incorrect/error termination. But you're raising a good point: what does it mean for a connection handshake to terminate incorrectly? Should we capture this in the specification?

docs/spec/connection-handshake/L1_2.md
adizere: If we don't fix names now, we'll probably never do it and remain stuck with them. Note that these are protocol-level modifications, so should be backwards-compatible with the Go implementation. But I have no strong preference, it is just a proposal to make things more consistent.

docs/spec/connection-handshake/L1_2.md
adizere: Will fix this, it is wrong indeed!

docs/spec/connection-handshake/L1_2.md
adizere: I'm not sure we need to introduce the term "module" to speak about connection handshake at this level of abstraction. I think we need two types of entities, not more: we can think of them as **client** (this is the user of the connection handshake, who initiates the protocol; the client is the external observer) and **server** (whoever it is that exposes the abstraction we're specifying here and implements the CH protocol).

In the current version of specification:
- client = application, and 
- server = chain = essentially an arbitrary process.

Should I modify this (also considering Anca's comment below) to be the following?

- client = actor, and
- server = module.

This would be more aligned with IBC specs terminology so would be useful! But in my view introduces unnecessary details that belong to implementation.

adizere: After thinking more about this, another option would be to have the following scheme:

- client = actor
- server = module
- and a chain is an external process, communicating with a module.

docs/spec/connection-handshake/L1_2.md
ebuchman: Bitcoin has randomness in its consensus algorithm, but the state machine it replicates is still a deterministic state machine (deterministic function of the txs).

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,416 @@
...
+
+#### Sequential problem statement
+
+###### Definitions
+When we speak of a party (either Alice or Bob), we identify three relevant abstractions:
+
+- a _module_: this is the process which actually carries the job of executing the CH protocol;
+- the _local state_ of the party;
+- the
```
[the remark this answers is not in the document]
adizere: Oops, the version you reviewed was a work-in-progress (I should not have pushed that). Will be fixed in the next version!

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,416 @@
...
+If a module establishes a `Connection` `c`, then `c` is a _valid_ connection; validity is defined by an application-level predicate.
+
+- **[CH/Proto/2/Uniqueness]**
+If a module establishes a connection `c`, no handshake has previously finished for `c`; uniqueness is defined by an application-level predicate.
+
+- **[CH/Proto/6/Termination]**
+Eventually, the connection handshake terminates, i.e., both parties mark their respective `ConnectionEnd` as `open`.
+
+- **[CH/Proto/7/Integrity]**
+If either Alice or Bob establishes a `Connection` `c`, then Alice proposed `c`.
```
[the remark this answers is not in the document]
adizere: That's a very good point. I only understood later that any party can be the proposer in a connection. This is going to be amended to the following version:

> If any two modules establish a connection `c`, then either one module or the other or both modules proposed `c`.

It might be redundant to say "or both" but keeping it like that for the moment.

docs/spec/connection-handshake/L1_2.md
adizere: After more thinking, I will incorporate the notion of "module" as a part of a chain. Hope this will resolve inconsistencies.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,416 @@
...
+type CommunicationLayer interface {
+    // A process `s` invokes this function to send a datagram `g` to
+    // a destination process `d`.
+    Send(g Datagram, d Process) null
+
+    // This callback function triggers at process `d` to deliver datagram `g`
+    // that process `s` previously sent.
+    Deliver(g Datagram, s Process) null
+}
+```
```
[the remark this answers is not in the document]
adizere: I think we cannot call these processes `i` and `r`. Alice `i` and Bob `r` can be playing any of the role of sender _or_ destination in _both_ of the Send and Deliver functions.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,416 @@
...
+
+2. A `ConnectionState` captures the state of a connection, and may have one of the following values. [TODO: not valid Go code below, keep?]
+
+```typescript
+enum ConnectionState {
+    INIT,
+    TRYOPEN,
+    OPEN,
+}
+```
```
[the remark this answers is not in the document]
adizere: As far as I know, nope.
Note that I'm using [this](https://github.com/cosmos/ics/tree/master/spec/ics-003-connection-semantics#data-structures) as a guide. Let me know otherwise!

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,416 @@
...
+
+- What is the interface between an actor and the chain? Seems that the main event for triggering a connection is `connOpenInit` which the Handler implements [ICS 25](https://github.com/cosmos/ics/tree/master/spec/ics-025-handler-interface).
+- Standard cryptographic assumptions (collision-resistant hash functions, public-key signatures)?
+- Unsure if terminology is clear enough, e.g., `data type`...
+- How to capture aborts?
+- Add support for versions in ConnectionEnd
+
+## Improvement suggestions
+
+- `ConnHandshakeTry` instead of `ConnOpenTry`. And similar renaming for the other handlers, replacing `Open` with `Handshake`, since this specification is for the connection _handshake_ problem, not the connection _open_ problem.
```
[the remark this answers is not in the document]
adizere: I'm going back to calling everything `ConnOpenTry` for the moment. Agree we have to come to an agreement regarding these terms as soon as possible!

docs/spec/connection-handshake/interfaces.md
```diff
@@ -57,17 +53,25 @@ function connOpenTry(
...
-  counterpartyClientIdentifier: Identifier
+  prefix: CommitmentPrefix
+}
+```
+
+- A `Connection` type:
+
+```typescript
+interface Connection {
+  state: ConnectionState
```
ancazamfir, line 72: This is the state of the remote/ counterparty end.

docs/spec/connection-handshake/interfaces.md
```diff
@@ -57,17 +53,25 @@ function connOpenTry(
...
+  localParty: Party
+  remoteParty: Party
   version: string | []string
```
ancazamfir, line 75: This is the version of the remote/ counterparty end.

docs/spec/connection-handshake/interfaces.md
```diff
@@ -81,125 +85,58 @@ interface Proof {
...
-    abortTransactionUnless(localConnEnd.verifyClientConsensusState(
+    abortTransactionUnless(e.remoteParty.verifyConnectionState(proofInit, e))
+    abortTransactionUnless(e.remoteParty.verifyClientConsensusState(
       proofConsensus, proofConsensus.height, expectedConsensusState))
-    previous = provableStore.get(connectionPath(localConnId))
+    previous = provableStore.get(connectionPath(e.remoteParty.connectionIdentifier))
     abortTransactionUnless(
       (previous === null) ||
-      (previous.state === INIT && previous === localConnEnd))
+      (previous === e))
```
ancazamfir, line 111: So this is a bit tricky wrt to the state. It happens that it works here. But in ConnOpenAck the remote state is expected to be TRYOPEN while the local could be INIT or TRYOPEN. If it is INIT this will fail. This is again because e.state is the remote state while previous.state is the local state.

docs/spec/connection-handshake/interfaces.md
```diff
@@ -81,125 +85,58 @@ interface Proof {
...
     state = TRYOPEN
-    provableStore.set(connectionPath(localConnEnd), connection)
-    addConnectionToClient(clientIdentifier, localConnEnd)
+    provableStore.set(connectionPath(identifier), connection)
+    addConnectionToClient(e.remoteParty.clientIdentifier, identifier)
 }
 ```
 
-### 3. Abstraction + no flipping
+Notice that this handler accesses predominantly the `remoteParty` fields. This happens because the relayer does not perform any "flipping" of parameters, so the perspective on this object is the one of the counterparty module. This can make it difficult to reason about, so below we discuss a version with flipping.
```
ancazamfir, line 119: Hmm, I am confused now. `e.localParty` is referring to the node running the handler above, no? And `e.remoteParty` to the other side. The relayer normally performs flipping.

docs/spec/connection-handshake/interfaces.md
```diff
@@ -0,0 +1,168 @@
...
+
+##### New Handler (version 2)
+
+Now we redefine `connOpenTry` handler as follows:
+
+```typescript
+function connOpenTry(
+  e: Connection,
+  proofInit: Proof,
+  proofConsensus: Proof) {
```
milosevic, line 96: I think that you are missing consensusHeight. ProofConsensus is a proof for that piece of data. Maybe it's better calling Connection ConnectionEnd.

docs/spec/connection-handshake/interfaces.md
```diff
@@ -0,0 +1,168 @@
...
+    abortTransactionUnless(validateConnectionIdentifier(e.remoteParty.connectionIdentifier))
+    abortTransactionUnless(proofConsensus.height <= getCurrentHeight())
+    expectedConsensusState = getConsensusState(proofConsensus.height)
```
milosevic, line 99: See above.

docs/spec/connection-handshake/interfaces.md
```diff
@@ -0,0 +1,168 @@
...
+
+    // basic validation of connection parties
+    abortTransactionUnless(e.state === INIT)
+    abortTransactionUnless(e.remoteParty.prefix === getCommitmentPrefix())
+
+    abortTransactionUnless(e.remoteParty.verifyConnectionState(proofInit, e))
+    abortTransactionUnless(e.remoteParty.verifyClientConsensusState(
+      proofConsensus, proofConsensus.height, expectedConsensusState))
```
milosevic, line 107: See comment above.

docs/spec/connection-handshake/interfaces.md
```diff
@@ -0,0 +1,168 @@
...
+    previous = provableStore.get(connectionPath(e.remoteParty.connectionIdentifier))
```
milosevic, line 108: I think that calling this previous is confusing. It is connectionEnd.

docs/spec/connection-handshake/interfaces.md
milosevic: I don't think that previous should be equal to e. They should correspond to the same connection, but each from its view. So e.localParty == previous.remoteParty, and vice versa.

docs/spec/connection-handshake/interfaces.md
```diff
@@ -0,0 +1,168 @@
...
+    abortTransactionUnless(
+      (previous === null) ||
+      (previous === e))
+    identifier = e.remoteParty.connectionIdentifier
+    state = TRYOPEN
+    provableStore.set(connectionPath(identifier), connection)
```
milosevic, line 114: What is connection here?

docs/spec/connection-handshake/interfaces.md
adizere: Yep. Tough I'm leaning towards removing the state field altogether..

docs/spec/connection-handshake/interfaces.md
milosevic: Alice and Bob are sending each other messages. Alice create connectionEnd and sends it to the Bob with a proof. As Alice initiates the connection, it chooses identifier for both side of the connection. Once Bob receives the message from Alice (connectionEnd with the proof), it verifies proof and creates the corresponding connectionEnd locally. I really don't understand why it is necessary talking about remoteParty view or flipping. Each party updates local state based on verified data it receives, and that's it.

docs/spec/connection-handshake/interfaces.md
adizere: I defined the `Proof` type (see on line 82 just above your comment) to include the height. Including that definition here:

```
interface Proof {
   content: CommitmentProof,
   height: uint64,
 }
```

docs/spec/connection-handshake/interfaces.md
adizere: That's a typo, sorry. Replace `connection` with `e` (the argument to this function).

docs/spec/connection-handshake/interfaces.md
adizere: No, this version has no flipping. See below for the version with flipping.

docs/spec/connection-handshake/interfaces.md
adizere: Note that in this version there is no flipping of _remote_ with _local_. So the module executing this handler sees everything in the `connection` object (and stores it) without flipping any field. For this reason, we can simply do equality check.

docs/spec/connection-handshake/interfaces.md
adizere: We removed all `ConnectionEnd` types and now we only have `Party` and `Connection`. In the previous version of this document we were using `ConnectionEnd` and it looked like there was too much redundancy in data we were transmitting. See [this commit comment](https://github.com/informalsystems/ibc-rs/commit/1842ed964b41cd0a0e83e8c10e1c2ba4e4791a2d#r38238027) for more info.

docs/spec/connection-handshake/L1_2.md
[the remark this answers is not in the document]
istoilkovska: I think "or both" should stay there, as it captures that *at least one* of the modules proposed `c`. You could also consider to formulate it as the contrapositive:
> If neither Alice nor Bob proposed a connection, then no connection between Alice and Bob is established.

docs/spec/connection-handshake/interfaces.md
ancazamfir: Ok, i am confused. In this version the flipping is done by the relayer (as it us currently) so when this handler receives `e` then `e.local` refers to its local state. Is this correct? If we agree can we be precise in what "flipping" means?

ancazamfir: ok, sorry i was looking at the wrong handler here...so let's show version3 before version2

adizere:
After discussions with @ancazamfir @milosevic @istoilkovska and @cwgoes, a new version of L1 & L2. Addressing #34 #35 #36. This spec depends on the relayer <-> chain interface proposal from #50.

Rendered file: [here](https://github.com/informalsystems/ibc-rs/blob/e720b30ca67097bb4a7794b530376a51c637fb2c/docs/spec/connection-handshake/L1_2.md).

adizere:
Missing elements from this spec:

- mention possibility of multiple relayers
- explanation for the `constructStateProof()` and `constructClientProof()`: we should clarify both here and in [ICS 018] how are proofs constructed
- the messages at L3 should be called **datagrams** (but not at L2, to be consistent with the TLA+ spec)

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+
+In addition to these four function, we also define a helper method to allow a relayer to obtain a list of all connections that a module stored.
+
+```golang
+    // A relayer can invoke this to `read` all connections stored by a chain.
+    public getAllConnections() // returns a `ConnectionEnd[]`
+```
+
+Refinement remarks (relation to L3):
+  - The `push` function is a higher-level version of `submitDatagram` [ICS018](#references).
```
ancazamfir, line 151: what does "higher-level" mean?

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+
+3. A `ConnectionEnd` is a data type that captures all the details of a connection at a party.
+One of the goals of the connection handshake protocol is to allocate an object of this type at each of Alice and Bob.
+
+```golang
+type ConnectionEnd interface {
+  state                             ConnectionState
+  remoteParty                       ConnectionParty
+  localParty                        ConnectionParty
+  version                           string | []string
```
ancazamfir, line 203:
```suggestion
  version                           []String
```

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+4. A `Proof` is a data type that abstracts over the details of a cryptographic proof that any module can create, and another module can verify.
+Intuitively, a proof is helpful for some module to guarantee that it is in a certain state.
+
+```golang
+type Proof struct {
+  content     CommitmentProof // the actual proof
+  height      uint64
+}
+
+```
```
ancazamfir, line 219: this needs fixing as per TLA+ spec discussion.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+  // was previously set at key `k`.
+  get(k) -> byte[]
+
+  // Sets the value at position `k` to be `v`, overwritting if this key is
+  // already present; always returns null (success).
+  set(k string, v byte[]) -> null
+}
+```
+
+Importantly, this store is shared with the relayer.
```
ancazamfir, line 237: What does this mean? The relayer does not have direct access to the store, only via queries. Queries can be made by anyone, relayer, user, scripts, etc.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+This property applies to _any two modules_ (there is no predefined "first" or "second" module in a connection).
+
+
+### 2.2 Protocol
+
+The ICS3 protocol comprises four steps, summarized below.
+
+1. An actor invokes the `ConnOpenInit` handler at the ICS3 module in one of the chains; this sets off the connection handshake protocol.
+In our example, we will use Alice as the party to execute `ConnOpenInit`.
+Once Alice executes this handler, she marks the connection on her end as initialized (i.e., `INIT`).
```
ancazamfir, line 282: it's not "she" but the handler that marks that.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+  connectionEnd.remote = remote
+  connectionEnd.version = getCompatibleVersions()
+
+  // Validity check. (Explained below in helper functions).
+  abortTransactionUnless(validate(connectionEnd))
+
+  // Uniqueness check. (Explained below in helper functions).
+  abortTransactionUnless(unique(connectionEnd))
+
+  // Now we save this connectionE in our local store.
```
ancazamfir, line 374: In general I would avoid "we" in specs.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+We first present the signatures of the four protocol handlers; the connection handshake module at each party implements and exposes these handlers.
+
+```golang
+func ConnOpenInit(local ConnectionParty,
+                  remote ConnectionParty)
+
+func ConnOpenTry(local ConnectionParty,
+                 remote ConnectionParty,
+                 remoteVersions string[],
+                 remoteStateProof Proof,
```
ancazamfir, line 338: this is ambiguous
```suggestion
                 remoteConnectionProof Proof,
```

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+
+Preconditions:
+  - the connection (obtained from the input `local` and `remote` arguments) is valid and unique
+
+Postconditions:
+  - the module executing this handler stores a `ConnectionEnd` (matching with the input arguments `local` and `remote`) in state `INIT`
+  - function produces no output; may abort if preconditions not met
+
+Implementation remarks:
+  - to ensure termination of the ICS3 protocol, this handler must execute at least once (either one of the chains or both)
```
ancazamfir, line 389:
```suggestion
  - to ensure termination of the ICS3 protocol, this handler must execute at least once (either on one of the chains or both)
```

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+
+
+###### ConnOpenTry
+
+This is step 2.b.
+
+```golang
+func ConnOpenTry(local ConnectionParty,
+                 remote ConnectionParty,
+                 remoteVersions string[],
```
ancazamfir, line 399:
```suggestion
                 remoteVersions []String,
```

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+  connectionEnd.state = TRYOPEN
+  connectionEnd.local = local
+  connectionEnd.remote = remote
+  connectionEnd.version = pickVersion(remoteVersions)
+
+  // Validity check.
+  abortTransactionUnless(validate(connectionEnd))
+
+  // Uniqueness check.
+  abortTransactionUnless(unique(connectionEnd))
```
ancazamfir, line 413: This should check that either there is no connection of if there is on it is in `INIT` state.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+  - the connection (obtained from inputs `local` and `remote`) is valid and unique
+  - the module running this handler supports at least one of the versions supplied in the input list `remoteVersions`
+  - the two proofs `remoteStateProof` and `remoteClientProof` are syntactically and semantically correct; this is necessary in connection to properties [ICS3-Proto-3-ConnectionIntegrity] and [ICS3-Proto-4-StateConsistency]
+
+Postconditions:
+  - the module executing this handler stores a `ConnectionEnd` (matching with the arguments `local` and `remote`) in state `TRYOPEN`
+  - no output; may abort if preconditions not met
+
+Implementation remarks:
+  - the `validate` check should also check the field `connectionEnd.version`.
```
ancazamfir, line 436: not sure what this check entails. I assume `validate` validates all fields.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+  abortTransactionUnless(connectionEnd.state == INIT
+                         || connectionEnd.state == TRYOPEN)
+
+  // Verification of proofs. See definition below in helper functions.
+  abortTransactionUnless(
+    verifyProofs(remoteStateProof, remoteClientProof, connectionEnd))
+
+  // Another integrity check: We should not open a connection unless it is for
+  // a version that we agree with.
+  abortTransactionUnless(getCompatibleVersions().indexOf(remoteVersion) !== -1)
```
ancazamfir, line 468: this is too detailed for this spec imo
```suggestion
  abortTransactionUnless(compatibleVersions(remoteVersion))
```

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+
+  // Termination (partial -- only at this end)
+  connectionEnd.state = OPEN
```
ancazamfir, line 471: also set version.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+  - The module executing this handler already has a `ConnectionEnd` stored locally, matching the `local` argument, and in state `INIT` or `TRYOPEN`.
+  - The supplied argument `remoteVersion` is supported (compatible)
+  - the two proofs `remoteStateProof` and `remoteClientProof` are syntactically and semantically correct; for guaranteeing properties [ICS3-Proto-3-ConnectionIntegrity] and [ICS3-Proto-4-StateConsistency]
+
+Postconditions:
+  - the module executing this handler stores a `ConnectionEnd` (matching with the arguments `local`) in state `OPEN`
+  - no output
+
+Implementation remarks:
+  - The `abortTransactionUnless(getCompatibleVersions())` check is very low-level. The `valid()` predicate should verify the `version` field, so it would be cleaner to invoke that predicate.
```
ancazamfir, line 489: what is the `valid()` predicate?

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+  - (none)
+
+
+##### Datagrams
+
+A correct relayer can pushe the following datagrams at a chain.
+Upon popping a datagram of a certain type, e.g., `ConnOpenConfirm`, a correct chain handles the datagram by invoking an eponymous handler.
+
+```golang
+type ConnOpenInit struct {
```
ancazamfir, line 540: I would use `ConnXHandler(X)` for handlers then as we don't want the same name.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+  local   ConnectionParty
+  remote  ConnectionParty
+}
+
+type ConnOpenTry struct {
+  local               ConnectionParty
+  remote              ConnectionParty
+  remoteVersions      string[]
```
ancazamfir, line 548:
```suggestion
  remoteVersions      []String
```

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+  remoteStateProof    Proof
+  remoteClientProof   Proof
+}
+
+type ConnOpenAck struct {
+  local               ConnectionParty
+  remoteVersion       string
```
ancazamfir, line 555:
```suggestion
  remoteVersion       String
```
in other places too

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+
+type ConnOpenConfirm struct {
+  local             ConnectionParty
+  remoteStateProof  Proof
+}
+```
+
+Notice that the fields in each datagram match with the corresponding protocol handler signature.
+
+##### Relayer Logic
```
ancazamfir, line 568: I think we should remove this section

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+  }
+}
+```
+
+#### Helper functions & application-level predicates
+
+###### `verifyProofs`
+
+```golang
+func verifyProofs(remoteStateProof Proof,
```
ancazamfir, line 650: I would make this function call a `verifyProof` for all proofs. Or even do two `verifyProof` calls in the callers/ handlers.
Then a `verifyProof` will do both what you call syntactical and semantical checks

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+                  remoteClientProof Proof,
+                  connectionEnd ConnectionEnd)
+  // Two syntactic check: for each proof, verify that it is authentic.
+  // This return either true (success) or false (failure).
+  // It is also acceptable that this function triggers an abort (e.g., due to
+  // malformed input).
+  t = verifyProofAuthenticity(remoteStateProof, connectionEnd)
+  if t == false {
+    return false
+  }
```
ancazamfir, line 660:
```suggestion
  if ~verifyProofAuthenticity(remoteStateProof, connectionEnd) {
    return false
  }
```

ancazamfir: what does `verifyProofAuthenticity(remoteStateProof, connectionEnd)` do exactly?

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+  if remoteClientProof != nil {
+    // This return either true (success) or false (failure). Aborting is also
+    // acceptable.
+    t = verifyProofAuthenticity(remoteClientProof, connectionEnd)
+    if t == false {
+      return false
+    }
```
ancazamfir, line 667:
```suggestion
    if ~verifyProofAuthenticity(remoteClientProof, connectionEnd) {
      return false
    }
```
and same everywhere this pattern occurs

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+  // specific connectionEnd. This helps guarantee the property
+  // [ICS3-Proto-3-ConnectionIntegrity].
+  // May trigger an abort if the verification fails. Usually returns true
+  // or false.
+  t = verifyRemoteState(remoteStateProof, ConnectionEnd)
+  if t == false {
+    return false
+  }
+  // 2. Verify that the client on the remote party is consistent with my state.
+  // It seems that remoteStateProof is necessary for this step. This also
```
ancazamfir, line 683: shouldn't be

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+  // checks, among others, that remoteClientProof.height <= getCurrentHeight()
+  // cf. [ICS 003].
+  if remoteClientProof != nil {
+    // May trigger an abort if the verification fails. Usually returns true
+    // or false.
+    t = verifyRemoteClient(
+        remoteClientProof [,remoteStateProof], ConnectionEnd)
+    if t == false {
+      return false
+    }
```
ancazamfir, line 693:
```suggestion
    if ~verifyRemoteClient(remoteClientProof, getLocalConsensusState(remoteClientProof.height) {
      return false
    }
```

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+  }
+  return true
+```
```
ancazamfir, line 696: I think the proof need rework, like we did in TLA spec. Given the current IBC spec, both proofs must be taken at the same height, `proof_height`. For the verification of the consensus proof we need, in addition, the `consensus_height`.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+  - `remoteClientProof` may be nil
+  - `remoteStateProof` and `connectionEnd` have to be well-formed and non-nil
+
+Postconditions:
+  - returns true upon success, false if any verification failed.
+  - if any of the preconditions are violated, the function may abort.
+  - no side-effects.
+
+
+###### `unique(ConnectionEnd c)`
```
ancazamfir, line 708: i think we can achieve this with `getConnection`. There are two places where this is called and the `tryOpen` one actually needs to get the connection and verify that either it doesn't exist or if it exist then it is in `INIT` state.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+
+### Open questions:
+
+- Standard cryptographic assumptions (collision-resistant hash functions, public-key signatures). We should mention these once we have more details about the implementation.
+
+- How to capture aborts or incorrect termination? See [issue raised by Anca](https://github.com/informalsystems/ibc-rs/pull/42#discussion_r397077901). Is this necessary? To inform this discussion, more implementation detials are neeeded.
+
+- **Verification of the unbonding period** in `ConnOpenTry`. See [ICS/#403](https://github.com/cosmos/ics/issues/403).
+
+## L3: Implementation
```
ancazamfir, line 740: We should remove this section.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+```
+
+Preconditions:
+  - the datagram `g` as well as Identifier `i` and Connection `e` (respectively for `push`, `setConnection` and `getConnection`) are non-null;
+
+Postconditions:
+  - for `setConnection`: the local store on the module implementing this function stores at position `i` the value `e`;
+  - for `push`: the `outstandingDatagrams` queue will append the Datagram `g`
+
+Error condition: none
```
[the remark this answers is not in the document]
adizere: Agree with `getConnection`. Added the following:

> Error condition: 
>   - `getConnection` may fail (returning `null`) if no `ConnectionEnd` exists for the input parameter `i`.

What are the conditions when `setConnection` may fail though?

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+
+Properties of the communication layer:
+
+- **[ICS3-CommAssumption-1-RWIntegrity]** If some process returns a connection `e` from `getConnection(i)` for some module, then the module previously invoked `setConnection(i, e)`.
+
+- **[ICS3-CommAssumption-2-QueueIntegrity]** If some module `d` returns a datagram `g` from `Pop("outstandingDatagrams")`, then a process previously invoked `Push("outstandingDatagrams", g)`.
+
+Neither of the functions in this interface are blocking.
```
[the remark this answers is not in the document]
adizere: This may be confusing. I meant to say that the functions guarantee to always return e.g., by timing-out or returning the adequate output (they never hang). But this would probably deserve more explanation then we want to go into spec. So I'll remove this detail.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+
+In addition to these four function, we also define a helper method to allow a relayer to obtain a list of all connections that a module stored.
+
+```golang
+    // A relayer can invoke this to `read` all connections stored by a chain.
+    public getAllConnections() // returns a `ConnectionEnd[]`
```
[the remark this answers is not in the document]
adizere: Good point, this was mean to be used in the "Relayer" sub-section, which we agree to remove. So we can remove this function as well.

docs/spec/connection-handshake/L1_2.md
adizere: "more abstract," or "farther away from actual implementation"

adizere: we can also say "analogous to"

docs/spec/connection-handshake/L1_2.md
adizere: From the point of view of this spec it does not seem relevant that users or scripts (in addition to relayers) can make these queries. I will add a side-note, however, to mention this detail if you believe it would help.

docs/spec/connection-handshake/L1_2.md
adizere: Got it, trimming down "our" as well.

docs/spec/connection-handshake/L1_2.md
adizere: There is a full description of this function further down in the spec:

> ###### `validate(ConnectionEnd c)` 
> This checks that the object `connectionEnd.local.connectionIdentifier` is valid.
Similar in functionality with `validateConnectionIdentifier` from [ICS 003]; additionally, this function should also validate the `version` field.

docs/spec/connection-handshake/L1_2.md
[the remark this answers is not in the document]
ancazamfir: For example when the client identified by `clientIdentifier` has not been created. It is true that the callers will do all the checks but here we just present an interface.

docs/spec/connection-handshake/L1_2.md
ancazamfir: "shared" might imply that the relayer can write there. There is no special access that the relayer requires as compared with other processes. The store is queryable by anyone. Maybe something like:
```suggestion
Importantly, queries to this store can be made.
```

docs/spec/connection-handshake/L1_2.md
ancazamfir: "abstract" or "analogous to" are great.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,777 @@
...
+  - for `setConnection`: the local store on the module implementing this function stores at position `i` the value `e`;
+  - for `push`: the `outstandingDatagrams` queue will append the Datagram `g`
+
+Error condition: none
+
+Properties of the communication layer:
+
+- **[ICS3-CommAssumption-1-RWIntegrity]** If some process returns a connection `e` from `getConnection(i)` for some module, then the module previously invoked `setConnection(i, e)`.
+
+- **[ICS3-CommAssumption-2-QueueIntegrity]** If some module `d` returns a datagram `g` from `Pop("outstandingDatagrams")`, then a process previously invoked `Push("outstandingDatagrams", g)`.
```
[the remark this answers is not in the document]
ancazamfir: yeah, not sure. it was "ditto" with a previous similar comment. re-reading it again, isn't it supposed to be reversed? meaning we do not lose messages? same question for the one above but there if `setConnection` has failed then get will not work.

docs/spec/connection-handshake/L1_2.md
adizere: It was a typo. I meant `validate`

docs/spec/connection-handshake/L1_2.md
adizere: I see your point, agree.

docs/spec/connection-handshake/L1_2.md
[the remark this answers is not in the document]
adizere: You're right, there should be a "no loss" property for the queue. This property would be similar to the one that @milosevic is trying to model in the Relayer English spec, which he calls 'Termination':

> Termination: In presence of correct relayer, message m sent by chain A to chain B is eventually received by chain B.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,626 @@
...
+```golang
+func ConnTryHandler(local ConnectionParty,
+                    remote ConnectionParty,
+                    remoteVersions []String,
+                    proofsHeight uint64,
+                    remoteConnectionProof ConnProof,
+                    remoteClientProof ClientProof)
+{ 
+  // Create local end of the connection.
+  connectionEnd = NewConnectionEnd(
```
ancazamfir, line 388:
```suggestion
  connectionEnd = newConnectionEnd(
```

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,626 @@
...
+
+func ConnAckHandler(local ConnectionParty,
+                 remoteVersion String,
+                 proofsHeight uint64,
+                 remoteConnectionProof ConnProof,
+                 remoteClientProof ClientProof)
+
+func ConnConfirmHandler(local ConnectionParty,
+                     proofsHeight uint64,
+                     remoteConnectionProof ConnProof)
```
ancazamfir, line 334: It's time to fix alignments :) You can indent more if you want.
```suggestion
func ConnInitHandler(
    local ConnectionParty,
    remote ConnectionParty)

func ConnTryHandler(
    local ConnectionParty,
    remote ConnectionParty,
    remoteVersions []String,
    proofsHeight uint64,
    remoteConnectionProof ConnProof,
    remoteClientProof ClientProof)

func ConnAckHandler(
    local ConnectionParty,
    remoteVersion String,
    proofsHeight uint64,
    remoteConnectionProof ConnProof,
    remoteClientProof ClientProof)

func ConnConfirmHandler(
    local ConnectionParty,
    proofsHeight uint64,
    remoteConnectionProof ConnProof)
```

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,626 @@
...
+  // Uniqueness check.
+  // Verify that either there is no connection OR it is in INIT state (i.e., 
+  // the Init handler previously executed with these same parameters).
+  currnt = getConnection(local.connectionIdentifier)
+  if currnt != nil {
+    abortTransactionUnless(currnt.state = 'INIT')
+  }
+  abortTransactionUnless(currnt == nil || matchingEnds(connectionEnd, currnt))
+
+  expectedConn = NewConnectionEnd('INIT', remote, local)
```
ancazamfir, line 400:
```suggestion
  expectedConn = newConnectionEnd('INIT', remote, local, remoteVersions)
```

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,626 @@
...
+  expectedClient = consensusState(remoteClientProof.height)
```
ancazamfir, line 401:
```suggestion
  expectedClient = localConsensusState(remoteClientProof.height)
```

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,626 @@
...
+  abortTransactionUnless(verifyProof(
+      localClientState, remoteConnectionProof,
+      connectionPath(local.connectionIdentifier), expectedConn))
+
+  // Client proof verification. Helps ensure state consistency: The remote 
+  // client should be consistent with the state of local party.
+  abortTransactionUnless(remoteClientProof.height <= getCurrentHeight())
+  abortTransactionUnless(verifyProof(localClientState, remoteClientProof,
+        clientPath(local.clientIdentifier, remoteClientProof.height),
+        expectedClient))
```
ancazamfir, line 419: to be consistently aligned with the above `ferifyProof`. I am just going over some but we should try to have nice alignment overall.
```suggestion
  abortTransactionUnless(
        verifyProof(localClientState, remoteClientProof,
        clientPath(local.clientIdentifier, remoteClientProof.height),
        expectedClient))
```

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,626 @@
...
+  // Update the connectionEnd in local state.
+  setConnection(local.connectionIdentifier, connectionEnd)
+}
+```
+
+
+Preconditions:
+  - The module executing this handler already has a `ConnectionEnd` stored locally, matching the `local` argument, and in state `INIT` or `TRYOPEN`.
+  - The `local` argument is valid (syntactic check), and the `remoteVersion` is supported, i.e., compatible with the local chain -- see `getCompatibleVersions()`
+  - The two proofs `remoteConnectionProof` and `remoteClientProof` are syntactically and semantically correct; for guaranteeing properties [ICS3-Proto-2-ConnectionIntegrity] and [ICS3-Proto-3-StateConsistency]
```
ancazamfir, line 489: Why isn't this in the `tryOpen` also? I also want to make sure it is clear what syntactically correct means (non-null proofs? non-zero height?) and semantically correct (the proofs verify?)

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,626 @@
...
+This is identical with the one in [ICS 003](#references).
+
+###### `getCompatibleVersions()`
+This is identical with the one in [ICS 003](#references).
+
+
+###### `constructConnectionProof()` and `constructClientProof()`
+
+These functions are not shown here, but they are part of the relayer logic so we mention them nonetheless.
+These are more abstract variants of the `proof()` method that a connection implements (see [ICS 018](#references) in particular).
```
ancazamfir, line 598: i think we can remove these as there is no detail and no one calls them.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,626 @@
...
+
+###### `verifyProof`
+
+This function handles the cryptographic verification of a proof.
+It expects four arguments: the local client state, a proof, a path, and the expected data of the proof.
+This function is analogous to `verifyMembership` of the [ICS 023](#references), except it must additionally perform two actions: serialize the expected data of the proof, and to extract from the client state the path to the root of the tree.
```
ancazamfir, line 604:
```suggestion
This function is analogous to `verifyMembership` of the [ICS 023](#references), except it must additionally perform two actions: serialize the expected data of the proof, and to extract from the client state the root of the tree.
```

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,626 @@
...
+###### `consensusState` and `clientConsensusState`
+
+These are analogous to `getConsensusState` and `queryClient` functions.
+
+
+### Open questions:
+
+- Standard cryptographic assumptions (collision-resistant hash functions, public-key signatures). We should mention these once we have more details about the implementation.
+
+- How to capture aborts or incorrect termination? See [issue raised by Anca](https://github.com/informalsystems/ibc-rs/pull/42#discussion_r397077901). Is this necessary? To inform this discussion, more implementation detials are neeeded.
```
ancazamfir, line 615: (spellcheck?)
```suggestion
- How to capture aborts or incorrect termination? See [issue raised by Anca](https://github.com/informalsystems/ibc-rs/pull/42#discussion_r397077901). Is this necessary? To inform this discussion, more implementation details are needed.
```

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,626 @@
...
+We now restate the connection handshake problem in a slightly more precise (or lower level) formulation:
+
+__A connection handshake enables two ICS3 modules to agree on a consistent view of their chain state, to verify each other's state, and to allocate a new connection.__
+
+
+###### Guarantees
+We refine the safety guarantee **[ICS3-Seq-1-Agreement]** (defined in the [sequential problem statement](#sequential-problem-statement)) by splitting this into three complementary safety properties:
+
+- **[ICS3-Proto-1-ConnectionUniqueness]**
+If a module establishes a `ConnectionEnd` `e`, no handshake has previously finished for `e`.
```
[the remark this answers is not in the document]
adizere: Indeed, there is something fishy here. I have to think more about it. I will partially fix this by removing the notion of "establish" and say "open" instead.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,626 @@
...
+If any two modules establish a connection `e`, then either one module or the other or both modules accepted (i.e., initialized with) `e`.
+
+- **[ICS3-Proto-3-StateConsistency]**
+If any two modules establish a connection, then the client in the first module is consistent with the state of the second module.
+
+The liveness property below is equivalent to **[ICS3-Seq-2-Termination]** (the livenes property in the sequential specification above):
+
+- **[ICS3-Proto-5-Termination]**
+If the two parties involved in a connection handshake are correct, then the connection handshake eventually terminates.
+Specifically, termination implies that each module allocates in the local state a new `ConnectionEnd` object with `state` field set to `OPEN`.
```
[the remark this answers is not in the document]
adizere: In connection to `clientProof` we have this property:

> If any two modules establish a connection, then the client in the first module is consistent with the state of the second module.

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,626 @@
...
+
+
+###### Guarantees
+We refine the safety guarantee **[ICS3-Seq-1-Agreement]** (defined in the [sequential problem statement](#sequential-problem-statement)) by splitting this into three complementary safety properties:
+
+- **[ICS3-Proto-1-ConnectionUniqueness]**
+If a module establishes a `ConnectionEnd` `e`, no handshake has previously finished for `e`.
+
+- **[ICS3-Proto-2-ConnectionIntegrity]**
+If any two modules establish a connection `e`, then either one module or the other or both modules accepted (i.e., initialized with) `e`.
```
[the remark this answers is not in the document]
adizere: Will fix "establish".

docs/spec/connection-handshake/L1_2.md
```diff
@@ -0,0 +1,626 @@
...
+Specifically, termination implies that each module allocates in the local state a new `ConnectionEnd` object with `state` field set to `OPEN`.
+
+
+_Remarks_:
+
+- Uniqueness property essentially provides a safeguard against overwritting a connection in the store with some new set of parameters.
+
+- One of the mechanisms that enforces the connection integrity property consists of cryptographic proofs, specifically the `ConnProof` type, which asserts that a certain chain stores a certain connection in a certain state.
+
+- The `ClientProof` type helps guarantee the state consistency property, by asserting that the client of a certain chain is updated to a certain height of the remote chain.
```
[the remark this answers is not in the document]
adizere: But as far as I'm aware there is no requirement in ICS3 on how __recent__ the consensus state on a client has to be. It would be useful to have a freshness guarantee, but is this enforced by `verifyClientConsensusState`, or where?

docs/spec/connection-handshake/L1_2.md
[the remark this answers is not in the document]
adizere: I think this is a more accurate statement of the uniqueness property we want:

> A module accepts (i.e., initializes on) a `ConnectionEnd` `e` at most once.

docs/spec/connection-handshake/L1_2.md
adizere: I removed the confusion around syntactic & semantic checks.

docs/spec/connection-handshake/L1_2.md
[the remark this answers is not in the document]
ancazamfir: I see. So it looks like `**[ICS3-Proto-3-StateConsistency]**` is property that holds after termination, while 1 holds at all handshake steps and 2 from `tryOpen` onwards. I am not suggesting adding this details there but trying to understand what exactly we want to specify here.

Issue #49, opened 2020-04-08
Perform full verification when verifying headers using the light client


romac:
## Summary

Perform full verification when verifying headers using the light client.

## Problem Definition

The current implementation does not follow the specification as it only performs simple validation of the headers pulled from the chain.

This is currently blocked because the light client task is not currently thread-safe for reasons currently unclear (might have to do with missing `Send+Sync` bounds, or perhaps a bug with the `async-trait` crate we depend on.

## Proposal

We should fix that by performing full verification of the pulled headers. Either by:

a) fixing the `Send+Sync` issue directly
b) using a single-thread executor and spawning local tasks
c) temporarily* working around it by blocking on the light client task.

\* ie. until we figure out the new event-based architecture and adapt to it.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #51, opened 2020-04-09
Perform bisection when verifying headers using the light client

romac:
Closes: #49

## Description

This PR implements proposal B described in #49.

All tasks spawned in the `start` command are now spawned on a local task set which executes them all on the same thread, ie. the main thread.

This seems like an acceptable solution to me given that the overall architecture of the relayer and the light client will likely change quite a bit in the coming weeks.
______

For contributor use:

- [ ] Unit tests written

### DIFF
--- a/relayer/cli/src/commands/start.rs
+++ b/relayer/cli/src/commands/start.rs
@@ -85,6 +85,8 @@ async fn update_client<C: Chain, S: Store<C>>(mut client: Client<C, S>) {
     let mut interval = tokio::time::interval(Duration::from_secs(3));
 
     loop {
+        interval.tick().await;
+
         info!(chain.id = %client.chain().id(), "updating client");
 
         let result = client.update(SystemTime::now()).await;
@@ -104,8 +106,6 @@ async fn update_client<C: Chain, S: Store<C>>(mut client: Client<C, S>) {
                 error!(chain.id = %client.chain().id(), "error when updating headers: {}", err)
             }
         }
-
-        interval.tick().await;
     }
 }
 
--- a/relayer/cli/src/commands/start.rs
+++ b/relayer/cli/src/commands/start.rs
@@ -27,60 +27,81 @@ impl Runnable for StartCmd {
 
         debug!("launching 'start' command");
 
-        block_on(async {
+        // Spawn all tasks on the same thread that calls `block_on`, ie. the main thread.
+        // This allows us to spawn tasks which do not implement `Send`,
+        // like the light client task.
+        let local = tokio::task::LocalSet::new();
+
+        block_on(local.run_until(async move {
             for chain_config in config.chains {
                 info!(chain.id = %chain_config.id, "spawning light client");
-
-                let _handle = tokio::spawn(async move {
-                    let client = create_client(chain_config).await;
-                    let trusted_state = client.last_trusted_state().unwrap();
-
-                    info!(
-                        chain.id = %client.chain().id(),
-                        "Spawned new client now at trusted state: {} at height {}",
-                        trusted_state.last_header().header().hash(),
-                        trusted_state.last_header().header().height(),
-                    );
-
-                    update_headers(client).await;
-                });
+                tokio::task::spawn_local(spawn_client(chain_config));
             }
 
-            start_relayer().await
-        })
+            info!("starting relayer");
+            relayer_task().await
+        }));
     }
 }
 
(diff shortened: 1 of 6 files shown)


ancazamfir:
I tried to run the relayer on this branch and I don't see the light client update header messages. Just these:
`Apr 10 10:26:56.933  INFO relayer_cli::relayer: Relayer is running`

If I remove the relay thread then the relayer finishes quietly after `spawning light client...` messages
```
ancas-macbook-pro:1ibc-rs ancaz$ cargo run --bin relayer -- -c simple_config.toml start
...
Apr 10 10:30:21.657  INFO relayer_cli::commands::start: spawning light client chain.id=ibc0
Apr 10 10:30:21.658  INFO relayer_cli::commands::start: spawning light client chain.id=ibc1
ancas-macbook-pro:1ibc-rs ancaz$ 
```
Maybe I am doing smthg wrong, it's the first time I try this; followed the instructions from: https://github.com/informalsystems/ibc-rs/blob/master/README.md

romac:
Yeah, I get the same output too. My bad, I should have mentioned that and marked the PR as a draft. Will let you know once(if) it works and is actually ready for review.

ancazamfir:
ah ok, thought is ready for review :)

romac:
I had forgotten to `.await` the client task 🤦‍♂️

This now works as expected but has not been rebased on master yet.

I rebased locally, but the `start` command now fails because of a deserialization error when fetching the `TrustOptions` from the store. Might be an issue with `tendermint-rs@v0.33`. Will investigate further!

romac:
Blocked by https://github.com/informalsystems/tendermint-rs/pull/196#issuecomment-611969478.

romac:
Unblocked now that https://github.com/informalsystems/tendermint-rs/pull/210 has been merged.

romac:
@ancazamfir Ready for review.

relayer/cli/src/commands/start.rs
```diff
@@ -27,60 +27,81 @@ impl Runnable for StartCmd {
...
+        info!(target: "relayer_cli::relayer", "relayer is running");
         interval.tick().await;
     }
 }
 
-async fn update_headers<C: Chain, S: Store<C>>(mut client: Client<C, S>) {
+async fn update_client<C: Chain, S: Store<C>>(mut client: Client<C, S>) {
     debug!(chain.id = %client.chain().id(), "updating headers");
 
     let mut interval = tokio::time::interval(Duration::from_secs(3));
```
ancazamfir, line 85: From the documentation: "Note that when it starts, it produces item too."  So the  `interval.tick().await;` should be the first in the loop. I looked at the logs when running with two chains and noticed two updates per chain without the wait when starting the relayer.

relayer/cli/src/commands/start.rs
```diff
@@ -27,60 +27,81 @@ impl Runnable for StartCmd {
...
 
     loop {
+        info!(chain.id = %client.chain().id(), "updating client");
```
ancazamfir, line 88:
```suggestion
        interval.tick().await;
        info!(chain.id = %client.chain().id(), "updating client");
```
and remove from end of loop.

relayer/cli/src/commands/start.rs
```diff
@@ -27,60 +27,81 @@ impl Runnable for StartCmd {
...
+
         let result = client.update(SystemTime::now()).await;
```
ancazamfir, line 90: I believe this is the only place we should upate client, i.e. remove from `client.rs`
```
--- a/relayer/relay/src/client.rs
+++ b/relayer/relay/src/client.rs
@@ -93,10 +93,6 @@ where
             client.init_with_trust_options(trust_options).await?;
         }

-        // Perform an update already, to make sure the client is up-to-date.
-        // TODO: Should we leave this up to the responsibility of the caller?
-        let _ = client.update(SystemTime::now()).await?;
-
         Ok(client)
     }```

relayer/cli/src/commands/start.rs
romac: Good catch!

relayer/cli/src/commands/start.rs
romac: 👍

romac:
Just addressed your comments :)

ancazamfir:
> Just addressed your comments :)

Looks good! Thanks @romac!

Issue #47, opened 2020-04-07
Create and submit a transaction to SDK/TM chain


ancazamfir:
## Summary
Create and submit an IBC transaction to SDK/TM chain.

## Problem Definition
relayer needs to submit transactions that include the IBC messages and packets.

## Proposal
Evaluate https://github.com/iqlusioninc/crates/tree/develop/stdtx for submitting an IBC transaction.  Could be for ClientCreate or ClientUpdate, or other. 
Keystore doesn't have to be part of this if a workaround is found. We will deal with this in another issue.

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ebuchman:
Let's start with a ConnOpenInit - https://github.com/cosmos/ics/tree/master/spec/ics-003-connection-semantics#opening-handshake

ebuchman:
We should open a new issue listing the set of transactions so we can start knocking them off

ebuchman:
Protobuf txs have not landed yet on the SDK so lets bump the milestone on this for now

tarcieri:
Related:

- [stdtx: tracking issue for Protobuf transaction support (#459)](https://github.com/iqlusioninc/crates/issues/459)
- [tmkms tx-signer: post-Stargate roadmap](https://github.com/iqlusioninc/tmkms/issues/96)

ebuchman:
This is effectively done in #242 but bringing it into the relayer-cli is now blocking on #243 or some other resolution of dependency hell.

andynog:
Got a basic keystore working. You can test against branch `andy/tx-sign`. Running the command below

```
 ./target/debug/relayer -c ./relayer-cli/tests/fixtures/two_chains.toml keys restore ibc0 test "mixture fatigue fold federal hour frequent atom seek apart diesel armed start lawn hello milk say steak peasant wonder inquiry vague mind reduce rib"
```

Should give the result:
```
key restore result:  "55d4ae3492a831453fc273aff431fc76094e9b1c"
```

Next step is to use the private key to sign the message in the transaction.

andynog:
I believe the tx signing is working but getting a `internal` error testing against stargate-4 chains (running with go relayer dev-env script)

```
[...]

PR #52, opened 2020-04-11
WIP: StdTx

ebuchman:
Took a first stab at #47. Very rough draft (some of these types will belong in tendermint-rs, etc.), just posting to show what I was able to do: 

- marshal a StdTx with the MsgCreateClient to JSON
- sign the JSON with gaiacli
- load the signed StdTx back into Rust
- almost fully Amino encode it

There are lots of rough edges, and I found some roughness in the Go too (eg. https://github.com/cosmos/cosmos-sdk/issues/5983)

Note the [Iqlusion StdTx](https://github.com/iqlusioninc/crates/tree/develop/stdtx) library was somewhat useful, but not fully:
- it expects Msgs to be specified in a toml file
- from what I can tell, Msgs are limited to certain "basic" fields, so you can't eg easily encode a Tendermint header in there
- the StdTx type in there expects the Msg to be already Amino encoded
- the JSON encoding isn't fully compliant with Tendermint

That said, the StdFee, StdSignature, and Address types from that lib were useful, after a few modifications. But we might want to consider just moving those types and logic, if not the whole crate, into here.

As for Amino encoding, it seems we still need to make some fixes to amino-rs to fully get there. I'm comparing the Amino encoding of my MsgCreateClient with one generated from the Go and it's still off by just a little (mostly seems to be around extra length prefixing). I included the Go script in here (amino.go), but it should be run from the gaia repo (ibc-alpha branch).

Meanwhile, I had to make changes to tendermint-rs, amino-rs, and stdtx crates to get here:
- tendermint-rs: export some serialization - https://github.com/informalsystems/tendermint-rs/pull/212
- amino-rs: don't always length encode amino types - https://github.com/tendermint/amino_rs/pull/29
- stdtx: various fixes for JSON encoding - https://github.com/iqlusioninc/crates/pull/391

Note the Cargo's were updated to use local paths to facilitate development - I'm not sure if there's a nicer way to be doing that without messing up the file :)

At a higher level, I'm not sure if any of this is even the right approach at all, especially since once we go full proto, we'll need to adjust. But I wanted to try and get something working with the current state of things. Hopefully the protobuf changes for Gaia and Tendermint will land soon; it probably doesn't make sense to go too much further here in Amino land unless we can keep it relatively simple. We might want to start looking at the development branches for the proto migration and see if it makes sense to work against those so we can get a feel for the ultimate state of things sooner.
______

For contributor use:

- [ ] Unit tests written


ebuchman:
Old and stale so closing. Proto txs have landed in the SDK so we can use that

Issue #53, opened 2020-04-12
Relayer clients initialization cleanup


ancazamfir:
After restarting the relayer that had previously sync-ed some header, there are some debug messages about hashes not matching and trusted options height < header height in store. I think we need to look into this a bit, not sure why they occur as this is seen with good providers/ chains. 
(see TODO logs below)
```
    Finished dev [unoptimized + debuginfo] target(s) in 0.39s
     Running `target/debug/relayer -v -c simple_config.toml start`
Apr 12 13:58:03.207 DEBUG abscissa_core::component::registry: registered component: abscissa_core::terminal::component::Terminal (v0.5.2)
Apr 12 13:58:03.208 DEBUG abscissa_core::component::registry: registered component: abscissa_core::trace::component::Tracing (v0.5.2)
Apr 12 13:58:03.208 DEBUG abscissa_core::component::registry: registered component: abscissa_tokio::TokioComponent (v0.5.1)
Apr 12 13:58:03.209 DEBUG relayer_cli::commands::start: launching 'start' command
Apr 12 13:58:03.209  INFO relayer_cli::commands::start: spawning light client chain.id=ibc1
Apr 12 13:58:03.210  INFO relayer_cli::commands::start: spawning light client chain.id=ibc0
Apr 12 13:58:03.210  INFO relayer_cli::commands::start: starting relayer
Apr 12 13:58:03.210  INFO relayer_cli::relayer: relayer is running
     Relayer spawning light client for chain ibc1
Apr 12 13:58:03.373 DEBUG relayer::client: initializing client from trusted store chain.id=ibc1
Apr 12 13:58:03.373 DEBUG relayer::client: restoring trusted state from store chain.id=ibc1
Apr 12 13:58:03.374 DEBUG relayer::client: found trusted state in store chain.id=ibc1 last_height=5321
Apr 12 13:58:03.374 DEBUG relayer::client: found last trusted state
Apr 12 13:58:03.374 DEBUG relayer::client: restoring trusted state from store chain.id=ibc1
Apr 12 13:58:03.374 DEBUG relayer::client: trusted options height is lesser than header height in trust store. TODO: rollback chain.id=ibc1 trust_options.height=806 trusted_header.height=5321
[...]

PR #54, opened 2020-04-14
Start light clients from trusted store by default

romac:
See: #53

## Description

- Start light clients from trusted store by default.
- Add a `--reset/-r` option to force the clients to start from the configured trust options.

______

For contributor use:

- [ ] Unit tests written


romac:
Right that's indeed a downside of the current approach. Let's fix this later as you say!

PR #55, opened 2020-04-15
[WIP] Relayer TLA+ specification, connection and channel handshake datagrams

istoilkovska:
## Description

WIP for #5 

Updated Relayer TLA+ specification, that in addition to client update datagrams, now handles connection and channel handshake datagrams. 

The whole specification now contains five modules: 
  - `Relayer.tla` (the main module)
  - `Environment.tla`
  - `ClientHandlers.tla`
  - `ConnectionHandlers.tla`
  - `ChannelHandlers.tla`

The module `Relayer.tla` contains the specification of the relayer algorithm. 
It instantiates the module `Environment.tla`, which takes care of the chain logic. 
The module `Environment.tla` extends the modules `ClientHandlers.tla`, 
`ConnectionHandlers.tla`, and `ChannelHandlers.tla`, which contain definition of 
operators that handle client, connection handshake, and channel handshake
datagrams, respectively.

The `README.md` file contains instructions about how to load the specification in 
the TLA+ Toolbox and check the specified invariants and properties with TLC.
______

For contributor use:

- [ ] Unit tests written


verification/spec/relayer/Relayer.tla
```diff
@@ -0,0 +1,769 @@
...
+          relayerChainHeights, \* a function that assigns a height to each chainID         
+          turn  
+          
+vars == <<chains, pendingDatagrams, relayerChainHeights, turn>>            
+envVars == <<chains>>           
+
+\* Instance of the module Environment, which encodes the chain logic    
+Env == INSTANCE Environment 
+       WITH chains <- chains, 
+            pendingDatagrams <- pendingDatagrams,
```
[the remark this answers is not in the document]
adizere: (learned about this in a more principled way today!)

verification/spec/relayer/Environment.tla
```diff
@@ -0,0 +1,334 @@
...
+        height : Heights \union {nullHeight},
+        counterpartyClientHeight : Heights \union {nullHeight},
+        connectionEnd : ConnectionEnds
+    ]
+    
+(******************************** Datagrams ********************************
+ A set of datagrams.
+ ***************************************************************************)
+Datagrams ==
+    [type : {"CreateClient"}, clientID : ClientIDs, height : Heights]
```
adizere, line 96: One question here:
Why is the `type` always a set? I understand everything is a set in TLA+. But could we also write `type : "CreateClient"` here and get the same behavior?

ancazamfir:
`Relay` calls `PendingDatagrams` for each `(srcChainID,  dstChainID)` combination in `chainIDs` but then inside `PendingDatagrams` we do both `(srcChainID,  dstChainID)` and `(dstChainID, srcChainID)` so it seems we do the work twice. I think we can simplify `PendingDatagrams` to only build the datagrams to `dstChainID`.

verification/spec/relayer/Environment.tla
```diff
@@ -0,0 +1,334 @@
...
+UpdateChain(chainID, datagrams) == 
+    LET chain == chains[chainID] IN
+    \* ICS 002: Client updates
+    LET lightClientsUpdated == LightClientUpdate(chainID, chain, datagrams) IN 
+    \* ICS 003: Connection updates
+    LET connectionsUpdated == ConnectionUpdate(chainID, lightClientsUpdated, datagrams) IN
+    \* ICS 004: Channel updates
+    LET channelsUpdated == ChannelUpdate(chainID, connectionsUpdated, datagrams) IN
+    
+    channelsUpdated
```
ancazamfir, line 241: the `chain.height` should be also increased.

verification/spec/relayer/ConnectionHandlers.tla
```diff
@@ -0,0 +1,168 @@
...
+    
+
+\* Handle "ConnOpenTry" datagrams
+HandleConnOpenTry(chainID, chain, datagrams) ==
+    \* get "ConnOpenTry" datagrams, with a valid connection ID and valid height
+    LET connOpenTryDgrs == {dgr \in datagrams : 
+                            /\ dgr.type = "ConnOpenTry"
+                            /\ dgr.desiredConnectionID = GetConnectionID(chainID)
+                            /\ dgr.consensusHeight <= chain.height
+                            /\ dgr.proofHeight <= chain.counterpartyClientHeight} IN
```
ancazamfir, line 74: I believe we need to have a `chain.counterpartyClientHeights` to simulate the full set of consensus states on this chain and then check that `dgr.proofHeight \in chain.counterpartyClientHeight`

verification/spec/relayer/ChannelHandlers.tla
```diff
@@ -0,0 +1,172 @@
...
+ ***************************************************************************)
+ 
+\* Handle "ChanOpenInit" datagrams
+HandleChanOpenInit(chainID, chain, datagrams) ==
+    \* get "ChanOpenInit" datagrams, with a valid channel ID
+    LET chanOpenInitDgrs == {dgr \in datagrams : 
+                            /\ dgr.type = "ChanOpenInit"
+                            /\ dgr.channelID = GetChannelID(chainID)} IN
+    
+    IF chanOpenInitDgrs /= {} /\ chain.connectionEnd.state /= "UNINIT"
```
milosevic, line 47: Should it be 'chain.connectionEnd.channelEnd.state'?

verification/spec/relayer/Relayer.tla
```diff
@@ -0,0 +1,769 @@
...
+           connectionID |-> dstConnectionEnd.connectionID, \* "connBtoA" (if srcChainID = "chainA", dstChainID = "chainB")
+           proofHeight |-> srcHeight]} 
+    ELSE {}
+
+\* Get the connection datagrams designated for srcChainID and dstChainID         
+ConnectionDatagrams(srcChainID, dstChainID) ==
+    \* connection datagrams from dstChainID to srcChainID 
+    LET srcConnectionDatagrams == ComputeConnectionDatagrams(dstChainID, srcChainID) IN
+    \* connection datagrams from srcChainID to dstChainID
+    LET dstConnectionDatagrams == ComputeConnectionDatagrams(srcChainID, dstChainID) IN
```
milosevic, line 222: If I am not wrong `ConnOpenInit` will be created for both source and destination chain in case `dstConnectionEnd.state = "UNINIT" /\ srcConnectionEnd.state = "UNINIT"`?

verification/spec/relayer/ClientHandlers.tla
```diff
@@ -0,0 +1,94 @@
...
+                            /\ dgr.type = "ClientUpdate"
+                            /\ dgr.clientID = GetCounterpartyClientID(chainID)} IN
+    \* get heights in datagrams with correct counterparty clientID for chainID
+    LET updateClientHeights == {h \in Heights : \E dgr \in updateClientDgrs : dgr.height = h} IN    
+
+    \* new chain record with clients updated
+    LET clientUpdatedChain == [
+            height |-> chain.height,
+            counterpartyClientHeight |-> IF updateClientHeights /= {}
+                                         THEN Max(updateClientHeights)
```
milosevic, line 78: We should store all installed heights, not just the maximum height.

verification/spec/relayer/ChannelHandlers.tla
ancazamfir: I think this check is for the state of the underlying connection, to make sure it is initiated.

verification/spec/relayer/ChannelHandlers.tla
```diff
@@ -0,0 +1,172 @@
...
+    \* if there are valid "ChanOpenInit" datagrams and the connection is not "UNINIT", 
+    \* create a new channel end and update the chain
```
ancazamfir, line 49: Can you move the comment above the `IF`. The comments should refer to lines after the comments. There are a number of places where they are after.

verification/spec/relayer/ClientHandlers.tla
istoilkovska: Indeed, installing only the maximum height violates liveness.

verification/spec/relayer/Relayer.tla
istoilkovska: This is fixed now, we are only sending in one direction, i.e., from `srcChainID` to `dstChainID`

verification/spec/relayer/ClientHandlers.tla
```diff
@@ -0,0 +1,106 @@
...
+   clientCreateChain
+ 
+\* Handle "ClientUpdate" datagrams
+HandleUpdateClient(chainID, chain, datagrams) ==      
+    \* get "ClientUpdate" datagrams with valid clientID
+    LET updateClientDgrs == {dgr \in datagrams : 
+                            /\ dgr.type = "ClientUpdate"
+                            /\ dgr.clientID = GetCounterpartyClientID(chainID)} IN
+    \* get heights in datagrams with correct counterparty clientID for chainID
+    LET updateClientHeights == {h \in Heights : \E dgr \in updateClientDgrs : dgr.height = h} IN
```
ancazamfir, line 81: Seems that here we don't know the order of the client updates and we update with all heights in the datagrams. If we want to catch the current problem with the chain (failing to update the client state with a lower height than the max it has) then we need to consider sequence instead of sets (?)

ancazamfir: As discussed, we will do these changes as part of multiple relayer version.

PR #58, opened 2020-04-21
Connection Handshake (ICS3) L2 TLA+ spec

adizere:
WIP for : #4 

Rendered version here: [ch.tla](https://github.com/informalsystems/ibc-rs/blob/adi/connection_handshake_tla/verification/spec/connection-handshake/ch.tla).
______

For contributor use:

- [ ] Unit tests written


verification/spec/connection-handshake/ch.tla
```diff
@@ -0,0 +1,188 @@
+--------------------------------- MODULE CH ---------------------------------
+
+
+VARIABLES
+    parties,            \* The state of each party.
```
milosevic, line 5: Maybe better to call it chModule (connectionHandshakeModule)?

verification/spec/connection-handshake/ch.tla
```diff
@@ -0,0 +1,188 @@
...
+    inboundDatagrams    \* For each party, a buffer with any incoming datagrams.
```
milosevic, line 6: Maybe better to call it inMsgs, as what they exchange is not really datagram (with respect to ICS03).

verification/spec/connection-handshake/ch.tla
```diff
@@ -0,0 +1,188 @@
...
+    - clientID -- a client identifier
+      Stores the client identifier associated with this connection end. 
+      
+    - counterpartyClientID -- a client identifier
+      Stores the counterparty client identifier associated with this connection end.
+ ***************************************************************************)
+ConnectionEnd ==
+    [
+        state : ConnectionStates,
+        connectionID : ConnectionIDs \union {Null},
```
milosevic, line 46: In current specs we were normally defining special null value for each "type". In this case you might want to have NilConnectionId and NilClientId. Although I am not sure that it is needed as once you create connectionEnd object you need to set all values to not-nil one. On the other side you will probably need to have NilConnectionEnd object as initially that would be the value of the module.

verification/spec/connection-handshake/ch.tla
```diff
@@ -0,0 +1,188 @@
...
+        clientID : ClientIDs \union {Null},
+        counterpartyConnectionID : ConnectionIDs \union {Null},
```
milosevic, line 48: We should probably introduce abstraction for clientId, and connectionId (like what is called Counterparty in Go implementation). Just I would suggest not to use that name of course :) In my view this abstraction should be called ConnectionEnd and what is currently called ConnectionEnd should be Connection. Connection is then defined by two connectionEnds (local and remote, or endAtA and endAtB, or something third.

verification/spec/connection-handshake/ch.tla
```diff
@@ -0,0 +1,188 @@
...
+    - counterpartyConnectionID -- a connection identifier
+      Stores the connection identifier of the counterparty connection end.
+    
+    - clientID -- a client identifier
+      Stores the client identifier associated with this connection end. 
+      
+    - counterpartyClientID -- a client identifier
+      Stores the counterparty client identifier associated with this connection end.
+ ***************************************************************************)
+ConnectionEnd ==
```
milosevic, line 43: As this defines a set of records it is normally named plural (ConnectionEnds), and then instance of this record will be connectionEnd.

verification/spec/connection-handshake/ch.tla
```diff
@@ -0,0 +1,188 @@
...
+        counterpartyConnectionID : ConnectionIDs \union {Null},
+        counterpartyClientID : ClientIDs \union {Null}
+    ]      
+
+(********************************** Helper functions
+ ***************************************************************************)
+
+\* Constructs a ConnectionEnd object for the given party in state UNINIT.
+getConnectioEnd(targetParty) ==
+    IF targetParty = "A"
```
milosevic, line 57: I think that we don't need this logic. The values should come from the outside, i.e., InitMessage will define what are values of identifiers.

verification/spec/connection-handshake/ch.tla
```diff
@@ -0,0 +1,188 @@
...
+                connectionID |-> "connBtoA",
+                clientID |-> "clientB",
+                counterpartyConnectionID |-> "connAtoB",
+                counterpartyClientID |-> "clientA" ]
+         ELSE Null    
+
+
+\* Constructs a 'Init' Datagram.
+\* Fields are hardcoded for the moment
+getInitDatagram(targetParty) ==
```
milosevic, line 76: InitDatatgram (better to call it InitMessage in my view) should be defines as one of InMsgs.

verification/spec/connection-handshake/ch.tla
milosevic: We also probably want to have in this spec only single module with an environment and then we will compose multi-module system in a separate spec by creating instances of this spec. So it should be a single module. In that case we don't need a variable for a party (module), but instead directly module state. You can start with connectionEnd field and a connectionProof. And we will probably need also lightClient with a proof.

verification/spec/connection-handshake/ch.tla
```diff
@@ -0,0 +1,188 @@
...
+         THEN "connBtoA"
+         ELSE Null  
+
+
+(********************************** Environment functions
+ ***************************************************************************)
+
+    
+\* The environment generates the ConnOpenInit datagram, sending it to 'party'
+GenerateConnOpenInit(targetParty) ==
```
milosevic, line 100: I think that environment should be super simple in this spec; it does not need any variable or helper functions. It just set inMsg variable to any value from the set of InMsgs that we need to define (InitConn, TryConnOpen, etc).  For example `inMsg \in InMsgs`.

verification/spec/connection-handshake/ch.tla
```diff
@@ -0,0 +1,188 @@
...
+
+EnvStep ==
+    \E targetParty \in Parties :
+        GenerateConnOpenInit(targetParty)
+
+
+(**************************** Party functions
+ ***************************************************************************)
+
+ConnectionEndInit ==
```
milosevic, line 117: We probably don't this.

verification/spec/connection-handshake/ch.tla
```diff
@@ -0,0 +1,188 @@
...
+        \/ CHModuleHandleInit(party)
+    
+
+(**************************** Main spec
+ ***************************************************************************)
+
+
+Init ==
+    \* Associate to each party an initial state.
+    /\ parties = [ p \in Parties |-> PartyInit ]
```
milosevic, line 165: See comments above.

verification/spec/connection-handshake/ch.tla
```diff
@@ -0,0 +1,188 @@
...
+     counterpartyClientID |-> Null ] 
+
+
+(**
+ * Connection Handshake Module handlers
+ **)
+
+CHModuleHandleInit(party) ==
+    /\ isPartyInState(party, "UNINIT")
+    /\  LET connOpenInitDgrs == {dgr \in inboundDatagrams[party] :
```
milosevic, line 131: You can simplify this. Handler should just consume an inMsg (see for example how we do it in fastSync spec), and you directly access it inside the handler.

verification/spec/connection-handshake/ch.tla
```diff
@@ -0,0 +1,188 @@
...
+                                /\ dgr.connectionID = getConnectionID(party)}
+        IN IF connOpenInitDgrs /= {} 
+           THEN LET connOpenInitDgr == CHOOSE dgr \in connOpenInitDgrs : TRUE 
+                IN LET connOpenInitConnectionEnd == [
+                     state |-> "INIT",
+                     connectionID |-> connOpenInitDgr.connectionID,
+                     clientID |-> connOpenInitDgr.clientID,
+                     counterpartyConnectionID |-> connOpenInitDgr.counterpartyConnectionID,
+                     counterpartyClientID |-> connOpenInitDgr.counterpartyClientID ] 
+                  IN parties' = [ parties EXCEPT
```
milosevic, line 142: It is good practice to return a value from a handler, for example updated state and messages that should be sent. And then you have multiple handlers for multiple message types, and at one place (for example ModuleStep) you just call a given handler and set variables (with ') with a return values.

verification/spec/connection-handshake/ch.tla
adizere: Yes, I noticed that there is a specific Null for each type. I'm sure there's a good reason to do so, but I didn't understand why (yet...) and so I didn't want to just copy/paste without making sure it is necessary & well-understood.

verification/spec/connection-handshake/chm.tla
```diff
@@ -0,0 +1,241 @@
...
+
+    - connection -- a connection with two ends
+      Stores the connection.
+
+    - chainHeight -- a height
+      Stores the height of the chain running alongside the CHM.
+ ***************************************************************************)
+ConnectionHandshakeModules ==
+    [
+        connectionState : ConnectionStates,
```
ancazamfir, line 84: why isn't `connectionState` part of `connection`?

verification/spec/connection-handshake/chm.tla
```diff
@@ -0,0 +1,241 @@
...
+ ***************************************************************************)
+ConnectionHandshakeMessages ==
+    [type : {"ConnOpenInit"}, 
+     connection : Connections]
+\*    \union
+\*    [type : {"ConnOpenTry"},
+\*        connection : Connections,
+\*        stateProof : Proofs, 
+\*        consensusHeight : Proofs]
+
```
ancazamfir, line 103: This should be uncommented and corrected as it is generated below.

verification/spec/connection-handshake/chm.tla
```diff
@@ -0,0 +1,241 @@
...
+                   connection |-> initMsg.connection]]
+                   \* TODO: Proofs go here.                   
+   \* Otherwise, do not update the chain
+   ELSE [initCHM |-> chm,
+         msg |-> noMsg ]
+
+handleMsgConnectionOpenInit ==
+ /\ inMsg.type = "ConnOpenInit"
+ /\ LET res == connectionOpenInit(chModule, inMsg) IN
+      /\ chModule' = res.initCHM
```
ancazamfir, line 126: we probably need this to be a sequence (one element per height) and append to it.

verification/spec/connection-handshake/chm.tla
```diff
@@ -0,0 +1,241 @@
...
+
+\* Advance the height of the chain if MaxHeight is not yet reached.
+advanceChainHeight ==
+    /\ chModule.chainHeight < MaxHeight
+    /\ chModule' = [chModule EXCEPT
+                    !.chainHeight = chModule.chainHeight + 1
+                 ]
+    /\ UNCHANGED <<outMsg, inMsg>>
+
+
```
ancazamfir, line 144: I think we should clearly separate the environment and the ch spec, including the variables and functionality. Right now they are mixed up. If a handler accepts a msg and changes state it should cause `advanceChainHeight`. We should keep track of connection info per height.  We will also need to have the environment send the client state updates and the handler to simulate proof verification. For this we need to keep track of client information per height.

verification/spec/connection-handshake/chm.tla
```diff
@@ -0,0 +1,241 @@
...
+  **)
+ConnectionInitValue ==
+    [localEnd |-> ConnectionEndInitValue,
+     remoteEnd |-> ConnectionEndInitValue]
+
+CHModuleInitValue ==
+    [chainHeight |-> 1,
+     connectionState |-> "UNINT",
+     connection |-> ConnectionInitValue ]
+
```
ancazamfir, line 164: will need to be a sequence ?

verification/spec/connection-handshake/chm.tla
adizere: Initially, it was part of `connection`.

But then I noticed a `connection` will be part of the function handler signatures, as follows:

```
ConnectionHandshakeMessages ==
    [type : {"ConnOpenInit"}, 
     connection : Connections]
    \union
    [type : {"ConnOpenTry"},
        connection : Connections,
        stateProof : Proofs, 
        consensusHeight : Proofs]]
```

Based on this, I decided it would be more elegant to pull out the `connectionState` field from `Connection`, because this field is irrelevant (and represents purely local information). But this is a cosmetic choice. If this is confusing, I'll put it back. Especially if we want to model multiple connection handshakes in this spec, then it will be important to put the field back into `Connection`.

verification/spec/connection-handshake/chm.tla
```diff
@@ -0,0 +1,241 @@
...
+      Stores the current state of this connection. It has one of the
+      following values: "UNINIT", "INIT", "TRYOPEN", "OPEN".
+
+    - connection -- a connection with two ends
+      Stores the connection.
+
+    - chainHeight -- a height
+      Stores the height of the chain running alongside the CHM.
+ ***************************************************************************)
+ConnectionHandshakeModules ==
```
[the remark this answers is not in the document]
adizere: Indeed, I am also confused a bit here. This is a provisional name. We should also break out the environment from the `chm.tla` file. File and module names would become clearer then.

Ideally, the CHM record should encapsulate all the information that a connection handshake module would need to carry out handshakes. We can also put all of this information as module-level fields, like the [earlier suggestion from Zarko](https://github.com/informalsystems/ibc-rs/pull/58#discussion_r412760571)?

verification/spec/connection-handshake/chm.tla
adizere: I guess this is related to your comment below "We should keep track of connection info per height." This also important also to be able to create/verify proofs, no?

ancazamfir: correct.

verification/spec/connection-handshake/chm.tla
adizere: Looking at fastsync.tla (tendermint-rs/blob/master/docs/spec/fastsync/fastsync.tla) the approach there is to have the network/environment/node and peers all bundled in a single .tla module.

In Ilina's relayer work (#55), in contrast, the environment is nicely separated: The relayer instantiates the environment and defines the the top-level spec.

For the sake of modularity, I agree we should have this separation between environment and CH spec. However, unlike the relayer spec, I think in our case the order will be reversed: the Environment would instantiate the CHM and would define the top-level spec.

verification/spec/connection-handshake/chm.tla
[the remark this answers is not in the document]
ancazamfir: Not entirely sure that this would contain all the info the module needs but rather the one that it produces, i.e. the connection state. For example, the light client state is maintained by the client module and this could be part of the environment. 
Regarding the name, the spec is already called `chm`. Everything in there doesn't need again `chm`.  In the spec we know we are in the connection handshake module.
An outside module would do something like:
```
chmA == INSTANCE chm 
       WITH ConnectionIDs <- {"connAtoB"}, 
            ...
            MaxHeight <- 4
                              
queryConnection(id) == getConnection(chmA!connectionStates, id)
```

verification/spec/connection-handshake/chm.tla
ancazamfir: maybe we should define a connection parameter.

verification/spec/connection-handshake/chm.tla
ancazamfir: yeah, same for fastsync.tla. But better to separate. And at least in the scheduler.tla the vars for scheduler and env are clearly separated. Sorry i can't remember how it's done in the fastsync.tla spec.

verification/spec/connection-handshake/chm.tla
adizere: Hmm, I think I like that.

We would end up with the following abstractions:

```tla
ConnectionEnds ==
    [
        connectionID : ConnectionIDs \union {NullConnectionId},
        clientID : ClientIDs \union {NullClientId}
        \* commitmentPrefix add later
    ]

ConnectionParameters ==
    [
        localEnd : ConnectionEnds,
        remoteEnd : ConnectionEnds
    ]

Connections ==
    [
        parameters : ConnectionParameters,
        state : ConnectionState,
        \* any other fields that are specific for a connection
    ]
```

The field `ConnectionParameters` uniquely identifies a connection. And `ConnectionParameters` go as parameters into all `ConnectionHandshakeMessages`.

verification/spec/connection-handshake/chm.tla
```diff
@@ -0,0 +1,237 @@
...
+    - clientID -- a client identifier
+      Stores the client identifier associated with this connection end.
+
+    Optional, add later:
+    - prefix
+    
+ ***************************************************************************)
+ConnectionEnds ==
+    [
+        connectionID : ConnectionIDs \union {Null},
```
milosevic, line 46: Is there a need having Null here as connectionEnd is never initialised with null fields?

verification/spec/connection-handshake/chm.tla
```diff
@@ -0,0 +1,237 @@
...
+    /\ chModule.chainHeight < MaxHeight
+    /\ chModule' = [chModule EXCEPT
+                        !.chainHeight = chModule.chainHeight + 1]
+    /\ UNCHANGED <<outMsg, inMsg>>
+
+
+(***
+  * Initial value for a connection end.
+  **)
+ConnectionEndInitValue ==
```
milosevic, line 145: This is probably not needed.

verification/spec/connection-handshake/chm.tla
```diff
@@ -0,0 +1,237 @@
...
+    [connectionID |-> Null,
+     clientID |-> Null]
+
+
+(***
+  * Initial values for a connection.
+  **)
+ConnectionInitValue ==
```
milosevic, line 153: The same here.

verification/spec/connection-handshake/chm.tla
```diff
@@ -0,0 +1,237 @@
...
+    [localEnd |-> ConnectionEndInitValue,
+     remoteEnd |-> ConnectionEndInitValue]
+
+CHModuleInitValue ==
+    [chainHeight |-> 1,
+     connectionState |-> "UNINT",
+     connection |-> ConnectionInitValue ]
```
milosevic, line 160: We should just define NilConnection as operator and that should be initial value of the connection field.

verification/spec/connection-handshake/chm.tla
```diff
@@ -0,0 +1,237 @@
...
+
+ InitEnv ==
+   /\ inMsg = noMsg
+
+ OnConnInitMsg ==
+   /\ inMsg' \in [type: {"ConnOpenInit"},
+                  connection : Connections]
+
+ NextEnv ==
+     \/ OnConnInitMsg
```
milosevic, line 182: We can probably simplify this with inMsg' \in ConnectionHandshakeMessages.

verification/spec/connection-handshake/chm.tla
```diff
@@ -0,0 +1,237 @@
...
+
+
+(******************************************************************************
+ Main spec.
+ The system comprises the connection handshake module & environment.
+ *****************************************************************************)
+
+Init ==
+    turn = "env" \* Initially, the environment takes a turn.
```
milosevic, line 191:
```suggestion
    /\ turn = "env" \* Initially, the environment takes a turn.
```

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,231 @@
...
+
+(***************************************************************************
+ Helper operators.
+ ***************************************************************************)
+
+(* Returns true if 'para' matches the parameters in the local connection,
+    and returns false otherwise.
+ *)
+CheckLocalParameters(para) ==
+     /\ store.connection.parameters.localEnd.connectionID = para.localEnd.connectionID
```
milosevic, line 62: This is hard to read. Maybe better define localEnd (store.connection.parameteres.localEnd with LET) and remoteEnd.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,231 @@
...
+    Expects as input a ConnectionParameter 'para' and returns true or false.
+    
+    This is a basic validation step, making sure that 'para' is valid with
+    respect to module-level constants ConnectionIDs and ClientIDs.
+    If there is a connection in the store, this also validates 'para'
+    against the parameters of an existing connection by relying on the
+    state predicate CheckLocalParameters.
+*)
+ValidConnectionParameters(para) ==
+    /\ para.localEnd.connectionID \in ConnectionIDs
```
milosevic, line 78: These kind of checks are normal part of TypeOK if I am not wrong.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,231 @@
...
+Init(chainID) ==
+    /\ store = [id |-> chainID,
+                height |-> 1,
+                connection |-> nullConnection,
+                client |-> nullClient]
+
+
+Next ==
+    LET m == Head(inBuf)
+    IN /\ inBuf /= <<>>        \* Enabled if we have an inbound msg.
```
milosevic, line 221: What is the reasoning behind modelling inBuf as a sequence?

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,233 @@
...
+              /\ storeChainB.connection.state = "INIT"
+
+
+\* Safety property.
+\* If the connections in the two chains are not null, then the
+\* connection parameters must always match.
+ConsistencyInv ==
+    \/ storeChainA.connection = chmA!nullConnection
+    \/ storeChainB.connection = chmB!nullConnection
+    \/ storeChainA.connection.parameters
```
milosevic, line 224: I guess we only want connection ends to match if connection is in OPEN state at both sides. In that case connection is open and established and it should satisfy this property. In theory we don't care if at some step they are not consistent: imagine a case where connections ends are simultaneously initialled with different ids. This is not problematic if later they converge to a consists ids.

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,233 @@
+---------------------------- MODULE Environment ----------------------------
```
milosevic, line 1: Not ideal name. Maybe it can be connection handshake and the other connection end, or something along those lines.

verification/spec/connection-handshake/Environment.tla
ancazamfir: There is currently no negotiation allowed in the IBC spec on IDs for connections and channels. If you make a mistake and mismatch the IDs, e.g. you send `openInit` on A and B without matching, they will be successful but will stay stuck in `INIT` state and another `openInit` will not fix this. The `connectionEnd` state stays in the store forever, they cannot be changed or removed.

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,233 @@
...
+
+VARIABLES
+    bufChainA,      \* A buffer (sequence) for messages inbound to chain A.
+    bufChainB,      \* A buffer for messages inbound to chain B.
+    storeChainA,    \* The local store of chain A.
+    storeChainB,    \* The local store of chain B.
+    maliciousEnv    \* If TRUE, environment interferes w/ CH protocol. 
+
+
+chainAParameters == [
```
milosevic, line 21: I thought this should come from ConnInit message?

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,233 @@
...
+    clientID |-> { "clientChainA" }
+]
+
+chainBParameters == [
+    connectionID |-> { "connBtoA" },
+    clientID |-> { "clientChainB" }
+]
+
+
+ClientIDs == { chainAParameters.clientID, chainBParameters.clientID }
```
milosevic, line 32: Maybe better if ClientIDs is a constant, i.e., it is defined as part of model.

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,233 @@
...
+ConnectionIDs == { chainAParameters.connectionID, chainBParameters.connectionID }
+
+chainAVars == <<bufChainA, bufChainB, storeChainA>>
```
milosevic, line 35: It's confusing seeing variables as part of both A and B vars.

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,233 @@
...
+
+Connections ==
+    [
+        parameters : ConnectionParameters,
+        state : ConnectionStates
+    ]
+
+Proofs ==
+    [
+        height : 1..MaxHeight
```
milosevic, line 76: We should explain what is the semantic we are modelling. As this is simplified view, we need to explain what property of proof (checking) we want to cover.

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,233 @@
...
+ make an explicit distinction to the low-level protocol. Message types
+ are as follows: CHMsgInit, CHMsgTry, CHMsgAck, and CHMsgConfirm. Notice that
+ the fields of each message are also different to the ICS 003 specification.
+ ***************************************************************************)
+ConnectionHandshakeMessages ==
+    [type : {"CHMsgInit"}, 
+     parameters : ConnectionParameters]
+    \union
+    [type : {"CHMsgTry"},
+     parameters : ConnectionParameters,
```
milosevic, line 95: I think that it should be Connection here as that is what is written to the store. So you need both parameters and the state.

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,233 @@
...
+     stateProof : Proofs,
```
milosevic, line 96: This proof is about connection. In my view it is better naming it like that, i.e., connectionProof for example.

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,233 @@
...
+     clientProof : Proofs]
```
milosevic, line 97: We are missing here clientState if I am not wrong. Otherwise what data clientProof corresponds to?

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,233 @@
...
+    \union
+    [type : {"CHMsgAck"},
+     parameters : ConnectionParameters,
```
milosevic, line 100: The same comment here.

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,233 @@
...
+     stateProof : Proofs,
+     clientProof : Proofs]
+     \union
+    [type : {"CHMsgConfirm"},
+     parameters : ConnectionParameters,
```
milosevic, line 105: The same comment here.

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,233 @@
...
+     stateProof : Proofs]
+
+
+(* The set of all Init messages, such that the local end is the
+    set 'le', and the remote end is set 're'. *)
+InitMsgs(le, re) ==
```
milosevic, line 111: Why you need this?You have already defined it at line 91.

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,233 @@
...
+
+(* TODO: find a more elegant way to do this. 
+    What we want, ideally, is to specify this:
+        z \in x
+    to obtain a random Sequence z out of a set of sequences x, obtaining:
+        bufChainA \in InitMsgs
+    instead of the awkward lines for building bufChainA and bufChainB below.
+ *)
+ChooseInitMsg(le, re) ==
+    CHOOSE x \in InitMsgs(le, re) : TRUE
```
milosevic, line 125: You normally don't want to do this as choose will deterministically pick one. This constrain model checker in a bad way. You probably want here to pick message from the set of all InitMsgs, for example \E m \in InitMsgs.

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,233 @@
+---------------------------- MODULE Environment ----------------------------
+
+EXTENDS Naturals, FiniteSets, Sequences
+
+
+CONSTANT MaxHeight,     \* Maximum height of any chain in the system.
+         MaxBufLen      \* Length (size) of message buffers.
```
ancazamfir, line 7: Can you add a `verification/spec/connection-handshake/README.md` and describe a bit the content and also recommended values for the constants above? I put in some numbers and my TLA+ toolbox is pretty much dead...will try to restart my laptop now. I wish some good engineers would redesign this and make it a modern and reliable tool :(

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,233 @@
...
+                arbitraryMsg \in ConnectionHandshakeMessages} 
+       /\ UNCHANGED bufChainB
+    \/ /\ Len(bufChainA) < MaxBufLen - 1
+       /\ bufChainB' \in
+            {Append(bufChainB, arbitraryMsg) :
+                arbitraryMsg \in ConnectionHandshakeMessages}
+       /\ UNCHANGED bufChainA
+
+
+NextEnv ==
```
milosevic, line 165: This is a bit confusing. What do you want to model with this?

verification/spec/connection-handshake/Environment.tla
adizere: I am not able to run it locally either, even with values `2` and `2` had issues with model checking.
I'm trying to refurbish the code to eliminate some of the non-determinism.

verification/spec/connection-handshake/Environment.tla
adizere: >  In theory we don't care if at some step they are not consistent: imagine a case where connections ends are simultaneously initialled with different ids. This is not problematic if later they converge to a consists ids.

I had an earlier version that was allowing "temporary inconsistency". But I ran into the problem that the connection ends would never reconcile in fact. To allow them to reconcile, we would need to modify our spec to capture multiple connections at the same time.

verification/spec/connection-handshake/Environment.tla
ancazamfir: yes, I tried the same and something is wrong. Never saw this before, the toolbox to just hang and crash. not sure at what point this started, it was fast last time I tried but it was days ago and there was only the `openInit` there.

verification/spec/connection-handshake/Environment.tla
adizere: Igor proposed `Medium`. I'll think more about this.

verification/spec/connection-handshake/Environment.tla
adizere: Quick fix when the toolbox hangs repeatedly for me was to delete the model and create another one!

verification/spec/connection-handshake/Environment.tla
milosevic: We are talking here about property it should ensure, so I am not saying protocol would be able to make it for a mistake (inconsistency). We care if this holds only if it's init at both side as at that point it becomes active (?), similar like we care about consensus only after we decide.

verification/spec/connection-handshake/Environment.tla
milosevic: We should probably switch to apalache from the command line.

ancazamfir: > Quick fix when the toolbox hangs repeatedly for me was to delete the model and create another one!

didn't work for me, tried many times

ancazamfir: > We should probably switch to apalache from the command line.

maybe. but i think it's a problem with the model. For this spec and small constants I expect local TLA toolbox to be fine. I can run fastsync with long chains and it's fine even if it takes ~10mins.

adizere: I think it's because I'm using the `ENABLED` operator. I'm trying to refactor that part now...

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
adizere: This is a module-specific check (not a global type invariant, which is captured in TypeOK). In this validation check, we make sure that the connection parameters are valid with respect to the connections and clients that this module expects.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
adizere: We wanted to permit multiple outstanding messages to each chain.

If we permit a single message (i.e., we model communication with a buffer/shared location), then we need to make sure that the environment (see `MaliciousNextEnv` action) never overwrites this location as long as the initialization message is there (see `InitEnv`), which is quite complicated. If the environment can overwrite the initialization message, then we get a deadlock, because the chains will never move out of the "UNINIT" state.

verification/spec/connection-handshake/Environment.tla
adizere: Pushed a new variant today that no longer has the hanging issue, after I removed the `ENABLED` operator.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
milosevic: In that case, \in conditions need to be more precise as valid connection and client ids are chain specific.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,294 @@
...
+
+(* Handles a "CHMsgTry" message.
+ *)
+HandleTryMsg(m) ==
+    \/ /\ PreconditionsTryMsg(m) = TRUE
+       /\ LET newCon == [parameters |-> m.parameters, 
+                         state |-> "INIT"]
+              sProof == GetStateProof
+              cProof == GetClientProof
+              replyMsg == [parameters |-> FlipConnectionParameters(m.parameters),
```
ancazamfir, line 196: Looks like the proof related spec is not finished? Here we need to check the proofs from message `m`. also include local proofs in replyMsg.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,294 @@
...
+(***************************************************************************
+ Connection Handshake Module main spec.
+ ***************************************************************************)
+
+
+Init(chainID) ==
+    /\ store = [id |-> chainID,
+                height |-> 1,
+                connection |-> nullConnection,
+                client |-> nullClient]
```
ancazamfir, line 278: i don't see where this is set to non-`nullClient`

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,354 @@
...
+
+ASSUME Cardinality(ConnectionIDs) >= 1
+ASSUME Cardinality(ClientIDs) >= 1
+
+
+VARIABLES
+(******************************* Store *****************************
+     The store record of a chain contains the following fields:
+     
+     - id -- a string
```
milosevic, line 45:
```suggestion
     - chainId -- a string 
```

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,354 @@
...
+
+
+VARIABLES
+(******************************* Store *****************************
+     The store record of a chain contains the following fields:
+     
+     - id -- a string 
+       Stores the identifier of the chain where this module executes.
+
+     - latestHeight -- a Nat
```
milosevic, line 48:
```suggestion
     - latestHeight -- 
```

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,353 @@
...
+    respect to module-level constants ConnectionIDs and ClientIDs.
+    If there is a connection in the store, this also validates 'para'
+    against the parameters of an existing connection by relying on the
+    state predicate CheckLocalParameters.
+*)
+ValidConnectionParameters(para) ==
+    /\ para.localEnd.connectionID \in ConnectionIDs
+    /\ para.localEnd.clientID \in ClientIDs
+    /\ \/ store.connection = NullConnection
+       \/ /\ store.connection /= NullConnection
```
milosevic, line 102: It is a bit mixing of concerns; maybe this check can be taken out and have this function only validates if ids are valid.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,353 @@
...
+
+    The connection proof is used to demonstrate to another chain that the
+    local store on this chain comprises a connection in a certain state. 
+ *)
+GetConnProof(connState) ==
+    [height |-> store.latestHeight,
+     connectionState |-> connState]
+
+
+(* Operator for construcing a client proof.
```
milosevic, line 127:
```suggestion
(* Operator for constructing a client proof.
```

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,353 @@
...
+ *)
+GetClientProof ==
+    [height |-> store.client.latestHeight]
```
milosevic, line 130: I think that you also need as with ConnProof to have additional field `height |-> store.latestHeight`. This effectively represents the proof and data in this case is clientState which we represent just with the `store.client.latestHeight`.

verification/spec/connection-handshake/ICS3Types.tla
```diff
@@ -0,0 +1,296 @@
...
+          AllClientIDs
+
+
+(******************************* InitClients *****************************
+     A set of records describing the possible initial values for the
+     clients on a chain.
+     
+     A client record contains the following fields:
+
+     - consensusStates -- a set of heights, each height being a Nat
```
milosevic, line 27: I find it a bit confusing to use term Nat here as we are constraining set of heights with MaxHeight. I would suggest just not to refer to "type" of the field, just say what it represents.

verification/spec/connection-handshake/ICS3Types.tla
```diff
@@ -0,0 +1,296 @@
...
+       Stores the set of all heights (i.e., consensus states) that this
+       client observed. At initialization time, the client only observes
+       the first height, so the only possible value for this record is
+       {1}.
+
+     - clientID -- a string
+       The identifier of the client. This is expected as a parameter, since
+       it is a chain-specific field at initialization time.
+       
+     - latestHeight -- a natural number
```
milosevic, line 37: As we can compute max out of consensusStates maybe it's better idea having just an operator that would compute it. If I understood well @istoilkovska, it might be more model checker friendly if we have an operator instead of additional field.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,353 @@
...
+     - latestHeight -- a Nat
+       Describes the current height of the chain.  
+
+     - connection -- a connection record.
+       Captures all the details of the connection on this chain.
+       For a full description of a connection record, see the
+       'Environment.Connections' set.
+
+     - client -- a client record.
+       Specifies the state of the client running on this chain.
```
milosevic, line 57: I think it would still be useful having "client record definition" here; even if it means duplicating text. The same for store. It simplifies reading functions after as you know what to expect there as fields.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,353 @@
...
+
+    This is a state predicate returning true if the following holds: the local
+    client on this chain stores the height reported in the proof. Note that this
+    condition should eventually become true, assuming a correct environment, which
+    should periodically update the client on each chain; see actions 'GoodNextEnv'
+    and 'UpdateClient'.
+
+ *)
+VerifyConnProof(cp) ==
+    /\ cp.height \in store.client.consensusStates
```
milosevic, line 143: Verifying proof consists of two parts; checking that height is installed on the destination, and second, that the proof corresponds to the data it accompanies. We are missing the second check here.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,353 @@
...
+
+
+(* Verification of a client proof.
+
+    This is a state predicate returning true if the following holds: the height
+    reported in the client proof must not exceed the current (latestHeight) of
+    this chain.
+ *)
+VerifyClientProof(cp) ==
+    cp.height <= store.latestHeight
```
milosevic, line 153: Here we miss the check that the corresponding height is installed at the destination.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,353 @@
...
+(* Handles a "ICS3MsgInit" message 'm'.
+    
+    Primes the store.connection to become initialized with the parameters
+    specified in 'm'. Also creates a reply message, enqueued on the outgoing
+    buffer.
+ *)
+HandleInitMsg(m) ==
+    LET newCon == [parameters |-> m.parameters,
+                   state      |-> "INIT"]
+        sProof == GetConnProof("INIT")
```
milosevic, line 180: Proof is for the whole connection object, not just for the state. I would suggest to pass newCon to the GetConnProof and to rename ConnProof field connState in conn.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,353 @@
...
+        cProof == GetClientProof
+        replyMsg == [parameters |-> FlipConnectionParameters(m.parameters),
```
milosevic, line 182: We need to agree what we want to model with messages. Either we will use exactly the same format as with ICS, or we will use more high level abstraction that would contain data and the corresponding proof. It feels we are with this version somewhere between the two and then it's a bit confusing. I tend to prefer more the former (although it's not very intuitive) as it would also make integration with the relayer spec easier.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,353 @@
...
+                     type |-> "ICS3MsgTry",
+                     connProof |-> sProof,
+                     clientProof |-> cProof] IN
+    IF /\ ValidConnectionParameters(m.parameters)
```
milosevic, line 186: This would led to calling `CheckLocalParameters` and I am not sure you want that check here. I would prefer inlining checks here and not delegating it to a function as checks are simpler and it's easier reading it if it's inlined. Only check for valid identifiers make sense to be a function.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,353 @@
...
+
+
+(* State predicate, guarding the handler for the Try msg.
+
+    If any of these preconditions does not hold, the message
+    is dropped. 
+ *)
+PreconditionsTryMsg(m) ==
+    /\ ValidConnectionParameters(m.parameters)
+    /\ \/ store.connection.state = "UNINIT"
```
milosevic, line 201: This logic is present both in `ValidConnectionParameters` and here: you probably want to remove it from `ValidConnectionParameters`.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,353 @@
...
+          /\ CheckLocalParameters(m.parameters)
+    /\ m.connProof.connectionState = "INIT"
+    /\ VerifyConnProof(m.connProof)
+    /\ VerifyClientProof(m.clientProof)
+
+
+(* Handles a "ICS3MsgTry" message.
+ *)
+HandleTryMsg(m) ==
+    LET newCon == [parameters |-> m.parameters,
```
milosevic, line 212: I think we can use EXCEPT semantic here to only modify state, as parameters should stay the same.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,353 @@
...
+
+
+(* State predicate, guarding the handler for the Ack msg. 
+ *)
+PreconditionsAckMsg(m) ==
+    /\ \/ store.connection.state = "INIT"
+       \/ store.connection.state = "TRYOPEN"
+    /\ CheckLocalParameters(m.parameters)
+    /\ m.connProof.connectionState = "TRYOPEN"
+    /\ VerifyConnProof(m.connProof)
```
milosevic, line 234: I think we also need `VerifyClientProof(m.clientProof)` here?

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,353 @@
...
+
+
+(* Handles a "ICS3MsgAck" message.
+ *)
+HandleAckMsg(m) ==
+    LET newCon == [parameters |-> m.parameters, 
+                   state |-> "OPEN"]
+        sProof == GetConnProof("OPEN")
+        cProof == GetClientProof
```
milosevic, line 243: This is probably not needed.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,353 @@
...
+
+(***************************************************************************
+ Connection Handshake Module (ICS3) main spec.
+ ***************************************************************************)
+
+
+Init(chainID, client, connection) ==
+    /\ store = [id |-> chainID,
+                latestHeight |-> 1,
+                connection |-> connection,
```
milosevic, line 328: Should it be initialised with nilConnection?

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,353 @@
...
+       /\ inBuf' = Tail(inBuf)    \* Strip the head of our inbound msg. buffer.
+    \/ /\ inBuf = <<>>
+       /\ ~ CanAdvance
+       /\ UNCHANGED<<moduleVars>>
+
+
+TypeInvariant ==
+    /\ inBuf \in Seq(ConnectionHandshakeMessages) \union {<<>>}
+    /\ outBuf \in Seq(ConnectionHandshakeMessages) \union {<<>>}
+    /\ store.connection \in Connections
```
milosevic, line 345: You can define for a store what type it should have: see for example how it was done in fastsync for complex record https://github.com/informalsystems/tendermint-rs/blob/02efca7eb50fc2ca6233038a7bd97b21ad6ab577/docs/spec/fastsync/fastsync.tla#L634.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
ancazamfir: It is also important to note that both proofs must have the same height. I would prefer to have it as a separate field in the message, that is you have 
```
{proof_height: <height_A>, 
 proofConn: <connection>, 
 proofConsensus: {
    consensus_height: <height_B>, 
    proofConsensus: <consensus>}
...
}
```

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
ancazamfir: This is going to be weird in the current model as it would mean looking up the remote chain.

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,402 @@
...
+    both chains finish locking on the same set of connection parameters.
+ *)
+InitEnv ==
+    /\ maliciousEnv = FALSE
+    /\ \/ /\ inBufChainA \in {<<msg>> : (* ICS3MsgInit to chain A. *)
+            msg \in InitMsgs(ChainAConnectionEnds, ChainBConnectionEnds)}
+          /\ inBufChainB = <<>>
+       \/ /\ inBufChainB \in {<<msg>> : (* ICS3MsgInit to chain B. *)
+            msg \in InitMsgs(ChainBConnectionEnds, ChainAConnectionEnds)}
+          /\ inBufChainB = <<>>
```
ancazamfir, line 170: ??
```suggestion
          /\ inBufChainA = <<>>
```

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
ancazamfir: I am not sure if it's a good idea to separate preconditions in a separate predicate.

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,402 @@
...
+
+
+(***************************************************************************
+ Environment actions.
+ ***************************************************************************)
+
+
+(* Environment initialization.
+    
+    This action kick-stars the ICS3 protocol by assigning an ICS3MsgInit
```
milosevic, line 116:
```suggestion
    This action kick-starts the ICS3 protocol by assigning an ICS3MsgInit
```

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,402 @@
...
+    msg to either of the two chains (or both).
+    
+    Initially, the environment is non-malicious. The environment starts
+    acting maliciously once the connection on both chains transitions out
+    of state "UNINIT" (the initials state). It is important to
+    initialize the protocol like this, otherwise the env. can provoke a
+    deadlock. This can happen with the following sequence of actions:
```
milosevic, line 123: I think that we don't want to constrain scenario that leads to deadlock (at least at the model level). We want to be able to point out to the problematic scenario. Good scenarios should then captured by (most probably) temporal properties.

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,402 @@
...
+    
+    3. Chain B processes the ICS3MsgInit (action HandleInitMsg) and
+    updates its store.connection with the parameters from step 2 above.
+    Chain B "locks onto" these parameters and will not accept any others.
+    At this step, chain B produces a ICS3MsgTry message with the local
+    parameters from its connection.
+    
+    Both chains will be locked on a different set of connection parameters,
+    and neither chain will accept their corresponding ICS3MsgTry, hence a
+    deadlock. To avoid this problem, we prevent the environment from
```
milosevic, line 159: This is really cool scenario; we should be able ideally to capture it with the model checker. Solution to this problem is not constraining environment as we can't do it in practice. We need to design protocol differently otherwise single faulty relayer can prevent us from establishing connections between chains.

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,402 @@
...
+    Without the first constraint, on the "Len(inBufChainA)" and
+    "Len(inBufChainB)", Env could fill buffers (DoS attack). This can
+    lead to a deadlock, because chains will simply be unable to reply
+    to each other.
+ *)
+MaliciousNextEnv ==
+    \/ /\ Len(inBufChainA) < MaxBufLen - 1
+       /\ inBufChainA' \in {Append(inBufChainA, arbitraryMsg) : 
+            arbitraryMsg \in {msg \in ConnectionHandshakeMessages : 
+            msg.type = "ICS3MsgTry"}}
```
milosevic, line 237: Why msg type is constrained here? We want to allow all messages to be created by adversary.

verification/spec/connection-handshake/Environment.tla
```diff
@@ -0,0 +1,402 @@
...
+    /\ \E clientA \in InitClients({ x.clientID : x \in ChainAConnectionEnds }) :
+            chmA!Init("chainA", clientA, NullConnection)
+    /\ \E clientB \in InitClients({ x.clientID : x \in ChainBConnectionEnds }) :
+            chmB!Init("chainB", clientB, NullConnection)
+    /\ InitEnv
+
+
+(* The two ICS3 modules and the environment alternate their steps. *)
+Next ==
+    \/ ICS3ReachTermination
```
milosevic, line 320: I think we normally want this logic to be part of precondition for something to happen: for example if it's not true then we make a step; otherwise we don't do anything. I think that it is semantically slightly different to what we have here.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,353 @@
...
+
+     - connection -- a connection record.
+       Captures all the details of the connection on this chain.
+       For a full description of a connection record, see the
+       'Environment.Connections' set.
+
+     - client -- a client record.
+       Specifies the state of the client running on this chain.
+       For a full description of a client and the initialization values
+       for this record, see the InitClients set below.
```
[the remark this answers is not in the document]
adizere: Good catch! I moved the `InitClient` operator to a separate file. Agree with the suggestion to duplicate the explanation of the client record.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
adizere: Fixed after syncing with @ancazamfir.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
adizere: Per @ancazamfir suggestion, the proof now includes the set of all heights, and added a verification against that set.

verification/spec/connection-handshake/ConnectionHandshakeModule.tla
adizere: Thanks for the pointer, added that.

verification/spec/connection-handshake/Environment.tla
adizere: I added this artificial constraint on "ICS3MsgTry" to make the checking more tractable. But even with this constraint, after 1 hour of execution (~50 milion distinct states), TLC didn't finish on my laptop.

I am removing this constraint.

verification/spec/connection-handshake/Environment.tla
adizere: Opened a separate issue #70 to track discussion on this.

Issue #71, opened 2020-05-21
Recreate the UpdateClient race condition in ICS3 TLA+ spec


adizere:
## Summary

A race condition is possible that can lead to deadlocking in the ICS3 protocol, caused by the way `UpdateClient` function is used.
## Problem Definition

Related issue #61 for a more thorough description.

## Proposal

We'll add a TLA+ spec that is a slight variation of the one in #58, exhibiting the deadlock issue.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #73, opened 2020-05-21
Added spec & execution trace for liveness problem scenario ref: #71

adizere:
Closes: #71 

## Description

Added specification plus execution trace with the UpdateClient race condition, resulting in a deadlock.
______

For contributor use:

- [ ] Unit tests written

### DIFF
--- a/verification/spec/connection-handshake/updateclient-deadlock/ConnectionHandshakeModule.tla
+++ b/verification/spec/connection-handshake/updateclient-deadlock/ConnectionHandshakeModule.tla
@@ -1,0 +1,383 @@
+--------------------- MODULE ConnectionHandshakeModule ---------------------
+
+(***************************************************************************
+
+    This module is part of the TLA+ specification for the
+    IBC Connection Handshake protocol (ICS3).
+    
+    This module captures the actions and operators of the ICS3 protocol.
+    Typically, it is an IBC module running on a chain that would implement
+    the logic in this TLA+ module, hence the name "ConnectionHandshakeModule"
+    sometimes abbreviated to "chModule" or "chm".
+
+    This module deals with a high-level spec of the ICS3 protocol, so it is
+    a simplification with respect to ICS3 proper in several regards:
+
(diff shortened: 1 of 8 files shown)


verification/spec/connection-handshake/updateclient-deadlock/ConnectionHandshakeModule.tla
```diff
@@ -0,0 +1,383 @@
...
+    buffer.
+ *)
+HandleInitMsg(m) ==
+    LET newCon == [parameters |-> m.parameters,
+                   state      |-> "INIT"]
+        myConnProof == GetConnProof(newCon)
+        myClientProof == GetClientProof
+        replyMsg == [parameters |-> FlipConnectionParameters(m.parameters),
+                     type |-> "ICS3MsgTry",
+                     proofHeight |-> store.latestHeight,
```
ancazamfir, line 210: Making the comment here but also valid for all other handles. I believe there is a problem here as the proof height is current height but then the store is updated below in `NewStore(newCon)` with +1 height. In real scenario, the store is updated to a height H (where the IBC msg event occurred) and then the relayer takes the proofs at height H. So we want to simulate this here.
If I make this change (and in all other proofs): 
```suggestion
                     proofHeight |-> store.latestHeight + 1,
```
the model stutters. Needs change to maxHeight 6 and then we run into a deadlock scenario somewhat similar to one described here. The trace shows even more clearly that the environment simulates in fact multiple relayers.

adizere: This is a great find, I was not aware that creating the proof at `store.latestHeight` is in a sense unrealistic.

adizere:
The work in this PR is irrelevant given the requirement of maintaining a _single_ ICS3 TLA+ spec. Reference: #108 (This PR proposes to have a separate spec that concerns itself specifically with the deadlock/liveness problem.) Consequently, I'll close this PR without merging, and open a separate PR to tackle both #108 and #71.

Issue #66, opened 2020-05-19
Implement Connection queries


ancazamfir:
## Summary
Implement Connection queries

## Problem Definition
In order to create connections handshake transactions the relayer needs to be able to query connections. Also needed for the user.

## Proposal
Add connection CLI, code the query request, response and connection structures, and the ConnOpenInit message.

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #77, opened 2020-05-23
Connection queries and ConnOpenInit message

Shivani912:
Closes: #66

## Description
Added connection CLI, connection query request, response and connection structures, and the ConnOpenInit message

______

For contributor use:

- [x] Unit tests written


Shivani912:
Closing for now, will do another PR

modules/src/ics03_connection/exported.rs
```diff
@@ -0,0 +1,113 @@
+use super::error;
+use crate::ics23_commitment::CommitmentPrefix;
+use anomaly::fail;
+use serde_derive::{Deserialize, Serialize};
+
+pub trait ConnectionI {
```
romac, line 6: `I` suffix for traits is not a common Rust practice in my experience. How about naming the trait `Connection` instead?

Ambiguities can always be resolver by renaming imports or using the qualified path to the trait.

modules/src/ics03_connection/exported.rs
```diff
@@ -0,0 +1,113 @@
...
+    type ValidationError: std::error::Error;
+
+    fn state(&self) -> State;
+    fn client_id(&self) -> String;
+    fn counterparty(&self) -> Box<dyn CounterpartyI<ValidationError = super::error::Error>>;
+    fn versions(&self) -> Vec<String>;
+    fn validate_basic(&self) -> Result<(), Self::ValidationError>;
+}
+
+pub trait CounterpartyI {
```
romac, line 16: `I` suffix for traits is not a common Rust practice in my experience. How about naming the trait `Counterparty` instead?

Ambiguities can always be resolver by renaming imports or using the qualified path to the trait.

modules/src/ics03_connection/connection.rs
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+}
+
+impl ConnectionI for ConnectionEnd {
+    type ValidationError = error::Error;
+
+    fn state(&self) -> State {
+        self.state.clone()
+    }
```
romac, line 32: It's usually a good practice to not clone data when it can be avoided, and let the call site perform the clone if it needs ownership of the returned data. Not a big deal in this case, as we can incur the penalty of doing a few clones here and there, so feel free to ignore.

```suggestion
    fn state(&self) -> &State {
        &self.state
    }
```

modules/src/ics03_connection/connection.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    fn client_id(&self) -> String {
+        String::from(self.client_id.as_str())
```
romac, line 35:
```suggestion
        self.client_id.as_str().into()
```

modules/src/ics03_connection/connection.rs
```diff
@@ -0,0 +1,107 @@
...
+            prefix,
+        })
+    }
+}
+
+impl CounterpartyI for Counterparty {
+    type ValidationError = error::Error;
+
+    fn client_id(&self) -> String {
+        String::from(self.client_id.as_str())
```
romac, line 93:
```suggestion
        self.client_id.as_str().into()
```

modules/src/ics03_connection/connection.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    fn connection_id(&self) -> String {
+        String::from(self.connection_id.as_str())
```
romac, line 97:
```suggestion
        self.connection_id.as_str().into()
```

modules/src/ics03_connection/connection.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    fn prefix(&self) -> CommitmentPrefix {
+        self.prefix.clone()
+    }
```
romac, line 102: Same as above, feel free to ignore.

```suggestion
    fn prefix(&self) -> &CommitmentPrefix {
        &self.prefix
    }
```

modules/src/ics03_connection/exported.rs
```diff
@@ -0,0 +1,113 @@
+use super::error;
+use crate::ics23_commitment::CommitmentPrefix;
+use anomaly::fail;
+use serde_derive::{Deserialize, Serialize};
+
+pub trait ConnectionI {
+    type ValidationError: std::error::Error;
+
+    fn state(&self) -> State;
```
romac, line 9: See above.

```suggestion
    fn state(&self) -> &State;
```

modules/src/ics03_connection/exported.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn versions(&self) -> Vec<String>;
+    fn validate_basic(&self) -> Result<(), Self::ValidationError>;
+}
+
+pub trait CounterpartyI {
+    type ValidationError: std::error::Error;
+
+    fn client_id(&self) -> String;
+    fn connection_id(&self) -> String;
+    fn prefix(&self) -> CommitmentPrefix;
```
romac, line 21: See above.

```suggestion
    fn prefix(&self) -> &CommitmentPrefix;
```

modules/src/ics03_connection/exported.rs
ancazamfir: I just made an attempt to fix this and the other comments you had. Also aligned connection and channel a bit more.

Shivani912:
Thank you for your valuable comments @romac and @ancazamfir 😊

Issue #74, opened 2020-05-22
IBC queries should not require items to query to be in the configuration file


ancazamfir:
## Summary of Bug
IBC queries should not require items to query to be in the configuration file. That is only used for filtering the relaying paths.

## Version
master branch
## For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

PR #78, opened 2020-05-25
Check query parameters but not against the config

ancazamfir:
Queries should be allowed for clients and channels that are not configured.
Also, fix parameter verification order to be consistent and simplify the code.
Fix regressions.

Closes: #74 

For contributor use:

- [X] Unit tests written

### DIFF
--- a/modules/src/ics02_client/query.rs
+++ b/modules/src/ics02_client/query.rs
@@ -88,19 +88,12 @@ where
         query: QueryClientFullState<CLS>,
         response: AbciQuery,
     ) -> Result<Self, error::Error> {
-        match (response.value, &response.proof) {
-            (Some(value), _) => {
-                let client_state = amino_unmarshal_binary_length_prefixed(&value)?;
-
-                Ok(ClientFullStateResponse::new(
-                    query.client_id,
-                    client_state,
-                    response.proof,
-                    response.height.into(),
-                ))
-            }
-            (None, _) => Err(error::Kind::Rpc.context("Bad response").into()),
-        }
+        Ok(ClientFullStateResponse::new(
+            query.client_id,
+            amino_unmarshal_binary_length_prefixed(&response.value)?,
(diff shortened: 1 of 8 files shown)


relayer/cli/src/commands/query/channel.rs
```diff
@@ -41,32 +42,42 @@ impl QueryChannelEndsCmd {
...
-                    _ => Err(format!("cannot find chain {} in config", chain_id)),
-                }
-            }
-            (None, _, _) => Err("missing chain identifier".to_string()),
-            (_, None, _) => Err("missing port identifier".to_string()),
-            (_, _, None) => Err("missing channel identifier".to_string()),
-        }
+        let chain_id = self
+            .chain_id
+            .ok_or_else(|| "missing chain configuration".to_string())?;
```
Shivani912, line 47: Can we change the error message so it's easier to identify the failing check?
```suggestion
            .ok_or_else(|| "missing chain id".to_string())?;
```

modules/src/ics02_client/query.rs
```diff
@@ -88,19 +88,12 @@ where
...
-                    client_state,
-                    response.proof,
-                    response.height.into(),
-                ))
-            }
-            (None, _) => Err(error::Kind::Rpc.context("Bad response").into()),
-        }
+        Ok(ClientFullStateResponse::new(
+            query.client_id,
+            amino_unmarshal_binary_length_prefixed(&response.value)?,
```
Shivani912, line 93: Ran this locally and it errored here on type check. This function expects a `&[u8]` but `response.value` is an `Option<vec<u8>>` How come this didn't fail in CI?  🤔

ancazamfir: still seeing this? `cargo update` maybe? loos like `response.value` still points to the old `Option<vec<u8>>` version of tm.

Issue #82, opened 2020-06-02
Create a 'disclosure' log


adizere:
## Summary

Create a document to record all bugs/issues we uncover while specifying & formally verifying the IBC protocols.

## Problem Definition

This feature addresses the problem of keeping a historical log of contentious matters in the IBC specs. We can share this log with external collaborators and disclose any issue we find in a transparent manner.

It is unclear if a markdown file is the right format to record these matters. Main disadvantage is that this document might be difficult to maintain over time.

## Proposal

- A basic markdown file `docs/disclosure-log.md` will help us keep track and disclose all the problems we find.
- Each entry in this file will consist of a brief description + a commit hash. The hash will usually point to the full details of that problem.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

cwgoes:
Thanks, this would be helpful. I'm open to any sort of process changes in the specification repository or procedures which you think would make this easier.

ebuchman:
Why not just open issues in cosmos/ics?

adizere:
We are opening issues in the other repo in addition to this disclosure log. The point of the log was to have a single place with the history of all issues we discussed so we can easily report/refer to them.

Issue #61, opened 2020-05-06
Clarify relayer algorithm with respect to consensus state changes via clientUpdates


ancazamfir:
## Summary
In the presence of multiple realyers or relayer threads, it can happen that a consensus update will fail if its height is smaller than the client state height.
This issue is here just to facilitate discussion and its resolution will be incorporated in the Relayer ADR.

## Problem Definition
The relayer scans IBC store for a source chain and creates datagrams for a destination chain. Many of these datagrams include proofs that must verify against a specified client consensus state.
Due to the fact that the `appHash` for block `n` is stored in `n+1,` a typical relayer action for X will include:
```
 queryX(X, h-1) -> {x, proof, h-1}
 getHeader(h) -> header
 sendMsgs(MsgClientUpdate{cl, h}, MsgX{X, proof, h})
```
One or both of the above messages may fail. This issue is concerned with failure of `MsgClientUpdate`, when it can happen and how it can be solved.

## Proposal
It is possible that there are two relayers scanning for different items, X vs Y, and they require consensus state (CS) updates on the same destination chains but at different heights. This could result on one of the relayer transactions to fail, as a `clientUpdate` for height `h` fails if the latest client height is `H >=h`.

Here is a proposal of overall handling for message X relaying in general that covers this scenario. 
We assume relaying for A->B path, the client on B has been created and maybe updated a number of times. The relayer in this example periodically scans the source chain A, but it would work the same for event driven relayer. Here are the steps:
- relayer gets the latest height `hx` on A and queries the client height on B, `ha`.
- takes the maximum of the two, `h = max(hx, ha)`
- queries for X at `h-1` and gets the proof `p`
- with `ha` being the latest trusted header, it gets header `h` and any intermediate headers
- at this point the relayer builds a number of `MsgUpdateClient(hi)` datagrams and the `MsgX(X, p, h)` datagram, and sends them in a transaction to B.
  - if `hx < ha` there will be no `MsgUpdateClient` datagrams
- if the transaction result indicates client update failure then(*):
  - it queries the (new) latest state of client on A, `hA` and queries X at `ha-1`
  - sends `MsgX(X, p, hA)` to A

If a failure is seen for the `MsgX..` transaction at either step, it is probably safe for the relayer to ignore it, as it could be caused by another "competing" relayer having submitting the same transaction for X. This needs more investigation but not part of this issue which is only for CS update failures.
(*) 
[...]

PR #83, opened 2020-06-02
Disclosure log to record ICS3 liveness problems

adizere:
Closes: #82 

Rendered [here](https://github.com/informalsystems/ibc-rs/blob/3294a438fb5aae1a6600551a1022df051c8f8fda/docs/disclosure-log.md)

## Description

Added a markdown file for recording and disclosing all the problems we find in IBC protocols.

______

For contributor use:

- [ ] Unit tests written

### DIFF
--- a/docs/disclosure-log.md
+++ b/docs/disclosure-log.md
@@ -40,15 +40,14 @@ This is difficult to guarantee in practice, however.
 Intuitively, the difficulty arises because of a combination of two factors:
 
 1. __Proof requirement:__ For chain `B` to accept item `X`, it must verify the authenticity of this item; this is done via the light client that chain `B` maintains.
-This light client requires, given item `X` at height `h` on `A`, a proof constructed at height `h-1`.
-Put simply, the light client needs: (1) the item `X` at height `h` (as part of what is called "consensus state"), plus (2) a proof constructed at height `h-1` for this item.
+Given an item `X` and a commitment proof for `X` constructed at height `h-1`, the light client requires the consensus state at height `h` that includes that commitment root required for verification.
 
-2. __Concurrency:__ Different relayers may update the same light client. Suppose a relayer `r1` wants to submit `X` at consensus state `h` plus the associated proof at `h-1`.
+2. __Concurrency:__ Different relayers may update the same light client. Suppose a relayer `r1` wants to submit a consensus state at height `h` and a message that includes `X` plus the associated proof at `h-1`.
 In the meantime, however, another relayer `r2` may update this same light client to height `h'`.
 It is important to note that `h'` is bigger than `h`.
 Now `r1` will be unable to update the light client with consensus state `h`, because the light client disallows updates with smaller heights than the current height `h'`; consequently, the relayer will be unable to submit `X`.
 
-To ensure eventual delivery, relayer `r1` would need to retry submitting item `X`, that is: resubmit the item at a larger height (e.g., at `h'`) with the associated proof (at `h'-1`).
+To ensure eventual delivery, relayer `r1` would need to retry submitting item `X`, that is: resubmit the consensus state at a larger height (e.g., at `h'`) followed by the message that includes the proof for `X` (at `h'-1`).
 This retry mechanism was adoped as a solution for the [current relayer implementation](https://github.com/informalsystems/ibc-rs/blob/master/docs/architecture/adr-002-ibc-relayer.md#ibc-client-consensus-state-vs-relayer-light-client-states-vs-chain-states).
 Note that it is also possible for relayer `r2` to have submitted the same item `X` successfully; in this case, the liveness problem does not actually surface.
 
--- a/docs/disclosure-log.md
+++ b/docs/disclosure-log.md
@@ -44,7 +44,7 @@ Given an item `X` and a commitment proof for `X` constructed at height `h-1`, th
 
 2. __Concurrency:__ Different relayers may update the same light client. Suppose a relayer `r1` wants to submit a consensus state at height `h` and a message that includes `X` plus the associated proof at `h-1`.
 In the meantime, however, another relayer `r2` may update this same light client to height `h'`.
-It is important to note that `h'` is bigger than `h`.
+Assume `h'` is bigger than `h`.
 Now `r1` will be unable to update the light client with consensus state `h`, because the light client disallows updates with smaller heights than the current height `h'`; consequently, the relayer will be unable to submit `X`.
 
 To ensure eventual delivery, relayer `r1` would need to retry submitting item `X`, that is: resubmit the consensus state at a larger height (e.g., at `h'`) followed by the message that includes the proof for `X` (at `h'-1`).
--- a/docs/disclosure-log.md
+++ b/docs/disclosure-log.md
@@ -55,38 +55,41 @@ Note that it is also possible for relayer `r2` to have submitted the same item `
 
 ##### TLA+ trace
 
-To obtain an execution in TLA+ that depicts the above liveness problem, it is sufficient to enable the `Concurrency` flag in the default TLA+ spec for ICS3.
+To obtain an execution in TLA+ that depicts the above liveness problem, it is sufficient to enable the `Concurrency` flag in the L2 default TLA+ spec for ICS3.
+This spec is located in [spec/connection-handshake/L2-tla/](./spec/connection-handshake/L2-tla/).
 In this spec we make a few simplifications compared to the real system, most importantly: to verify an item at height `h`, a light client can use the consensus state at the same height `h` (no need for smaller height `h-1`).
 Below we summarize the parameters as well as the sequence of actions that lead to the liveness problem.
 
 ###### Parameters:
 
 - `MaxBufLen <- 2`
-- `MaxHeight <- 6`
-- `Concurrency <- True`
+- `MaxHeight <- 8`
+- `Concurrency <- TRUE`
 - Behavior spec: Temporal formula `Spec`
 - Check for `Deadlock`, Invariants `TypeInvariant` and `ConsistencyProperty`, as well as Property `Termination`
 
 ###### Trace:
 
-Both chains `A` and `B` start at height `1`, and the light client on each chain has consensus state `1`.
+Both chains `A` and `B` start at height `1`, and the light client on each chain has consensus state for height `1`.
 
 1. The environment submits a `ICS3MsgInit` message to chain `A`.
 
-2. The environment triggers the `AdvanceChainHeight` action of chain `A`, so this chain transitions from height `1` to height `2`.
+2. Chain `A` processes the `ICS3MsgInit`, advances to height `2`, and prepares a `ICS3MsgTry` message destined for chain `B`.
+The proof in this message is for height `2`.
 
-3. Chain `A` processes the `ICS3MsgInit`, advances to height `3`, and prepares a `ICS3MsgTry` message destined for chain `B`.
-The proof in this message is for height `3`.
+3. The environment triggers the `AdvanceChainHeight` action of chain `B`, so this chain transitions from height `1` to height `2`.
 
-4. The environment triggers the `AdvanceChainHeight` action of chain `A`, so this chain transitions from height `3` to height `4`.
+4. The environment triggers the `AdvanceChainHeight` action of chain `A`, so this chain transitions from height `2` to height `3`.
 
-5. __Concurrency:__ The environment triggers the `UpdateClient` action on chain `B`: the light client on this chain is updated with height `4` (that is, the latest height of chain `A`), and chain `B` also transitions from height `1` to height `2`.
+5. The environment triggers the `AdvanceChainHeight` action of chain `A`, so this chain transitions from height `3` to height `4`.
 
-6. The environment passes (i.e., relays) the `ICS3MsgTry` message to chain `B`.
-Recall that this message has proofs for height `3`; consenquently, the environment also attempts to trigger `UpdateClient` action on chain `B` for consensus state `3`.
-This action does not enable because the light client on `B` has a more recent consensus state `4`.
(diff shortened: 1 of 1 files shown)


docs/disclosure-log.md
```diff
@@ -0,0 +1,21 @@
...
+```
+
+The correct code should include both the cases when a single chain is in state `INIT` and the case when both chains are in state `INIT`, as specified here: [Relayer.tla](https://github.com/informalsystems/ibc-rs/blob/master/docs/spec/relayer/Relayer.tla#L174)
+
+
+##### 2. ICS3 deadlock due to `UpdateClient` semantics
+
+Related issues: [#71](https://github.com/informalsystems/ibc-rs/issues/71) and [#61](https://github.com/informalsystems/ibc-rs/issues/61).
+
+Code: [#73](https://github.com/informalsystems/ibc-rs/pull/73).
```
ancazamfir, line 21: Not sure what ICS3 deadlock means here, also above. In both cases it seems that we deal with a relayer that has some implementation bugs. Is this (#2) also about a relayer using ICS-18 algorithm? 

Here I believe we need a bit more detail on the consequences of a relayer not handling `UpdateClient` failure properly and describe when and how the deadlock can occur (e.g. presence of another relayer that somehow manages to do updates at precise times). Then we need to propose a change to ICS-18 like you do in #1.

Also let's try to close https://github.com/informalsystems/ibc-rs/pull/73 first before we merge this. Have you managed to show the deadlock here?

cwgoes: To me https://github.com/informalsystems/ibc-rs/issues/61 describes a race condition, which is quite a different thing than a "deadlock" ("deadlock" brings to mind badly written concurrent code using mutual exclusion locks that can reach a state from which it is impossible to progress, which is not what we're talking about here)

docs/disclosure-log.md
```diff
@@ -0,0 +1,21 @@
+# Disclosure Log for IBC Protocols
+
+This document is a record of all the bugs or issues we uncovered while specifying & formally verifying the IBC protocols.
+
+
+##### 1. ICS3 deadlock due to relayer algorithm
```
cwgoes, line 6: As I understand, this is a mistake in the example relayer algorithm described in ICS 18, not a mistake in ICS 3 itself - am I understanding correctly? Or is there code in ICS 3 which needs to be fixed as well?

ancazamfir: Just needs fix in ICS-18

docs/disclosure-log.md
ancazamfir: Yes, I agree "deadlock" is not the best word, we will rephrase. Also in 1. In both cases the handshake cannot progress. While in 1 it is clear why, here the scenarios for "cannot progress" have to be detailed.

adizere: Hmm, by "deadlock" I meant, loosely, "loss of liveness." I agree this can be confusing.

To prevent confusion, I removed "deadlock" and replaced with "liveness problem."

docs/disclosure-log.md
adizere: Added a remark clarifying that only ICS018 requires modifications.

For completeness, I added a separate issue in cosmos/ics for tracking the required changes in ICS018: https://github.com/cosmos/ics/issues/436

docs/disclosure-log.md
adizere: > While in 1 it is clear why, here the scenarios for "cannot progress" have to be detailed.

The "cannot progress" details are given in the referenced code (#73).
More specifically, see the "problematic execution" here: https://github.com/informalsystems/ibc-rs/blob/e33f32db95314461d18fbdb13574ddeabafedad4/verification/spec/connection-handshake/updateclient-deadlock/README.md#specification-details

I'm not sure it's worth detailing here (in addition to the execution in #73). @ancazamfir but if you think it's relevant for completeness sake, then I can describe the issue.

docs/disclosure-log.md
```diff
@@ -0,0 +1,38 @@
...
+
+##### 2. ICS3 liveness problem due to `UpdateClient` semantics
+
+Related issues: [#71](https://github.com/informalsystems/ibc-rs/issues/71) and [#61](https://github.com/informalsystems/ibc-rs/issues/61).
+
+Code for obtaining an execution trace in TLA+: [#73](https://github.com/informalsystems/ibc-rs/pull/73).
+Briefly, this trace comprises the following sequence of actions, assuming two chains `A` and `B` and a relayer (as a simplification, the chains in this model are in charge of creating messages, a responsibility that the relayer typically implements):
+
+
+0. Chain `A` creates a message `M` that has a proof for height `h`, chain `A` also advances its height to `h+1`.
```
ancazamfir, line 30:
```suggestion
0. Chain `A` creates a message `M` that includes a proof at height `h`, chain `A` also advances its height to `h+1`.
```

docs/disclosure-log.md
```diff
@@ -0,0 +1,38 @@
...
+
+1. The relayer updates the client on chain `B` with the latest height of chain A, namely `h+1` (call to `UpdateClient`).
```
ancazamfir, line 32:
```suggestion
1. Another relayer updates the client on chain `B` with the latest height of chain A, namely `h+1` (call to `UpdateClient`).
```

docs/disclosure-log.md
```diff
@@ -0,0 +1,38 @@
...
+
+2. The relayer sends message `M` that chain `A` created (at step 0) to chain B, and also attempt to update the client on chain `B` with height `h` (the height in the proof of this message), but is unable to do so because `h` is smaller than `h +1` (the latest height of client on chain B).
```
ancazamfir, line 34:
```suggestion
2. The relayer sends two messages to chain `B`, possibly in a single transaction:
- an `updateClient` message for height `h` and,
- message `M` that was vreated created at step 0
But when processed on `B` the `updateClient(h)` fails as the client's state is at a higher height (`h+1`) than `h`.
```

docs/disclosure-log.md
```diff
@@ -0,0 +1,38 @@
...
+
+3. Chain `B` will be unable to verify message `M` since it lacks the adequate height information `h`, so chain `B` will drop this message, hence ICS3 can no longer progress.
```
ancazamfir, line 36:
```suggestion
3. Chain `B`is unable to verify message `M` since it lacks the consensus state at height `h`. Chain `B` drops this message, hence ICS3 can no longer progress.
```

adizere: > It is possible that chain B has a consensus state at height h from a previous update by the "other" relayer. So even if the updateClient(h) failed above, M could still be accepted.

That's true.. but this is not really the scenario in this execution. The trace here has a single relayer. I'm not sure if it helps to add more information about other potential problematic scenarios involving multiple relayers.

docs/disclosure-log.md
adizere: >  The relayer sends two messages to chain `B`, possibly in a single transaction:

This sentence is not really accurate w.r.t to the trace in #73. (There is no `updateClient` __message__, as updates do not rely on messages.)

docs/disclosure-log.md
adizere: As described below, this trace does not comprise multiple relayers. The loss of liveness happens due to solely one relayer and the way `updateClient` is used.

milosevic: It is more complicated explaining this scenario with only single relayer as it would not update client unless there is a corresponding message to send. The exception would be the case where the highest header is about to expire. But this is not the scenario explained here. Maybe the confusion come from the fact that in the used model we don't have explicit relayer, and relayer is being used to describe the problem.

docs/disclosure-log.md
```diff
@@ -0,0 +1,38 @@
...
+###### Channel handshake (ICS4) liveness problem
+
+The same issue (and fix) seems to exist for the channel handshake datagrams.
+
+##### 2. ICS3 liveness problem due to `UpdateClient` semantics
+
+Related issues: [#71](https://github.com/informalsystems/ibc-rs/issues/71) and [#61](https://github.com/informalsystems/ibc-rs/issues/61).
+
+Code for obtaining an execution trace in TLA+: [#73](https://github.com/informalsystems/ibc-rs/pull/73).
+Briefly, this trace comprises the following sequence of actions, assuming two chains `A` and `B` and a relayer (as a simplification, the chains in this model are in charge of creating messages, a responsibility that the relayer typically implements):
```
milosevic, line 27: This sentence is confusing as it says that modules sent directly messages and then there is a relayer also. In case we want to stay in the model of #73 we can say we have a model in which modules send messages to each other but network can also inject faulty messages (simulating concurrent correct and faulty relayers). Then scenario is explained where network inject UpdateClient message with h+1, which leads to failure of M to be delivered.

milosevic:
We should remove reference to a TLA+ spec (for a time being) that generates counter example and just leave description in English with two concurrent relayers stepping at each other foot, and close this PR.

adizere:
> We should remove reference to a TLA+ spec (for a time being) that generates counter example and just leave description in English with two concurrent relayers stepping at each other foot, and close this PR.

This PR is ready for review with a TLA+ spec at the moment. Have a look, and if this proves problematic, then I'll remove the spec and keep just a simple explanation (similar to the description in #61).

docs/disclosure-log.md
```diff
@@ -0,0 +1,95 @@
...
+Assuming some source chain `A`, destination chain `B`, and an IBC item `X` (e.g., connection, channel, or packet) on chain `A`, we can define this property as follows:
+
+> For any IBC item `X` on chain `A` destined for chain `B`, eventually, a correct relayer will submit item `X` to chain `B`.
+
+This is difficult to guarantee in practice, however.
+Intuitively, the difficulty arises because of a combination of two factors:
+
+1. __Proof requirement:__ For chain `B` to accept item `X`, it must verify the authenticity of this item; this is done via the light client that chain `B` maintains.
+This light client requires, given item `X` at height `h` on `A`, a proof constructed at height `h-1`.
+Put simply, the light client needs: (1) the item `X` at height `h` (as part of what is called "consensus state"), plus (2) a proof constructed at height `h-1` for this item.
```
ancazamfir, line 44:
```suggestion
Given an item `X` and a commitment proof for `X` constructed at height `h-1`, the light client requires the consensus state at height `h` that includes that commitment root required for verification.
```

docs/disclosure-log.md
```diff
@@ -0,0 +1,95 @@
...
+
+2. __Concurrency:__ Different relayers may update the same light client. Suppose a relayer `r1` wants to submit `X` at consensus state `h` plus the associated proof at `h-1`.
```
ancazamfir, line 46:
```suggestion
2. __Concurrency:__ Different relayers may update the same light client. Suppose a relayer `r1` wants to submit a consensus state at height `h` and a message that includes `X` plus the associated proof at `h-1`.
```

docs/disclosure-log.md
```diff
@@ -0,0 +1,95 @@
...
+In the meantime, however, another relayer `r2` may update this same light client to height `h'`.
+It is important to note that `h'` is bigger than `h`.
```
ancazamfir, line 48:
```suggestion
Assume `h'` is bigger than `h`.
```

docs/disclosure-log.md
```diff
@@ -0,0 +1,95 @@
...
+Now `r1` will be unable to update the light client with consensus state `h`, because the light client disallows updates with smaller heights than the current height `h'`; consequently, the relayer will be unable to submit `X`.
+
+To ensure eventual delivery, relayer `r1` would need to retry submitting item `X`, that is: resubmit the item at a larger height (e.g., at `h'`) with the associated proof (at `h'-1`).
```
ancazamfir, line 51:
```suggestion
To ensure eventual delivery, relayer `r1` would need to retry submitting item `X`, that is: resubmit the consensus state at a larger height (e.g., at `h'`) followed by the message that includes the proof for `X` (at `h'-1`).
```

ancazamfir:
Can we can formulate 2 such that we don't talk about `h` vs `h-1`? Just mention that in general a proof requires a consensus state at height `h` and an update of the on-chain client with that may fail.

adizere:
> Can we can formulate 2 such that we don't talk about h vs h-1? Just mention that in general a proof requires a consensus state at height h and an update of the on-chain client with that may fail.

By "2" you mean the concurrency part?
If I understood correctly, the [latest commit](https://github.com/informalsystems/ibc-rs/pull/83/commits/c3115aaacd6a0aec7213c744a0724994e767e577) should be simpler and hopefully achieve this.

docs/disclosure-log.md
```diff
@@ -0,0 +1,96 @@
...
+Intuitively, the difficulty arises because of a combination of two factors:
+
+1. __Proof requirement:__ For chain `B` to accept item `X`, it must verify the authenticity of this item; this is done via the light client that chain `B` maintains.
+Given an item `X` and a commitment proof for `X` constructed at height `h-1`, the light client requires the consensus state at height `h` that includes that commitment root required for verification.
+
+2. __Concurrency:__ Different relayers may update the same light client.
+Suppose a relayer `r1` wants to submit a consensus state at height `h`.
+In the meantime, however, another relayer `r2` may update this same light client to height `h'`.
+Assume `h'` is bigger than `h`.
+Now `r1` will be unable to update the light client with consensus state `h`, because the light client disallows updates with smaller heights than the current height `h'`.
```
ancazamfir, line 49: This is true for TM light client. Might be different for others.
```suggestion
If the light client disallows updates with heights smaller than the current height `h'` then `r1`'s update fails .
```

docs/disclosure-log.md
```diff
@@ -0,0 +1,96 @@
...
+Consequently, the relayer will be unable to submit consensus state at height `h`.
```
ancazamfir, line 50:
```suggestion
Consequently, the relayer will be unable to submit message for `X`.
```

adizere: `X` is not mentioned in (2/Concurrency), so it would be confusing to talk about it here.

Issue #87, opened 2020-06-09
Subscribe to TM events and push IBC events to the relayer


ancazamfir:
## Summary
Implement event monitor that subscribes to TM events, extracts IBC events and pushes them to the main relayer handler.

## Problem Definition
The main relay handler will be driven by external and internal events. Implement the event monitor that is instantiated per chain, subscribes to Tx and Block events and extracts IBC events
____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

PR #88, opened 2020-06-09
[WIP] Add main event handler and per chain monitor tasks

ancazamfir:
Closes: #87

## Description
Starting with https://github.com/informalsystems/ibc-rs/pull/59 as base, extract attributes specific to IBC events.
Add monitor tasks, one per chain, that register for Tx and Block events with tendermint nodes. 
Push the IBC events to the main handler task.

There are a few things to clarify:
- can the IBC events be determined from "message.action" list? or a try for all possible events is needed (see https://github.com/cosmos/cosmos-sdk/issues/6358)
- need access to `TMEventData::EventDataNewBlock()` that is an Option; need to extract block height. I had to change tendermint for this. 
______

For contributor use:

- [ ] Unit tests written

### DIFF
--- a/modules/Cargo.toml
+++ b/modules/Cargo.toml
@@ -17,13 +17,15 @@ paths-cosmos = []
 paths-ics = []
 
 [dependencies]
-tendermint = { git = "https://github.com/informalsystems/tendermint-rs.git", branch = "master" }
+tendermint = { path = "/Users/ancaz/rust/tendermint-rs/tendermint"}
 
 anomaly = "0.2.0"
 thiserror = "1.0.11"
 serde_derive = "1.0.104"
 serde = "1.0.104"
+serde_json = "1"
+tracing = "0.1.13"
 
 [dev-dependencies]
-serde_json = "1"
+tokio = { version = "0.2", features = ["macros"] }
 
--- a/modules/src/error.rs
+++ b/modules/src/error.rs
@@ -8,6 +8,10 @@ pub enum Kind {
     /// RPC error (typically raised by the RPC client or the RPC requester)
     #[error("RPC error")]
     Rpc,
+
+    /// Event error (raised by the event monitor)
+    #[error("Bad Notification")]
+    EventErr,
 }
 
 impl Kind {
--- a/modules/src/events.rs
+++ b/modules/src/events.rs
@@ -1,0 +1,248 @@
+use crate::ics02_client::events as ClientEvents;
+use crate::ics02_client::events::NewBlock;
+use crate::ics03_connection::events as ConnectionEvents;
+use crate::ics04_channel::events as ChannelEvents;
+use crate::ics20_fungible_token_transfer::events as TransferEvents;
+use serde_derive::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::convert::{TryFrom, TryInto};
+use tendermint::block;
+use tendermint::rpc::event_listener::{ResultEvent, TMEventData};
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub enum IBCEvent {
+    NewBlock(NewBlock),
+
+    CreateClient(ClientEvents::CreateClient),
+    UpdateClient(ClientEvents::UpdateClient),
+    ClientMisbehavior(ClientEvents::ClientMisbehavior),
+
+    OpenInitConnection(ConnectionEvents::OpenInit),
+    OpenTryConnection(ConnectionEvents::OpenTry),
+    OpenAckConnection(ConnectionEvents::OpenAck),
+    OpenConfirmConnection(ConnectionEvents::OpenConfirm),
+
+    OpenInitChannel(ChannelEvents::OpenInit),
+    OpenTryChannel(ChannelEvents::OpenTry),
+    OpenAckChannel(ChannelEvents::OpenAck),
+    OpenConfirmChannel(ChannelEvents::OpenConfirm),
+    CloseInitChannel(ChannelEvents::CloseInit),
+    CloseConfirmChannel(ChannelEvents::CloseConfirm),
+
+    SendPacketChannel(ChannelEvents::SendPacket),
+    ReceivePacketChannel(ChannelEvents::ReceivePacket),
+    AcknowledgePacketChannel(ChannelEvents::AcknowledgePacket),
+    CleanupPacketChannel(ChannelEvents::CleanupPacket),
+    TimeoutPacketChannel(ChannelEvents::TimeoutPacket),
+
+    TimeoutTransfer(TransferEvents::Timeout),
+    PacketTransfer(TransferEvents::Packet),
+    ChannelClosedTransfer(TransferEvents::ChannelClosed),
+}
+
+impl IBCEvent {
+    pub fn to_json(&self) -> String {
+        serde_json::to_string(self).unwrap()
+    }
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct TryObject {
+    pub height: block::Height,
+    pub action: String,
+    pub idx: usize,
+    pub events: HashMap<String, Vec<String>>,
+}
+
+impl TryObject {
+    pub fn new(
+        height: block::Height,
+        action: String,
+        idx: usize,
+        events: HashMap<String, Vec<String>>,
+    ) -> TryObject {
+        TryObject {
+            height,
+            action,
+            idx,
+            events,
+        }
+    }
+}
+
+pub fn extract_events<'a, S: ::std::hash::BuildHasher>(
+    events: &'a HashMap<String, Vec<String>, S>,
+    action_string: &'a str,
+) -> Result<&'a HashMap<String, Vec<String>, S>, Box<dyn std::error::Error>> {
+    if let Some(message_action) = events.get("message.action") {
+        if message_action.contains(&action_string.to_owned()) {
+            return Ok(events);
+        }
+        return Err("Missing action string".into());
+    }
+    Err("Incorrect Event Type".into())
+}
+
+//Takes events in the form
+//  "events":{
+//      "connection_open_init.client_id":["testclient","testclientsec"],
+//      "connection_open_init.connection_id":["ancaconnonetest","ancaconnonetestsec"],
+//      "connection_open_init.counterparty_client_id":["testclientsec","testclientsecsec"],
+//      "create_client.client_id":["testclientthird"],
+//      "create_client.client_type":["tendermint"],
+//      "message.action":["connection_open_init","create_client","connection_open_init"],
+//      "message.module":["ibc_connection","ibc_client",ibc_connection"],
+//      "message.sender":["cosmos187xxg4yfkypl05cqylucezpjvycj24nurvm8p9","cosmos187xxg4yfkypl05cqylucezpjvycj24nurvm8p9","cosmos187xxg4yfkypl05cqylucezpjvycj24nurvm8p9","cosmos187xxg4yfkypl05cqylucezpjvycj24nurvm8p9"],"tm.event":["Tx"],"transfer.amount":["5000stake"],"transfer.recipient":["cosmos17xpfvakm2amg962yls6f84z3kell8c5lserqta"],"tx.hash":["A9E18AE3909F22232F8DBDB1C48F2FECB260A308A2D157E8832E901D45950605"],
+//      "tx.height":["35"]
+// and returns:
+// [{"connection_open_init", 0}, {"create_client", 0}, {"connection_open_init", 1}]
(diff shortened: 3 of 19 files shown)


modules/Cargo.toml
```diff
@@ -17,13 +17,15 @@ paths-cosmos = []
 paths-ics = []
 
 [dependencies]
-tendermint = { git = "https://github.com/informalsystems/tendermint-rs.git", branch = "master" }
+tendermint = { path = "/Users/ancaz/rust/tendermint-rs/tendermint"}
```
romac, line 20: - you can use a relative path here instead of the absolute one
- iirc one can specify both `git` and `path`, and Cargo will prioritize the `path` dependency over `git` if it exists

modules/src/error.rs
```diff
@@ -8,6 +8,10 @@ pub enum Kind {
     /// RPC error (typically raised by the RPC client or the RPC requester)
     #[error("RPC error")]
     Rpc,
+
+    /// Event error (raised by the event monitor)
+    #[error("Bad Notification")]
+    EventErr,
```
romac, line 14: For consistency with other error cases, we can drop the `Err` suffix here.

modules/src/events.rs
```diff
@@ -0,0 +1,248 @@
...
+        TryObject {
+            height,
+            action,
+            idx,
+            events,
+        }
+    }
+}
+
+pub fn extract_events<'a, S: ::std::hash::BuildHasher>(
```
romac, line 73: Is this explicit lifetime actually needed here? I would expect rustc to be able to infer it.

modules/src/events.rs
```diff
@@ -0,0 +1,248 @@
...
+    events: &'a HashMap<String, Vec<String>, S>,
+    action_string: &'a str,
+) -> Result<&'a HashMap<String, Vec<String>, S>, Box<dyn std::error::Error>> {
```
romac, line 76: Do we really need to return the given HashMap here? Since this function seem to actually perform validation rather than extraction, it could be renamed and its signature changed to `Result<(), Box<dyn std::error::Error>>`.

modules/src/ics02_client/client_type.rs
```diff
@@ -1,8 +1,10 @@
 use super::error;
 use anomaly::fail;
+use serde_derive::{Deserialize, Serialize};
 
 /// Type of the consensus algorithm
-#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+
```
romac, line 7: Nitpick: attributes are typically on the very previous line to the item they annotate. Might be fixed by running rustfmt.

modules/src/ics02_client/events.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+impl NewBlock {
+    pub fn new(h: block::Height) -> NewBlock {
+        NewBlock { height: h }
+    }
+}
+
+impl TryFrom<&ResultEvent> for NewBlock {
+    type Error = Box<dyn std::error::Error>;
```
romac, line 24: Putting this here, but applies in general: since we are using `Box<dyn std::error::Error>` in a bunch of places we might want to introduce a type alias for it, eg. `BoxError`, or just use the one defined in [`anomaly::BoxError`](https://docs.rs/anomaly/0.2.0/anomaly/type.BoxError.html) (though the latter is a bit more restrictive as it enforces thread-safety by having `Send + Sync + 'static` bounds as well.

modules/src/events.rs
```diff
@@ -0,0 +1,248 @@
...
+}
+
+impl IBCEvent {
+    pub fn to_json(&self) -> String {
+        serde_json::to_string(self).unwrap()
+    }
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct TryObject {
```
romac, line 50: Perhaps `RawObject` is a better name here, for consistency with this pattern in `tendermint-rs`?

modules/Cargo.toml
ancazamfir: thanks, will keep that in mind. I switched back to git as my local changes were just committed.

modules/src/events.rs
ancazamfir: I added it because the compiler was complaining without it...but i can remove the lifetime once I change the return value to (). Here is the old error:
```
   Compiling relayer-modules v0.1.0 (/Users/ancaz/rust/2ibc-rs/modules)
error[E0106]: missing lifetime specifier
  --> modules/src/events.rs:76:13
   |
74 |     events: &HashMap<String, Vec<String>, S>,
   |             --------------------------------
75 |     action_string: &str,
   |                    ----
76 | ) -> Result<&HashMap<String, Vec<String>, S>, Box<dyn std::error::Error>> {
   |             ^ expected named lifetime parameter
   |
   = help: this function's return type contains a borrowed value, but the signature does not say whether it is borrowed from `events` or `action_string`
help: consider introducing a named lifetime parameter
```

modules/src/ics02_client/client_type.rs
ancazamfir: fixed.. rustfmt didn't catch this.

modules/src/events.rs
ancazamfir: changed

modules/src/events.rs
ancazamfir: Good catch! changed

modules/src/events.rs
```diff
@@ -0,0 +1,250 @@
...
+        RawObject {
+            height,
+            action,
+            idx,
+            events,
+        }
+    }
+}
+
+pub fn extract_events<S: ::std::hash::BuildHasher>(
```
milosevic, line 74: It does not look like extracting events?

modules/src/events.rs
```diff
@@ -0,0 +1,250 @@
...
+//      "connection_open_init.connection_id":["ancaconnonetest","ancaconnonetestsec"],
+//      "connection_open_init.counterparty_client_id":["testclientsec","testclientsecsec"],
+//      "create_client.client_id":["testclientthird"],
+//      "create_client.client_type":["tendermint"],
+//      "message.action":["connection_open_init","create_client","connection_open_init"],
+//      "message.module":["ibc_connection","ibc_client",ibc_connection"],
+//      "message.sender":["cosmos187xxg4yfkypl05cqylucezpjvycj24nurvm8p9","cosmos187xxg4yfkypl05cqylucezpjvycj24nurvm8p9","cosmos187xxg4yfkypl05cqylucezpjvycj24nurvm8p9","cosmos187xxg4yfkypl05cqylucezpjvycj24nurvm8p9"],"tm.event":["Tx"],"transfer.amount":["5000stake"],"transfer.recipient":["cosmos17xpfvakm2amg962yls6f84z3kell8c5lserqta"],"tx.hash":["A9E18AE3909F22232F8DBDB1C48F2FECB260A308A2D157E8832E901D45950605"],
+//      "tx.height":["35"]
+// and returns:
+// [{"connection_open_init", 0}, {"create_client", 0}, {"connection_open_init", 1}]
```
milosevic, line 99: What is the reasoning behind this format of return value?

modules/src/events.rs
```diff
@@ -0,0 +1,250 @@
...
+            }
+        }
+        _ => {}
+    }
+    Ok(vals)
+}
+
+pub fn build_event(object: RawObject) -> Result<IBCEvent, BoxError> {
+    match object.action.as_str() {
+        "create_client" => Ok(IBCEvent::from(ClientEvents::CreateClient::try_from(
```
milosevic, line 153: I am wondering if action string can somehow be defined as part of the corresponding IBCEvent so we don't hard code it here.

modules/src/events.rs
```diff
@@ -0,0 +1,250 @@
...
+        }
+
+        TMEventData::EventDataTx(_tx) => {
+            let events = &result.events.ok_or("missing events")?;
+            let height = events.get("tx.height").ok_or("tx.height")?[0]
+                .parse::<u64>()
+                .map_err(|e| e.to_string())?;
+            let actions_and_indices = extract_helper(&events)?;
+            for action in actions_and_indices {
+                let ev = build_event(RawObject::new(
```
milosevic, line 136: Why we need RawObject?We can't directly build IBCEvent?

modules/src/events.rs
```diff
@@ -0,0 +1,250 @@
...
+        // TODO not clear what the message.action for this is
+        "cleanup_packet" => Ok(IBCEvent::from(ChannelEvents::CleanupPacket::try_from(
+            object,
+        )?)),
+
+        _ => Err("Incorrect Event Type".into()),
+    }
+}
+
+pub fn extract_block_height(result: &ResultEvent) -> Result<block::Height, BoxError> {
```
milosevic, line 214: Can parameter be `TMEventData::EventDataNewBlock` instead of `ResultEvent`?

modules/src/ics02_client/events.rs
```diff
@@ -0,0 +1,110 @@
...
+
+impl From<CreateClient> for IBCEvent {
+    fn from(v: CreateClient) -> Self {
+        IBCEvent::CreateClient(v)
+    }
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub struct UpdateClient {
+    pub height: block::Height,
```
milosevic, line 66: Does it make sense introducing "asbtract" event type with height, client_id and client_type as these three events types have the same fields?

relayer/relay/src/event_monitor.rs
```diff
@@ -0,0 +1,94 @@
...
+
+impl EventMonitor {
+    /// Create an event monitor, connect to a node and subscribe to queries.
+    pub async fn create(
+        chain_id: ChainId,
+        rpc_addr: net::Address,
+        channel_to_handler: Sender<(ChainId, Vec<IBCEvent>)>,
+    ) -> Result<Self, Box<dyn std::error::Error>> {
+        let mut event_listener = event_listener::EventListener::connect(rpc_addr.clone()).await?;
+        // TODO move them to config file(?)
```
milosevic, line 31: Can we create issue for this?

relayer/relay/src/event_monitor.rs
```diff
@@ -0,0 +1,94 @@
...
+                }
+            }
+        }
+    }
+
+    /// get a TM event and extract the IBC events
+    pub async fn collect_events(&mut self) -> Result<(), TMError> {
+        if let Some(tm_event) = self.event_listener.get_event().await? {
+            if let Ok(ibc_events) = relayer_modules::events::get_all_events(tm_event) {
+                // TODO - send_timeout()?
```
milosevic, line 86: Can we create issue for this?

modules/src/events.rs
zmanian: https://doc.rust-lang.org/edition-guide/rust-2018/trait-system/associated-constants.html

modules/src/events.rs
ancazamfir: Indeed, it builds a "helper" for later extraction.

modules/src/events.rs
ancazamfir: To help extract events in the order of their occurrence. The `message.action` is a vector of events e.g. 
`message.action = [x, y,  x]` 
where you have two events of type `x`. 
For each event type, each attribute is an array, eg: 
```
x.attr1 = ["a", "b"]
x.attr2 = ["A", "B"]
y.attr1 = ["c"]
```
 So if you would want to get all events of type `x` you iterate over the attribute arrays `x.attr1` and `x.attr2`
But if you want them in order then you first extract x then y then x. The helper tells you at what index in the x and y vectors you should take the values from.

I was thinking to merge this with `build_event()` as it can all be done in one pass.

modules/src/events.rs
ancazamfir: I thought about it a bit but couldn't find quickly a good solution. The other problem was that these I was not easy to find these strings so this was a sandbox for figuring this out. I will raise another PR to finish the packet events and cleanup the code.

modules/src/events.rs
ancazamfir: Yes, `ev` is an `IBCEvent`. `build_event()` uses `RawObject` to create `ev`.

modules/src/events.rs
ancazamfir: Oh, actually this is not used anymore. will remove.

modules/src/ics02_client/events.rs
ancazamfir: Yes, eventually we will. Hopefully for most of the channel and connection also.

Issue #1, opened 2020-02-26
Repository structure 


ancazamfir:
Describe the repository structure and guidelines in:
[`docs/architecture/adr-001-repo.md`](https://github.com/informalsystems/ibc-rs/blob/master/docs/architecture/adr-001-repo.md)

ebuchman:
Let's keep the english and TLA+ specs under `docs/spec` for now

PR #100, opened 2020-06-11
ADR001

adizere:
Closes: #1 

## Description

ADR 001 describing the repository structure.
______

For contributor use:

- [ ] Unit tests written


ancazamfir:
My initial thought was to use this issue to propose a layout, to go a bit more into detail and explain the layout of `modules` and `relayer`, e.g. typical files in `ics0X` like `msgs.rs`, `query.rs`, what is chain specific and what is generic, what goes in `relayer/cli/src, etc. At this point we may leave it like this or still describe it in more detail and review with everyone and make any changes now if needed. Thoughts?

adizere:
The details of the layout are not very clear to me at this point, so describing this would be very valuable to clarify! I'm in favor of providing more details (as much as we can, it's ok if some parts are debatable still) + reviewing with everyone (perhaps during the call next Tuesday).

ebuchman:
Lets just merge for now

Issue #108, opened 2020-06-18
Consolidation of ICS3 TLA+ spec


adizere:
## Summary

This repository should maintain a single version of the ICS3 TLA+ spec.

## Problem Definition

We were planning to maintain two versions of the ICS3 TLA+ spec: (1) a default version of the spec, plus (2) a version that exhibits a liveness problem.

The motivation for the first version was to model (and guide the implementation of) a correct connection handshake protocol. The motivation for the second version was to provide the context for certain protocol changes surrounding `UpdateClient`.

Having two versions, however, puts additional maintenance burden and may be unnecessary.

### Current state of the work on ICS3 TLA+ spec:

- we have the default version in [docs/spec/connection-handshake/L2-tla](https://github.com/informalsystems/ibc-rs/tree/master/docs/spec/connection-handshake/L2-tla);
- issue #71 states the necessity of capturing the liveness problem surrounding `UpdateClient` and PR #73 captures the work for this issue.

### Future, ideal state [! to be clarified]:
We'd like to keep a single version of the ICS3 TLA+ spec. This version should:

- provide the context for the liveness problem, hence it will deadlock,
- include the `store.latestHeight + 1` fix [highlighted here](https://github.com/informalsystems/ibc-rs/pull/73/files/e33f32db95314461d18fbdb13574ddeabafedad4#r440705958).
- model check & document results with heights >= 6.

____

#### For Admin Use

- [x] Not duplicate issue
- [X] Appropriate labels applied
- [x] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

ebuchman:
> provide the context for the liveness problem, hence it will deadlock,

What does this mean? 

Also, in my understanding, this UpdateClient relayer race isn't a problem for the connection itself, it's a problem for a particular relayer algorithm - ie. the existence of the race is a low-level detail of how a relayer might be implemented that the Connection shouldn't care about. I would think we'd want a representation of the Connection that doesn't depend on such low level details of individual relayers, but rather on global properties of the relayer network, that wouldn't even be able to see the existence of such races, so long as properties like "the client is eventually updated, but we don't know with which height" are upheld.

That said I haven't reviewed the TLA+ spec enough yet.

adizere:
Both versions of the ICS3 (Connection) TLA+ specs assume a high-level relayer abstraction that works like this, roughly speaking: 
1. some module pushes a message `m` to its outgoing buffer (here `m` may include a proof that the module created);
[...]

PR #109, opened 2020-06-22
ICS3 TLA+ specs consolidation

adizere:
Closes: #71 #108 .

_____

For contributor use:

- [ ] Unit tests written

### DIFF
--- a/docs/spec/connection-handshake/L2-tla/ICS3Module.tla
+++ b/docs/spec/connection-handshake/L2-tla/ICS3Module.tla
@@ -1,4 +1,4 @@
---------------------- MODULE ConnectionHandshakeModule ---------------------
+-------------------------- MODULE ICS3Module ------------------------------
 
 (***************************************************************************
 
@@ -7,7 +7,7 @@
     
     This module captures the actions and operators of the ICS3 protocol.
     Typically, it is an IBC module running on a chain that would implement
-    the logic in this TLA+ module, hence the name "ConnectionHandshakeModule"
+    the logic in this TLA+ module, hence the name "ICS3Module".
     sometimes abbreviated to "chModule" or "chm".
 
     This module deals with a high-level spec of the ICS3 protocol, so it is
@@ -48,10 +48,10 @@ VARIABLES
 
      The store record of a chain contains the following fields:
 
-     - chainID -- a string 
+     - chainID -- a string.
        Stores the identifier of the chain where this module executes.
 
-     - latestHeight -- a natural number in the range 1..MaxHeight
+     - latestHeight -- a natural number in the range 1..MaxHeight.
        Describes the current height of the chain.  
 
      - connection -- a connection record.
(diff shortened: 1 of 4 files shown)


adizere:
As a historical reference, the current PR supersedes and includes some of the work we did in #73 (but the latter PR remained unmerged since it proposed to keep a separate spec for documenting the liveness problem).

docs/spec/connection-handshake/L2-tla/ICS3Module.tla
```diff
@@ -242,7 +252,7 @@ HandleTryMsg(m) ==
         myClientProof == GetClientProof
         replyMsg == [parameters |-> FlipConnectionParameters(m.parameters),
                      type |-> "ICS3MsgAck",
-                     proofHeight |-> store.latestHeight,
+                     proofHeight |-> store.latestHeight + 1,
```
milosevic, line 255: Can you explain why it's +1 here? I guess processing a message requires to increase chain height, but it is confusing seeing +1 here?Maybe we can increase chain height in `ProcessMsg` and then have actions for each handler just read the latest height?

adizere: This `+1` is a fix for a [bug that Anca found](https://github.com/informalsystems/ibc-rs/pull/73#discussion_r440705958).

Explanation: We use `+1` to reflect that the chain advances its height before the proof is created. We advance the chain height via `NewStore`: 

https://github.com/informalsystems/ibc-rs/blob/eb0f21019778c0bf40f767d59f8d2f2002587ead/docs/spec/connection-handshake/L2-tla/ICS3Module.tla#L192-L194
that is part of every `Handle*` action, at the same time with computing `proofHeight |-> store.latestHeight + 1`. In the current spec we cannot do these two steps (advancing the height in `NewStore`, and computing the `proofHeight`) separately, so we do them at the same time, and that's why we see this `+1` in two places.

adizere: > Maybe we can increase chain height in ProcessMsg and then have actions for each handler just read the latest height?

Not sure how to acheive this. It would require that we split the `Next` action of the ICS3 module into two parts, one for increasing the height & processing a message, and another for computing proof height. This could work but it's a more major modification.

milosevic: We shouldn't be increasing height as part of message handlers (including NewStore), but outside. If there is a message to be processed then we increase height and then process message. Proof height is always the latest height, i.e., the height at the moment you are reading the state. I don't get why you need to split next action into two parts.

adizere: > If there is a message to be processed then we increase height and then process message. 

Note that all three actions happen in the same TLA+ step: (i) increasing the height, (ii) processing the message, and (iii) creating the proof. To do it otherwise than this, the `Next` action of the module would need to have separate high-level actions for each of these. (Major changes would be required for that.)

After a sync with @ancazamfir we made some small modifications that might make it better. WDYT? See [here](https://github.com/informalsystems/ibc-rs/pull/109/commits/1d988dc7dcdd14302003fc03341601daa357ff44).

Issue #84, opened 2020-06-05
Relayer algorithm specification


milosevic:
Write detailed technical specification of the relayer algorithm. The specification in ICS 18 (https://github.com/cosmos/ics/tree/master/spec/ics-018-relayer-algorithms) are not covering all details as it is meant to serve more as a relayer example. Protocol description from ADR (https://github.com/informalsystems/ibc-rs/blob/master/docs/architecture/adr-002-ibc-relayer.md) should be a starting point and it should be most probably moved to some extent outside ADR. 

As part of this task we should understand what are the properties of IBC correct relayer(s) should provide. In ICS 18, the following sentence provides hints about expectations: "as long as each pair of chains has at least one correct & live relayer and the chains remain live, all packets flowing between chains in the network will eventually be relayed." We should formalise this property, and clarify the expectations in the presence of multiple correct relayers and some faulty relayers. Furthermore, More precisely, we should understand the relation with exactly-once property specified in the IBC specification. It seems that this property defines guarantees at the channel/packet level, but it does not specify what are guarantees IBC assumes at the level of a single IBC datagram.

Furthermore, we might want to separate the relayer algorithm that is responsible for establishing of connections/channels based on relayer configuration that initially takes place, from the algorithm for relaying IBC packets that happens after connections/channels are established.

ebuchman:
Just a note this work should contribute directly to the ICS repo yes?

PR #110, opened 2020-06-23
The first version of relayer algorithm.

milosevic:
Closes: #84 

## Description

This PR defines formally requirements correct relayers should provide. Furthermore, it gives 
a relayer algorithm that satisfies these requirements. 
______

For contributor use:

- [ ] Unit tests written

### DIFF
--- a/docs/spec/relayer/Relayer.md
+++ b/docs/spec/relayer/Relayer.md
@@ -45,52 +45,107 @@ type MembershipProof struct {
 We assume the existence of the following helper functions:
 
 ```go
-queryClientConsensusState(chainA, datagramProofHeight) (ClientState, MembershipProof)
+// returns ClientState for the targetHeight if it exists; otherwise returns ClientState at the latest height
+queryClientConsensusState(chainA, targetHeight) (ClientState, MembershipProof)
 verifyClientStateProof(clientStateAonB, membershipProof, sh.appHash) boolean
 pendingDatagrams(height, chainA, chainB) IBCDatagram[]
 verifyProof(datagrams, sh.appHash) boolean
+createUpdateClientDatagrams(shs) IBCDatagram[]
+submit(datagrams) error
 ```
 
-The pseudocode of the main relayer event loop is the following: 
+The main relayer event loop is a pipeline of three stages. Assuming some IBC event at height `h` on `chainA`, 
+the relayer:
 
-```go
+1. Updates (on `chainB`) the IBC client for `chainA` to a certain height `H` where `H >= h+1`.
+2. Create IBC datagrams at height `H-1`.
+3. Submit the datagrams from stage (2) to `chainB`.
+
+Note that an IBC event at height `h` corresponds to the modifications to the data store made as part of executing
+block at height `h`. The corresponding proof (that data is indeed written to the data store) can be verified using
+the data store root hash that is part of the header at height `h+1`.
+
+Once stage 1 finishes correctly, stage 2 should succeed assuming that `chainB` has not already processed the event. The 
+interface between stage 1 and stage 2 is just the height `H`. Once stage 2 finishes correctly, stage 3 should 
+succeed. The interface between stage 2 and stage 3 is a set of datagrams.
+
+```golang
 func handleEvent(ev, chainA, chainB) {
-    datagramsProofHeight = ev.height + 1
-    
-    // query client state of chain A on chain B for proof height
-    clientStateAonB, membershipProof = chainB.queryClientConsensusState(chainA, datagramProofHeight)
-      
-    // verify client state
-    sh = chainB.lc.getHeader(membershipProof.Height + 1)
-    if !verifyClientStateProof(clientStateAonB, membershipProof, sh.appHash) then return error
-      
-    while datagramsProofHeight > clientStateAonB.Height {
-        // we need to update client for A on B before submitting datagram
-        shs = chainA.lc.get_minimal_set(clientStateAonB.Height, datagramsProofHeight)
-        chainB.submit(createUpdateClientDatagrams(shs)) // might fail to be processed because of conncurent client update
-                
-        // query client state of chain A on chain B for proof height
-        clientStateAonB, membershipProof = chainB.queryClientConsensusState(chainA, datagramProofHeight)
-                      
-        // verify client state
-        sh = chainB.lc.getHeader(membershipProof.Height + 1)
-        if !verifyClientStateProof(clientStateAonB, membershipProof, sh.appHash) then return error
+    // NOTE: we don't verify if event data are valid at this point. We trust full node we are connected to
+    // until some verification fails. 
+   
+    // Stage 1.
+    // Update on `chainB` the IBC client for `chainA` to height `>= targetHeight`.
+    targetHeight = ev.height + 1
+    // See the code for `updateIBCClient` below.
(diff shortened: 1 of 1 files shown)


docs/spec/relayer/Relayer.md
```diff
@@ -0,0 +1,96 @@
...
+
+```go
+func handleEvent(ev, chainA, chainB) {
+    datagramsProofHeight = ev.height + 1
+    
+    // query client state of chain A on chain B for proof height
+    clientStateAonB, membershipProof = chainB.queryClientConsensusState(chainA, datagramProofHeight)
+      
+    // verify client state
+    sh = chainB.lc.getHeader(membershipProof.Height + 1)
```
adizere, line 64: I don't understand the different between `membershipProof.Height` and `clientStateAonB.Height`. Intuitively, they should be the same. Is this correct?

docs/spec/relayer/Relayer.md
```diff
@@ -0,0 +1,96 @@
...
+verifyProof(datagrams, sh.appHash) boolean
+```
+
+The pseudocode of the main relayer event loop is the following: 
+
+```go
+func handleEvent(ev, chainA, chainB) {
+    datagramsProofHeight = ev.height + 1
+    
+    // query client state of chain A on chain B for proof height
```
adizere, line 60:
```suggestion
    // RPC call to chain B, querying for the state of IBC client of chain A
```

docs/spec/relayer/Relayer.md
```diff
@@ -0,0 +1,96 @@
...
+        chainB.submit(createUpdateClientDatagrams(shs)) // might fail to be processed because of conncurent client update
+                
+        // query client state of chain A on chain B for proof height
+        clientStateAonB, membershipProof = chainB.queryClientConsensusState(chainA, datagramProofHeight)
+                      
+        // verify client state
+        sh = chainB.lc.getHeader(membershipProof.Height + 1)
+        if !verifyClientStateProof(clientStateAonB, membershipProof, sh.appHash) then return error
+    }
+
```
adizere, line 79:
```suggestion

// The IBC client on chain B has a more recent state than datagramsProofHeight
```

docs/spec/relayer/Relayer.md
milosevic: membershipProof.height is the height of chain B at the point of query. clientStateAonB.Height is a data written to the store. By coincidence in this case it also has a semantic of height, but it would be exactly the same with other kind of data. In this case semantic of this field is latest installed header for chain A on chain B. Does this clarify the difference?

adizere: Thanks, in fact this makes the whole algorithm clear to me!

adizere:
I re-read the algorithm and tried to break it down into three separate stages. I also removed the recursive function call. This is more of an exercise to check I understood everything, but I hope it will also help towards getting a more modular implementation of the main event loop.

As I understand it, the main event loop is a pipeline of three stages. Assuming some IBC event at height `h` on `chainA`, the relayer:

1. Updates (on `chainB`) the IBC client for `chainA` to a certain height `H` where `H >= h+1`.
2. Create IBC datagrams at height `H-1`.
3. Submit the datagrams from stage (2) to `chainB`.

Once stage 1 finishes correctly, stage 2 should succeed assuming that `chainB` has not already processed the event. The interface between stage 1 and stage 2 is just the height `H`.
Once stage 2 finishes correctly, stage 3 should succeed. The interface between stage 2 and stage 3 is a set of datagrams.

```golang
func handleEvent(ev, chainA, chainB) {
    // Stage 1.
    // Update on `chainB` the IBC client for `chainA` to height `targetHeight`.
    targetHeight = ev.height + 1
    // See the code for `updateIBCClient` below.
    installedHeight, error := updateIBCClient(chainB, chainA, targetHeight)
    if error != nil {
        // Updating the IBC client failed because of verification error.
        // Connect to a different full node of `chainB` and retry.
    }

    // Stage 2.
    // Create the IBC datagrams including `ev` & verify them.
    datagrams = pendingDatagrams(installedHeight - 1, chainA, chainB)
    sh = chainA.lc.get_header(installedHeight)
    if !verifyProof(datagrams, sh.appHash) then return error;

    // Stage 3.
    // Submit datagrams.
    chainB.submit(datagrams)
}


// Perform an update on `dest` chain for the IBC client for `src` chain.
//    Preconditions:
//      - `src` chain has height greater or equal to `targetHeight`
//    Postconditions:
//      - returns the installedHeight >= targetHeight
//      - return error if verification of client state fails
func updateIBCClient(dest, src, targetHeight) -> {installedHeight, error} {
    installedHeight = 0;

    do {
        // Check if targetHeight exists already on destination chain.
        // Query state of IBC client for `src` on chain `dest`.
        clientState, membershipProof = dest.queryClientConsensusState(src, targetHeight)

        // Verify the result of the query.
        sh = dest.lc.get_header(membershipProof.Height + 1)
        if !verifyClientStateProof(clientState, membershipProof, sh.appHash) then return {nil, error}

        if clientState.Height < targetHeight {
            // Installed height is smaller than the target height.
            // Do an update to IBC client for `src` on `dest`.
            shs = src.lc.get_minimal_set(clientState.Height, targetHeight)
            // Might fail due to conncurent client updates.
            dest.submit(createUpdateClientDatagrams(shs))
            
            // Query (again) the state of IBC client for `src` on chain `dest`.
            clientState, membershipProof = dest.queryClientConsensusState(src, targetHeight)

            // Verify (again) the query response.
            sh = dest.lc.get_header(membershipProof.Height + 1)
            if !verifyClientStateProof(clientState, membershipProof, sh.appHash) then return {nil, error}
        }

        installedHeight = clientState.Height

    } while (installedHeight < targetHeight);

    return {installedHeight, nil}
}
```

docs/spec/relayer/Relayer.md
```diff
@@ -0,0 +1,240 @@
...
+For the purpose of this specification we assume existence of two on chain modules A and B, that executes
+IBC protocol. We say that a module A (or B) sends an IBC datagram m to a module B (or A) when a correct
+relayer can construct valid datagram m by scanning the state of the chain A. We say that a module A receives
+an IBC datagram m, when m was processed by the module A on chain. We assume that modules
+are correct.    
+
+Correct relayers need to ensure the following properties:
+
+**[ICS18-Delivery]**: If a module A sends an IBC datagram m to a module B, then m is
```
adizere, line 21: I wonder if it would help to be more accurate here. I see two issues: 

1. The notion of "send" for a module is not well-defined. Instead, I guess what happens is: a module A emits IBC events, and some of these events indicate state transitions that are relevant for another module B.

2. Perhaps we should refer to chains here, not modules. The function signature `func handleEvent(ev, chainA, chainB)` refers to chains, and there isn't reference to modules anywhere except in these properties. Indeed, relaying is done at chain-level. Relayer granularity seems impossible at module-level.

Proposal for refining the properties based on these ideas:

**[ICS18-Delivery]**: If any module on `chainA` emits an IBC event `m` relevant for some module on `chainB`, then `chainB` eventually receives an IBC datagram matching event `m`.

**[ICS18-Validity]**: If some module on `chainB` receives an IBC datagram for an event `m` from a module on `chainA`, then a module on `chainA` emitted an event signalling event `m` to a module on `chainB`.

This is not perfect because we need to define what it means for an event to be "relevant" for a module. We can clarify this with specific examples.

Issue #113, opened 2020-06-24
Dependency on tendermint-rs causes compilation failure


adizere:
## Summary of Bug

Compilation (`cargo build`) on a vanilla clone fails with errors:

```
error[E0433]: failed to resolve: could not find `rpc` in `tendermint`
  --> modules/src/events.rs:11:17
   |
11 | use tendermint::rpc::event_listener::{ResultEvent, TMEventData};
   |                 ^^^ could not find `rpc` in `tendermint`

error[E0433]: failed to resolve: could not find `rpc` in `tendermint`
 --> modules/src/ics02_client/query.rs:2:17
  |
2 | use tendermint::rpc::endpoint::abci_query::AbciQuery;
  |                 ^^^ could not find `rpc` in `tendermint`

error[E0433]: failed to resolve: could not find `rpc` in `tendermint`
 --> modules/src/ics03_connection/query.rs:1:17
  |
1 | use tendermint::rpc::endpoint::abci_query::AbciQuery;
  |                 ^^^ could not find `rpc` in `tendermint`
```

This seems to happen since `tendermint-rs` release 0.14.1, which would explain the error, since this release included a separation of the `rpc` module into its own crate: https://github.com/informalsystems/tendermint-rs/commit/e2335c40b1c5e1f7d47ee28ae5f9cc679730b7a2.

## Version

```
rustc --version
rustc 1.44.1 (c7087fe00 2020-06-17)
```

ibc-rs commit: `commit 7a49bca2c34cb9964c66137a9edaa4c5f7a9b6c1`

## Steps to Reproduce

```
git clone git@github.com:informalsystems/ibc-rs.git
cd ibc-rs
cargo build
```

## Solutions

1. Wherever we use the `tendermint::rpc` create, we need to replace that with `tendermint_rpc`. For instance:
```use tendermint_rpc::event_listener::{ResultEvent, TMEventData};```
instead of
```use tendermint::rpc::event_listener::{ResultEvent, TMEventData};```

Alternatively, for a nicer read `use tendermint_rpc as rpc;`.

2. A quicker -- temporary fix -- would be to rely on a specific release as dependency in our .toml files:

```
tendermint = { git = "https://github.com/informalsystems/tendermint-rs.git", tag = "0.14.0" }
```
____

## For Admin Use

- [x] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

[...]

PR #115, opened 2020-06-25
Fixes for tendermint rpc refactored in its own crate (#113)

andynog:
Closes: #113 

## Description

Added a new dependency for the 'tendermint-rpc' crate. Fixed the dependency path (from tendermint::rpc to tendermint_rpc).

Opening this as a draft PR since it's still not compiling because the 'client' module is not public in tendermint-rs so this build is failing. Need to make ['client' module in tendermint-rs public](https://github.com/informalsystems/tendermint-rs/blob/81f7eb3aac9aac24ac83a02ba7e91ac477bf57fc/rpc/src/lib.rs). 
______

For contributor use:

- [ ] Unit tests written

### DIFF
--- a/modules/src/events.rs
+++ b/modules/src/events.rs
@@ -8,7 +8,7 @@ use serde_derive::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::convert::{TryFrom, TryInto};
 use tendermint::block;
-use tendermint::rpc::event_listener::{ResultEvent, TMEventData};
+use tendermint_rpc::client::event_listener::{ResultEvent, TMEventData};
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub enum IBCEvent {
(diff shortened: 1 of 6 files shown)


modules/src/events.rs
```diff
@@ -8,7 +8,7 @@ use serde_derive::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::convert::{TryFrom, TryInto};
 use tendermint::block;
-use tendermint::rpc::event_listener::{ResultEvent, TMEventData};
+use tendermint_rpc::client::event_listener::{ResultEvent, TMEventData};
```
liamsi, line 11: I think you don't need access to the `client` module directly but to the `event_listener` instead (as previously).

I'll try to hack together a PR to confirm this.

liamsi: I don't seem to have perms to push a branch anymore. I'll use a fork instead. Not sure CI will trigger though.

liamsi:
See: https://github.com/informalsystems/ibc-rs/pull/116

adizere:
Thanks both for the discussion and fixes! You move fast. @andynog perhaps we can close this PR since the fix was merged in #116.

Issue #89, opened 2020-06-09
Relayer TLA+ specification: Multiple relayers and modular ICS protocols


istoilkovska:
## Summary

The idea is to refactor and restructure the existing modules of the Relayer TLA+ specification such that a system where multiple relayers (at least two correct and one faulty) can be specified in a straightforward way. 

## Problem Definition

In the existing Relayer TLA+ specification, we assume that we have a single correct relayer, that relays datagrams between two chains, that constitute its environment. The relayer logic follows the naive relayer algorithm, proposed in [ICS 018](https://github.com/cosmos/ics/tree/master/spec/ics-018-relayer-algorithms). 

By refactoring and restructuring the Relayer TLA+ specification, we aim at producing a more modular specification, where the modules can be composed easily, and the properties of a system containing instances of several modules can be checked efficiently. Additionally, we aim at providing the possibility to check the properties of the modules in isolation as well, and thus avoid duplicate specifications of the same protocols.

## Proposal

We propose the following modular structure of the Relayer TLA+ specification:
- module `Relayer.tla`: 
  - contains the relayer logic and encodes a process running the relayer algorithm,
  - stores heights for participating chains, and produces datagrams,
  - defines Boolean constants, which flag which kinds of datagrams are being created and which properties are being checked (e.g., one can restrict the relayer to only create connection handshake datagrams, and check properties about eventual delivery of connection handshake datagrams);
- module `Chain.tla`: 
  - contains the chain logic,
  - stores a chain height, a set of counterparty client heights, connection and channel ends (in the current specification, this is maintained by the environment),
  - defines chain actions, such as, increasing the chain height, or processing incoming datagrams;
- modules `ProtocolHandlers.tla`, where `Protocol` is one of `Client`, `Connection`, `Channel`:
  - defines operators that are used by a chain to handle `Protocol` datagrams;
- module `RelayerDefinitions.tla`:
  - defines operators shared by all the modules;
- **main module** `RelayerSystem.tla`:
  - specifies a system with one or more relayers, two or more communicating chains,
  - defines system variables and creates instances of `Relayer.tla` and `Chain.tla`, by ensuring the variables are  correctly shared,
  - defines the system initial state predicate and next state action by reasonably composing the initial state predicates and next state actions of the instantiated modules.

[...]

PR #117, opened 2020-06-26
Relayer TLA+ refactoring

istoilkovska:
WIP: #89

## Description

The specification has seven modules: 
  - `ICS18Environment.tla` (*new*, the main module)
  - `Relayer.tla` (*modified*, to allow for instantiation) 
  - `Chain.tla` (*new*, created to specify the chain logic, previously specified in `Environment.tla`)
  - `ClientHandlers.tla` (handler logic as in the older version)
  - `ConnectionHandlers.tla` (handler logic as in the older version)
  - `ChannelHandlers.tla` (handler logic as in the older version)
  - `RelayerDefinitions.tla` (*new*, created to hold definitions of various operators used across all modules)

The module `ICS18Environment.tla` creates instances of `Relayer.tla` and 
`Chain.tla`. Currently, we have 
 - two instances of `Relayer.tla`, specifying the behavior of two correct relayers, `Relayer1` and `Relayer2`,
 - two instances of `Chain.tla`, specifying the behaviors of two chains, `ChainA` and `ChainB`.

We should specify three kinds of properties for the relayer algorithm:
- **ICS18Safety**: Bad datagrams are not used to update the chain stores.
- **ICS18Validity**: If `ChainB` receives a datagram from `ChainA`, then the datagram was sent by `ChainA` 
- **ICS18Delivery**: If `ChainA` sends a datagram to `ChainB`, then `ChainB` eventually receives the datagram

Currently, we specify **ICS18Delivery** for all datagrams, and **ICS18Safety** for `ClientUpdate` datagrams.
- **WIP**: specification of **ICS18Safety** for other datagrams.
______

For contributor use:

- [ ] Unit tests written


docs/spec/relayer-v2/Chain.tla
```diff
@@ -0,0 +1,146 @@
...
+
+(***************************************************************************
+ Client update operators
+ ***************************************************************************)
+\* Update the clients on chain with chainID, 
+\* using the client datagrams generated by the relayer      
+\* (Handler operators defined in ClientHandlers.tla)
+LightClientUpdate(chainID, chain, datagrams) == 
+    \* create clients
+    LET clientCreatedChain == HandleCreateClient(chainID, chain, datagrams) IN
```
milosevic, line 22: I didn't know we can use operators from different files like this. Is there a way we can prefix it somehow so it is more clear where it comes from?

docs/spec/relayer-v2/Chain.tla
```diff
@@ -0,0 +1,146 @@
...
+    
+(***************************************************************************
+ Connection update operators
+ ***************************************************************************)
+\* Update the connections on chain with chainID, 
+\* using the connection datagrams generated by the relayer
+\* (Handler operators defined in ConnectionHandlers.tla)
+ConnectionUpdate(chainID, chain, datagrams) ==
+    \* update chain with "ConnOpenInit" datagrams
+    LET connOpenInitChain == HandleConnOpenInit(chainID, chain, datagrams) IN
```
milosevic, line 36: Can order in which we apply those operators matter?I guess that in theory it shouldn't but maybe it could be relevant for some implementations.

docs/spec/relayer-v2/Chain.tla
```diff
@@ -0,0 +1,146 @@
...
+
+vars == <<chainStore, incomingDatagrams>>          
+
+(***************************************************************************
+ Client update operators
+ ***************************************************************************)
+\* Update the clients on chain with chainID, 
+\* using the client datagrams generated by the relayer      
+\* (Handler operators defined in ClientHandlers.tla)
+LightClientUpdate(chainID, chain, datagrams) ==
```
milosevic, line 20: Is second parameter chain or chainStore?

docs/spec/relayer-v2/Chain.tla
```diff
@@ -0,0 +1,146 @@
...
+    \* ICS 002: Client updates
+    LET lightClientsUpdatedStore == LightClientUpdate(chainID, chainStore, datagrams) IN 
+    \* ICS 003: Connection updates
+    LET connectionsUpdatedStore == ConnectionUpdate(chainID, lightClientsUpdatedStore, datagrams) IN
+    \* ICS 004: Channel updates
+    LET channelsUpdatedStore == ChannelUpdate(chainID, connectionsUpdatedStore, datagrams) IN
+    
+    \* update height
+    LET updatedChainStore == 
+        IF /\ chainStore /= channelsUpdatedStore
```
milosevic, line 79: In principle, if incomingDatagrams is not empty, we should increase the height. This would model the fact that IBC datagrams are processed on chain, although they don't necessarily need to lead to state change. On the other side, that would potentially significantly increase the state space, so increasing height only when we effectively make state change by processing datagrams is more meaningful. Maybe we should just add comment explaining this design decision.

docs/spec/relayer-v2/RelayerDefinitions.tla
```diff
@@ -0,0 +1,282 @@
...
+    
+    - counterpartyClientHeights : a set of integers between 1 and MaxHeight
+      Stores the heights of the client for the counterparty chain.
+
+    - connectionEnd : a connection end record 
+      Stores data about the connection with the counterparty chain
+ ***************************************************************************)
+Chains ==    
+    [
+        height : Nat,
```
milosevic, line 95: Why not 1..MaxHeight?

docs/spec/relayer-v2/Chain.tla
istoilkovska: This is because `Chain.tla` extends the module `ClientHandlers.tla`. Then the operators defined in `ClientHandlers.tla` are incorporated in `Chain.tla`, and are used as if they were defined in `Chain.tla`.

docs/spec/relayer-v2/Chain.tla
istoilkovska: The output chain of every operator is given as input to the next operator. I think it makes a difference in case there are faulty relayers. If there are `ConnOpenInit` (sent by correct relayer) and `ConnOpenAck` (sent by faulty relayer) in incoming datagrams for `chainA`, and `chainA` is in state `UNINIT`, in this order, after applying `ConnectionUpdate`, the connection end at `chainA` will be in state `OPEN`. If the order of the operators is swapped, then the chain first processes `ConnOpenAck` (discards it), and then processes `ConnOpenInit`, which results in a connection end in state `INIT` at `chainA`.

docs/spec/relayer-v2/RelayerDefinitions.tla
istoilkovska: This module contains definitions of operators which are reused across other modules (they extend `RelayerDefinitions.tla`, and thus get access to the definitions). `MaxHeight` is a constant of the environment, and its value cannot be transferred to an extended module. This definition of `Chains` is only used in the type invariant, to check if a given chain store is a record from this set. TLC does not use the definition of `Chains` to compute states, hence it does not complain. The chain stores restrict height to be in the set `1..MaxHeight`, so the type invariant ensures that the height (between `1..MaxHeight`) is a natural number.

docs/spec/relayer-v2/Chain.tla
istoilkovska: I agree, increasing the height will imply that we need to assign higher value to `MaxHeight`, which in turn increases the state space. I'll add a comment.

docs/spec/relayer-v2/Chain.tla
milosevic: I guess there is no way to somehow namespace differently imported operators: for example clientHandlers.HandleCreateClient as it is imported from the file ClientHandlers.tla?

docs/spec/relayer/Relayer.tla
```diff
@@ -5,111 +5,31 @@
  It instantiates the module Environment, which takes care of the chain logic.
```
milosevic, line 5: ICS18Environment instead of Environment.

docs/spec/relayer/Relayer.tla
```diff
@@ -155,14 +77,17 @@ LightClientUpdates(srcChainID, dstChainID, relayer) ==
...
+    LET srcConnectionEnd == GetConnectionEnd(srcChain) IN
+    LET dstConnectionEnd == GetConnectionEnd(dstChain) IN
 
     LET srcConnectionID == GetConnectionID(srcChainID) IN
     LET dstConnectionID == GetConnectionID(dstChainID) IN
 
-    LET srcHeight == GetLatestHeight(srcChainID) IN
-    LET srcConsensusHeight == GetMaxCounterpartyClientHeight(srcChainID) IN
+    LET srcHeight == GetLatestHeight(srcChain) IN
+    LET srcConsensusHeight == GetMaxCounterpartyClientHeight(srcChain) IN
```
milosevic, line 90: Shouldn't it be dstChain?Also maybe better using the same name as above (srcClientHeight).

docs/spec/relayer/Relayer.tla
```diff
@@ -155,14 +77,17 @@ LightClientUpdates(srcChainID, dstChainID, relayer) ==
...
     
     IF dstConnectionEnd.state = "UNINIT" /\ srcConnectionEnd.state = "UNINIT" THEN 
          {[type |-> "ConnOpenInit",
```
milosevic, line 93: I am not sure if we want to create ConnOpenInit the same way as other datagrams as this is the initial trigger that should come from outside, for example from the config file. Don't have better proposal, but leaving comment as a reminder to discuss.

docs/spec/relayer/Relayer.tla
```diff
@@ -233,25 +161,34 @@ ChannelDatagrams(srcChainID, dstChainID) ==
...
 \*  -  ICS 004: Channel handshake
 \* TODO: Support the remaining datagrams
-PendingDatagrams(srcChainID, dstChainID) ==
+ComputeDatagrams(srcChainID, dstChainID) ==
     \* ICS 002 : Clients
     \* - Determine if light clients needs to be updated 
-    LET clientDatagrams == LightClientUpdates(srcChainID, dstChainID, relayerChainHeights) IN
+    LET clientDatagrams == 
+        IF GenerateClientDatagrams 
+        THEN LightClientUpdates(srcChainID, dstChainID, relayerHeights)
```
milosevic, line 176: Where relayerHeights come from?I guess idea here is to update relayer's height for chainA and chainB even if we don't generate clientDatagrams?

docs/spec/relayer/Relayer.tla
```diff
@@ -267,478 +204,68 @@ PendingDatagrams(srcChainID, dstChainID) ==
  ***************************************************************************)   
 \* Update the height of the relayer client for some chainID
 UpdateRelayerClients(chainID) ==
-        /\ relayerChainHeights[chainID] < GetLatestHeight(chainID)
-        /\ relayerChainHeights' = [relayerChainHeights EXCEPT 
-                                        ![chainID] = GetLatestHeight(chainID)
+        /\ relayerHeights[chainID] < GetLatestHeight(GetChainByID(chainID))
```
milosevic, line 207: Extract GetLatestHeight(GetChainByID(chainID)) into a variable so conditions read nicer.

docs/spec/relayer/Relayer.tla
```diff
@@ -5,111 +5,31 @@
...
+
+CONSTANTS Heights \* set of possible heights of the chains in the system 
+
+VARIABLES chainAstore, \* store of ChainA
+          chainBstore, \* store of ChainB
           outgoingDatagrams, \* a function that assigns a set of pending datagrams 
                              \* outgoing from the relayer to each chainID 
-          relayerChainHeights, \* a function that assigns a height to each chainID         
-          turn  
+          relayerHeights \* a function that assigns a height to each chainID
```
milosevic, line 24: Relayer in principle does not maintain latest height (header) for each chain. It actually needs to send client update for the corresponding datagrams. For this it needs to be able to create clientDatagrams and for this is needs access to trusted headers. This is what light client will provide. We should think how we want to model this, but for now I would suggest to remove relayerHeights as it does not really add anything useful to the model as we create client updates based on src chain height and src client height on the destination chain.

docs/spec/relayer/Relayer.tla
```diff
@@ -118,8 +38,10 @@ IsChannelOpen(chainID) == Env!IsChannelOpen(chainID)
 \* Some client updates might trigger an update of the height that 
 \* the relayer stores for srcChainID
 LightClientUpdates(srcChainID, dstChainID, relayer) ==
-    LET srcChainHeight == GetLatestHeight(srcChainID) IN    
-    LET srcClientHeight == GetMaxCounterpartyClientHeight(dstChainID) IN
+    LET srcChain == GetChainByID(srcChainID) IN
+    LET dstChain == GetChainByID(dstChainID) IN
+    LET srcChainHeight == GetLatestHeight(srcChain) IN
```
milosevic, line 43: We should probably parameterise datagram generation functions with a height for which we are creating datagrams. If datagrams are created for height h, then the light client updates are created for the height h+1 as root hash for datagrams created at height h is stored in the block h+1.

docs/spec/relayer-old/README.md
```diff
@@ -0,0 +1,26 @@
+A TLA+ specification of the Relayer algorithm.
```
adizere, line 1: Do we want to keep the old version of ICS18 in the repository? If yes, then I suggest we call it `v1` and locate it inside `docs/spec/relayer/` so we'll end up with a structure as follows:

```
docs/spec/relayer:
  --- v1:
         |-- ChannelHandlers.tla
         |-- Environment.tla 
         |-- README.md
         ....
  --- v2
         |-- ChannelHandlers.tla 
         |-- Chain.tla
         ....
```

I'm not a fan of calling anything "old" because that's relative (so the "refactored" new spec might become "old" at some point).

docs/spec/relayer/README.md
```diff
@@ -1,26 +1,46 @@
-A TLA+ specification of the Relayer algorithm.
+A refactored TLA+ specification of the ICS18 Relayer algorithm.
```
adizere, line 1:
```suggestion
A refactored (v2) TLA+ specification of the ICS18 Relayer algorithm.
```

docs/spec/relayer/Relayer.tla
```diff
@@ -267,478 +204,68 @@ PendingDatagrams(srcChainID, dstChainID) ==
...
+    \E chainID \in ChainIDs : UpdateRelayerClients(chainID)
     
-\* Environment
-\*    The environment takes the action Env!Next
-\*    and leaves the relayer variable unchanged
-Environment ==
-    /\ Env!Next
-    /\ UNCHANGED relayerChainHeights    
+CreateDatagrams ==
+    \E srcChainID \in ChainIDs : \E dstChainID \in ChainIDs : Relay(srcChainID, dstChainID)
```
milosevic, line 231: Can it happen that chosen srcChainID and dstChainID are the same, so Relay action does not trigger? I am wondering if it's simpler just picking srcChainID as we then know dstChainID, at least in a model with two chains.

docs/spec/relayer/Relayer.tla
```diff
@@ -267,478 +204,68 @@ PendingDatagrams(srcChainID, dstChainID) ==
...
+\*        - relays datagrams between two chains, or
+\*        - does nothing
 Next ==
-    \/ /\ turn = "relayer"
-       /\ Relayer
-       /\ turn' = "environment"
-    \/ /\ turn = "environment"
-       /\ Environment
-       /\ turn' = "relayer"     
+    \/ UpdateRelayer
```
milosevic, line 249: We probably don't need this (see comments above). We should eventually model client update as a consequence of trusted period expiration. But logic for it would probably also be part of CreateDatagrams.

docs/spec/relayer/Relayer.tla
adizere: I think the "comment header" of the Relayer module is outdated. The Relayer module no longer instantiates the Environment: it's the other way around now, so we should probably update this comment to reflect the new situation.

docs/spec/relayer/Relayer.tla
adizere: Would be a nice improvement to parametrize datagram generation this way!

docs/spec/relayer-old/README.md
milosevic: We will not keep it after this PR is merged. It is just there as a stable version.

docs/spec/relayer/ICS18Environment.tla
```diff
@@ -0,0 +1,314 @@
...
+               incomingDatagrams <- incomingDatagramsChainB
+
+(***************************************************************************
+ Component actions
+ ***************************************************************************)
+
+\* RelayerAction: either correct relayer takes a step, leaving the other 
+\* variables unchanged 
+RelayerAction ==
+    \/ /\ Relayer1!Next
```
milosevic, line 78: We will probably need to make relayer steps more fine grained, i.e., smaller atomic steps. In this version a relayer comes, query state of both chains, create datagrams and submit them, and only then the second relayer can do something. This constrains unrealistically system concurrency. But that is probably out of scope of this PR.

docs/spec/relayer/ICS18Environment.tla
```diff
@@ -0,0 +1,314 @@
...
+       /\ UNCHANGED relayerVars
+    \/ /\ ChainB!Next  
+       /\ UNCHANGED chainAvars
+       /\ UNCHANGED relayerVars  
+
+(***************************************************************************
+ ICS18Environment actions
+ ***************************************************************************)
+\* Submit datagrams from relayers to chains
+SubmitDatagrams ==
```
milosevic, line 101: I guess the idea with this action is to separate datagram computation from submission so we enable more concurrency between relayers?

docs/spec/relayer/ICS18Environment.tla
istoilkovska: It is a bit more fine-grained than that. Currently, in a relayer step, defined by Relayer!Next (lines 249-252 of Relayer.tla), we have that a relayer either increases the client height of one chain (action `UpdateRelayerClients(chainID)`, lines 206-212 of Relayer.tla), or it scans srcChain and creates datagrams for dstChain (action `Relay(srcChainID, dstChainID)`, lines 216-226 of Relayer.tla). These datagrams are only created, not delivered. The environment delivers them using `SubmitDatagrams`, adding more concurrency. I assume we can go even more fine-grained if we define separate actions for creating datagrams of each protocol (client, connection, channel...), as well as submitting datagrams one by one, rather than in a batch.

I agree that the current model does not capture the concurrency in the real world, but I think we should be careful when adding more behaviors, in order not to make model checking infeasible.

docs/spec/relayer/ICS18Environment.tla
istoilkovska: Yes. Once there is a faulty relayer, which would just add datagrams non-deterministically to `outgoingDatagrams`, the environment action `SubmitDatagrams` would also deliver datagrams from the faulty relayer.

docs/spec/relayer/Relayer.tla
istoilkovska: The `LightClientUpdates` may trigger the relayer to update its client heights, in case they are not up to date with the height of the chain involved in the creation of client datagrams. When the relayer does not create client datagrams, these we don't update the relayer's client heights, that is, we output the heights that were given as input (by setting `relayerUpdate |-> relayerHeights`).

docs/spec/relayer/README.md
```diff
@@ -1,26 +1,46 @@
...
+- **ICS18Delivery**: If `ChainA` sends a datagram to `ChainB`, then `ChainB` eventually receives the datagram
+
+Currently, we specify **ICS18Delivery** for all datagrams, and 
+**ICS18Safety** for `ClientUpdate` datagrams.
 
 To import the specification in the TLA+ toolbox and run TLC:
-  - add the spec `Relayer.tla` in TLA+ toolbox
-  - add `Environment.tla`, `ClientHandlers.tla`, `ConnectionHandlers.tla`, and 
-    `ChannelHandlers.tla` as modules in the spec
+  - add the spec `ICS18Environment.tla` in TLA+ toolbox
```
adizere, line 41:
```suggestion
  - add the spec (root-module file) `ICS18Environment.tla` in TLA+ toolbox
```

docs/spec/relayer-v2/Chain.tla
istoilkovska: `chainStore` is an already defined variable, so there's a name clash. I can rename the variable to `store`, if that's more intuitive.

docs/spec/relayer/Relayer.tla
ancazamfir: Yes I tried before, in spec and code to find a good name for this:
"the height of the latest consensus of clientB running on chain A" or
"the height of the latest consensus of dest client running on src chain".
First thing that comes to my mind here is that this is a height of chain B that you obtain from its client on A. So indeed `srcConsensusHeight` might make someone think that this is a height of chain A.

docs/spec/relayer-old/ChannelHandlers.tla
```diff
@@ -0,0 +1,172 @@
...
+                  chain EXCEPT !.connectionEnd = chanOpenConfirmConnectionEnd
+              ] IN
+              
+              chanOpenConfirmChain                
+         \* otherwise, do not update the chain
+         ELSE chain
+    \* otherwise, do not update the chain     
+    ELSE chain 
+    
+=============================================================================
```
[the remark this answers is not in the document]
istoilkovska: I've just created an issue for specifying the channel closing handshake.

docs/spec/relayer-old/ChannelHandlers.tla
```diff
@@ -0,0 +1,172 @@
...
+    ELSE chain
+
+\* Handle "ChanOpenTry" datagrams
+HandleChanOpenTry(chainID, chain, datagrams) ==
+    \* get "ChanOpenTry" datagrams, with a valid channel ID
+    LET chanOpenTryDgrs == {dgr \in datagrams : 
+                            /\ dgr.type = "ChanOpenTry"
+                            /\ dgr.channelID = GetChannelID(chainID)
+                            /\ dgr.proofHeight \in chain.counterpartyClientHeights} IN
+
```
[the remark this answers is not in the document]
istoilkovska: Yes, that's correct, the `datagrams` is a set and there is no order in which the messages are received currently. However, the way in which the operator `ChannelUpdate` in Chain.tla is defined, the `openInit` is processed before `openTry`, since the `HandleChanOpenInit` operator is applied first, and then `HandleChanOpenTry` is applied next (and `HandleChanOpenConfirm`, `HandleChanOpenAck`, in that order). 

I suppose that when we specify the event-based relayer, we would switch from sets to queues to store the incoming datagrams. This would induce a minor change in the handlers, exactly in these lines.

Issue #112, opened 2020-06-23
First release + more mature repo standards


adizere:
## Summary

Preparations for the first release.

## Proposal

Minimal list of prerequisites:

- CHANGELOG
- README update
- release not necessarily aligned with milestones, e.g., this release fill fall between milestones [5](https://github.com/informalsystems/ibc-rs/milestone/5) and [6](https://github.com/informalsystems/ibc-rs/milestone/6).

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ancazamfir:
> release not necessarily aligned with milestones, e.g., this release fill fall between milestones 5 and 6.

Yes, this was on purpose. Milestones are mid-month to allow for release preparation by end of month. But now with releases every every month not sure it makes sense anymore.

PR #120, opened 2020-06-29
Release 0.0.1 of ibc-rs

ancazamfir:
Closes: #112

## Description
Added CHANGELOG.md in preparation for 0.0.1 release of ibc-rs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
-## [0.0.1] - 2020-07-01
+## [0.0.1]
+
+* July 1st, 2020*
 
 - Initial release
 ### Added
@@ -11,4 +13,4 @@
 
     - Note: the event handler just displays the events.
 - IBC Modules partial implementation for datastructures, messages and queries.
-- Some specifications and models for Connection Handshake and Naive relayer algorithm.
+- Some English and TLA+ specifications for Connection & Channel Handshake as well as naive relayer algorithm.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,14 @@
+## [0.0.1] - 2020-07-01
+
+- Initial release
+### Added
+- Configuration file definition and validation
+- Client state, consensus state, connection, channel queries.
+
+    - Note: deserialization is unimplemented as it has dependency on migration to protobuf for ABCI queries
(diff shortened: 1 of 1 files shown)


milosevic:
I guess we will add CHANGELOG_PENDING file and more structure to the files (BREAKING CHANGES, FEATURES, IMPROVEMENTS, BUG FIXES) after the initial release.

CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
+## [0.0.1] - 2020-07-01
+
+- Initial release
+### Added
```
milosevic, line 4: Instead of added, we should maybe just summarise what this initial release contains. This is basically what follows, but we can instead of Added maybe explain that we cut initial release to be able to track more carefully changes to the project with the following releases, and that at this point we don't have minimal set of features so it can be used in any reasonable scenario.

CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
+## [0.0.1] - 2020-07-01
```
adizere, line 1:
```suggestion
## [0.0.1]

* July 1st, 2020*
```

In keeping with tendermint-rs [changes.md](https://github.com/informalsystems/tendermint-rs/blob/master/CHANGES.md), we should consider writing the date between parentheses. The "Github Repo Standards", however, has a different proposal: to write the release date in italics on a separate line (see merge_requests/47 of the handbook). I guess there is no right or wrong way, and this is not a critical aspect, but it would be helpful to have a single, consistent format across our projects. I would vote for the repo standards format.

CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
...
+- Configuration file definition and validation
+- Client state, consensus state, connection, channel queries.
+
+    - Note: deserialization is unimplemented as it has dependency on migration to protobuf for ABCI queries
+- Per chain light clients threads are created and headers are periodically retrieved and verified.
+- Per chain IBC event monitor threads are spawned and main event handler that receives them.
+
+    - Note: the event handler just displays the events.
+- IBC Modules partial implementation for datastructures, messages and queries.
+- Some specifications and models for Connection Handshake and Naive relayer algorithm.
```
adizere, line 14:
```suggestion
- Some English and TLA+ specifications for Connection & Channel Handshake as well as naive relayer algorithm.
```

Issue #130, opened 2020-07-07
query connection given connection_id


ebuchman:
Given a connection id, query the connection.

Should be able to run this from the relayer CLI.

ancazamfir:
Proto definitions for connection queries and responses (see `ConnectionEnd`, `Counterparty`, etc) are here:
https://github.com/cosmos/cosmos-sdk/tree/master/proto/ibc/connection

andynog:
As per our discussion today, I'll create a proto folder and store the proto files in there following the same structure as the  cosmos sdk proto folder.

liamsi:
BTW, something I forgot to mention but you've probably seen already is that if you want to generate the rust types during build time, you also need a dependency on [`prost-build`](https://docs.rs/prost-build/0.6.1/prost_build/):
```
[build-dependencies]
prost-build = "0.6"
```
and a `build.rs` file. In tendermint-rs we never used that because we never generated the types (this might change in the near future too). Here is an example:
https://github.com/libp2p/rust-libp2p/blob/master/core/build.rs
https://github.com/libp2p/rust-libp2p/blob/master/core/src/keys.proto
https://github.com/libp2p/rust-libp2p/blob/6ad05b0ff1cb11ceec37cb191adb0d13feaceeb4/core/src/lib.rs#L38-L40

andynog:
- Added a new branch **andy/proto**
- Implemented the initial proto files required for the connection. 
- The proto files are in the modules folder. 
- Copied proto files from cosmos-sdk and tendermint repos, but it seems they are not in sync, e.g. a cosmos proto file reference an import from tendermint that was not valid. Noticed the tendermint proto files just had a refactoring a few hours ago. Looks like this proto work on the crates are pretty much in flux for now.
- There's a new build.rs in there that generate the files in the output folder (e.g. /target/debug/build/relayer-modules-[HASH]/ibc.connection.rs).
- Then tried to replace the connections mod (modules/src/ics03_connection/mod.rs) with generated file as per prost_build instructions but got bunch of failures probably because there are dependencies that the original Rust connection had and it's not in the generated proto files. 

Next steps are trying to make it compile again using the new proto generated file. We might need to chat about this. Seems to me that a lot of logic in the proto files are somewhat duplicate with things already implemented in the Rust code.

ebuchman:
[...]

PR #136, opened 2020-07-09
WIP Connection queries for Relayer

adizere:
Closes: #130 

## Description

______

For contributor use:

- [ ] Unit tests written


modules/src/ics03_connection/query.rs
```diff
@@ -105,6 +119,24 @@ impl IdentifiedConnectionEnd {
     }
 }
 
+fn proto_unmarshal(bytes: Vec<u8>) -> Result<ConnectionEnd, error::Error> {
+    let buf = Bytes::from(bytes);
+    let decoded = proto_ibc_connection::ConnectionEnd::decode(buf);
+    match decoded {
+        Ok(conn) => {
+            let client_id = ClientId::from_str(&conn.client_id).unwrap();
```
ancazamfir, line 127: We should `.map_err` here (also for the other `unwrap()`s) and send the error up. For a command like this:
`cargo run --bin relayer -- -c ./relayer/relay/tests/config/fixtures/relayer_conf_example.toml query connection end chain_A connectionnotthere` we should get a friendlier output.

modules/src/ics03_connection/query.rs
```diff
@@ -105,6 +119,24 @@ impl IdentifiedConnectionEnd {
...
+            let connection_end = ConnectionEnd::new(client_id, counterparty, conn.versions).unwrap();
+            Ok(connection_end)
+        },
+        Err(e) => {
+            println!("Error proto un-marshall: {:?}", e);
+            todo!()
+        }
+    }
+}
+
```
ancazamfir, line 139: You can remove the `amino` func below.

adizere:
### Before merging:


Because of the code which `prost` generates (based on the .proto file), we had to disable `unused_qualifications` warnings [here](https://github.com/informalsystems/ibc-rs/blob/949ad61e38e1bee2638e858e08bc011ce8d64e4b/modules/src/lib.rs). 
The explicit warning originates from the .rs file generated by prost:

```
  Compiling relayer-modules v0.0.1 (/Users/adi/Hammers/ibc-rs/modules)
error: unnecessary qualification
  --> /Users/adi/Hammers/ibc-rs/target/debug/build/relayer-modules-4a039a021af40008/out/connection.rs:12:13
   |
12 |     pub id: std::string::String,
   |             ^^^^^^^^^^^^^^^^^^^
   |
note: the lint level is defined here
  --> modules/src/lib.rs:9:5
   |
9  |     unused_qualifications,
   |     ^^^^^^^^^^^^^^^^^^^^^
```


We have two options:

1. keep the warnings disabled, or
2. re-enable the warnings and bypass the prost compilation step, i.e., we put in our codebase not the .proto files but the final .rs files, modified by hand to pass the `unused_qualifications` warning.

modules/src/ics03_connection/exported.rs
```diff
@@ -42,6 +42,15 @@ impl State {
             Self::Open => "OPEN",
         }
     }
+
+    pub fn from_i32(nr: i32) -> Self {
```
adizere, line 46: @ancazamfir was not sure this is accurate, please double-check.

modules/src/ics03_connection/connection.rs
```diff
@@ -25,6 +29,34 @@ impl ConnectionEnd {
...
+    pub fn set_state(&mut self, new_state: State) {
+        self.state = new_state;
+    }
+
+    pub fn from_proto_connection(pc: proto_connection::ConnectionEnd) -> Result<Self, Error> {
+        // The Counterparty field is an Option, may be missing.
+        match pc.counterparty {
+            Some(cp) => {
+                let mut conn = ConnectionEnd::new(
+                    ClientId::from_str(&pc.client_id).unwrap(),
```
ancazamfir, line 42: Could we get rid of the `unwrap()`s? If the user issues a query command and gets back a few pages of traces it does not help. I had this comment earlier but I guess it got lost. Try to query with non-exsiting connection id.

modules/src/ics23_commitment/mod.rs
```diff
@@ -32,6 +32,10 @@ impl CommitmentProof {
 */
 
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
-pub struct CommitmentPrefix;
+pub struct CommitmentPrefix(Vec<u8>);
```
ancazamfir, line 35: Wondering if we should have a formatter for this. The output of the query shows: `prefix: CommitmentPrefix([112, 114, 101, 102, 105, 120]) }` (decimal), in the genesis and curl it's `cHJlZml4` (base64), ascii is `"prefix"`

adizere:
> I hate to be "that guy" but cherry-picking the proto files for each use-case will become a hassle very quickly. I tried importing them all to see the interconnectedness and in my experience, it's worth keeping proto definitions together.
> 
> See the ibc-proto crate for the results or https://github.com/informalsystems/ibc-proto for the source code.
> 
> I suggest we use that crate or at least import its findings for a more complete proto experience.

Module `ibc_proto::ibc` is private [E0603] :( can we have a quick fix for this @greg-szabo ?

greg-szabo:
Why would you need `ibc_proto::ibc`? It's redundant.
`use ibc_proto::connection` or similar works. Check the documentation: https://docs.rs/ibc-proto/0.1.0/ibc_proto/

modules/src/ics03_connection/connection.rs
greg-szabo: I'd like to make this into a From trait, if possible.

ebuchman: Also I guess this should be verifying all the fields, not just 2? And should we really have an error per field or maybe start with a more general InvalidConnectionEnd error of some kind and include what's missing as a string? Not sure ...

modules/src/ics03_connection/exported.rs
ancazamfir: Looks good.

modules/src/ics03_connection/connection.rs
adizere: > I'd like to make this into a From trait, if possible.

Either option could work, the question is if we want to keep validation inside the `from` or outside. The current code uses the `FromStr` trait, with the signature `fn from_str(s: &str) -> Result<Self, Self::Err>` so we can do validation inside there and return error if validation fails. With `From`, the sig is `fn from(T) -> Self` so we'd have to do validation after calling `from`. Is it generally speaking, or specifically in this case, better to do validation outside of `from`?

> Also I guess this should be verifying all the fields, not just 2?

Agree this deserves a discussion to find a more principled approach.

modules/src/ics23_commitment/mod.rs
andynog: Implemented a fmt::Debug trait to properly format the CommitmentPrefix output in this [commit](https://github.com/informalsystems/ibc-rs/pull/136/commits/f8c0721250f770bc5bdd67b8cdd6ea92532369c6)

Issue #122, opened 2020-06-30
Relayer Message Builder logic


ancazamfir:
## Summary
Describe and implement the message builder algorithm and FSM

## Problem Definition
The Message Builder is a Relayer module that contains most of the logic of how IBC datagrams are built. For an event based model, this should be implemented as an FSM. 
There should be clearly defined events and actions (routines). 
An analysis of the action to be taken for each (state, event) combination should be done. Given ~15 events and ~10 states, 
 this may seem like an overwhelming task. However, many combinations can be ignored (e.g. all packet events for a connection builder FSM). 
For a builder of message related to some object X, only the IBC events related to X and its flipped version (if one exists) and to the client(s) involved should be considered. 
It will probably be the event handler that decides to remove an existing message builder when a "superior" event occurs for same X (e.g. a builder is in progress for `ConnOpenTry` to B message and a notification for `ConnOpenTry` from chain B is received for the same object)

Consideration should be given to having an FSM framework that presents events as generic or trait objects, with concrete implementations limited to object specific details.  In the same time, it should be implemented such that it is very clear what is the action taken in a given state and for a given event.

## Proposal
Here is one draft proposal, final decision and implementation may look differently. Only main "success" path events and actions are shown. Shown in italics are states valid for builders of messages `connOpenTry` and `connOpenAck` only. 
![message_builder (1)](https://user-images.githubusercontent.com/6361427/86183896-4e52d900-bb33-11ea-8dfd-7dc6f2ed7cb5.png)

Details should be included in an ADR and also documented in the implementation.
(lower level issues to be raised soon)

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

ebuchman:
One way of thinking about the relayer might be that, for each chain it's connected to, it synchronizes the chain's entire IBC store (ie. the whole `store/ibc/key` namespace).  It could literally maintain a local copy of the whole ibc kv-store.  This could be done using events+querying, but it could also be done instead by processing the IBC transaction messages directly, which would effectively amount to an implementation of a good chunk of IBC handler logic. (And bonus, if it used an IAVL tree for the local kv store, it could avoid having to query for merkle proofs all together, since it could do them itself).

[...]

PR #137, opened 2020-07-09
FSM English description of connection handshake (ICS3)

adizere:
Closes (eventually): #122

This is just the FSM for the connection handshake (the relayer's perspective). Other FSM are needed (channel, packet, client).

______

For contributor use:

- [ ] Unit tests written


docs/architecture/fsm-1.md
```diff
@@ -0,0 +1,72 @@
...
+
+
+## FSM transition table:
+
+
+| State|           Input        | Next State |  Output  |
+|:----:|:----------------------:|:----------:|:--------:|
+| Z    | OpenInit \| OpenTry    |   Q[d]     | Query destination chain headers via relayer client |
+| Z    | OpenAck \| OpenConfirm |      W     | _ |
+| Q[d] | RelayerClientResponse  |   U[s]     | UpdateClient on source chain |
```
ancazamfir, line 47: Depending on the LC response, next state here could also be `W` if the source client doesn't need to be updated.

docs/architecture/fsm-1.md
```diff
@@ -0,0 +1,72 @@
...
+| U[s] | UpdateClient           |   Q[s]     | _ |
+| W    | NewBlock               |   Q[s]     | Query source chain headers via relayer client |
+| Q[s] | RelayerClientResponse  |   U[d]     | UpdateClient on destination chain |
```
ancazamfir, line 50: Depending on the LC response, next state here could also be W if the destination client doesn't need to be updated.

docs/architecture/fsm-1.md
```diff
@@ -0,0 +1,72 @@
...
+- __PC__ = creating the connection state proof
+
+
+## FSM transition table:
+
+
+| State|           Input        | Next State |  Output  |
+|:----:|:----------------------:|:----------:|:--------:|
+| Z    | OpenInit \| OpenTry    |   Q[d]     | Query destination chain headers via relayer client |
+| Z    | OpenAck \| OpenConfirm |      W     | _ |
```
ancazamfir, line 46: Next state on `OpenConfirm` should be X.

docs/architecture/fsm-1.md
```diff
@@ -0,0 +1,72 @@
...
+| Q[d] | RelayerClientResponse  |   U[s]     | UpdateClient on source chain |
+| U[s] | UpdateClient           |   Q[s]     | _ |
+| W    | NewBlock               |   Q[s]     | Query source chain headers via relayer client |
+| Q[s] | RelayerClientResponse  |   U[d]     | UpdateClient on destination chain |
+| U[d] | UpdateClient           |    P       | [internal] Create proofs |
+| P    | OpenInit \| OpenTry    |   PCC      | Query the IBC client state and the connection state on source chain |
```
ancazamfir, line 52: I am not sure we need `P`. Also the FSM should be created with a connection event that dictates what message the FSM is building. In principle the FSM should not "see" events for  same or other connections. This should be dealt with by the entity that manages the message builder FSMs

docs/architecture/fsm-1.md
```diff
@@ -0,0 +1,72 @@
...
+- Height of source chain
+- Height of destination chain
+- Height of IBC client on source chain
+- Height of IBC client on destination chain
+- RPC to destination chain (for state __U[d]__)
+- RPC to source chain (for state __U[s]__)
+
+Limitations to this description:
+
+- this is a sequential description
```
ancazamfir, line 71: What does sequential mean?

docs/architecture/fsm-1.md
adizere: The same can be the case for the IBC client on the destination chain? Or is it the case that the IBC client on the destination chain pretty much _always_ has to be updated?

docs/architecture/fsm-1.md
adizere: I agree, the internal event is awkward and only necessary because I made the FSM memory-free. If the FSM had memory, then it would know which event "dictates" the state transitions, and we can also get rid of the internal event `P`.

docs/architecture/fsm-1.md
adizere: Sequential, as opposed to concurrent. What I had in mind here was that certain tasks could be done in parallel. But for the level of detail that we have in this description I can see how this can be confusing.

docs/architecture/fsm-1.md
```diff
@@ -0,0 +1,78 @@
+# Connection Handshake FSM
+
+
+## Input Events:
+
```
ancazamfir, line 5: We should fill this in. The events into the FSM that are relevant are:
- the chain event that creates the FSM (`openInit`, `openTry` or `openAck`)
- response to LC query
- response to connection query
- other chain events:
  - new block
  - on-chain client updated

All the others will be dealt with/ filtered by the event handler

docs/architecture/fsm-1.md
```diff
@@ -0,0 +1,78 @@
...
+| W    | NewBlock               |   Q[s]     | Query source chain headers via relayer client |
+| Q[s] | RelayerClientResponse(y) |   U[d]     | Submit UpdateClient on destination chain |
+| Q[s] | RelayerClientResponse(n) |   P     | [internal] Create proofs |
+| U[d] | UpdateClient           |    P       | [internal] Create proofs |
+| P    | OpenInit \| OpenTry    |   PCC      | Query the IBC client state and the connection state on source chain |
+| P    | OpenAck     |    PC      | Query the connection state on source chain |
+| P    | OpenConfirm     |    X      | FSM Exit |
+| PC   | RelayerClientResponse(n) |    S      | Submit datagram + proofs to destination chain |
+| PCC  | RelayerClientResponse(n) |    S      | Submit datagram + proofs to destination chain |
+|  S   | NewBlock               |   X       | FSM Exit |
```
ancazamfir, line 62: Why exit on new block?

ancazamfir, line 62: I think we need the connection query response events.

adizere: This NewBlock event signals that the datagram and proof (submitted in state __S__) arrived and processes at the destination chain. Is there further processing we'd like to do as part of the FSM?

adizere: The connection query response events should occur as part of state __PC__ (Query the IBC client state and the connection state on source chain), no?

I think there was an error in my understanding. The connection query response is encoded as RelayerClientResponse (line 60, 61) but it should be different than that. Maybe we can call it QueryResponse.

PR #139, opened 2020-07-09
Complete example for compiling and decoding ConnectionEnd with Prost.

adizere:
Closes: #130 

## Description

______

For contributor use:

- [ ] Unit tests written

### DIFF
--- a/modules/src/ics03_connection/query.rs
+++ b/modules/src/ics03_connection/query.rs
@@ -1,6 +1,7 @@
 use tendermint_rpc::endpoint::abci_query::AbciQuery;
 
 use tendermint::abci;
+use prost::Message;
 
 use crate::ics23_commitment::{CommitmentPath, CommitmentProof};
 use crate::ics24_host::identifier::ConnectionId;
@@ -10,6 +11,11 @@ use crate::ics03_connection::connection::ConnectionEnd;
 use crate::path::{ConnectionPath, Path};
 use crate::query::{IbcQuery, IbcResponse};
 use crate::Height;
+use tracing::info;
+
+pub mod items {
+    include!(concat!(env!("OUT_DIR"), "/items.rs"));
+}
 
 pub struct QueryConnection {
     pub chain_height: Height,
@@ -76,17 +82,31 @@ impl ConnectionResponse {
 
 impl IbcResponse<QueryConnection> for ConnectionResponse {
     fn from_abci_response(
-        query: QueryConnection,
-        response: AbciQuery,
+        _query: QueryConnection,
+        _response: AbciQuery,
     ) -> Result<Self, error::Error> {
-        let connection = amino_unmarshal_binary_length_prefixed(&response.value)?;
+        use bytes::Bytes;
 
-        Ok(ConnectionResponse::new(
-            query.connection_id,
-            connection,
-            response.proof,
-            response.height.into(),
-        ))
+        let buf = Bytes::from(_response.value);
+
+        let conn_end:items::ConnectionEnd = items::ConnectionEnd::decode(buf).unwrap();
+
+        info!("the connection: {:?}", conn_end);
+        // info!("the ACTUAL connection: {:?}", std::str::from_utf8(&*_response.value));
+
+        // @ADI: That's the spot.
+        // let connection = proto_unmarshal(&response.value)?;
+
+        // debug!("done with 'query connection end' command: {:?}", connection);
+
+        todo!()
+
+        // Ok(ConnectionResponse::new(
+        //     query.connection_id,
+        //     connection,
+        //     response.proof,
+        //     response.height.into(),
+        // ))
     }
 }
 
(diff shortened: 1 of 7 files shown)


modules/src/ics03_connection/query.rs
```diff
@@ -76,17 +82,31 @@ impl ConnectionResponse {
...
-            connection,
-            response.proof,
-            response.height.into(),
-        ))
+        use bytes::Bytes;
+
+        let buf = Bytes::from(_response.value);
+
+        let conn_end:items::ConnectionEnd = items::ConnectionEnd::decode(buf).unwrap();
+
```
adizere, line 93: @andynog This is perhaps the most important part for decoding.

adizere:
One major part we should decide is how to handle the `gogoproto` annotations that are Go-specific.

The .proto file in this commit is as follows (everything that starts with `gogo` is irrelevant for Rust code):


```proto
syntax = "proto3";

package items;

import "proto/gogo.proto";
//import "ibc/commitment/commitment.proto";

// ConnectionEnd defines a stateful object on a chain connected to another separate
// one.
// NOTE: there must only be 2 defined ConnectionEnds to establish a connection
// between two chains.
message ConnectionEnd {
  option (gogoproto.goproto_getters) = false;
  // connection identifier.
  string id = 1 [(gogoproto.customname) = "ID", (gogoproto.moretags) = "yaml:\"id\""];
  // client associated with this connection.
  string client_id = 2 [(gogoproto.customname) = "ClientID", (gogoproto.moretags) = "yaml:\"client_id\""];
  // opaque string which can be utilised to determine encodings or protocols for
  // channels or packets utilising this connection
  repeated string versions = 3;
  // current state of the connection end.
  State state = 4;
  // counterparty chain associated with this corparty = 5 [(gogoprotonnection.
//  Counterparty counte.nullable) = false];
}

// State defines if a connection is in one of the following states:
// INIT, TRYOPEN, OPEN or UNINITIALIZED.
enum State {
  option (gogoproto.goproto_enum_prefix) = false;

  // Default State
  STATE_UNINITIALIZED_UNSPECIFIED = 0 [(gogoproto.enumvalue_customname) = "UNINITIALIZED"];
  // A connection end has just started the opening handshake.
  STATE_INIT = 1 [(gogoproto.enumvalue_customname) = "INIT"];
  // A connection end has acknowledged the handshake step on the counterparty chain.
  STATE_TRYOPEN = 2 [(gogoproto.enumvalue_customname) = "TRYOPEN"];
  // A connection end has completed the handshake.
  STATE_OPEN = 3 [(gogoproto.enumvalue_customname) = "OPEN"];
}
```

The resulting .rs file is as follows:
```rust
/// ConnectionEnd defines a stateful object on a chain connected to another separate
/// one.
/// NOTE: there must only be 2 defined ConnectionEnds to establish a connection
/// between two chains.
#[derive(Clone, PartialEq, ::prost::Message)]
pub struct ConnectionEnd {
    /// connection identifier.
    #[prost(string, tag="1")]
    pub id: std::string::String,
    /// client associated with this connection.
    #[prost(string, tag="2")]
    pub client_id: std::string::String,
    /// opaque string which can be utilised to determine encodings or protocols for
    /// channels or packets utilising this connection
    #[prost(string, repeated, tag="3")]
    pub versions: ::std::vec::Vec<std::string::String>,
    /// current state of the connection end.
    ///
    ///  // counterparty chain associated with this connection.
    ///  Counterparty counterparty = 5 [(gogoproto.nullable) = false];
    #[prost(enumeration="State", tag="4")]
    pub state: i32,
}
/// State defines if a connection is in one of the following states:
/// INIT, TRYOPEN, OPEN or UNINITIALIZED.
#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, ::prost::Enumeration)]
#[repr(i32)]
pub enum State {
    /// Default State
    UninitializedUnspecified = 0,
    /// A connection end has just started the opening handshake.
    Init = 1,
    /// A connection end has acknowledged the handshake step on the counterparty chain.
    Tryopen = 2,
    /// A connection end has completed the handshake.
    Open = 3,
}
```

TODOs: 

- if we want to keep the `.proto` files in this repo:
   - figure out if we can compile without the `import gogo...`
   - figure out if all the `gogo` annotations can go away, basically trim the .proto file to its bare minimum
- preferable alternative: use in Rust the compiled version directly (and forget about the `.proto`)

modules/src/lib.rs
```diff
@@ -6,7 +6,7 @@
     trivial_casts,
     trivial_numeric_casts,
     unused_import_braces,
-    unused_qualifications,
+    // unused_qualifications,
```
romac, line 9: What's the reason for commenting this out?

adizere: It has to do with the generated Rust code which `prost` generated, for instance: `pub versions: ::std::vec::Vec<std::string::String>`. This seems to be the offending part that triggers the unused_qualifications warning, but I have not digged into it.

Any thoughts or prior experience with unused_qualifications ? I almost forgot about this. We should probably fix & enable the warning again it before merge.

Issue #32, opened 2020-03-17
Infrastructure to support running integration tests


romac:
## Summary

Implement integration tests for IBC queries.

## Problem Definition

As we implement more and more IBC queries, we need a way of making sure they are specified and executed properly, to avoid painstaking manual reviews like @ancazamfir did for #30.

## Proposal

This is just the initial issue, for kick-starting integration tests.

- Write integration tests for at least one IBC query or tx
- Set up the CI infrastructure need to run such a test automatically
- Allow specifying release (e.g. stargate-4) related to #204 
____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

andynog:
Working on creating images to support integration test on CI/CD and local machines (devs)

PR #133, opened 2020-07-08
add chain data

ebuchman:
## Description

Addresses https://github.com/informalsystems/ibc-rs/issues/129

This adds chain data for running an IBC enabled chain with some initial IBC state that can be queried.

Data was generated from https://github.com/cosmos/cosmos-sdk/pull/6637

See the included README here for how to use it.

Big thanks to @greg-szabo for working through this with me :)

______

For contributor use:

- [ ] Unit tests written
- [ ] Added test to CI if applicable 
- [ ] Updated CHANGELOG_PENDING.md
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Updated relevant documentation (`docs/`) and code comments
- [ ] Re-reviewed `Files changed` in the Github PR explorer

PR #140, opened 2020-07-10
integration tools

greg-szabo:
~~Closes: #32~~ Well, it gives us a framework but we should keep #32 open for the actual integration tests that we still have to implement.

## Description
This is a set of scripts and GitHub Actions that implement a framework for integration testing for IBC and the relayer.

- Script to create gaiad nodes docker image for the "two-chainz" relayer
- Script to create a simd node docker image with existing IBC state on chain
- Genesis.json (and simd configuration) with initial IBC state filled (courtesy of @ebuchman in #133, expanded with more state)

There is a new "integration-test" GitHub Action that sets up two chains and the relayer from source and tries to query things. Currently, it's failing because we're still implementing queries. This is fine, as soon as we have something to query, we can improve it. Or we can disable it for now...

The docker images are created and uploaded from dev machines, they are not automated. The images are used for development work and they change rapidly as we implement new features and need new tests... But the scripts are important to share, so we can easily create/recreate Docker images with the fresh code.

______

For contributor use:

- [ ] Unit tests written


ci/simapp/README.md
```diff
@@ -0,0 +1,34 @@
...
+This will start an IBC enabled blockchain with some initial IBC state that you
+can then query. For instance:
+
+```
+curl 'localhost:26657/abci_query?path="store/ibc/key"&data="connections/connectionidone"'
+curl 'localhost:26657/abci_query?path="store/ibc/key"&data="clients/ethbridge"'
+curl 'localhost:26657/abci_query?path="store/ibc/key"&data="channels/firstchannel"'
+```
+
+will return the relevent proto encoded connection data.
```
ebuchman, line 31:
```suggestion
will return the relevent proto encoded data.
```

Issue #126, opened 2020-07-03
Specify channel closing handshake in TLA+


istoilkovska:
## Summary

Specify creation and handling of channel closing datagrams in the relayer TLA+ specification.

## Proposal

Add support for channel closing datagrams.

On the chain side:
 - add handlers for processing channel closing handshake datagrams

On the relayer side:
 - define operator for creating  channel closing handshake datagrams

Properties: 
 - ensure that closed channels cannot be reopened

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

adizere:
Please don't forget to cleanup the relayer spec by deleting the [relayer-old](https://github.com/informalsystems/ibc-rs/tree/master/docs/spec/relayer-old) directory, according to this [comment](https://github.com/informalsystems/ibc-rs/pull/117#discussion_r448324773).

PR #141, opened 2020-07-10
Channel closing handshake

istoilkovska:
Closes: #126 

## Description

A small PR that deals with creation and handling of channel closing datagrams in the relayer TLA+ specification.

Also, relayer-old is deleted.

## Key points 

1. the environment non-deterministically sets two Boolean flags, `closeChannelA` and `closeChannelB`. 
2. a relayer reads the value of these flags, and sends `ChanCloseInit` to the appropirate chain once the flag is true.
3. a relayer reads that a chain set its channel end state to `CLOSED` and that its counterparty channel end is not closed yet, and sends `ChanCloseConfirm` to the its counterparty chain
4. a chain handles `ChanCloseInit` and `ChanCloseConfirm` by setting its channel end state to `CLOSED`

## Properties
- `ChannelCloseSafety` ensures that once a channel is closed, it is never open again.
- `ChanCloseInitDelivery` ensures that once `ChanCloseInit` is sent, eventually both channel ends are closed.

______

For contributor use:

- [ ] Unit tests written


docs/spec/relayer/ChannelHandlers.tla
```diff
@@ -140,7 +140,64 @@ HandleChanOpenConfirm(chainID, chain, datagrams) ==
     \* otherwise, do not update the chain     
     ELSE chain 
     
+\* Handle "ChanCloseInit" datagrams
```
adizere, line 143: The relayer algorithm for handling channel closing does not have an English specification in [ICS 018](https://github.com/cosmos/ics/tree/master/spec/ics-018-relayer-algorithms) (at least in the master branch). Any idea where is the English high-level spec for this algorithm? I wanted to cross-check it with your TLA+ code.

istoilkovska: You're right, it's missing in the ICS018, which is why channel closing wasn't added to the relayer spec until now. I've reverse engineered the datagram creation from the channel closure handlers.

docs/spec/relayer/ChannelHandlers.tla
```diff
@@ -140,7 +140,64 @@ HandleChanOpenConfirm(chainID, chain, datagrams) ==
...
+HandleChanCloseInit(chainID, chain, datagrams) ==
+    \* get "ChanCloseInit" datagrams, with a valid channel ID 
+    LET chanCloseInitDgrs == {dgr \in datagrams : 
+                              /\ dgr.type = "ChanCloseInit"
+                              /\ dgr.channelID = GetChannelID(chainID)} IN
+    
+    \* if there are valid "ChanCloseInit" datagrams
+    IF /\ chanCloseInitDgrs /= {} 
+    \* and the channel end is neither UNINIT nor CLOSED
+       /\ chain.connectionEnd.channelEnd.state \notin {"UNINIT", "CLOSED"}
```
milosevic, line 153: This is different from ICS04. I guess the reason is that they don't have UNINIT state?Is this the condition used in the Go implementation?

istoilkovska: Both the ICS004 and the Go implementation checks if the channel end can be found in the store, i.e. if it is non-null. In the TLA+ specification, each channel end is in UNINIT in the initial state of the system (c.f. [Chain.tla](https://github.com/informalsystems/ibc-rs/blob/master/docs/spec/relayer/Chain.tla#L110) and [RelayerDefinitions.tla](https://github.com/informalsystems/ibc-rs/blob/ilina/closing_handshake/docs/spec/relayer/RelayerDefinitions.tla#L160)), hence a channel (also connection) end in state UNINIT denotes a null channel (connection) end.

PR #162, opened 2020-07-21
ADR:  Domain Decomposition

brapse:
## Domain Decomposition
[to address #326]

The objective here is to outline the dependency graph for the relayer that will facilitate
* Reasonable error handling and failure cases
* Testability
* Correct by construction data types
  * Only produce verified data so downstream data consumers don't have to deal with proof

Still a work in progress, probably needs to be scoped down. But the ideas are ready for critique ☺️ 

[rendered](https://github.com/informalsystems/ibc-rs/blob/brapse/adr-domain-decomposition/docs/architecture/adr-003-domain-decomposition.md)

______

For contributor use:

- [ ] Unit tests written


docs/architecture/adr-003-domain-decomposition.md
```diff
@@ -0,0 +1,165 @@
...
+        // come in
+        return Subscription::new(self.clone(), ...);
+    }
+}
+```
+
+### Connection
+
+```rust
+trait Connection {
```
ebuchman, line 100: Should these really be traits? IBC objects like Connection, Channel, Packet are all concrete types that don't vary across chains. Clients are more of an interface and do vary depending on the chain, so we end up with concrete types like the Tendermint Client, Grandpa Client, Local Client, etc.

romac: Agreed, if those don't vary across chain then they should definitely be plain structs, and we leave it to the traits to abstract over operations, etc. rather than data.

brapse:
This is implemented, can we go ahead and merge this @ancazamfir?

ancazamfir (approved):
Looks good at this point. I think we should update this along with the implementation.

Issue #148, opened 2020-07-15
query channel given channel id and port id


ebuchman:
Following the success of #130, we should do the same thing for channels - given a channel id and a port id, fetch the ChannelEnd.

ABCIQuery.Path = `store/ibc/key`
ABCIQuery.Data = `channelEnds/ports/{port identifier}/channels/{channel identifier}`

See [ICS 24](https://github.com/cosmos/ics/tree/master/spec/ics-024-host-requirements#path-space).

Note this query requires both a channel id and port id.

adizere:
This issue is (partially?) covered in Greg's PR #152.

greg-szabo:
Yes, it's partially covered. As soon as #152 is merged, someone can start working on implementing a decent validation. The validation in #152 is only a "happy path" or "example" (no error handling) so the underlying code can work.

Issue #118, opened 2020-06-26
[modules] 'tests' folder should not be inside 'src'


adizere:
## Summary

For the sake of being consistent across this project (and consistent with `tendermint-rs`), we should not keep helper files for tests (e.g., JSON files) inside `src`.

## Problem Definition

In `relayer/cli`, and `relayer/relay` we find that `tests` folder is separated from the `src` folder. This separation does not hold for `modules` where `tests` are bundled [inside](https://github.com/informalsystems/ibc-rs/tree/master/modules/src/tests) `src`. 

## Proposal

Move the `tests` folder from `modules/src/tests` to `modules/tests` and update the tests accordingly.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #163, opened 2020-07-21
Channel query validation & tests

adizere:
Closes: #148
Closes: #118 

## Description

______

For contributor use:

- [ ] Unit tests written


modules/src/ics04_channel/channel.rs
```diff
@@ -22,30 +22,47 @@ pub struct ChannelEnd {
...
         let connection_hops = value
             .connection_hops
             .into_iter()
-            .map(|e| ConnectionId::from_str(e.as_str()).unwrap())
-            .collect();
+            .map(|conn_id| ConnectionId::from_str(conn_id.as_str()))
+            .collect::<Result<Vec<_>, _>>()
+            .map_err(|e| Kind::IdentifierError.context(e))?;
+
+        // No explicit validation is necessary for the version attribute (opaque field).
```
greg-szabo, line 54: Then what's the point of it? :)
Is it checked in some other context?

modules/src/ics07_tendermint/client_state.rs
```diff
@@ -103,14 +103,15 @@ mod tests {
 
     #[test]
     fn serialization_roundtrip_no_proof() {
-        let json_data = include_str!("../tests/query/serialization/client_state.json");
+        let json_data = include_str!("../../tests/support/query/serialization/client_state.json");
```
greg-szabo, line 106: Did this not work before or something change in the folder structure?

greg-szabo: Never mind, I found the change.

greg-szabo:
Ah, one more thing! If you merge master into your branch, you can get rid of the nightly coverage error. If others are picky about it, it's an easy solution.

modules/src/ics24_host/validate.rs
```diff
@@ -9,6 +9,7 @@ macro_rules! bail {
 
 /// Path separator (ie. forward slash '/')
 const PATH_SEPARATOR: char = '/';
+const VALID_SPECIAL_CHARS: &str = "._+-# []<>";
```
andynog, line 12: In the ics24 spec, it seems a 'space' is not a valid char. Does that space between '#' and '[' cause some issues ? For example a identifier like 'abc xyz' would be considered valid by the validate_identifier function.

modules/src/ics04_channel/exported.rs
```diff
@@ -40,6 +40,18 @@ impl State {
             Self::Closed => "CLOSED",
         }
     }
+
+    // Parses the State out from a i32.
+    pub fn from_i32(s: i32) -> Result<Self, error::Error> {
+        match s {
+            0 => Ok(Self::Uninitialized),
```
andynog, line 47: Just out of curiosity, is the Uninitialized state an implementation requirement ? Just asking because in the ics4 spec the Channel state only has 

`enum ChannelState {
  INIT,
  OPENTRY,
  OPEN,
  CLOSED,
}`

I know the Golang implementation has the uninitialized state too, so maybe the spec should be updated ?

modules/src/ics04_channel/channel.rs
adizere: The full details on the version field are here:
https://github.com/cosmos/ics/tree/master/spec/ics-004-channel-and-packet-semantics#versioning

The version field may be checked in other contexts (e.g., during a handshake, but as far as I can tell, it is not). But the point is that this field may contain fairly arbitrary data, according to the ICS4 spec. Looking through the cosmos-sdk codebase, however, I see that there is some basic validation on this field:

https://github.com/cosmos/cosmos-sdk/blob/e6a56226a88ec4e8761b1388a3a666af627d9e2e/x/ibc/04-channel/types/channel.go#L72

I'm adding a TODO to clarify this.

modules/src/ics24_host/validate.rs
adizere: Yup that blank space is an intruder. That's a very good catch!

modules/src/ics04_channel/exported.rs
adizere: I think this is an oversight in the cosmos/ICS repo and should be fixed in ICS4. I pinged Chris about it, will update here with details!

modules/src/ics04_channel/channel.rs
adizere: It seems that there might be a mistake in cosmos-sdk code. We should permit an empty version string for ChannelEnd, so basically no validation is required here.

https://github.com/cosmos/cosmos-sdk/issues/6830

modules/src/ics04_channel/exported.rs
adizere: Opened an issue in https://github.com/cosmos/ics/issues/452.

Issue #151, opened 2020-07-15
query connections given client id


ebuchman:
Given a client id, query the list of connection identifiers.

This is an easy one that just returns a list of connection identifiers.

ABCIQuery.Path = `store/ibc/key`
ABCIQuery.Data = `clients/{identifier}/connections`

greg-szabo:
#152 changes a lot of the underlying code and gives it the basics but validation is incomplete. There is a todo in the code to finish that.

andynog:
Implemented a good chunk of the logic to query client connections. But got stuck trying to figure out how to parse the result returned by the simapp. If I run simapp (docker version) and do the following command:

`curl 'localhost:26657/abci_query?path="store/ibc/key"&data="clients/clientidone/connections"'`

I get this result:

```
{
  "jsonrpc": "2.0",
  "id": -1,
  "result": {
    "response": {
      "code": 0,
      "log": "",
      "info": "",
      "index": "0",
      "key": "Y2xpZW50cy9jbGllbnRpZG9uZS9jb25uZWN0aW9ucw==",
      "value": "Chtjb25uZWN0aW9ucy9jb25uZWN0aW9uaWRvbmU=",
      "proof": null,
      "height": "1177",
      "codespace": ""
    }
  }
}
```
and the value decoded is `connections/connectionidone` (the connection identifier). So following the pattern we have to convert the result, I need to create an array of this connections as the response. But in my code I currently have

`let res: Result<ConnectionEnd, Error> = block_on(query(&chain, opts));`

where it should be more like something

`let res: Result<Vec<ConnectionId>, Error> = block_on(query(&chain, opts));`

but then had some issues trying to implement the TryFromRaw trait for this. Need to put a bit more thinking on it. The current implementation compiles but if you run:

```
./target/debug/relayer -c ./relayer/relay/tests/config/fixtures/relayer_conf_example.toml query client connections chain_A ethbridge -h 4 -p false
```

you get an empty response value

```
 Options QueryClientConnectionsOptions { client_id: ClientId("ethbridge"), height: 4, proof: false }
query client connections error Empty response value
```

andynog:
Found a way to parse the client connections response. If you use the command below (with docker simapp running):

``` 
./target/debug/relayer -c ./relayer/relay/tests/config/fixtures/relayer_conf_example.toml query client connections chain_A clientidone -h 4 -p false
```

It shows the client connections:

```
[...]

PR #169, opened 2020-07-27
Implement query for client connections

andynog:
Closes: #151 

## Description

Implemented query for client connections. Response is parsed as array of strings (which is the ABCI query result).
______

For contributor use:

- [X] Unit tests written
- [X] Re-reviewed `Files changed` in the Github PR explorer


modules/src/ics02_client/query.rs
```diff
@@ -3,10 +3,34 @@ use std::marker::PhantomData;
...
 
 //use crate::ics02_client::state::{ClientState, ConsensusState};
+use crate::ics03_connection::error::Kind;
 use crate::ics24_host::identifier::ClientId;
 use crate::path::{ClientStatePath, ConsensusStatePath};
+use crate::try_from_raw::TryFromRaw;
 use crate::Height;
 
+//TODO: This might need to be migrated to ibc-proto crate. But ClientConnections (as array of strings)
+// might not be part of an official proto file
```
greg-szabo, line 13: I forgot about this comment here. So, currently, it seems that the ClientConnections are not created as proto files because they are not in the provable store. This might change but until then we should keep this implementation here. The ibc-proto crate is auto-generated while this needs manual attention if something changes in the spec.

Homework: I believe that Ethan opened an issue about proto-ing this in the Cosmos SDK repo. Find the issue and replace this comment with a link to the issue. Alternatively, if you can't find the issue, open a new one. :)

modules/src/ics02_client/query.rs
```diff
@@ -3,10 +3,34 @@ use std::marker::PhantomData;
...
+
+impl TryFromRaw for Vec<String> {
+    type RawType = RawClientConnections;
+    type Error = anomaly::Error<Kind>;
+    fn try_from(value: RawClientConnections) -> Result<Self, Self::Error> {
+        let mut client_connections: Vec<String> = vec![];
+        for conn in value.connections {
+            client_connections.push(conn);
+        }
+        Ok(client_connections)
```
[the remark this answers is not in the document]
romac: That should work, yes!

andynog:
> Ok, this might be a very high level feedback, but running the command gives me an empty value. (I'm running simd in docker on the side.)
> 
> ```
> cargo run --bin relayer -- -c relayer/relay/tests/config/fixtures/simple_config.toml query client connections ibc-test ethbridge --height 3 -p false
> ```
> 
> ```
> query client connections error Empty response value
> ```

Using the simapp version in this branch, the command I've used to get the client connections is:

```
cargo run --bin relayer -- -c ./relayer/relay/tests/config/fixtures/relayer_conf_example.toml query client connections chain_A clientidone -h 4 -p false
```

the chain_id used is chain_A and client_id clientidone.

greg-szabo:
Ah cool, thanks! I was a bit lost there. Can you update the comment in the code to reflect that?
I know it's hairsplitting but I'm going to use that comment to write our first tests for testing queries and it helps me tremendously if the comment is correct.

modules/src/ics02_client/query.rs
```diff
@@ -3,10 +3,34 @@ use std::marker::PhantomData;
...
+        // If no connections return error
```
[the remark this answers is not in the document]
adizere: There's also more commented lines above (the `//use`) and in other files, and would be nice to clean them up.

relayer/cli/src/commands/query/client.rs
```diff
@@ -8,11 +8,20 @@ use relayer::config::{ChainConfig, Config};
...
 use relayer::chain::tendermint::TendermintChain;
+use relayer::query::{query, Request};
+use relayer_modules::error::Error;
 use relayer_modules::ics24_host::error::ValidationError;
+use relayer_modules::path::{ClientConnectionsPath, Path};
+use std::str::FromStr;
+use tendermint::abci::Path as TendermintPath;
 use tendermint::chain::Id as ChainId;
 
+// *****************************************************************
```
adizere, line 21: Following our last call, was wondering if we can have more idiomatic comments. I like the stars `**` here and how well they stand out (we used them in TLA+ specs also), but not sure if this is typical for Rust codebases. Perhaps we can get inspired from:

https://doc.rust-lang.org/rust-by-example/meta/doc.html
https://github.com/informalsystems/tendermint-rs/blob/master/light-client/src/fork_detector.rs

modules/src/ics02_client/query.rs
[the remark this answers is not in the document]
andynog: I can do some clean up but for the integration testing is this something part of you're going to work on @greg-szabo ?

modules/src/ics02_client/query.rs
andynog: Are you talking about this [issue](https://github.com/cosmos/cosmos-sdk/issues/6736) ? I think this is for ClientState. Couldn't find anything for client connections.

modules/src/ics02_client/query.rs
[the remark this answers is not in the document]
andynog: Fixed in 4c5523f21c6f228df2fc1a06635a5a144f3f8723

relayer/cli/src/commands/query/client.rs
andynog: Sure, should have used the '///' comment because that will help with docs. I'll fix that.

andynog: Fixed in b4dfcf3cd8466ed4652a6ca9af88e200de3e2533

andynog:
> If we supply a height that is too advanced, the reported error is a bit misleading saying something along the lines of `proof is unexpectedly empty; ensure height has not been pruned:`. See below:
> 
> ```
>      Running `target/debug/relayer -c ./relayer/relay/tests/config/fixtures/relayer_conf_example.toml query client connections chain_A clientidone -h 1000`
>      Options QueryClientConnectionsOptions { client_id: ClientId("clientidone"), height: 1000, proof: true }
> query client connections error RPC error: proof is unexpectedly empty; ensure height has not been pruned: invalid request
> ```

Actually noticed this on another issue and provided [comments](https://github.com/informalsystems/ibc-rs/issues/130#issuecomment-657788967). I think someone mentioned that has to do with pruning I believe but can't find that thread.

modules/src/ics02_client/query.rs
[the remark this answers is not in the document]
andynog: Fixed in 2c4771be2d54d7f57783fddb4953452f195dd24d

modules/src/ics02_client/query.rs
greg-szabo: Yes, that's the one I was thinking about! But you're right, that's slightly similar but not exactly the same.

All in all we're not going to move this into ibc-proto but instead we'll wait for the SDK team to create proto files for it. (Based on the issue, currently, it's pending on Tendermint Go.) If you could update the comment to reflect that, then we can leave it here until that time comes.

greg-szabo:
Hint: the file `modules/src/ics02_client/query.rs` should completely disappear after this PR. Ping me if you need any pointers on how to achieve that.

modules/src/ics02_client/query.rs
andynog: Ok, agree once Tendermint has proto types for this we can refactor and move the code to ibc-proto. So this is good for now.

adizere:
> Actually noticed this on another issue and provided comments. I think someone mentioned that has to do with pruning I believe but can't find that thread.

I'm not sure this is about pruning. Pruning would interfere with a query if the supplied height `-h` would be very old (which was pruned). But in this case I was supplying a height (1000) that does not exist yet (it's a height in the future). At the least, I think the error message is not right.

> query client connections error RPC error: proof is unexpectedly empty; ensure height has not been pruned: invalid request

But there might a bug in the validation/handling.

andynog:
> I just noticed that you implemented the TryFromRaw in the obsolete `query.rs`. Please clean up the unused code and move things around like you can see in the connection and channel code.

This is fixed in 1ca727d302e1eaa27aab5b44ca6cc21025116ee6

relayer/cli/src/commands/query/client.rs
```diff
@@ -2,17 +2,15 @@ use crate::prelude::*;
...
 use relayer::config::{ChainConfig, Config};
-//use relayer::query::client::{query_client_consensus_state, query_client_full_state};
-//use relayer::query::{query, Request};
 
-//use relayer_modules::ics02_client::query::QueryClientFullState;
-use relayer_modules::ics24_host::identifier::ClientId;
-
-//use crate::commands::utils::block_on;
-use relayer::chain::CosmosSDKChain;
+use relayer::chain::tendermint::TendermintChain;
```
greg-szabo, line 6:
```suggestion
use relayer::chain::CosmosSDKChain;
```
as of #176

relayer/cli/src/commands/query/client.rs
```diff
@@ -214,9 +213,95 @@ fn validate_common_options(
...
+        let (chain_config, opts) = match self.validate_options(&config) {
+            Err(err) => {
+                status_err!("invalid options: {}", err);
+                return;
+            }
+            Ok(result) => result,
+        };
+        status_info!("Options", "{:?}", opts);
+
+        let chain = TendermintChain::from_config(chain_config).unwrap();
```
greg-szabo, line 291:
```suggestion
        let chain = CosmosSDKChain::from_config(chain_config).unwrap();
```

Issue #173, opened 2020-07-28
Why is this code checking the "Path" instead of the request for proof?


greg-szabo:
https://github.com/informalsystems/ibc-rs/blob/9545ccddfad78582532f1eb13f7ac201ccdac72f/relayer/relay/src/query.rs#L68-L72

This code survived a few refactors and we should implement it - or at least make it easier to understand.

As far as I understand, if the request came in with `prove=true`, it would be nice, if we checked that proof. But this code seems to suggest, that based on the Path there's an indication if the response should be checked.

I guess this must relate to the provableStore idea in the ICS. But as far as I'm aware, all request Paths are going to come from `store/ibc/key`. Or maybe I'm missing something.

Is it possible that the code is just slightly wrong and it should check the data field to see if the request contains something that's in the provableStore? (In that case I know how to deal with that.)

greg-szabo:
Here's my assumption for now:
- we're going to assume we query the IBC store
- a path is provable if it's in the provableStore, as indicated in ICS-024.

ebuchman:
Just a note.

>  // Data that is not from trusted node or subspace query needs verification. 

So obviously if we trust the node/ we set prove=false we don't verify.

The other type, `subspace query`, is when we query a whole subspace, which I mentioned here: https://github.com/informalsystems/ibc-rs/issues/128#issuecomment-659094702

ie. if we use a path like `store/ibc/subspace` we get all the keys prefixed with whatever we pass in data

I believe the idea re proofs here is that if you get a full subspace (ie. a complete merkle sub-tree), you don't need to worry about proofs, because you have all the data to construct them yourself. But you'd still need some proof to verify that the root of your subtree is in the larger tree ...

Issue #172, opened 2020-07-28
Simplify Path type


ebuchman:
I wonder if the Path trait and all those path structs in path.rs even buys us anything? I would think the only purpose of those structs is to call to_string() on them to get the actual path, so why bother with the constructing the structs at all when we could just call the path functions directly?

eg. https://github.com/informalsystems/ibc-rs/blob/13da6b7f26b31e591bf39cdd14423c802389d8a6/relayer/cli/src/commands/query/connection.rs#L45

we might as well just call `paths::connection_path(conn_id)` ...

_Originally posted by @ebuchman in https://github.com/informalsystems/ibc-rs/issues/170#issuecomment-664730005_

ebuchman:
Just a note that @romac raise the possibility that we may want Path to be a trait so we can add an Associated Type and ensure that only that type can be stored under that path, which may be nice. But we can come back to that in the future as work evolves on the store

Issue #157, opened 2020-07-18
Generalize client queries to enable multiple chain types


greg-szabo:
## Summary
This is a braindump from a meeting I had with Anca where she mentioned that other chain types (apart from the currently implemented TendermintChain) might connect through IBC and the relayer should be ready to implement those different chain queries without major rewrites.

Notes:
* The current query implementation takes the Chain type as an input and calls its generic rpc_client for queries. That's good. But, it goes one step further and within that rpc_client, it calls the Tendermint specific abci_query method. Maybe there should be a generic `query()` method in `Chain`?
* The input parameters for the current Tendermint Abci query are four: `path`, `data`, `height`, `prove`. According to Anca, only path is Tendermint specific, the other three are supposed to be universal across chains. This would allow us to introduce some kind of `chain_specific_options()` query in the `Chain` trait (which in Tendermint's case would return the `path`) and keep the other three as input parameters for the generic `query()`. Or, if the generic `query()` just calls into a `Chain.query()` then I guess it can mostly just move into that. Needs some thoughts around this.
* The current Chain trait and TendermintChain implementation has dependencies toward the light client, Tendermint, and the relayer. This is a bit overwhelming for a generic trait. Possibly, current work happening on the light client will allow us to get away from the light client dependency, leaving on the the specific chain (Tendermint) and the relayer as dependencies. I think this is a prerequisite to be able to write a decent generic trait for chains to implement to be able to use the relayer easily.

## Proposal
Anca says, we can open this up for discussion after some more work has been done on the light client.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

greg-szabo:
Note: https://github.com/informalsystems/ibc-rs/issues/128#issuecomment-663560636

PR #174, opened 2020-07-28
Refactored and completed Path; added query to Chain trait; refactored query code into TendermintChain

greg-szabo:
Closes: #157
Closes: #172
Closes: #173

## Description
Apologies for the monstre changes again, but the changes are unfortunately dependent on each other.

* Refactored Path into an enum, implemented all paths. Result is a smaller codebase.
* Paths now know if they are in the provableStore or not and a check was added to the query to not try to prove something in the privateStore.
* Implemented query in line with #162 : Chain trait and TendermintChain. This gives us (somewhat) a template on how to implement subscription and transactions. At least it helped me understand the necessities, so I can start working on those in the relayer.
* Deleted a bunch more obsolete code, where I could.

Tests are lacking, but maybe soon, in another issue. (Local execution worked.)
______

For contributor use:

- [ ] Unit tests written


relayer/relay/src/chain.rs
```diff
@@ -35,6 +38,19 @@ pub trait Chain {
     /// Type of RPC requester (wrapper around low-level RPC client) for this chain
     type Requester: tmlite::Requester<Self::Commit, Self::Header>;
 
+    /// Error types defined by this chain
+    type Error: Into<Box<dyn Error + Send + Sync + 'static>>;
+
+    /// Perform a generic `query`, and return the corresponding deserialized response data.
+    // This is going to be a blocking request.
```
romac, line 45: Not very important right now, but just pointing out that async methods are somewhat supported on stable Rust via the [`async-trait`](https://docs.rs/async-trait) crate. This crate sometimes yield weird compile-time errors and incurs a [boxing of the returned `Future`](https://docs.rs/async-trait/0.1.36/async_trait/#explanation) but has otherwise worked pretty well for us in the past.

modules/src/ics24_host/path.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    /// into_bytes implementation
+    pub fn into_bytes(self) -> Vec<u8> {
+        self.to_string().into_bytes()
+    }
+}
+
+/// The Display trait adds the `.to_string()` method to the Path struct
+/// This is where the different path strings are constructed
+impl Display for Path {
```
romac, line 45: We could perhaps consider using the [`derive_more`](https://docs.rs/derive_more) crate to make this more readable. Downside is that it pulls in quite a few dependencies, though they are compile-time only. This can be mitigated a bit by only using the [features](https://github.com/JelteF/derive_more/blob/ece3773ef924e5d2691a61797d4d40fdfbfff56b/Cargo.toml#L41-L65) we need.

relayer/relay/src/chain.rs
greg-szabo: Cool! That's the kind of thing I need. I'll take this info away and use it for implementations for subscription and/or transactions.
I want to stop "improving" queries and if this stays as a non-async implementation for ow, it still works out fine.
Thanks for the info and thanks for understanding!

modules/src/ics24_host/path.rs
greg-szabo: I'll look into this in a new issue. Compile-time dependencies are not necessarily good, because Ethan is already complaining about the compile time, but readability is always welcome!

modules/src/ics24_host/path.rs
```diff
@@ -0,0 +1,86 @@
...
+/// https://github.com/cosmos/ics/tree/master/spec/ics-024-host-requirements#path-space
+/// Some of these are implemented in other ICSs, but ICS-024 has a nice summary table.
+use crate::ics24_host::identifier::{ChannelId, ClientId, ConnectionId, PortId};
+use std::fmt::{Display, Formatter, Result};
+
+/// IBC Query Path is hard-coded
+pub const IBC_QUERY_PATH: &str = "store/ibc/key";
+
+/// The Path enum abstracts out the different sub-paths
+pub enum Path {
```
adizere, line 11: Oooo nice!

relayer/relay/src/error.rs
```diff
@@ -28,6 +28,17 @@ pub enum Kind {
...
     Store,
+
+    /// Event error (raised by the event monitor)
+    #[error("Bad Notification")]
+    Event,
+
+    /// Response parsing error
+    #[error("Could not parse/unmarshall response")]
+    ResponseParsing,
+
```
adizere, line 39: Nitpick but still important: all other error types (above `ResponseParsing`) have a comment. Can we have one here as well?

greg-szabo: Good find, fixed.

Issue #160, opened 2020-07-19
ICS03 - Connection handshake


ebuchman:
Re #119. We already have the ICS03 message types, so we should build out the logic for processing them - ie. the connection handshake.

I think this can start with an abstract KV-Store and Client, but it should ultimately leverage #158. 

We should be able to test it simply by running two instances of the connection handlers against each other, potentially even without the machinery of #158.

PR #188, opened 2020-07-31
WIP: ICS3 connection handshake message processing logic (core & context)

adizere:
Closes: #160 

## Description

______

For contributor use:

- [ ] Unit tests written


modules/src/ics03_connection/error.rs
```diff
@@ -10,6 +10,21 @@ pub enum Kind {
     #[error("connection state unknown")]
     UnknownState,
 
+    #[error("connection exists (was initialized) already")]
+    ConnectionExistsAlready,
```
romac, line 14: It would be useful to store some more metadata (like the identifier of the connection) in the errors themselves:

```suggestion
    #[error("connection exists (was initialized) already: {0}")]
    ConnectionExistsAlready(ClientId),
```

modules/src/ics03_connection/exported.rs
```diff
@@ -68,6 +46,27 @@ impl std::str::FromStr for State {
...
 }
 
+/// Function required by ICS 03. Returns the list of all possible versions that the connection
+/// handshake protocol supports.
+/// TODO: What are the precise values for the versions which this function returns? Perhaps encode the versions as constants.
+pub fn get_compatible_versions() -> Vec<String> {
+    let mut result = Vec::with_capacity(1);
+    result[0] = String::from("test");
+
+    result
```
romac, line 56:
```suggestion
    vec!["test".to_string()]
```

modules/src/ics03_connection/error.rs
adizere: Whenever I use `Kind::` syntax, I usually pass a context. For example:

```rust
Err(Kind::InvalidProof.context(String::from("consensus state proof is invalid"))
                        .into())
```

The alternative, based on your suggestion, would be to have:

```rust
Err(Kind::InvalidProof(String::from("consensus state proof is invalid"))
                        .into())
```
Would these be these the same thing? If so, not clear to me which is preferable or idiomatic.

modules/src/address.rs
```diff
@@ -0,0 +1 @@
+pub struct AccAddress(Vec<u8>;
```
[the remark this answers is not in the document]
ancazamfir: Yes, we need to make this change. Didn't do it here because it touches all messages. We don't do anything with this field in `Msg` in IBC modules. It is going to be handled up the stack once we have auth, accounts, etc. Can you open an issue for this?

modules/src/context.rs
```diff
@@ -0,0 +1,30 @@
+use crate::ics02_client::client_def::AnyConsensusState;
+use crate::mock_client::header::MockHeader;
+use crate::mock_client::state::MockConsensusState;
+use serde_derive::{Deserialize, Serialize};
+use tendermint::block::Height;
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub enum SelfHeader {
```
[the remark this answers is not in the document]
ancazamfir: Yes, correct.

modules/src/ics03_connection/error.rs
romac: The purpose of `.context` is to provide additional context as to why the error was raised at this specific place / in these specific circumstances, not to provide an error message or to attach metadata to the error. The error message is defined by the `Display` impl automatically derived by the `thiserror` crate via it's `#[display]` annotation, while metadata that is specific to the error itself, as opposed to the circumstances of where and why it was raised, should be embedded within the error itself, much like the `ClientId` above.

Does that make thing clearer?

modules/src/ics23_commitment/commitment.rs
```diff
@@ -0,0 +1,54 @@
...
+
+impl From<Vec<u8>> for CommitmentPrefix {
+    fn from(v: Vec<u8>) -> Self {
+        Self { 0: v }
+    }
+}
+
+impl fmt::Debug for CommitmentPrefix {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let converted = String::from_utf8(self.clone().0);
```
romac, line 48:
```suggestion
        let converted = str::from_utf8(&self.0);
```

romac (changes requested):

modules/src/ics02_client/client_def.rs
```diff
@@ -19,7 +23,7 @@ pub trait ClientDef: Clone {
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)] // TODO: Add Eq
 #[allow(clippy::large_enum_variant)]
 pub enum AnyHeader {
-    Mock(mocks::MockHeader),
+    Mock(MockHeader),
```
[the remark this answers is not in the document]
romac: Marked the whole module as `#[cfg(test)]` instead.

romac (approved):

Issue #185, opened 2020-07-31
Implement decoding into ConnectionId


greg-szabo:
https://github.com/informalsystems/ibc-rs/blob/9c733d7af1ee7c2dee07086c24f2a94f4d169825/modules/src/ics02_client/client.rs#L12

I think at the query of client Connection ID, we should decode the incoming connection details into a `Vec<ConnectionId>` instead of just keeping them as strings.

The reason is that here, we should use a ConnectionId type for easy handling of ConnectionIDs.

https://github.com/informalsystems/ibc-rs/blob/9c733d7af1ee7c2dee07086c24f2a94f4d169825/relayer-cli/src/commands/query/client.rs#L292-L294

PR #196, opened 2020-08-05
Replaced Vec<String> with Vec<ConnectionId>

andynog:
Closes: #185

## Description

Refactored the response for the query client connections to return a collection of ConnectionId. Better than just a string collection.
______

For contributor use:

- [ ] Unit tests written
- [X] Re-reviewed `Files changed` in the Github PR explorer

### DIFF
--- a/modules/src/ics02_client/client.rs
+++ b/modules/src/ics02_client/client.rs
@@ -1,5 +1,7 @@
 use crate::ics03_connection::error::Kind;
+use crate::ics24_host::identifier::ConnectionId;
 use crate::try_from_raw::TryFromRaw;
+use std::str::FromStr;
 
 //TODO: This might need to be migrated to ibc-proto crate. But ClientConnections (as array of strings)
 // might not be part of an official proto file
@@ -9,12 +11,20 @@ pub struct RawClientConnections {
     pub connections: ::std::vec::Vec<String>,
 }
 
-impl TryFromRaw for Vec<String> {
+impl TryFromRaw for Vec<ConnectionId> {
     type RawType = RawClientConnections;
     type Error = anomaly::Error<Kind>;
     fn try_from(value: RawClientConnections) -> Result<Self, Self::Error> {
         if !value.connections.is_empty() {
-            Ok(value.connections)
+            let mut connections: Vec<ConnectionId> = vec![];
+            for value in value.connections {
+                let conn_id = ConnectionId::from_str(&value.replace("connections/", ""));
+                match conn_id {
+                    Ok(c) => connections.push(c),
+                    Err(_e) => return Err(Kind::IdentifierError.into()),
+                }
+            }
+            Ok(connections)
         } else {
             Err(Kind::ConnectionNotFound.into())
         }
--- a/relayer-cli/src/commands/query/client.rs
+++ b/relayer-cli/src/commands/query/client.rs
@@ -4,7 +4,7 @@ use abscissa_core::{Command, Options, Runnable};
 use relayer::config::{ChainConfig, Config};
 
 use modules::ics24_host::error::ValidationError;
-use modules::ics24_host::identifier::ClientId;
+use modules::ics24_host::identifier::{ClientId, ConnectionId};
 use modules::ics24_host::Path::ClientConnections;
 use relayer::chain::Chain;
 use relayer::chain::CosmosSDKChain;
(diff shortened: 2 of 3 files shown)


modules/src/ics02_client/client.rs
```diff
@@ -9,12 +11,20 @@ pub struct RawClientConnections {
...
 }
 
-impl TryFromRaw for Vec<String> {
+impl TryFromRaw for Vec<ConnectionId> {
     type RawType = RawClientConnections;
     type Error = anomaly::Error<Kind>;
     fn try_from(value: RawClientConnections) -> Result<Self, Self::Error> {
         if !value.connections.is_empty() {
-            Ok(value.connections)
+            let mut connections: Vec<ConnectionId> = vec![];
```
brapse, line 19: Can this be done with a `.map`?

andynog: it could, but because this can return an error (in case a connection id is not valid, eg. longer than max characters) this is implementation is not too bad. I had the assumption that for some reason there are multiple connections and one of them is not valid then an error is returned right away. There might be patterns with map you can accomplish that with things like ok_or_else and fold but sometimes I think this is not as intuitive as some explicit code sometimes. If you think it's worth it I can refactor it.

andynog:
> Just found a small issue that's probably related to the recent changes.
> 

Good catch @adizere. Just fixed it.

relayer-cli/src/commands/query/client.rs
```diff
@@ -289,8 +289,11 @@ impl Runnable for QueryClientConnectionsCmd {
         status_info!("Options", "{:?}", opts);
 
         let chain = CosmosSDKChain::from_config(chain_config).unwrap();
-        let res =
-            chain.query::<Vec<String>>(ClientConnections(opts.client_id), opts.height, opts.proof);
+        let res = chain.query::<Vec<ConnectionId>>(
+            ClientConnections(opts.client_id),
+            opts.height,
+            opts.proof,
```
ancazamfir, line 295: Should remove `opts.proof`, this is querying the private store (see https://github.com/cosmos/ics/tree/master/spec/ics-024-host-requirements).
Have you tried to query with proof? What is the result?

relayer-cli/tests/integration.rs
```diff
@@ -105,12 +105,15 @@ fn query_channel_id() {
 fn query_client_id() {
```
ancazamfir, line 105: It would be nice to have a test/ simapp with two connections

modules/src/ics02_client/client.rs
andynog: I'll just resolve the conversation for now. If we think we should refactor we can open a separate issue for that.

relayer-cli/src/commands/query/client.rs
andynog: if you specify `-p true` you get a proper error saying it's a privateStore `client connections query error store error: requested proof for privateStore path` so I believe this is OK. I could hard code the value in the function call instead of using `opts.proof` as argument and set it to `false` but I think the way it is probably better.

relayer-cli/tests/integration.rs
andynog: I believe in order to have two connections some configuration should be changed in the simapp. I looked into that and couldn't figure out where to change it maybe need @greg-szabo help. I see that in /simapp/config/genesis.json the connections are configured there but not sure how is the configuration passed as an argument to relayer (-c relayer/test/config/fixtures/relayer_conf_example.toml) also has [[connections]] so not sure how these two play out.

relayer-cli/src/commands/query/client.rs
ancazamfir: From an UX experience I find it strange that we offer an option that we know it is not valid. My preference is to just remove it all together and hardcode `false`. Or at least change the default to `false`. I realize the client state falls under the same debate. Maybe the best thing is to open a separate issue and discuss there with the whole team.

relayer-cli/tests/integration.rs
ancazamfir: > not sure how is the configuration passed as an argument to relayer (-c relayer/test/config/fixtures/relayer_conf_example.toml) also has [[connections]] so not sure how these two play out.

For the queries the config should only be used to extract the rpc address. We should be able to query connections that are not in the relayer configuration.

andynog: Added separate issue #204 for this

relayer-cli/src/commands/query/client.rs
andynog: Open a separate issue to handle this discussion #205

andynog:
This should be good to merge now

Issue #194, opened 2020-08-04
ADR for Handlers


ebuchman:
We started making some progress towards the handlers (#119), ie. #188.

The architecture being pursued came out of some synchronous discussions and spikes but we should now document it in a formal ADR

PR #197, opened 2020-08-07
ADR 003: Handler implementation

romac:
Closes: #194

## Description

This PR introduces a initial version of ADR 003 which provides recommendations for the implementation of IBC handlers.

[Rendered](https://github.com/informalsystems/ibc-rs/blob/romain/ics02-client-handlers/docs/architecture/adr-003-handler-implementation.md)

______

For contributor use:

- [ ] Unit tests written


docs/architecture/003-handler-implementation.md
```diff
@@ -0,0 +1,552 @@
...
+
+## Context
+
+> This section contains all the context one needs to understand the current state, and why there is a problem. It should be as succinct as possible and introduce the high level idea behind the solution.
+
+TODO
+
+## Decision
+
+In this chapter, we provide recommendation for implementing IBC handlers within the `ibc-rs` crate.
```
adizere, line 14: ADR :P not chapter.

docs/architecture/003-handler-implementation.md
```diff
@@ -0,0 +1,552 @@
...
+    pub fn new(key: String, value: String) -> Self;
+}
+
+pub enum EventType {
+    Message,
+    Custom(String),
+}
+
+pub struct Event {
+    tpe: EventType,
```
adizere, line 42:
```suggestion
    type: EventType,
```

docs/architecture/003-handler-implementation.md
```diff
@@ -0,0 +1,552 @@
...
+Concepts are introduced in the order given by a topological sort of their dependencies on each other.
+
+### Events
+
+IBC handlers must be able to emit events which will then be broadcasted via the node's pub/sub mechanism,
+and eventually picked up by the IBC relayer.
+
+A generic interface for events is provided below, where an event is represented
+as a pair of an event type and a list of attributes. An attribute is simply a pair
+of a key and a value, both represented as strings. TODO: Describe event type.
```
adizere, line 24: Instead of this TODO here, maybe we can link to here for an overview (though they are not precisely the same)?

https://github.com/informalsystems/ibc-rs/blob/master/modules/src/events.rs#L15

docs/architecture/003-handler-implementation.md
```diff
@@ -0,0 +1,552 @@
...
+}
+
+impl Event {
+    pub fn new(tpe: EventType, attrs: Vec<(String, String)>) -> Self;
+}
+```
+
+### Logging
+
+IBC handlers must be able to log information for introspectability and ease of debugging.
```
adizere, line 53: Not sure if introspectability is right (does that mean self-examination?). Maybe we mean here to say that the handlers log information for exporting it in a simple/plain format?

docs/architecture/003-handler-implementation.md
```diff
@@ -0,0 +1,552 @@
+# ADR 003: Handler implementation
```
adizere, line 1: I propose we use the term "protocol" to replace "handlers".
The "message processing logic" for a given codebase is effectively implementing the "protocol".

The term "protocol" is still imperfect, but easier to talk about than "message processing logic" and less contentious than "handler". A protocol consists of processing functions `process_msg_x`, `process_msg_y` etc. so that any input message is processed by such a function.

romac:
This PR also provides an implementation of the "Create Client" handler together with unit tests for it.

docs/architecture/003-handler-implementation.md
romac: `type` is not a valid identifier. We could use [`r#type`](https://doc.rust-lang.org/edition-guide/rust-2018/module-system/raw-identifiers.html) but `tpe` looks better to me.

shonfeder: In OCamlland we often use `typ` for this. The advantage over `tpe` is in the phonetic consistency :)

romac: Oh I like ‘typ’ as well! Will use that then. In Stainless we used tpe so I just carried it over but typ looks better indeed :)

modules/src/ics02_client/handler/create_client.rs
```diff
@@ -0,0 +1,286 @@
...
+                self.consensus_state
+            } else {
+                None
+            }
+        }
+    }
+
+    #[test]
+    fn test_create_client_ok() {
+        let mock = MockClientContext {
```
ancazamfir, line 200: As shown below, this is missing the `client_id`. 
What is the expectation on the caller that needs to give a context to `process()` in addition to the message? Specifically for clients, if the client does not exist, the context should still include a valid `client_id` and this test passes as long as one is given. Outside tests, is the caller expected to check if this is a `MsgCreateClient` create a context with `None` for state and type if the client doesn't exist or return existing if it does? We briefly talked about this last week and this is still an aspect we should clarify. For connections the context/ keeper passed to `process()` needs both client and connection context, for channels we need client, connection and channel, etc. Things will become more complex when setting up all the mocks for tests.

But the bigger concern is that the IBC handler (as per ICS) logic is split somewhat between `handle()` and the caller.

romac: The expectations are captured by the `CientContext` trait:

```rust
pub trait ClientContext<CD>
where
    CD: ClientDef,
{
    fn client_type(&self, client_id: &ClientId) -> Option<ClientType>;
    fn client_state(&self, client_id: &ClientId) -> Option<CD::ClientState>;
    fn consensus_state(&self, client_id: &ClientId, height: Height) -> Option<CD::ConsensusState>;
}
```

The fact that the mock requires a client id is an implementation detail of this particular mock struct.
One could imagine implementing the `ClientContext` trait in a different way, but the way it's done here is imho pretty straightforward as one can exercise all code paths in the `process` function just by constructing different values of the mock struct.

> But the bigger concern is that the IBC handler (as per ICS) logic is split somewhat between handle() and the caller.

Could you expand on that, in the context of what I wrote above, if that still makes sense?

romac:
> My doubts are still around the context and keeper in general, thinking also about the context for connection, channel and packets.

Agreed, definitely something we need to take a closer look at, by eg. writing tests for the module-level handler/dispatcher.

modules/src/ics02_client/msgs.rs
```diff
@@ -1,25 +1,25 @@
...
-
-    fn client_id(&self) -> &ClientId;
-    fn client_type(&self) -> ClientType;
-    fn consensus_state(&self) -> Self::ConsensusState;
+/// A type of message that triggers the update of an on-chain (IBC) client with new headers.
+#[derive(Clone, Debug)]
+pub struct MsgUpdateClient<CD: ClientDef> {
+    pub client_id: ClientId,
+    pub header: CD::Header,
 }
```
ancazamfir, line 24: The structure of `MsgCreateClient` is different depending on the client type. Same for `MsgUpdateClient`. For example for `tendermint` client, the `MsgCreateClient` includes fields like `trusting_period`, `unbonding_period`, etc. While for `loopback` (not implemented) or other client types, these are not present while others might be. Not sure how we can implement the ics02 messages without traits.

ancazamfir: Actually, I think this will work if we create the concrete `ClientState`, `ConsensusState`, etc, from the wire/ raw message and then we call `dispatch` -> `process`. So some of the logic will be in this pre-processing step. I will give it a try.

docs/architecture/adr-003-handler-implementation.md
```diff
@@ -0,0 +1,679 @@
...
+        } else {
+            None
+        }
+    }
+}
+```
+
+#### Keeper
+
+Once a message processor executes successfully, some data will typically need to be persisted in the chain state
```
ebuchman, line 229: Should mention the need for atomicity here

docs/architecture/adr-003-handler-implementation.md
```diff
@@ -0,0 +1,679 @@
...
+    type ClientState = MockClientState;
+    type ConsensusState = MockConsensusState;
+}
+```
+
+Since the actual type of client can only be determined at runtime, we cannot encode
+the type of client within the message itself.
+
+Because of some limitations of the Rust type system, namely the lack of proper support
+for existential types, it is currently impossible to define `Reader` and `Keeper` traits
```
ebuchman, line 483: Can I read more about this somewhere?

Issue #193, opened 2020-08-03
Cleanup exported types in connection


adizere:
> The exported module brings up a related problem I believe. Parts of this module (and its intended functionality) may be unnecessary and we can remove it..

yes, we should remove the traits from the `exported.rs` files (in both connection and channel folders).

_Originally posted by @ancazamfir in https://github.com/informalsystems/ibc-rs/issues/186#issuecomment-667995580_

More specifically, we can remove the traits `Connection` and `ConnectionCounterparty`:

https://github.com/informalsystems/ibc-rs/blob/574371a280a282bdfaa1045ca8a6e628d612b79d/modules/src/ics03_connection/exported.rs#L6-L25

PR #203, opened 2020-08-12
Remove Connection and ConnectionCounterparty traits

ancazamfir:
Closes: #193

## Description
Remove `Connection` and `ConnectionCounterparty` traits, remove `exported.rs` file

______

For contributor use:

- [ ] Unit tests written


modules/src/ics03_connection/connection.rs
```diff
@@ -176,3 +165,103 @@ pub fn validate_version(version: String) -> Result<String, String> {
...
+            _ => Self::Uninitialized,
+        }
+    }
+}
+
+impl std::str::FromStr for State {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
```
andynog, line 203: Should we expect that when calling this function the words will be always uppercase ? Will there be cases where a value like `Init` or 'init` could be passed ? If so, an case-sensitive is not important maybe have `match s.to_uppercase()` and only fail if the keyword for state passed is something totally unknown e.g. `CLOSED`

ancazamfir: Actually I believe that we don't need the `from_str()` functions, they are currently used from tests only. Will double check and remove from both connection and channel.

PR #210, opened 2020-08-17
[DNM] Relayer Spike

brapse:
[includes sketched implementation for #328]
Do Not Merge: Relayer Spike.

The goal here is to sketch out the dependency graph needed to implement the relayer logic. Still under active development and no need for review just yet. Putting it here for provide a bit of transparency into the process.


brapse:
This was merged in #335

Issue #221, opened 2020-09-04
Extend client context mock to support more realistic unit testing


adizere:
## Summary of Bug

Sub-issue of #218.

Currently, the ICS02 mock context supports only 1 client and 1 consensus state for that client:
https://github.com/informalsystems/ibc-rs/blob/cf6c815eb03ddbee811c5220935f8b96c6b48440/modules/src/ics02_client/context_mock.rs#L11-L16

Consequently, this mock permits very limited testing scenarios.

## Rough roadmap:

- extend struct `MockClientContext` to permit storing multiple `consensus_state`, 
- extend the same struct to support multiple `client_state` (indexed by their `client_id` most likely)
- add additional testing scenarios to both in ICS02 (create & update client) and ICS03 (init & try handlers)

____

## For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #222, opened 2020-09-08
ICS02 mock extensions + more tests in ICS02 and ICS03

adizere:
Closes: #221 

______

For contributor use:

- [ ] Unit tests written

### DIFF
--- a/modules/src/ics02_client/context_mock.rs
+++ b/modules/src/ics02_client/context_mock.rs
@@ -1,3 +1,4 @@
+use crate::ics02_client::client_def::AnyConsensusState::Mock;
 use crate::ics02_client::client_def::{AnyClientState, AnyConsensusState};
 use crate::ics02_client::client_type::ClientType;
 use crate::ics02_client::context::{ClientKeeper, ClientReader};
@@ -5,6 +6,7 @@ use crate::ics02_client::error::Error;
 use crate::ics24_host::identifier::ClientId;
 use crate::mock_client::header::MockHeader;
 use crate::mock_client::state::{MockClientState, MockConsensusState};
+use std::collections::HashMap;
 use tendermint::block::Height;
 
 #[derive(Clone, Debug, PartialEq)]
@@ -12,7 +14,7 @@ pub struct MockClientContext {
     pub client_id: ClientId,
     pub client_state: Option<MockClientState>,
     pub client_type: Option<ClientType>,
-    pub consensus_state: Option<MockConsensusState>,
+    pub consensus_states: HashMap<Height, MockConsensusState>,
 }
 
 impl Default for MockClientContext {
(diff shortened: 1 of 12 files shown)


modules/src/ics02_client/context_mock.rs
```diff
@@ -5,73 +5,73 @@ use crate::ics02_client::error::Error;
...
+/// A mock implementation of client context. Implements traits `ClientReader`
+/// and `ClientKeeper`, and is therefore useful for testing the ICS02 handlers (create, update
+/// client) and other dependent ICS handlers (e.g., ICS03).
 #[derive(Clone, Debug, PartialEq)]
 pub struct MockClientContext {
-    pub client_id: ClientId,
-    pub client_state: Option<MockClientState>,
+    /// The set of all clients, as well as their states, which this context comprises.
+    pub client_states: HashMap<ClientId, MockClientState>,
+    /// All clients in this context have this (same) type.
```
ancazamfir, line 18: We should build the mock context to allow multiple clients with different types. It should be a collection of items very similar to what used to be `MockClientContext` (e.g. a `HashMap` with key `ClientId` and value that includes type, state and a collection of consensus states).

modules/src/mock_client/state.rs
```diff
@@ -22,7 +22,7 @@ impl MockClientState {
         header: AnyHeader,
     ) -> Result<(MockClientState, MockConsensusState), Box<dyn std::error::Error>> {
         match header {
-            #[cfg(test)]
+            #[cfg(test)] // this is weird...
```
romac, line 25: Agreed, it looks a bit weird. To clarify just in case, this directive tells the compiler emit this branch when compiling under the test profile, which is required as the `AnyHeader::Mock` variant is similarly only defined under the testing profile.

romac (approved):
LGTM! 👍

modules/src/mock_client/state.rs
adizere: Got it, thanks for the explanation! I had that comment specifically to remember to ask you hehe.

modules/src/ics03_connection/context_mock.rs
```diff
@@ -11,11 +11,12 @@ use crate::ics24_host::identifier::{ClientId, ConnectionId};
 use std::collections::HashMap;
 use tendermint::block::Height;
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug)]
 pub struct MockConnectionContext {
     chain_context: MockChainContext,
     client_context: MockClientContext,
     connections: HashMap<ConnectionId, ConnectionEnd>,
+    client_connections: HashMap<ClientId, ConnectionId>,
```
ancazamfir, line 19: I think this should be in the client context. We can discuss and change it in a different PR.

ancazamfir (approved):
Looks good, one minor comment.

Issue #218, opened 2020-09-03
Follow-up on message processor (handler) implementations


adizere:
## Summary of Bug

As a follow-up to #160, the short term plan for maturing the handlers is sketched below, in the order of their priority.

- [x] Wrap-up the work for #160 (PR #188) as soon as possible
(Anca, Romain, Adi)
- [x] Align testing across the handlers of ICS02 and ICS03; tentative strategy: test `process` as well as `keep` separately, then test `dispatch` by comparing its output to the ones of `process` and `keep`.
(primarily Romain) -> deferred to #269 
- [x] Adapt handlers to newer protobuf definitions, which include client state in the Try message
(primarily Anca) -- #226 
- [x] Extend ICS02 client mock to permit multiple consensus states, and two client states, thus reflecting more realistic testing scenarios
(primarily Adi) -- #221 
- [x] Cleanup Any* related code
(Romain) -- #225 
- [x] Finish all ICS03 messages: Ack and Confirm left (Adi) -- #223
- [X] Finish all ICS02/ ICS07/ mock message processing (ICS02 done with #270, ICS07 deferred to later)
- [x] Investigate whether we can replace `TryFromRaw` with the more standard `TryFrom`, with the goal of removing the former, see e.g., ICS 03 msgs.rs
(Adi) -- this was made obsolete by Greg's #252 

We'll expand and document each point (except for the first one) in a separate full-fledged issue as needed.

____

## For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #225, opened 2020-09-10
Cleanup Any* wrappers related code

romac:
See: #218

## Description

This PR improves the ergonomics of the way we deal with multiple client types at runtime,
ie. the various `Any*` wrappers and the associated traits.

- Removes uses of `dyn ClientState` and `dyn ConsensusState`
- Move validation functions out of the `ClientState` trait and into the `ClientDef` trait
- Add method to get a `AnyClient` from a `ClientType`
- Remove type parameter on messages structs, enums, `process` and `keep` functions
______

For contributor use:

- [ ] Unit tests written

### DIFF
--- a/modules/src/ics02_client/client_def.rs
+++ b/modules/src/ics02_client/client_def.rs
@@ -189,8 +189,6 @@ impl ClientDef for AnyClient {
         client_state: AnyClientState,
         header: AnyHeader,
     ) -> Result<(AnyClientState, AnyConsensusState), Box<dyn std::error::Error>> {
-        // We have to split and nest the following match because patterns containing
-        // both by-move and by-ref bindings are still unstable (feature: move_ref_pattern).
         match self {
             Self::Tendermint(client) => {
                 let (client_state, header) = downcast!(
--- a/modules/src/ics02_client/client_def.rs
+++ b/modules/src/ics02_client/client_def.rs
@@ -1,6 +1,8 @@
 use serde_derive::{Deserialize, Serialize};
 
+use crate::downcast;
 use crate::ics02_client::client_type::ClientType;
+use crate::ics02_client::error;
 use crate::ics02_client::header::Header;
 use crate::ics02_client::state::{ClientState, ConsensusState};
 use crate::ics03_connection::connection::ConnectionEnd;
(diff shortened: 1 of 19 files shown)


romac:
@ancazamfir @adizere I wonder what you think of this item?

> - Move validation functions out of the ClientState trait and into the ClientDef trait

To me, it seemed more appropriate to have this live in the `ClientDef` implementation as this is client-specific (ie. the client state is mostly data, but it's the client definition that defines how to validate the client state).

I don't feel strongly either way though, thoughts welcome!

adizere (commented):
Didn't finish the review. Just submitting very basic comments here until I manage to compile the branch..

adizere:
> > * Move validation functions out of the ClientState trait and into the ClientDef trait
> 
> To me, it seemed more appropriate to have this live in the `ClientDef` implementation as this is client-specific (ie. the client state is mostly data, but it's the client definition that defines how to validate the client state).

Agree. Your suggestion makes sense and I prefer it to the older solution where ClientState was overloaded. I think, however, the older solution was that way because of the influence of [ICS02](https://github.com/cosmos/ics/tree/master/spec/ics-002-client-semantics#definitions), which states:

> ClientState is an opaque type representing the state of a client. A ClientState __must expose__ query functions to verify membership or non-membership of key/value pairs in state at particular heights and to retrieve the current ConsensusState.

That being said, if it affords us having this cleaner implementation, I would vote that it's ok to depart a bit from ICS02 and go forward with the suggestion in this branch.

modules/src/ics02_client/client_def.rs
```diff
@@ -216,8 +167,152 @@ pub enum AnyClient {
...
+        match client_type {
+            ClientType::Tendermint => Self::Tendermint(TendermintClient),
+
+            #[cfg(test)]
+            ClientType::Mock => Self::Mock(MockClient),
+        }
+    }
+}
+
+// ⚠️  Beware of the awful boilerplate below ⚠️
```
brapse, line 181: cuuuuute

romac:
@adizere 

> I'm not able to compile locally.

I suspect you need to update your version of Rust. You can do so by running:

```
$ rustup update
```

Indeed, `Option::zip` has only been [stabilized](https://doc.rust-lang.org/stable/src/core/option.rs.html#931) since Rust v1.46.0.

My bad, I should have mentioned this in the PR. For my defense, it's sometimes hard to keep track of new APIs and their stabilization version.

modules/src/lib.rs
```diff
@@ -9,7 +9,7 @@
     unused_qualifications,
     rust_2018_idioms
 )]
-#![allow(dead_code)]
+#![allow(dead_code, unused_imports, unused_variables)]
```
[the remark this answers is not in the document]
romac: Right, forgot to remove them after the refactoring.

romac:
> Your suggestion makes sense and I prefer it to the older solution where ClientState was overloaded.

Yeah that was my feeling as well, hence the departure from ICS 02.

> That being said, if it affords us having this cleaner implementation, I would vote that it's ok to depart a bit from ICS02 and go forward with the suggestion in this branch.

It's actually orthogonal to the rest of the PR, just a matter of which trait these functions should live in. So if we feel we'd rather stick to the spec more closely, we can always move them back into the `ClientState` trait.

brapse (approved):
Great to have this moving forward. A bit weary of the roll your own inheritance system here but I can definitely appreciate the difficulties of supporting an unknown number of clients at runtime. Might be something we circle back on later but looks for as  next step 👍

modules/src/ics02_client/client_def.rs
```diff
@@ -216,8 +167,152 @@ pub enum AnyClient {
...
+        proof: &CommitmentProof,
+        client_id: &ClientId,
+        consensus_height: Height,
+        expected_consensus_state: &AnyConsensusState,
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        match self {
+            Self::Tendermint(client) => {
+                let (client_state, expected_consensus_state) = downcast!(
+                    client_state => AnyClientState::Tendermint,
+                    expected_consensus_state => AnyConsensusState::Tendermint,
```
ancazamfir, line 242: The `expected_consensus_state` should be of the same type as the local chain (`Tendermint` in production or `Mock` in testing). This is different than the `ClientDef` type. For example`self` could be a client for `Substrate` while the `expected_consensus_state` stored by "our" client on the substrate chain is `Tendermint`. Not entirely sure of the best way to get the local chain info. One possibility is to pass the `Reader` context as a parameter and get the local chain type from there. Didn't have a chance to review this before it was merged, only looked at it now while merging master to `anca/proto_update`. Here we have a similar problem for the new `verify_full_client_state()` method. Let's try to take care of this in https://github.com/informalsystems/ibc-rs/pull/227

Issue #226, opened 2020-09-10
Migrate handlers to newer protobuf definitions


ancazamfir:
## Summary
Migrate to ibc_proto v0.3.0
This is part of the follow up issue https://github.com/informalsystems/ibc-rs/issues/218

## Problem Definition
There have been major changes in the IBC proto definitions especially in connection messages and with the addition of a new `Height` type.

## Proposal
Make the minimal changes such that the code compiles
____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

PR #227, opened 2020-09-10
Migration to ibc_proto v.0.3.0

ancazamfir:
Closes: #226

## Description
Migrate to ibc_proto v0.3.0.

Includes minimal changes such that the code compiles and tests pass. In particular:

- the client state in the handshake messages is not deserialized and therefore further ignored
- the epoch in the new type `Height` is ignored and the deserialization code only extracts the `EpochHeight` in the current `Height` type which is u64.

______

For contributor use:

- [ ] Unit tests written

### DIFF
--- a/modules/src/context_mock.rs
+++ b/modules/src/context_mock.rs
@@ -1,10 +1,10 @@
-use crate::context::{ChainKeeper, ChainReader, HistoricalInfo, SelfHeader};
+use crate::context::{ChainKeeper, ChainReader, HistoricalInfo, SelfChainType, SelfHeader};
 use crate::mock_client::header::MockHeader;
 use serde_derive::{Deserialize, Serialize};
 use std::error::Error;
 use tendermint::block::Height;
 
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Default)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, Default)]
 pub struct MockChainContext {
     pub max_size: usize,
     pub latest: Height,
@@ -40,32 +40,37 @@ impl MockChainContext {
         }
     }
 
-    /// Used for testing
     pub fn validate(&self) -> Result<(), Box<dyn Error>> {
-        // check that the number of entries is not higher than max_size
-        if self.history.len() > self.max_size {
-            return Err("too many entries".to_string().into());
-        }
-
-        // check latest is properly updated with highest header height
-        let SelfHeader::Mock(lh) = self.history[self.history.len() - 1].header;
-        if lh.height() != self.latest {
-            return Err("latest height is not updated".to_string().into());
-        }
-
-        // check that all headers are in sequential order
-        for i in 1..self.history.len() {
-            let SelfHeader::Mock(ph) = self.history[i - 1].header;
-            let SelfHeader::Mock(h) = self.history[i].header;
-            if ph.height().increment() != h.height() {
-                return Err("headers in history not sequential".to_string().into());
-            }
-        }
+        // TODO
         Ok(())
+        //         // check that the number of entries is not higher than max_size
+        //         if self.history.len() > self.max_size {
+        //             return Err("too many entries".to_string().into());
+        //         }
+        //
+        //         // check latest is properly updated with highest header height
+        //         let SelfHeader::Mock(lh) = self.history[self.history.len() - 1].header;
+        //         if lh.height() != self.latest {
+        //             return Err("latest height is not updated".to_string().into());
+        //         }
+        //
+        //         // check that all headers are in sequential order
+        //         for i in 1..self.history.len() {
+        //             let SelfHeader::Mock(ph) = self.history[i - 1].header;
+        //             let SelfHeader::Mock(h) = self.history[i].header;
+        //             if ph.height().increment() != h.height() {
+        //                 return Err("headers in history not sequential".to_string().into());
+        //             }
+        //         }
+        //         Ok(())
     }
 }
 
 impl ChainReader for MockChainContext {
+    fn chain_type(&self) -> SelfChainType {
+        SelfChainType::Mock
+    }
+
     fn self_historical_info(&self, height: Height) -> Option<&HistoricalInfo> {
         let l = height.value() as usize;
         let h = self.latest.value() as usize;
--- a/modules/src/context_mock.rs
+++ b/modules/src/context_mock.rs
@@ -8,7 +8,7 @@ use tendermint::block::Height;
 pub struct MockChainContext {
     pub max_size: usize,
     pub latest: Height,
-    pub history: Vec<HistoricalInfo>,
+    pub history: Vec<MockHeader>,
 }
 
 impl MockChainContext {
@@ -17,9 +17,7 @@ impl MockChainContext {
             max_size,
             latest: n,
             history: (0..n.value())
-                .map(|i| HistoricalInfo {
-                    header: SelfHeader::Mock(MockHeader(Height(i).increment())),
-                })
+                .map(|i| MockHeader(Height(i).increment()))
                 .collect(),
         }
     }
@@ -71,7 +69,7 @@ impl ChainReader for MockChainContext {
         SelfChainType::Mock
     }
 
-    fn self_historical_info(&self, height: Height) -> Option<&HistoricalInfo> {
+    fn self_historical_info(&self, height: Height) -> Option<HistoricalInfo> {
         let l = height.value() as usize;
         let h = self.latest.value() as usize;
 
@@ -79,7 +77,9 @@ impl ChainReader for MockChainContext {
             // header with height not in the history
             None
         } else {
-            Some(&self.history[h - l])
+            Some(HistoricalInfo {
+                header: SelfHeader::Mock(self.history[h - l]),
+            })
         }
     }
 }
@@ -89,16 +89,17 @@ impl ChainKeeper for MockChainContext {
         if height != self.latest.increment() {
             return;
         }
-        let mut history = self.history.clone();
-        if history.len() >= self.max_size {
-            history.rotate_left(1);
-            history[self.max_size - 1] = info;
-        } else {
-            history.push(info);
+        if let SelfHeader::Mock(header) = info.header {
+            let mut history = self.history.clone();
+            if history.len() >= self.max_size {
+                history.rotate_left(1);
+                history[self.max_size - 1] = header;
+            } else {
+                history.push(header);
+            }
+            self.history = history;
+            self.latest = height;
         }
-        //history.insert(height, info);
-        self.history = history;
-        self.latest = height;
     }
 }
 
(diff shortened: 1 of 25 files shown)


modules/src/ics03_connection/error.rs
```diff
@@ -35,6 +35,12 @@ pub enum Kind {
     #[error("invalid address")]
     InvalidAddress,
 
+    #[error("missing consensus proof height in opentTry")]
+    MissingProofHeight,
+
+    #[error("missing consensus proof height in opentTry")]
+    MissingConsesusHeight,
```
romac, line 42:
```suggestion
    MissingConsensusHeight,
```

modules/src/ics03_connection/error.rs
```diff
@@ -35,6 +35,12 @@ pub enum Kind {
     #[error("invalid address")]
     InvalidAddress,
 
+    #[error("missing consensus proof height in opentTry")]
+    MissingProofHeight,
+
+    #[error("missing consensus proof height in opentTry")]
```
romac, line 41:
```suggestion
    #[error("missing consensus proof height in openTry")]
```

modules/src/ics03_connection/error.rs
```diff
@@ -35,6 +35,12 @@ pub enum Kind {
     #[error("invalid address")]
     InvalidAddress,
 
+    #[error("missing consensus proof height in opentTry")]
```
romac, line 38:
```suggestion
    #[error("missing consensus proof height in openTry")]
```

modules/src/ics03_connection/handler/verify.rs
```diff
@@ -23,6 +25,21 @@ pub fn verify_proofs(
...
+            ctx,
+            connection_end,
+            state,
+            proofs.height(),
+            proofs
+                .client_proof()
+                .as_ref()
+                .ok_or_else(|| Kind::NullClientProof)?,
+        )?,
+    };
```
romac, line 41:
```suggestion
    if let Some(state) = client_state {
        verify_client_proof(
            ctx,
            connection_end,
            state,
            proofs.height(),
            proofs
                .client_proof()
                .as_ref()
                .ok_or_else(|| Kind::NullClientProof)?,
        )?;
    }
```

modules/src/ics03_connection/msgs.rs
```diff
@@ -207,13 +214,38 @@ impl Msg for MsgConnectionOpenTry {
...
-        let consensus_proof_obj =
-            ConsensusProof::new(msg.proof_consensus.into(), msg.consensus_height)
-                .map_err(|e| Kind::InvalidProof.context(e))?;
+        let proof_height = msg
+            .proof_height
+            .ok_or_else(|| Kind::MissingProofHeight)?
+            .epoch_height;
+        let consensus_height = msg
+            .consensus_height
+            .ok_or_else(|| Kind::MissingConsesusHeight)?
```
romac, line 240:
```suggestion
            .ok_or_else(|| Kind::MissingConsensusHeight)?
```

modules/src/ics03_connection/msgs.rs
```diff
@@ -310,20 +351,35 @@ impl TryFromRaw for MsgConnectionOpenAck {
...
```
romac, line 360:
```suggestion
            .ok_or_else(|| Kind::MissingConsensusHeight)?
```

romac (commented):
Looks good, modulo some typos and a couple idioms :) Will leave it to @adizere to approve the PR as I am not super familiar with ICS 03 yet.

adizere:
Heads-up, once we merge #220 we might get the following error:

```
$ cargo clippy --all-targets
    Checking ibc v0.0.3 (/Users/adi/Hammers/ibc-rs/modules)
error: large size difference between variants
  --> modules/src/ics03_connection/msgs.rs:51:5
   |
51 |     ConnectionOpenTry(MsgConnectionOpenTry),
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this variant is 352 bytes
   |
note: the lint level is defined here
  --> modules/src/lib.rs:2:9
   |
2  | #![deny(clippy::all)]
   |         ^^^^^^^^^^^
   = note: `#[deny(clippy::large_enum_variant)]` implied by `#[deny(clippy::all)]`
note: and the second-largest variant is 144 bytes:
  --> modules/src/ics03_connection/msgs.rs:50:5
   |
50 |     ConnectionOpenInit(MsgConnectionOpenInit),
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
help: consider boxing the large fields to reduce the total size of the enum
   |
51 |     ConnectionOpenTry(Box<MsgConnectionOpenTry>),
   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error

error: could not compile `ibc`.```

File: modules/src/ics02_client/state.rs [hunk 7]
```diff
@@ -43,6 +43,17 @@ pub trait ClientState: Clone + std::fmt::Debug {
 
     /// Verification functions as specified in:
     /// https://github.com/cosmos/ics/tree/master/spec/ics-002-client-semantics
+    ///
+    fn verify_client_full_state(
```
adizere, line 47: The other two `verify_*` functions here appear in [ICS02](https://github.com/cosmos/ics/tree/master/spec/ics-002-client-semantics#required-functions) but this `verify_client_full_state ` one does not ring a bell. All implementations of `verify_client_full_state` are also `unimplemented` so it's not really clear what the function does by just looking at the code. Can we have a comment here (describing the purpose of the function like we do for `verify_client_consensus_state` or `verify_connection_state`)?

adizere (approved):
I think `clippy` will bite us with next commits, so we'll need to address that (see my earlier comment). Otherwise looks good!

modules/src/ics03_connection/msgs.rs
```diff
@@ -662,8 +744,8 @@ mod tests {
                 want_pass: false,
             },
         ]
-        .into_iter()
-        .collect();
+            .into_iter()
```
brapse, line 747: Why the indentation here?

ancazamfir:
Another need to fix before merge:

When running the Rust relayer query with Gaia and Go Relayer, the connection query fails. 
```
cargo run --bin relayer -- -c simple_config.toml query connection end ibc0 ibconeconnection
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
     Running `target/debug/relayer -c simple_config.toml query connection end ibc0 ibconeconnection`
     Options QueryConnectionOptions { connection_id: ConnectionId("ibconeconnection"), height: 0, proof: true }
[relayer/src/chain/cosmos.rs:67] "Todo: implement proof verification." = "Todo: implement proof verification."
connection query error Could not parse/unmarshall response: failed to decode Protobuf message: ConnectionEnd.state: invalid wire type: LengthDelimited (expected Varint)
```
The failure is here:
https://github.com/informalsystems/ibc-rs/blob/6f59e2f047777e881fdff79cb8181ed30f76c733/relayer/src/chain/cosmos.rs#L72

The query works on master (with ibc-proto v.0.1.0), the protobuf for `ConnectionEnd` hasn't changed between the two versions.

The channel query that gets a `ChannelEnd` with an almost identical proto for `Sate` works fine.

modules/src/ics03_connection/msgs.rs
ancazamfir: I think it's fixed now but no idea. Just ran `cargo fmt`, didn't notice it.

modules/src/context_mock.rs
```diff
@@ -40,32 +40,37 @@ impl MockChainContext {
...
-        for i in 1..self.history.len() {
-            let SelfHeader::Mock(ph) = self.history[i - 1].header;
-            let SelfHeader::Mock(h) = self.history[i].header;
-            if ph.height().increment() != h.height() {
-                return Err("headers in history not sequential".to_string().into());
-            }
-        }
+        // TODO
         Ok(())
+        //         // check that the number of entries is not higher than max_size
```
adizere, line 46: Is this buggy or problematic for compilation? Or how come it is commented out?

modules/src/ics03_connection/connection.rs
```diff
@@ -18,13 +20,12 @@ pub struct ConnectionEnd {
 }
 
 impl TryFromRaw for ConnectionEnd {
-    type RawType = RawConnectionEnd;
+    type RawType = RawIdentifiedConnection;
     type Error = anomaly::Error<Kind>;
-    fn try_from(value: RawConnectionEnd) -> Result<Self, Self::Error> {
+    fn try_from(value: RawIdentifiedConnection) -> Result<Self, Self::Error> {
```
adizere, line 25: I looked for where is this conversion used and could not find. Can you point me to where do we use the decoding of RawIdentifiedConnection -> ConnectionEnd ? Is it the call to `query` from `CosmosSDKChain`?

modules/src/ics03_connection/msgs.rs
```diff
@@ -139,46 +143,52 @@ impl Msg for MsgConnectionOpenInit {
...
-        match self.proofs.consensus_proof() {
-            None => Height(0),
-            Some(p) => p.height(),
-        }
-    }
-
-    /// Getter for accesing the whole counterparty of this message. Returns a `clone()`.
-    pub fn counterparty(&self) -> Counterparty {
-        self.counterparty.clone()
+    /// Getter for accessing the connection identifier of this message.
```
adizere, line 154: Did you reorder these functions? I'm just trying to understand why is this diff so complex, because nothing seems to have changed except for the order of functions (and there is `client_state` getter new).

modules/src/mock_client/state.rs
```diff
@@ -74,6 +91,65 @@ impl ClientState for MockClientState {
         // TODO
         false
     }
+
```
adizere, line 94: Let's purge this (or move it in the correct spot, which is probably `mock_client/client_def.rs`).

modules/src/ics03_connection/connection.rs
ancazamfir: yes it was from the query to fix the connection query with Gaia/ Go Relayer. But as it turns out I was using a slightly older version and once I updated the connection query worked fine. I just reverted this change.

modules/src/ics03_connection/msgs.rs
ancazamfir: Yes, I ordered the getters consistent with the order in the struct.

modules/src/context_mock.rs
ancazamfir: Just fixed it.

modules/src/ics03_connection/msgs.rs
adizere: Good idea, I like this. I'll try to do the same in the future.

adizere (approved):

Issue #223, opened 2020-09-09
ICS03 Ack and Confirm message processors


adizere:
Subtask of #218.
We should add the logic for processing ICS3 messages of type ACK and Confirm + tests.

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #228, opened 2020-09-10
Msg processors for ICS03 Ack and Confirm messages

adizere:
Closes: #223

ready for review!

______

For contributor use:

- [x] Unit tests written
- [x] Updated relevant documentation (`docs/`) and code comments


modules/src/ics03_connection/handler/verify.rs
```diff
@@ -154,28 +154,27 @@ pub fn verify_consensus_proof(
...
         < (ctx.chain_current_height().value() - ctx.chain_consensus_states_history_size() as u64)
     {
-        return Err(Kind::StaleConsensusHeight
-            .context(claimed_height.to_string())
-            .into());
+        // Fail if the consensus height is too old (outside of trusting period).
+        Err(Kind::StaleConsensusHeight(claimed_height).into())
+    } else {
+        // Height check is within normal bounds, check passes.
+        Ok(())
```
ancazamfir, line 178: Curious on the Rust idiomatic way for this. In other languages this is frowned upon:
```
if <expr1> {
    return <err1>
} else if <expr2> {
    return <err2>
...
} else  {
    return <err99>
}
```
And instead:
```
if <expr1> {
    return <err1>
}
if <expr2> {
    return <err2>
}
...
return <err99>
```
I also prefer the latter.

ancazamfir (approved):
Looks good! Thanks Adi!

modules/src/ics03_connection/handler/verify.rs
romac: I would say that in Rust the former is more idiomatic since in Rust if statements are expressions of their own, but both are fine by me.

adizere: https://doc.rust-lang.org/rust-by-example/flow_control/if_else.html but I have no preference.

romac: But if this were happening in the middle of the function (for eg. an early return with explicit `return` statements, then yeah I would also prefer the latter.

ancazamfir: > https://doc.rust-lang.org/rust-by-example/flow_control/if_else.html but I have no preference.

yes but there is no return in the if arms

ancazamfir: I am asking because it is very typical that we start a function body with a number of checks where we return errors if they fail. If we have a lot of checks we end up with a long if-then-elseif-... and I find that harder to read.

ancazamfir: More importantly though I would like to have consistency in our code base. So I will follow whatever we decide :)

romac: Yeah for such checks I agree that the version with `return` is better.

Issue #229, opened 2020-09-11
English specification of SendPacketEvent handling


milosevic:
This task should specify handling of SendPacketEvent as part of relayer processing. 
https://github.com/cosmos/ics/tree/master/spec/ics-018-relayer-algorithms is under-specified so 
we need to fill missing details. 

Related to https://github.com/informalsystems/ibc-rs/issues/124.

ancazamfir:
Solved with  https://github.com/informalsystems/ibc-rs/pull/236 and  https://github.com/informalsystems/ibc-rs/pull/230

PR #230, opened 2020-09-11
English spec for SendPacketEvent handling

milosevic:
Addresses https://github.com/informalsystems/ibc-rs/issues/229.

### DIFF
--- a/docs/spec/relayer/Packets.md
+++ b/docs/spec/relayer/Packets.md
@@ -152,11 +152,20 @@ func handleSendPacketEvent(ev, chainA) {
     // if packet commitment is empty, then packet is already received by the counter party
     if packetCommitment == null return
     if packetCommitment != hash(ev.data, ev.timeoutHeight, ev.timeoutTimestamp {
-        panic  // invalid data; probably
+        panic  // invalid data; probably fork
     }
 
     // we now check if this packet is already received by the destination chain
-    
+    if (channel.order === ORDERED) {
+          // TODO: All get call should either return from the relayer state, or if needed
+          // query full node of the chain. In the latter case, proof verification should be done 
+          // in the call.     
+          nextSequenceRecv = GetNextSequenceRecv(chainB, ev.destPort, ev.destChannel) 
+          if ev.sequence != nextSequenceRecv return // packet has already been delivered by another relayer
+    } else {
+        packetAcknowledgement = GetPacketAcknowledgement(ev.destPort, ev.destChannel, ev.sequence)
+        if packetAcknowledgement != null return
+    }
     
     packetData = Packet{logEntry.sequence, logEntry.timeoutHeight, logEntry.timeoutTimestamp,
     localEnd.portIdentifier, localEnd.channelIdentifier,
--- a/docs/spec/relayer/Packets.md
+++ b/docs/spec/relayer/Packets.md
@@ -1,0 +1,327 @@
+# IBC packet handling
+
+This document specifies relayer's logic for packet handling.
+
+## Data Types
+
+```go
+type Packet {
+    sequence           uint64
+    timeoutHeight      Height
+    timeoutTimestamp   uint64
+    sourcePort         Identifier
+    sourceChannel      Identifier
+    destPort           Identifier
+    destChannel        Identifier
+    data               bytes	
+}
+```
+
+```go
+type PacketRecv {
+     packet          Packet
+     proof           CommitmentProof
+     proofHeight     Height
+}
+```
+
+```go
+type SendPacketEvent {
+    height             Height
+    sequence           uint64
+    timeoutHeight      Height
+    timeoutTimestamp   uint64
+    sourcePort         Identifier
+    sourceChannel      Identifier
+    destPort           Identifier
+    destChannel        Identifier
+    data               bytes	
+}
+```
+
+```go
+type ChannelEnd {
+    state                           ChannelState
+    ordering                        ChannelOrder
+    counterpartyPortIdentifier      Identifier
+    counterpartyChannelIdentifier   Identifier
+    connectionHops                  [Identifier]
+    version                         string
+}
+
+enum ChannelState {
+  INIT,
+  TRYOPEN,
+  OPEN,
+  CLOSED,
+}
+
+enum ChannelOrder {
+  ORDERED,
+  UNORDERED,
+}
+```
+
+```go
+type ConnectionEnd {
+    state                               ConnectionState
+    counterpartyConnectionIdentifier    Identifier
+    counterpartyPrefix                  CommitmentPrefix
+    clientIdentifier                    Identifier
+    counterpartyClientIdentifier        Identifier
+    version                             []string
+}
+
+enum ConnectionState {
+  INIT,
+  TRYOPEN,
+  OPEN,
+}
+```
+
+```go
+type ClientState {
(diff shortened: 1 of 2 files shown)


istoilkovska (commented):
Looks good in general. 

I'm missing a bit of context and explanations on why it is necessary for the relayer to do verification before creating and submitting the datagrams. Also, it seems that the [last paragraph](https://github.com/informalsystems/ibc-rs/blob/zarko/packets-english-spec/docs/spec/relayer/Relayer.md#data-verification) of `Relayer.md` is incomplete.

docs/spec/relayer/Packets.md
```diff
@@ -0,0 +1,250 @@
...
+  upgradeCommitmentPrefix       CommitmentPrefix
+  upgradeKey                    []byte
+  maxClockDrift                 uint64
+  proofSpecs                    []ProofSpec
+}
+```
+## Helper functions
+
+We assume the existence of the following helper functions:
+
```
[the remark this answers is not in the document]
milosevic: I will provide more context; maybe write a psedocode of one of those functions.

docs/spec/relayer/Packets.md
```diff
@@ -0,0 +1,250 @@
...
+    channel, proof = GetChannel(chainB, ev.destPort, ev.destChannel, LATEST_HEIGHT)
+    if proof != nil { return nil }
+    
+    if channel == null OR
+       channel.state != OPEN OR
+       ev.sourcePort != channel.counterpartyPortIdentifier OR
+       ev.sourceChannel != channel.counterpartyChannelIdentifier { return nil }
+    
+    connectionId = channel.connectionHops[0]
+    connection, proof = GetConnection(chainB, connectionId, LATEST_HEIGHT)
```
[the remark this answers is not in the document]
milosevic: It is a signal that the query should be made for the latest height. In Tendermint APIs it is normally achieved by putting 0 in case we are not interested in a particular height.

istoilkovska (approved):

docs/spec/relayer/Packets.md
```diff
@@ -0,0 +1,267 @@
...
+    if (channel.ordering === ORDERED) {    
+        nextSequenceRecv, proof = GetNextSequenceRecv(chainB, ev.destPort, ev.destChannel, LATEST_HEIGHT) 
+        if proof != nil { return nil }
+        
+        if ev.sequence != nextSequenceRecv { return nil } // packet has already been delivered by another relayer
+    
+    } else {
+        packetAcknowledgement, proof = 
+            GetPacketAcknowledgement(chainB, ev.destPort, ev.destChannel, ev.sequence, LATEST_HEIGHT)
+        if proof != nil { return nil }
```
ancazamfir, line 245:
```suggestion
        if proof == nil { return nil }
```

docs/spec/relayer/Packets.md
```diff
@@ -0,0 +1,267 @@
...
+
+        if packetAcknowledgement != nil { return nil }
```
ancazamfir, line 247: I believe the proof returned by `GetPacketAcknowledgment` is a proof of non-membership if `packetAcknowledgment` is `nil` and we should be able to verify that.

docs/spec/relayer/Relayer.md
```diff
@@ -87,22 +91,15 @@ func handleEvent(ev, chainA, chainB) {
        return error
     }
 
-    // Stage 2.
+    // Stage 3.
     // Create the IBC datagrams including `ev` & verify them.
+    datagram = createDatagram(ev, chainA, chainB, installedHeight)
```
ancazamfir, line 96: I think this used to be `pendingDatagram`, there is still a reference to this above, could you replace it with `createDatagram`?

ancazamfir: I think we will also need to review again now that the app state queries return the poof height already incremented.
For example in these lines further up (don't know of a way to comment on lines that are not part of the diffs):
```
   // Verify the result of the query
   sh = dest.lc.get_header(membershipProof.Height + 1)
```

ancazamfir (commented):
Looks great, a few comments, opened for debate :)

docs/spec/relayer/Packets.md
```diff
@@ -0,0 +1,267 @@
...
+       packetCommitment != hash(concat(ev.data, ev.timeoutHeight, ev.timeoutTimestamp)) { return nil }
+            
+    // Stage 2 
+    // Execute checks IBC handler on chainB will execute
+    
+    channel, proof = GetChannel(chainB, ev.destPort, ev.destChannel, LATEST_HEIGHT)
+    if proof != nil { return nil }
+    
+    if channel == null OR
+       channel.state != OPEN OR
```
ancazamfir, line 222: I believe on this one we need to retry the handling of the event in order to deal with the opportunistic channel opening case. We have a similar case when we handle the channel related events and where we check if the associated connection is opened.

docs/spec/relayer/Packets.md
```diff
@@ -0,0 +1,267 @@
...
+    
+    // Stage 1 
+    // Verify if packet is committed to chain A and it is still pending (commitment exists)
+    
+    proofHeight = installedHeight - 1
+    packetCommitment, packetCommitmentProof = 
+        GetPacketCommitment(chainA, ev.sourcePort, ev.sourceChannel, ev.sequence, proofHeight)     
+    if packetCommitmentProof != nil { return nil }
+        
+    if packetCommitment == null OR
```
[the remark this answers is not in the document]
milosevic: Nope; we should just pick one and be consistent :)

docs/spec/relayer/Packets.md
milosevic: This is a very good point. I assume that we will need to deal with this at a different level; i.e., at the caller level. Even handler should probably just provide more information in a return value (we should return errors with semantics instead of nil) so caller can decide if event should be dropped or it should be retried later.

docs/spec/relayer/Packets.md
```diff
@@ -0,0 +1,267 @@
...
+    
+    if connection == null OR connection.state != OPEN { return nil } 
+    
+    if ev.timeoutHeight != 0 AND GetConsensusHeight(chainB) >= ev.timeoutHeight { return nil }
+    if ev.timeoutTimestamp != 0 AND GetCurrentTimestamp(chainB) >= ev.timeoutTimestamp { return nil }
+    
+    // we now check if this packet is already received by the destination chain
+    if (channel.ordering === ORDERED) {    
+        nextSequenceRecv, proof = GetNextSequenceRecv(chainB, ev.destPort, ev.destChannel, LATEST_HEIGHT) 
+        if proof != nil { return nil }
```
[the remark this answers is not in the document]
milosevic: We trust light client, so everything checked by the light client (via header's app hash) we also trust. I will add more on error handling in the next version, but in general we should probably switch to different rpc node within GetX functions, but handler ideally don't need to care about that. But this need to be clarified.

docs/spec/relayer/Packets.md
milosevic: Very good point; I will double check this.

docs/spec/relayer/Relayer.md
milosevic: Wasn't aware this was already decided/changed. I will correct this.

PR #233, opened 2020-09-15
Decode ClientState from Protobuf's Any type

romac:
See: #218

## Description

Decode both Tendermint and Mock client states from Protobuf's Any type.

______

For contributor use:

- [ ] Unit tests written

### DIFF
--- a/modules/src/ics02_client/client_def.rs
+++ b/modules/src/ics02_client/client_def.rs
@@ -1,15 +1,16 @@
 use serde_derive::{Deserialize, Serialize};
 
 use crate::ics02_client::client_type::ClientType;
+use crate::ics02_client::error::{Error, Kind};
 use crate::ics02_client::header::Header;
 use crate::ics02_client::state::{ClientState, ConsensusState};
-use crate::ics23_commitment::commitment::{CommitmentPrefix, CommitmentProof, CommitmentRoot};
-use crate::Height;
-
 use crate::ics03_connection::connection::ConnectionEnd;
 use crate::ics07_tendermint as tendermint;
 use crate::ics07_tendermint::client_def::TendermintClient;
+use crate::ics07_tendermint::client_state::ClientState as TendermintClientState;
+use crate::ics23_commitment::commitment::{CommitmentPrefix, CommitmentProof, CommitmentRoot};
 use crate::ics24_host::identifier::{ClientId, ConnectionId};
+use crate::Height;
 
 #[cfg(test)]
 use crate::mock_client::client_def::MockClient;
@@ -53,7 +54,7 @@ impl Header for AnyHeader {
 
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub enum AnyClientState {
-    Tendermint(crate::ics07_tendermint::client_state::ClientState),
+    Tendermint(TendermintClientState),
 
     #[cfg(test)]
     Mock(MockClientState),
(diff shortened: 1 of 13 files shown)


modules/src/ics02_client/client_def.rs
```diff
@@ -1,24 +1,30 @@
...
 
 use ::tendermint::block::Height;
 
 #[cfg(test)]
-use crate::mock_client::client_def::MockClient;
-#[cfg(test)]
-use crate::mock_client::header::MockHeader;
-#[cfg(test)]
-use crate::mock_client::state::{MockClientState, MockConsensusState};
+use {
```
adizere, line 22: Ahh this is such a treat to the eyes.

modules/src/ics02_client/client_def.rs
```diff
@@ -106,12 +113,39 @@ impl Header for AnyHeader {
...
+    Tendermint(TendermintClientState),
 
     #[cfg(test)]
     Mock(MockClientState),
 }
 
+impl AnyClientState {
+    pub fn from_any(any: prost_types::Any) -> Result<Self, Error> {
+        match any.type_url.as_str() {
+            "ibc.tendermint.ClientState" => {
```
adizere, line 125: Minor:

Maybe we should consider keeping these strings (lines 125 `"ibc.tendermint.ClientState"` and 135 `"ibc.mock.ClientState"`) hardcoded in some enum as opposed to hardcoded here? I can imagine they could even go in the ibc-proto repo, not here, but not sure what would be best.

To keep things simpler and get this merged fast, we can also just keep the hardcoded values as they are and just add a todo.

modules/src/ics03_connection/msgs.rs
```diff
@@ -14,23 +14,26 @@
 //! TODO: Separate the Cosmos-SDK specific functionality from canonical ICS types. Decorators?
 
 #![allow(clippy::too_many_arguments)]
```
adizere, line 16: Looking at the code, I think it's safe to remove this now.

modules/src/ics23_commitment/merkle.rs
```diff
@@ -1,5 +1,6 @@
 use crate::ics23_commitment::commitment::CommitmentPrefix;
-use ibc_proto::commitment::MerklePath;
+
+use ibc_proto::ibc::commitment::MerklePath;
```
adizere, line 3: Seems we always import types from ibc proto as `use ibc_proto::ibc::X`, so I'm wondering why is the `ibc` namespace in the middle necessary. We could just say `use ibc_proto::X`.

adizere (approved):
Looks great! I only had relatively minor comments.

modules/src/ics02_client/client_def.rs
ancazamfir, line 125: Wondering if we should define these `type_url` strings in a separate place so they can also be used for encoding. We can do it later also.

ancazamfir (approved):
Looks great! thanks!

modules/src/ics02_client/client_def.rs
romac: Yeah good point! Those should probably be defined as constants somewhere.

modules/src/ics23_commitment/merkle.rs
romac: See this PR for a (poorly explained) rationale: https://github.com/informalsystems/ibc-proto/pull/9

romac: Also, the ibc_proto crate exports other types which are not ibc-specific so imho it makes sense to qualify their usage, though we could te-export them at the top level but that may cause similar issues to the one mentioned in the PR linked above.

modules/src/ics02_client/client_def.rs
romac: Agreed! See my reply to Adi’s comment above.

Issue #231, opened 2020-09-15
Uncomment and fix client query commands


ancazamfir:
## Summary of Bug
The client query commands for state and consensus are currently commented out as the initial query infrastructure was changed. 
The code should be uncommented and fixed so it works at least up to the deserialization.

____

- [X] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ebuchman:
Maybe let's not bother with things that "work up to deserialization" and just fix it once we can go all the way through, including deserialization? Should be soon given progress on tendermint-rs

ancazamfir:
We will need this for the relayer MVP where we short-circuit the serialization/ deserialization. It's also a good validation that the way we model the multiple client types and their states, consensus, etc. is good.

adizere:
This is a sub-issue of #128.

The query logic used to be there, unsatisfactory, so it was mostly cut out in #174 where the infrastructure around queries changes. What's left can be seen in [`relayer/src/query/client.rs`](https://github.com/informalsystems/ibc-rs/blob/4877ee9e56b000e34cdfe5f4ec37388c889c5ad6/relayer/src/query/client.rs#L1).

PR #235, opened 2020-09-16
Fix for client query subcommands

adizere:
Closes: #231 

~~Waiting for #233 to get merged to get this to compile...~~ ready for review!

______

For contributor use:

- [ ] Unit tests written


modules/src/ics24_host/path.rs
```diff
@@ -11,7 +11,7 @@ pub const IBC_QUERY_PATH: &str = "store/ibc/key";
 pub enum Path {
     ClientType(ClientId),
     ClientState(ClientId),
```
ancazamfir, line 13: Could you change the client state to be provable in `is_provable()` below? This was changed in the last few versions (protobuf, sdk, etc).

modules/src/ics02_client/client_def.rs
```diff
@@ -119,11 +122,14 @@ pub enum AnyClientState {
...
-impl AnyClientState {
-    pub fn from_any(any: prost_types::Any) -> Result<Self, Error> {
-        match any.type_url.as_str() {
+impl TryFromRaw for AnyClientState {
+    type RawType = prost_types::Any;
+    type Error = Error;
+
+    fn try_from(value: Self::RawType) -> Result<Self, Self::Error> {
+        match value.type_url.as_str() {
             "ibc.tendermint.ClientState" => {
```
ancazamfir, line 131: Could you change all `type_url` strings to start with `/`?
```suggestion
            "/ibc.tendermint.ClientState" => {
```
Made that change and tried the client state query on latest gaia chains with clients created by the Go relayer and I get this which is nice:
```
rrly -c simple_config.toml query client state ibc0 ibconeclient -p false
    Finished dev [unoptimized + debuginfo] target(s) in 0.13s
     Running `target/debug/relayer -c simple_config.toml query client state ibc0 ibconeclient -p false`
     Options QueryClientStateOptions { client_id: ClientId("ibconeclient"), height: 0, proof: false }
client state query result:  Tendermint(ClientState { chain_id: "ibc1", trusting_period: 1209600s, unbonding_period: 1814400s, max_clock_drift: 60s, latest_height: block::Height(112), frozen_height: block::Height(0) })
```

For the consensus state I cannot get it to work even from curl or browser. But at least the client state looks good :)

ancazamfir (commented):
Looks great! Added a few comments and questions.

modules/src/ics02_client/client_def.rs
```diff
@@ -119,11 +122,14 @@ pub enum AnyClientState {
...
-                let raw = RawTendermintClientState::decode(any.value.as_ref())
+                let raw = RawTendermintClientState::decode(value.value.as_ref())
```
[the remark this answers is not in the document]
adizere: It was necessary for `AnyClientState` to implement the `TryFromRaw` trait, otherwise the call below would fail:

https://github.com/informalsystems/ibc-rs/blob/106c459cab6cc1a6359759c7bfc2d0d62e004686/relayer-cli/src/commands/query/client.rs#L80-L81

The failure occurs because the function `chain.query::<T>` requires the trait bound: `T: TryFromRaw`.

> Also value.value.as_ref() doesn't ready very well :)

Agree! Didn't notice that, will fix.

modules/src/ics24_host/path.rs
adizere: Changed to true. I was not aware of this change, good to know!

modules/src/ics02_client/client_def.rs
adizere: WOhoo!

> Could you change all type_url strings to start with /?

Yep, changed all of them.

modules/src/ics07_tendermint/consensus_state.rs
```diff
@@ -12,6 +14,15 @@ pub struct ConsensusState {
     pub next_validators_hash: Hash,
 }
 
+impl TryFromRaw for ConsensusState {
+    type RawType = ibc_proto::ibc::tendermint::ConsensusState;
+    type Error = Error;
+
+    fn try_from(_raw: Self::RawType) -> Result<Self, Self::Error> {
+        unimplemented!()
```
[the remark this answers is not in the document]
adizere: Done, not tested though!

ancazamfir:
Opened issue https://github.com/informalsystems/ibc-rs/issues/245 to track updating the mock consensus state in the proto file.

ancazamfir (approved):
Looks great! Thanks Adi!

Issue #234, opened 2020-09-16
Extend IBC packet event handling spec with error handling 


milosevic:
This issue is a follow up of https://github.com/informalsystems/ibc-rs/pull/230. The plan is extend the spec to address error handling concerns and to potentially add few more event types so additional concerns are exposed.

PR #236, opened 2020-09-17
Add error handling logic to relayer spec

milosevic:
Fixes https://github.com/informalsystems/ibc-rs/issues/234 and https://github.com/informalsystems/ibc-rs/issues/237.


docs/spec/relayer/Definitions.md
```diff
@@ -0,0 +1,267 @@
...
+
+    while(true) {
+        // Query provable store exposed by the full node of chain. 
+        // The path for the channel end is at channelEnds/ports/{portId}/channels/{channelId}".
+        // The membership proof returned is read at height proofHeight. 
+        channel, proof, error = QueryChannel(chain.provider, portId, channelId, proofHeight) 
+        if error != nil {
+            // elect a new provider from the peer list
+            if !ReplaceProvider(chain) { return (nil, nil) }  // return if fail to elect new provider         
+        }
```
ancazamfir, line 240: shouldn't we retry `QueryChannel` once we replace the provider? If we tall through here we have a bad `channel`.
```suggestion
            continue
        }
```
Or return some retry error?

docs/spec/relayer/Definitions.md
```diff
@@ -0,0 +1,267 @@
...
+    
+        header, error = GetHeader(chain.lc, proofHeight) // get header for height proofHeight using light client
+        if error != nil { return (nil, nil) }  // return if light client can't provide header for the given height
```
ancazamfir, line 243: wondering if the function should also return an error to distinguish from the case where there is no channel created.

docs/spec/relayer/Packets.md
```diff
@@ -39,212 +18,67 @@ type SendPacketEvent {
...
-       channel.state != OPEN OR
-       ev.sourcePort != channel.counterpartyPortIdentifier OR
-       ev.sourceChannel != channel.counterpartyChannelIdentifier { return nil }
+    if channel != nil AND
+       (channel.state == CLOSED OR
+        ev.sourcePort != channel.counterpartyPortIdentifier OR
+        ev.sourceChannel != channel.counterpartyChannelIdentifier) { (nil, Error.DROP) } 
     
+    if channel == nil OR channel.state != OPEN  { (nil, Error.RETRY) } 
+    // TODO: Maybe we shouldn't even enter handle loop for packets if the corresponding channel is not open!
```
ancazamfir, line 57: Yes but I think here is the first time when we can make that determination.

docs/spec/relayer/Relayer.md
```diff
@@ -220,20 +71,123 @@ More details about light client assumptions and protocols can be found
...
-`verifyMembership(header.AppHash, h, proof, path, d)` that returns `true` if data was committed by the corresponding
+`VerifyMembership(header.AppHash, h, proof, path, d)` that returns `true` if data was committed by the corresponding
 chain at height *h*. The trusted header is provided by the corresponding light client. 
   
+## Relayer algorithm
+
+The main relayer event loop is a pipeline of four stages. Assuming some IBC event at height `h` on `chainA`, 
+the relayer:
+
+1. Determine destination chain (`chainB`)
```
ancazamfir, line 82:
```suggestion
1. Determines destination chain (`chainB`)
```

docs/spec/relayer/Relayer.md
```diff
@@ -220,20 +71,123 @@ More details about light client assumptions and protocols can be found
...
+2. Updates (on `chainB`) the IBC client for `chainA` to a certain height `H` where `H >= h+1`.
+3. Create IBC datagram at height `H-1`.
```
ancazamfir, line 84:
```suggestion
3. Creates IBC datagram at height `H-1`.
```

docs/spec/relayer/Relayer.md
```diff
@@ -220,20 +71,123 @@ More details about light client assumptions and protocols can be found
...
+4. Submit the datagram from stage (2) to `chainB`.
```
ancazamfir, line 85:
```suggestion
4. Submits the datagram from stage (2) to `chainB`.
```

docs/spec/relayer/Relayer.md
```diff
@@ -220,20 +71,123 @@ More details about light client assumptions and protocols can be found
...
+    // Update on `chainB` the IBC client for `chainA` to height `>= targetHeight`.
+    targetHeight = ev.height + 1
+    // See the code for `updateIBCClient` below.
+    installedHeight, error := updateIBCClient(chainB, chainA, targetHeight)
+    if error != nil { return Error.RETRY }
+
+    // Stage 3.
+    // Create the IBC datagrams including `ev` & verify them.
+    datagram, error = CreateDatagram(ev, chainA, chainB, installedHeight)
+    if error != nil { return Error.RETRY }
```
ancazamfir, line 132: Need to tell the caller if it should drop or retry? So just pass the error through
```suggestion
    if error != nil { return error }
```

docs/spec/relayer/Definitions.md
```diff
@@ -0,0 +1,267 @@
...
+           proofHeight Height) (ChannelEnd, CommitmentProof) {
+
+    while(true) {
+        // Query provable store exposed by the full node of chain. 
+        // The path for the channel end is at channelEnds/ports/{portId}/channels/{channelId}".
+        // The membership proof returned is read at height proofHeight. 
+        channel, proof, error = QueryChannel(chain.provider, portId, channelId, proofHeight) 
+        if error != nil {
+            // elect a new provider from the peer list
+            if !ReplaceProvider(chain) { return (nil, nil) }  // return if fail to elect new provider
```
ancazamfir, line 239: This is a bit different than the comment earlier where the caller replaces the provider. I guess it is a bit hard to figure the best way until we start coding this. But the higher lever function, `handleEvent()`, function could bubble up `drop`, `provider_error`, `retry`, etc. And the caller would be the one to take the action, like change the provider node, buffer the event for later retry, drop, etc.

docs/spec/relayer/Relayer.md
```diff
@@ -220,20 +71,123 @@ More details about light client assumptions and protocols can be found
...
+    
+    // Stage 4.
+    // Submit datagrams.
+    error = Submit(chainB, datagram)
+    if error != nil { return Error.RETRY }
```
ancazamfir, line 137:
```suggestion
    if error != nil { return error }      
```

docs/spec/relayer/Relayer.md
```diff
@@ -220,20 +71,123 @@ More details about light client assumptions and protocols can be found
...
+We assume that the corresponding light client is correctly installed on each chain.
+
+The semantic of the returned error is to signal to the caller whether there was an error in the event processing that 
+could be transient (for example not able to establish connection to a correct full node), in which case
+the event can be reprocessed later (`Error.RETRY`), or the event should be dropped as it has already been 
+received by the destination chain due to activity of a concurrent relayer (`Error.DROP`).
+
+Data structures and helper function definitions are provided 
+[here](https://github.com/informalsystems/ibc-rs/blob/master/docs/spec/relayer/Definitions.md). 
+
```
istoilkovska, line 104: Do we need to precisely specify the way in which retry errors are handled? Or is this out of the scope of this specification?

docs/spec/relayer/Definitions.md
milosevic: Yep; I realised I was not consistent with throwing errors and handling them. I will modify it to return error and then we can have handleEvent() caller dealing with it. We might want to sketch also that functionality (it's a bit like light client supervisor or fast sync scheduler) as it will need to implement error handling policies, store events that need to be retried, manage timeout triggers needed by packets handling, etc.

docs/spec/relayer/Definitions.md
milosevic: I have added error as a return value.

docs/spec/relayer/Packets.md
milosevic: What I meant here is basically having supervisor logic that would decide when to call handleEvent based on event type, i.e., there would be some preconditions that are based on relayer state. For example, packet related events are preconditioned on channel and connections being open, all events are preconditioned on light client correctly installed, etc.

docs/spec/relayer/Relayer.md
milosevic: As I mentioned above, we will probably specify some kind of supervisor that would be calling handleEvent and dealing with errors.

docs/spec/relayer/Packets.md
ancazamfir: Does this assume that the relayer keeps state for channel and connections? And the supervisor does not perform any queries to determine that. Trying to save on the number of queries :)

docs/spec/relayer/Packets.md
```diff
@@ -98,4 +115,82 @@ func CreateDatagram(ev SendPacketEvent,
...
+    channelA, proof, error = GetChannel(chainA, ev.port, ev.channel, ev.height)
+    if error != nil { return (nil, error) }
+    
+    channelB, proof, error = 
+        GetChannel(chainB, channelA.counterpartyPortIdentifier, channelA.counterpartyChannelIdentifier, LATEST_HEIGHT)
+    if error != nil { return (nil, error) }
+    
+    if channelB == nil OR channel.state != OPEN  { (nil, Error.DROP) } 
+    // Note that we checked implicitly above that counterparty identifiers match each other
+
```
ancazamfir, line 152: Are these channel retrievals and checks the same as for the SendPacket event above? If yes can we make a single function for this? If not maybe we cover the differences and why they are.

docs/spec/relayer/Packets.md
```diff
@@ -98,4 +115,82 @@ func CreateDatagram(ev SendPacketEvent,
...
+    if error != nil { return (nil, error) }
+    
+    if channelB == nil OR channel.state != OPEN  { (nil, Error.DROP) } 
+    // Note that we checked implicitly above that counterparty identifiers match each other
+               
+    connectionId = channelB.connectionHops[0]
+    connection, proof, error = GetConnection(chainB, connectionId, LATEST_HEIGHT) 
+    if error != nil { return (nil, error) }
+        
+    if connection == nil OR connection.state != OPEN { return (nil, Error.DROP) }
```
ancazamfir, line 157: same question here.

docs/spec/relayer/Packets.md
```diff
@@ -98,4 +115,82 @@ func CreateDatagram(ev SendPacketEvent,
...
+        GetPacketCommitment(chainB, channelA.counterpartyPortIdentifier, 
+                            channelA.counterpartyChannelIdentifier, ev.sequence, LATEST_HEIGHT)     
+    if error != nil { return (nil, error) }
+            
+    if packetCommitment == nil OR
+       packetCommitment != hash(concat(ev.data, ev.timeoutHeight, ev.timeoutTimestamp)) { 
+            // invalid event; bad provider
+            return (nil, Error.BADPROVIDER) 
+    }
+
```
ancazamfir, line 170: same here

ancazamfir (commented):
Looks great! A few comments looking for modularizing a bit the checks so the pseudocode is easier to read, if possible.

docs/spec/relayer/Packets.md
milosevic: Exactly. Supervisor will need anyway to keep some state (for example for pending packets so it can trigger timeouts), so why not keeping status of connections and channels, current block height and time, etc. It can save queries and potentially operates faster.

docs/spec/relayer/Packets.md
milosevic: Checks are similar, but error message is different. With recv packets we can retry on error (if connection and channel are not open)  while for acks we should drop event in this case. We can make a function for this (and we should at the code level) but at the spec level maybe it's better making it more verbose so reading it is easier.

ancazamfir (approved):
Looks good, thanks Zarko!

Issue #142, opened 2020-07-14
Move ibc-proto source code into ibc-rs


greg-szabo:
## Summary
https://github.com/informalsystems/ibc-proto can be moved into ibc-rs.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ebuchman:
Is this blocking on anything? Should we move it in for v0.0.2?

greg-szabo:
It’s just waiting for me to get to it. We can schedule it into 0.0.2.

________________________________

PR #239, opened 2020-09-17
the ibc crate now uses the local ibc-proto crate

greg-szabo:
Closes: #142 

## Description
The ibc crate (modules folder) now uses the ibc-proto crate defined in this repository. (proto folder).

I also updated the version in html_url for ibc-proto but it's too late for v0.3.0. We should make sure we update it in the next release.

______

For contributor use:

- [ ] Unit tests written

### DIFF
--- a/modules/Cargo.toml
+++ b/modules/Cargo.toml
@@ -19,7 +19,7 @@ tendermint = "0.15.0"
 tendermint-rpc = { version = "0.15.0", features = ["client"] }
 
 # Proto definitions for all IBC-related interfaces, e.g., connections or channels.
-ibc-proto = "0.3.0"
+ibc-proto = { path = "../proto" }
 
 anomaly = "0.2.0"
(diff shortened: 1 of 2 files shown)


modules/Cargo.toml
```diff
@@ -19,7 +19,7 @@ tendermint = "0.15.0"
 tendermint-rpc = { version = "0.15.0", features = ["client"] }
 
 # Proto definitions for all IBC-related interfaces, e.g., connections or channels.
-ibc-proto = "0.3.0"
+ibc-proto = { path = "../proto" }
```
romac, line 22: We need to specify a version number as well, otherwise we won't be able to publish the crate.

```suggestion
ibc-proto = { version = "0.3.0", path = "../proto" }
```

romac (changes requested):

romac (approved):

adizere (approved):

PR #242, opened 2020-09-20
Signed Tx against latest Gaia

ebuchman:
Closes: #47 

This shouldn't be merged as it's just for demonstration but here we have a working signed protobuf tx in rust that is accepted by the latest gaiav3.0 branch of gaia. 

This has all the protobuf stuff copied in for convenience, and is using the MsgSend. We should instead apply these changes to Andy's `andy/cli-open-conn` branch and get it working with ConnOpenInit.

Only problem with merging it into relayer-cli is it uses the latest version of eg. the k256 crate for signing and we need to update tendermint-rs to v0.16 (#243) to be compatible 

## Description

______

For contributor use:

- [ ] Unit tests written

### DIFF
--- a/main.rs
+++ b/main.rs
@@ -1,0 +1,382 @@
+use rand_core::OsRng;
+// use secp256k1::{Message, PublicKey, Secp256k1, SecretKey};
+use hex;
+use std::num::ParseIntError;
+
+use prost::bytes::Buf;
+
+use k256::{
+    ecdsa::{signature::Signer, signature::Verifier, Signature, SigningKey, VerifyKey},
+    EncodedPoint, SecretKey,
+};
+
+// Use this for account
+// use tendermint::amino_types::message::AminoMessage;
+// use tendermint::account::Id;
+// use tendermint::public_key::Ed25519 as TMEd25519;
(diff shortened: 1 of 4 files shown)


main.rs
```diff
@@ -0,0 +1,382 @@
...
+/// Single is the mode info for a single signer. It is structured as a message
+/// to allow for additional fields such as locale for SIGN_MODE_TEXTUAL in the future
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct Single {
+    /// mode is the signing mode of the single signer
+    #[prost(enumeration = "SignMode", tag = "1")]
+    pub mode: i32,
+}
+
+/// CompactBitArray is an implementation of a space efficient bit array.
```
brapse, line 184: Was this comment and accompanying code copied from the go codebase?

andynog: I think this PR might be closed. This code was just added as a reference for the tx signing. It will be properly implemented when tendermint-rs upgrade lands on ibc-rs.

adizere:
https://github.com/informalsystems/ibc-rs/pull/345 took the work in this PR to completion. Closing this, as I think it's not necessary anymore.

Issue #124, opened 2020-07-02
Specify packet datagram creation and handling


istoilkovska:
## Summary

Specify packet datagrams creation and handling in the relayer TLA+ specification. 

## Proposal
Add support for packet datagrams. 

On the chain side:
 - define logs for sent packets 
 - define logs for received packets
 - add action for sending packets 
 - add action for processing packets

On the relayer side:
 - define operator for creating packet datagrams by scanning the sent packets log
 - define operator for creating packet acknowledgment datagrams by scanning the received packets log

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

Issue #138, opened 2020-07-09
Relayer TLA+ spec enhancements/features originating from PR reviews


istoilkovska:
## Summary

This issue summarizes questions raised and suggestions proposed during pull request reviews of the Relayer TLA+ specification, which may need further discussion/clarification.

## Problem Definition

1. **Relayer client heights.** Currently, each relayer process stores the latest height of each chain locally.  
   **Questions:**
    - Should the relayer store client heights? 
    - Should it have action that updates the heights of its clients?

2. **Datagram transmission.** Currently, the datagrams the relayers create are stored in a set, thus losing the order in which they were created.
   **Questions:**
    - Should the outgoing datagrams from the relayer and the incoming datagrams to the chains be stored in a set or sequence?

3. **Creation of client datagrams.** It would be nice to paramterize the operators for creating client datagrams by height. If the datagrams are created for height `h`, then the light client updates are created for height `h+1`, as the root hash for datagrams created at height `h` is stored in the block `h+1`.

4. **`openInit` datagrams**. Currently, the relayer creates `ConnOpenInit` and `ChanOpenInit` datagrams. In [ICS003](https://github.com/cosmos/ics/tree/master/spec/ics-003-connection-semantics) and [ICS004](https://github.com/cosmos/ics/tree/master/spec/ics-004-channel-and-packet-semantics), it is noted that the chains are initiators of these datagrams
   **Questions:**
     - Should the relayer trigger the creation of `ConnOpenInit` and `ChanOpenInit` datagrams, or should it be done by the environment using non-deterministically assigned Boolean flags? 

5. **Packet datagrams creation**. The TLA+ specification at the moment follows [ICS018](https://github.com/cosmos/ics/tree/master/spec/ics-018-relayer-algorithms), where the relayer does not do any checks locally on the validity of the data it uses to create `PacketRecv` and `PacketAck` datagrams. It is assumed that the handlers on the chain side do these checks and discard the irrelevant datagrams. It would be good to add these checks in order to model the cases when it is desirable to not even send invalid datagrams to the chains (e.g., in the case when a fee has to be paid). Necessary checks:
             a. check that the counterparty has not yet received the packet when creating `PacketRecv` datagrams
             b. check that the packet commitment has been deleted when creating `PacketAck` datagrams
____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

[...]

PR #248, opened 2020-09-22
Basic packet flow in TLA+

istoilkovska:
Closes: #124 

## Description

Added support for handling the basic packet flow in the Relayer TLA+ specification. Currently, the spec handles sending, receiving, and acknowledgements of packets.

The following basic packet flow is assumed:
 1. `ChainA` sends a packet by writing a sent packet log entry to the packet log and writing a packet commitment
 2. A relayer reads the packet log, picks the sent packet log entry, and creates `PacketRecv` datagram for `ChainB`
 3. `ChainB` processes the `PacketRecv` datagram, and adds the received packet to its set of packets that should be acknowledged (asynchronous acknowledgements) 
 4. `ChainB` writes the acknowledgement by writing a log entry to the packet log
 5. A relayer reads the packet log, picks the write ack log entry, and creates `PacketAck` datagram for `ChainA`
 6. `ChainA` processes the `PacketAck` datagram by deleting the packet commitment of the original sent packet
 
The packets and packet logs are sequences, rather than sets. This is to ensure that there is order in sending, receiving, and acknowledgements of packets.
______

For contributor use:

- [ ] Unit tests written
- [x] Updated relevant documentation (`docs/`) and code comments
- [x] Re-reviewed `Files changed` in the Github PR explorer


ancazamfir (commented):
Very nice spec! I have a few comments after the first pass. Struggling with TLC, one relayer and minimal config, it still runs after almost 1hr.

docs/spec/relayer/PacketHandlers.tla
```diff
@@ -0,0 +1,125 @@
...
+HandlePacketRecv(chainID, chain, packetDatagram, log) ==
+    \* get chainID's connection end
+    LET connectionEnd == chain.connectionEnd IN
+    \* get chainID's channel end
+    LET channelEnd == chain.connectionEnd.channelEnd IN
+    \* get packet
+    LET packet == packetDatagram.packet IN
+    
+    IF \* if the channel and connection ends are open for packet transmission
+       /\ channelEnd.state /= "UNINIT"
```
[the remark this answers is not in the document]
istoilkovska: In the ICS, the check is whether `channelEnd != null` and `channelEnd.state = OPEN`. So I was trying to capture that with this check, since in the TLA spec, `channelEnd == null` is modeled by `channelEnd.state /= UNINIT`.

docs/spec/relayer/Relayer.tla
```diff
@@ -233,19 +234,45 @@ ChannelDatagrams(srcChainID, dstChainID) ==
...
+    
+    LET packetData(logEntry) == AsPacket([sequence |-> logEntry.sequence, 
+                                 timeoutHeight |-> logEntry.timeoutHeight,
+                                 srcChannelID |-> srcChannelID,
+                                 dstChannelID |-> dstChannelID]) IN
+    
+    IF packetLogEntry.type = "PacketSent"
+    THEN AsDatagram([type |-> "PacketRecv",
+          packet |-> packetData(packetLogEntry),  
+          proofHeight |-> srcHeight])
```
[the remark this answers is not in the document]
istoilkovska: This spec is aligned with ICS018, where the relayer does not do any checks. In a discussion with @milosevic, we agreed that for now it is enough to model the packet datagram creation without doing these checks. I have documented the need for adding checks in #138.

docs/spec/relayer/Chain.tla
```diff
@@ -10,9 +10,12 @@ CONSTANTS MaxHeight, \* maximal chain height
 
 VARIABLES chainStore, \* chain store, containing client heights, a connection end, a channel end 
           incomingDatagrams, \* set of incoming datagrams
-          history \* history variable
+          incomingPacketDatagrams, \* sequence of incoming packet datagrams
+          history, \* history variable
```
milosevic, line 14: What is the purpose of history variable?

docs/spec/relayer/Chain.tla
```diff
@@ -78,12 +81,34 @@ ChannelUpdate(chainID, store, datagrams) ==
...
 
+(***************************************************************************
+ Packet update operators
+ ***************************************************************************)
+\* Update the chain store of the chain with chainID and the packet log, 
+\* using the packet datagrams generated by the relayer
+\* (Handler operators defined in PacketHandlers.tla)    
+PacketUpdate(chainID, store, packetDatagrams, log) ==
+    \* if the sequence of packet datagrams is not empty
+    IF packetDatagrams /= AsSeqPacketDatagrams(<<>>)
```
milosevic, line 92: Why the structure is not the same as with others, i.e., it contains `HandlePacketRecv` and `HandlePacketAck` that internally check for the datagram type?

docs/spec/relayer/Chain.tla
```diff
@@ -78,12 +81,34 @@ ChannelUpdate(chainID, store, datagrams) ==
...
+    ELSE [chainStore |-> store, packetLog |->log]
+
 (***************************************************************************
  Chain update operators
  ***************************************************************************)
 \* Update chainID with the received datagrams
-\* Supports ICS2 (Clients), ICS3 (Connections), and ICS4 (Channels).
-UpdateChainStore(chainID, datagrams) == 
+\* Supports ICS2 (Clients), ICS3 (Connections), and ICS4 (Channels & Packets).
+UpdateChainStoreAndPacketLog(chainID, datagrams, packetDatagrams, log) ==
```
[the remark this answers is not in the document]
milosevic: Agree with Anca that orders matter. We can have either sequence or handle only single datagram at the time.

docs/spec/relayer/Chain.tla
```diff
@@ -94,15 +119,62 @@ UpdateChainStore(chainID, datagrams) ==
...
+                            {packetAcknowledgement},
+                       !.packetsToAcknowledge = 
+                            Tail(chain.packetsToAcknowledge)]                         
+    
+    \* remove the packet from the set of packets to acknowledge
+    ELSE [chain EXCEPT !.packetsToAcknowledge = 
+                            Tail(chain.packetsToAcknowledge)] 
+
+\* log acknowledgements to packet Log
+LogAcknowledgement(chain, log, packet) ==
```
milosevic, line 164: I am not sure if we need to model event log as it is not necessarily, i.e., current ICS18 is not using it. It might be useful as part of MBT, but in the initial version I would prefer having the minimal set of functionalities needed.

docs/spec/relayer/Chain.tla
```diff
@@ -111,29 +183,98 @@ UpdateChainStore(chainID, datagrams) ==
...
+    \* connection end is initialized
+    /\ connectionEnd.state /= "UNINIT"
+    \* timeout height has not passed
+    /\ \/ packet.timeoutHeight = 0 
+       \/ latestClientHeight < packet.timeoutHeight
+    /\ \* if the channel is ordered, check if packetSeq is nextSendSeq, 
+       \* add a packet committment in the chain store, and increase nextSendSeq
+       \/ /\ channelEnd.order = "ORDERED"
+          /\ packet.sequence = channelEnd.nextSendSeq
+          /\ packet.sequence <= MaxPacketSeq
```
milosevic, line 214: This is a model constraint, not the handler one; ideally they should not be mixed and model constraint can be a precondition at the higher level.

docs/spec/relayer/Chain.tla
```diff
@@ -156,6 +299,8 @@ Init ==
 Next ==
     \/ AdvanceChain 
     \/ HandleIncomingDatagrams
+    \/ SendPacket
```
milosevic, line 302: There is a TODO above that can be removed :)

docs/spec/relayer/Chain.tla
```diff
@@ -156,6 +299,8 @@ Init ==
...
+    \/ AcknowledgePacket
```
milosevic, line 303: Should we add timeout handling here? Then it seems complete.

milosevic: Also, what about channel closing? Is it also triggered "externally"?

docs/spec/relayer/PacketHandlers.tla
```diff
@@ -0,0 +1,123 @@
...
+       /\ channelEnd.state = "OPEN"
+       /\ connectionEnd.state /= "UNINIT"
+       /\ connectionEnd.state = "OPEN" 
+       \* if the packet has not passed the timeout height
+       /\ \/ packet.timeoutHeight = 0 
+          \/ chain.height < packet.timeoutHeight  
+       \* if the "PacketRecv" datagram can be verified 
+       /\ packetDatagram.srcChannelID = GetCounterpartyChannelID(chainID)
```
milosevic, line 32: This should be checked against channelEnd object, and not from "global" knowledge.

docs/spec/relayer/PacketHandlers.tla
```diff
@@ -0,0 +1,123 @@
...
+EXTENDS Integers, FiniteSets, RelayerDefinitions    
+
+(***************************************************************************
+ Packet datagram handlers
+ ***************************************************************************)
+
+\* Handle "PacketRecv" datagrams
+HandlePacketRecv(chainID, chain, packetDatagram, log) ==
+    \* get chainID's connection end
+    LET connectionEnd == chain.connectionEnd IN
```
milosevic, line 17: I am wondering if we are missing important aspect here as we don't use chain and connection id from packetDatagram to fetch local connection and channel end.

docs/spec/relayer/Relayer.tla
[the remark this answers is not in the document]
ancazamfir: Ok, as long as we add these checks eventually. This is something we will do in the relayer and it's good to include the checks in the spec.

milosevic (approved):
LGTM. There are few comments that could be addressed and few more general concerns that we can address later.

docs/spec/relayer/Chain.tla
ancazamfir: channel closing is triggered by the application so yes the trigger is "external"

docs/spec/relayer/Chain.tla
istoilkovska: I've refactored this action a bit.

ancazamfir (approved):
Looks good! Thanks Ilina!

Issue #207, opened 2020-08-14
Relayer CLIs for client create messages


ancazamfir:
## Summary
We should be able to send transactions from the relayer CLI for all IBC client related messages.

##  Proposal
The following CLIs should be implemented by the relayer:

1. `relayer tx raw create-client <to-chain-B> <chain-A> <client-id-A> ` 
      - creates a client of `chain-A`, with id `client-id-A`, on chain `to-chain-B`:
        - sends request to local relayer light client for `chain-A` to retrieve latest header `header_A` 
        - sends a `MsgCreateClient` to `chain-B` with:
```
    client_id = client-id-A,
    header=header_A,
    trusting_period: get_trusting_period(chain-A),
    bonding_period: get_bonding_period(chain-A),
    signer: get_account(chain-B),
``` 

2. `relayer tx raw update-client <to-chain-B> <chain-A> <client-id-A>  -h <A-height>` 
This sub-issue was moved here: https://github.com/informalsystems/ibc-rs/issues/277

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ancazamfir:
`update-client` still needs to be implemented.

Issue #90, opened 2020-06-09
Clean up light client related code 


milosevic:
Review the relayers use of the light-client and tendermint and clean up any lose ends. Minimize the code the relayer needs to setup and manage the light-client by adding more functionality to the light-client crate and to ics07 module as necessary. Any light-client specific logic should be handled in the light-client Eg. light-client needs to properly handle initialization (https://github.com/informalsystems/tendermint-rs/issues/484).

ebuchman:
Some cleanup can probably already be done but this also blocks on https://github.com/informalsystems/tendermint-rs/issues/484

ebuchman:
Related to #243 - need to update to tendermint-rs v0.16 and replace all use of the now removed `lite` module

PR #251, opened 2020-09-23
Relayer CLIs for client messages

ancazamfir:
Closes: #207

## Description
Implements the relayer CLI to create a client. 

## Notes
Currently light client fails to get the header from the source chain (https://github.com/informalsystems/ibc-rs/issues/90)
Also `chain.send()` is not fully implemented (https://github.com/informalsystems/ibc-rs/issues/47)

For contributor use:

- [ ] Unit tests written


modules/src/ics02_client/client_def.rs
```diff
@@ -14,8 +14,8 @@ use crate::ics23_commitment::commitment::{CommitmentPrefix, CommitmentProof, Com
 use crate::ics24_host::identifier::{ClientId, ConnectionId};
 
 use ::tendermint::block::Height;
-
 use prost_types::Any;
+
 use std::convert::TryFrom;
 use tendermint_proto::{DomainType, Error, Kind};
 #[cfg(test)]
```
romac, line 21: Small nitpick: Can we group the imports as follows?

```rust
use prost_types::Any;
use std::convert::TryFrom;

use ::tendermint::block::Height;
use tendermint_proto::{DomainType, Error, Kind};

 #[cfg(test)]
```

and perhaps move the `std` and `prost` imports at the top of the file together with the `serde_derive` import.

modules/src/ics02_client/handler/create_client.rs
```diff
@@ -62,24 +63,29 @@ pub fn keep(keeper: &mut dyn ClientKeeper, result: CreateClientResult) -> Result
...
     use super::*;
     use crate::ics02_client::context_mock::MockClientContext;
+    use crate::ics03_connection::msgs::test_util::get_dummy_account_id;
+    use crate::ics07_tendermint::client_state::ClientState;
     use crate::ics07_tendermint::header::test_util::get_dummy_header;
-    use crate::ics07_tendermint::msgs::create_client::MsgCreateClient;
     use crate::mock_client::header::MockHeader;
     use crate::mock_client::state::{MockClientState, MockConsensusState};
-    use std::str::FromStr;
+    use std::str::{from_utf8, FromStr};
```
romac, line 71: I would rather have `use std::str::{self, FromStr}` and then use `str::from_utf` at call site to make it clearer what we are converting the utf8-encoded bytes into.

```suggestion
    use std::str::{self, FromStr};
```

modules/src/ics02_client/handler/create_client.rs
```diff
@@ -62,24 +63,29 @@ pub fn keep(keeper: &mut dyn ClientKeeper, result: CreateClientResult) -> Result
...
     use std::time::Duration;
+    use tendermint::account::Id as AccountId;
     use tendermint::block::Height;
 
     #[test]
     fn test_create_client_ok() {
         let client_id: ClientId = "mockclient".parse().unwrap();
         let ctx = MockClientContext::default();
 
+        let signer = AccountId::from_str(from_utf8(&get_dummy_account_id()).unwrap()).unwrap();
```
romac, line 81: See above.

```suggestion
        let signer = AccountId::from_str(str::from_utf8(&get_dummy_account_id()).unwrap()).unwrap();
```

But since this seems to be a common pattern (at least in tests), perhaps that warrants having a `from_utf8` static method on `AccountId`? This would get rid of the two unwraps as well:

```suggestion
        let signer = AccountId::from_utf8(&get_dummy_account_id()).unwrap();
```

modules/src/ics02_client/handler/create_client.rs
```diff
@@ -113,6 +119,8 @@ mod tests {
     fn test_create_client_existing_client_type() {
         let height = Height(42);
         let client_id: ClientId = "mockclient".parse().unwrap();
+        let signer = AccountId::from_str(from_utf8(&get_dummy_account_id()).unwrap()).unwrap();
```
romac, line 122:
```suggestion
        let signer = AccountId::from_str(str::from_utf8(&get_dummy_account_id()).unwrap()).unwrap();
```

or

```suggestion
        let signer = AccountId::from_utf8(&get_dummy_account_id()).unwrap();
```

modules/src/ics02_client/handler/create_client.rs
```diff
@@ -137,6 +146,8 @@ mod tests {
     #[test]
     fn test_create_client_existing_client_state() {
         let client_id: ClientId = "mockclient".parse().unwrap();
+        let signer = AccountId::from_str(from_utf8(&get_dummy_account_id()).unwrap()).unwrap();
```
romac, line 149:
```suggestion
        let signer = AccountId::from_str(str::from_utf8(&get_dummy_account_id()).unwrap()).unwrap();
```

or

```suggestion
        let signer = AccountId::from_utf8(&get_dummy_account_id()).unwrap();
```

modules/src/ics02_client/handler/create_client.rs
```diff
@@ -160,6 +172,7 @@ mod tests {
     #[test]
     fn test_create_client_ok_multiple() {
         let existing_client_id: ClientId = "existingmockclient".parse().unwrap();
+        let signer = AccountId::from_str(from_utf8(&get_dummy_account_id()).unwrap()).unwrap();
```
romac, line 175:
```suggestion
        let signer = AccountId::from_str(str::from_utf8(&get_dummy_account_id()).unwrap()).unwrap();
```

modules/src/ics02_client/handler/create_client.rs
```diff
@@ -218,24 +234,27 @@ mod tests {
 
     #[test]
     fn test_tm_create_client_ok() {
-        use tendermint::account::Id as AccountId;
         let client_id: ClientId = "tendermint".parse().unwrap();
+        let signer = AccountId::from_str(from_utf8(&get_dummy_account_id()).unwrap()).unwrap();
```
romac, line 238:
```suggestion
        let signer = AccountId::from_str(str::from_utf8(&get_dummy_account_id()).unwrap()).unwrap();
```

or

```suggestion
        let signer = AccountId::from_utf8(&get_dummy_account_id()).unwrap();
```

modules/src/ics02_client/handler/update_client.rs
```diff
@@ -68,18 +72,24 @@ mod tests {
...
+    use crate::ics03_connection::msgs::test_util::get_dummy_account_id;
     use crate::mock_client::header::MockHeader;
+    use std::str::{from_utf8, FromStr};
+    use tendermint::account::Id as AccountId;
     use tendermint::block::Height;
 
     #[test]
     fn test_update_client_ok() {
         let client_id: ClientId = "mockclient".parse().unwrap();
+        let signer = AccountId::from_str(from_utf8(&get_dummy_account_id()).unwrap()).unwrap();
```
romac, line 84: Same comments as for `create_client.rs`.

modules/src/ics02_client/msgs.rs
```diff
@@ -7,6 +7,15 @@
 use crate::ics02_client::client_def::{AnyClientState, AnyConsensusState, AnyHeader};
 use crate::ics02_client::client_type::ClientType;
 use crate::ics24_host::identifier::ClientId;
+use crate::tx_msg::Msg;
+use tendermint::account::Id as AccountId;
+
+use crate::ics02_client::error;
+use ibc_proto::ibc::client::MsgCreateClient;
+use std::convert::TryFrom;
+use tendermint_proto::{DomainType, Error, Kind};
```
romac, line 16:
```suggestion
use crate::ics02_client::client_def::{AnyClientState, AnyConsensusState, AnyHeader};
use crate::ics02_client::client_type::ClientType;
use crate::ics02_client::error;
use crate::ics24_host::identifier::ClientId;
use crate::tx_msg::Msg;

use ibc_proto::ibc::client::MsgCreateClient;
use tendermint::account::Id as AccountId;
use tendermint_proto::{DomainType, Error, Kind};

use std::convert::TryFrom;
```

modules/src/ics02_client/msgs.rs
```diff
@@ -15,17 +24,155 @@ pub enum ClientMsg {
 }
 
 /// A type of message that triggers the creation of a new on-chain (IBC) client.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
```
romac, line 27: It's usually a good practice to include the `Eq` bound when deriving `PartialEq`, unless there is a good reason not to do so (eg. equality is not an equivalence relation).

```suggestion
#[derive(Clone, Debug, PartialEq, Eq)]
```

modules/src/ics02_client/msgs.rs
```diff
@@ -15,17 +24,155 @@ pub enum ClientMsg {
...
+            None => Err(Kind::DecodeMessage.context(error::Kind::InvalidRawConsensusState)),
+            Some(raw_consensus_state) => Ok(raw_consensus_state),
+        }?;
+
+        Ok(MsgCreateAnyClient {
+            client_id: raw.client_id.parse().unwrap(),
+            client_type,
+            client_state: AnyClientState::try_from(raw_client_state).unwrap(),
+            consensus_state: AnyConsensusState::try_from(raw_consensus_state).unwrap(),
+            signer: AccountId::new(<[u8; 20]>::try_from(&raw.signer[..20]).unwrap()),
```
romac, line 112: Something like this should work as well:

```suggestion
            signer: AccountId::new(raw.signer[..20].try_into()).unwrap()),
```

Perhaps this could be added as a static method on `AccountId`?

```rust
fn from_bytes(bytes: &[u8]) -> Result<Self, std::array::TryFromSliceError>> {
  Ok(Self::new(raw.signer[..20].try_into()?))
}
```

I now see that the type lives in `tendermint-rs` so we can definitely leave this up as a future improvement.

relayer-cli/src/commands/tx.rs
```diff
@@ -0,0 +1,21 @@
...
+use crate::commands::tx::client::TxCreateClientCmd;
+use abscissa_core::{Command, Options, Runnable};
+
+mod client;
+
+/// `tx` subcommand
+#[derive(Command, Debug, Options, Runnable)]
+pub enum TxCmd {
+    /// The `tx raw` subcommand
+    #[options(help = "tx raw")]
```
romac, line 12: Perhaps give a short sentence which explains what the command does?

relayer-cli/src/commands/tx.rs
```diff
@@ -0,0 +1,21 @@
...
+    Raw(TxRawCommands),
+}
+
+#[derive(Command, Debug, Options, Runnable)]
+pub enum TxRawCommands {
+    /// The `tx raw client-create` subcommand
+    #[options(help = "tx raw create-client")]
```
romac, line 19: Same here.

relayer-cli/src/commands/tx/client.rs
```diff
@@ -0,0 +1,177 @@
...
+        any_client_state,
+        any_consensus_state,
+        signer,
+    );
+
+    // Create a proto any message
+    let mut proto_msgs: Vec<Any> = Vec::new();
+    let any_msg = Any {
+        // TODO - add get_url_type() to prepend proper string to get_type()
+        type_url: "/ibc.client.MsgCreateClient".to_ascii_lowercase(),
```
romac, line 171: Let's extract this into a constant at some later point?

relayer/src/chain.rs
```diff
@@ -58,6 +59,9 @@ pub trait Chain {
     /// The naming is foreshadowing for an upcoming `grpc_query` function in the future.
     fn abci_query(&self, data: Path, height: u64, prove: bool) -> Result<Vec<u8>, Self::Error>;
 
+    /// send a transaction with `msgs` to chain.
+    fn send(&self, msgs: Vec<Any>) -> Result<(), Self::Error>;
```
romac, line 63: This should probably be a `&[Any]` or Clippy will complain ;)

romac (commented):
Great stuff! Some minor nitpicks inline, feel free to ignore/postpone. Not approving yet since it's still marked WIP.

relayer-cli/src/commands/tx/client.rs
ancazamfir: Yes, this comes up in every PR now ;) I will create an issue for this.

modules/src/ics02_client/error.rs
```diff
@@ -48,6 +48,12 @@ pub enum Kind {
 
     #[error("mismatch between client and arguments types, expected: {0:?}")]
     ClientArgsTypeMismatch(ClientType),
+
+    #[error("unknown client message type: {0}")]
```
adizere, line 52: Hmm seems like `UnknownClientMessageType` and `InvalidSigner ` are not used anywhere.

modules/src/ics02_client/handler/create_client.rs
```diff
@@ -62,24 +63,29 @@ pub fn keep(keeper: &mut dyn ClientKeeper, result: CreateClientResult) -> Result
...
     use std::time::Duration;
+    use tendermint::account::Id as AccountId;
     use tendermint::block::Height;
 
     #[test]
     fn test_create_client_ok() {
         let client_id: ClientId = "mockclient".parse().unwrap();
         let ctx = MockClientContext::default();
 
+        let signer = AccountId::from_str(str::from_utf8(&get_dummy_account_id()).unwrap()).unwrap();
```
adizere, line 81: Not very important but caught my eyes:

There's a lot of places where we unwrap & unwrap the `AccountId::from_str()`. Might be worth refactoring all those calls into a method. An idea is to rename `get_dummy_account_id` into `get_dummy_account_id_bytes`, which would return `Vec<u8>`, and then introduce a second method `get_dummy_account_id` that returns a `AccountId` (and hence eliminate the duplicate code calling `AccountId::from_str()`).

modules/src/ics02_client/msgs.rs
```diff
@@ -3,10 +3,19 @@
...
 use crate::ics02_client::client_type::ClientType;
+use crate::ics02_client::error;
 use crate::ics24_host::identifier::ClientId;
+use crate::tx_msg::Msg;
+
+use ibc_proto::ibc::client::MsgCreateClient;
+use tendermint::account::Id as AccountId;
+use tendermint_proto::{DomainType, Error, Kind};
+
+pub const TYPE_MSG_CREATE_CLIENT: &str = "create_client";
```
adizere, line 18: Maybe this doesn't need to be public (there's the getter `get_type` below).

modules/src/ics02_client/msgs.rs
```diff
@@ -15,17 +24,155 @@ pub enum ClientMsg {
...
+impl DomainType<MsgCreateClient> for MsgCreateAnyClient {}
+
+impl TryFrom<MsgCreateClient> for MsgCreateAnyClient {
+    type Error = Error;
+
+    fn try_from(raw: MsgCreateClient) -> Result<Self, Self::Error> {
+        let raw_client_state = match raw.client_state {
+            None => Err(Kind::DecodeMessage.context(error::Kind::InvalidRawClientState)),
+            Some(raw_state) => Ok(raw_state),
+        }?;
```
adizere, line 91: I think your approach is more explicit (I tend to prefer it), but learning from Romain's practices I notices this more compact alternative:


```suggestion
        let raw_client_state = raw
            .client_state
            .ok_or_else(|| Kind::DecodeMessage.context(error::Kind::InvalidRawClientState))?;
```

modules/src/ics02_client/msgs.rs
```diff
@@ -15,17 +24,155 @@ pub enum ClientMsg {
...
+        let client_type = match raw_client_state.type_url.as_str() {
+            "/ibc.tendermint.ClientState" => Ok(ClientType::Tendermint),
+
+            _ => Err(
+                Kind::DecodeMessage.context(error::Kind::UnknownConsensusStateType(
+                    raw_client_state.clone().type_url,
+                )),
+            ),
+        }?;
+        let raw_consensus_state = match raw.consensus_state {
```
adizere, line 102: Same comment as above regarding the alternative use of `ok_or_else` which might be more idiomatic.

modules/src/ics02_client/msgs.rs
```diff
@@ -15,17 +24,155 @@ pub enum ClientMsg {
...
+        prost::Message::encode(&raw_msg, &mut buf).unwrap();
+        buf
+    }
+
+    fn get_signers(&self) -> Vec<AccountId> {
+        vec![self.signer]
+    }
+}
+
+impl DomainType<MsgCreateClient> for MsgCreateAnyClient {}
```
adizere, line 82: Found it a bit confusing that the raw type here is called `MsgCreateClient`. To be consistent with other places, maybe import this type `as RawMsgCreateClient`.

modules/src/ics07_tendermint/error.rs
```diff
@@ -5,6 +5,9 @@ pub type Error = anomaly::Error<Kind>;
 
 #[derive(Clone, Debug, Error)]
 pub enum Kind {
+    #[error("identifier error")]
```
adizere, line 8: Not used.

modules/src/ics07_tendermint/error.rs
```diff
@@ -25,6 +28,9 @@ pub enum Kind {
 
     #[error("invalid raw client consensus state")]
     InvalidRawConsensusState,
+
+    #[error("invalid signer")]
+    InvalidSigner,
```
adizere, line 33: Not used either.

modules/src/ics07_tendermint/consensus_state.rs
```diff
@@ -78,6 +83,15 @@ impl ConsensusState {
             next_validators_hash,
         }
     }
+
+    pub fn new_from_header(header: SignedHeader<TMCommit, TMHeader>) -> Self {
```
adizere, line 87: Minor: maybe we can use the `From` trait to implement this? Not sure what the trade-off is.

adizere (approved):
Wohoo awesome! I mostly had minor suggestions.

ancazamfir:
Thanks for the reviews @romac and @adizere! I tried to address all of them, please take another look if you have time.

relayer/src/tx/client.rs
```diff
@@ -0,0 +1,95 @@
...
+
+use tendermint::block::Height;
+
+use crate::chain::cosmos::block_on;
+use crate::chain::{query_latest_header, Chain, CosmosSDKChain};
+use crate::config::ChainConfig;
+use crate::error::{Error, Kind};
+
+#[derive(Clone, Debug)]
+pub struct CreateClientStateOptions {
```
adizere, line 19: Nitpick: Was wondering why wasn't this called "CreateClientOptions". (The name "ClientState" appears in my mind as a contrast to "Client Consensus State".)

adizere (approved):
Looks good, thanks Anca!

Issue #249, opened 2020-09-22
Implement DomainType


greg-szabo:
## Summary
Implement the DomainType trait, introduced with Tendermint protobuf. 

## Problem Definition
This makes the code slightly easier to read and enables us to focus on implementing domain knowledge, instead of jumping through the hoops of encoding/decoding protobuf.

## Proposal
Implement the tendermint_proto::DomainType trait and remove the TryFromRaw trait.
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #252, opened 2020-09-24
Removed TryFromRaw, implemented DomainType where needed

greg-szabo:
Closes: #249 

## Description
TryFromRaw was removed and replaced with the DomainType trait. Tests were updated and I found that the ConnectionID integration test fails because the simd application protobuf is outdated. (Commented out for now.)

______

For contributor use:

- [X] Unit tests written
- [X] Updated relevant documentation (`docs/`) and code comments


romac (approved):
Great stuff, thanks :)

relayer-cli/tests/integration.rs
```diff
@@ -47,14 +47,18 @@ fn simd_chain() -> CosmosSDKChain {
 #[test]
 #[ignore]
 fn query_connection_id() {
+    /* the current informaldev/simd image has an incompatible (old?) protobuf implementation
```
romac, line 50: What’s the path forward for re-enabling this test? Do we just need to wait for the image to be updated to the new protobuf implementation? What does that mean actually? New protobuf message descriptors?

modules/src/ics03_connection/connection.rs
```diff
@@ -36,6 +37,23 @@ impl TryFromRaw for ConnectionEnd {
...
+            client_id: value.client_id.to_string(),
+            versions: value.versions,
+            state: value.state as i32,
+            counterparty: Some(ibc_proto::ibc::connection::Counterparty {
+                client_id: value.counterparty.client_id.as_str().to_string(),
+                connection_id: value.counterparty.connection_id.to_string(),
+                prefix: Some(ibc_proto::ibc::commitment::MerklePrefix {
+                    key_prefix: value.counterparty.prefix.0,
+                }),
+            }),
```
adizere, line 52:
```suggestion
            counterparty: Some(RawCounterparty::from(value.counterparty)),
```

adizere (approved):
Cool cool! Thanks, Greg.

relayer-cli/tests/integration.rs
greg-szabo: Yes, we need a new docker image with an updated simd application. The protobuf message was updated in the SDK and an extra field added. We have more recent protobuf structs than the simd. (The simd is from the times of game-of-zones phase 1 when a bunch of the protobuf looked different.)

Issue #253, opened 2020-09-24
Split the `msg.rs` in per message modules


ancazamfir:
## Summary
Split the `msg.rs` in per message modules

## Problem Definition
These files/ modules have grown quite a bit in size, making difficult to manage. 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #186, opened 2020-07-31
Clean up unhelpful extra folder names in crate libs


greg-szabo:
## Summary

The modules crate's exported features are in an unnecessary long path.

For example in the CLI acceptance.rs file, here's how module components are imported:
```
use relayer_modules::ics03_connection::connection::ConnectionEnd;
use relayer_modules::ics03_connection::exported::Connection;
use relayer_modules::ics03_connection::exported::State as ConnectionState;
use relayer_modules::ics04_channel::channel::ChannelEnd;
use relayer_modules::ics04_channel::exported::Channel;
use relayer_modules::ics04_channel::exported::Order;
use relayer_modules::ics04_channel::exported::State as ChannelState;
use relayer_modules::ics23_commitment::CommitmentPrefix;
use relayer_modules::ics24_host::identifier::{ChannelId, ClientId, ConnectionId, PortId};
use relayer_modules::ics24_host::Path::{ChannelEnds, ClientConnections, Connections};
```

### Proposal:
Remove the extra path where unnecessary, like `exported`, `connection`, etc.

For example, updated list:
```
use relayer_modules::ics03_connection::ConnectionEnd;
use relayer_modules::ics03_connection::Connection;
use relayer_modules::ics03_connection::State as ConnectionState;
use relayer_modules::ics04_channel::ChannelEnd;
use relayer_modules::ics04_channel::Channel;
use relayer_modules::ics04_channel::Order;
use relayer_modules::ics04_channel::State as ChannelState;
use relayer_modules::ics23_commitment::CommitmentPrefix;
use relayer_modules::ics24_host::{ChannelId, ClientId, ConnectionId, PortId};
use relayer_modules::ics24_host::Path::{ChannelEnds, ClientConnections, Connections};
```
(Makes easier to couple them too.)

To do this, the `mod.rs` file has to be updated. For example, instead of
```
pub mod connection;
```
it will read:
```
mod connection;
pub use connection::ConnectionEnd;
```

It's a bit of extra work for us since all published components need to be listed in the `mod.rs` file (unless there's some globbing option I'm not aware of), but it's beneficial for the developers: it reads more easily.

Please comment your opinions here. This is an opinionated idea, we may or may not want to do it.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
[...]

PR #268, opened 2020-09-29
connection msgs.rs split

ancazamfir:
Closes: #253

## Description
Split the `ics03_connection/msgs.rs`, added `From` for all connection messages and tests 

______

For contributor use:

- [ ] Unit tests written


modules/src/ics03_connection/handler/conn_open_try.rs
```diff
@@ -5,7 +5,7 @@ use crate::ics03_connection::error::{Error, Kind};
 use crate::ics03_connection::handler::verify::{check_client_consensus_height, verify_proofs};
 use crate::ics03_connection::handler::ConnectionEvent::ConnOpenTry;
 use crate::ics03_connection::handler::ConnectionResult;
-use crate::ics03_connection::msgs::MsgConnectionOpenTry;
+use crate::ics03_connection::msgs::conn_open_try::MsgConnectionOpenTry;
```
adizere, line 8: In a separate, later PR, we can consider using the trick Greg used in ibc-proto, to shorten the import paths. So instead of

```
use crate::ics03_connection::msgs::conn_open_try::MsgConnectionOpenTry;
```

we could get 
```
use crate::ics03_connection::msgs::MsgConnectionOpenTry;
```

by using in msgs.rs something like:
```
mod conn_open_try;
pub use conn_open_try::*;
```

This improvement could benefit multiple other parts of our codebase, not just msgs. I will open an issue for this later on.

adizere (approved):
Looks much cleaner like this! Looks good to me.

modules/src/ics03_connection/handler/conn_open_try.rs
ancazamfir: Agreed. I tried this briefly but couldn't make it work and ran out of time.

adizere: Ah in fact there is already an issue open for this: https://github.com/informalsystems/ibc-rs/issues/186. We'll solve this later!

PR #280, opened 2020-10-01
fix tendermint dependencies

ancazamfir:
## Description
tendermint dependencies need update after merge of #255

### DIFF
--- a/modules/Cargo.toml
+++ b/modules/Cargo.toml
@@ -35,13 +35,11 @@ dyn-clonable = "0.9.0"
 [dependencies.tendermint]
 version = "0.16.0"
 git = "https://github.com/informalsystems/tendermint-rs"
-branch = "romac/light-client-ibc"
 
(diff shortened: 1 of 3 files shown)


andynog (commented):
Do we really need to keep the git property or should we just use version? Was the git introduced in order to specify the brach? In the past we had

tendermint = "0.15.0"

ancazamfir:
> Do we really need to keep the git property or should we just use version? Was the git introduced in order to specify the brach? In the past we had
> 
> tendermint = "0.15.0"

We will move to 0.17 once it's out, hopefully in a few days. But now we need the fix from  https://github.com/informalsystems/tendermint-rs/pull/583 (on top of 0.16.0). Also maybe other commits that will go in master before 0.17 is released.

andynog:
Makes sense since there's a dependency on #583 until 0.17 release

andynog (approved):
Looks good!

Issue #286, opened 2020-10-06
query client consensus command fails 


ancazamfir:
## Summary of Bug
query client consensus command fails

## Version
master

## Steps to Reproduce
issue the command against the network setup by go relayer:
```
$ cargo run --bin relayer -- -c simple_config.toml query client consensus ibc0 ibconeclient 23
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
     Running `target/debug/relayer -c simple_config.toml query client consensus ibc0 ibconeclient 23`
     Options QueryClientConsensusOptions { client_id: ClientId("ibconeclient"), consensus_height: 23, height: 0, proof: true }
client consensus state query error:  Error(Context { kind: Query, source: Some(Error(Context { kind: EmptyResponseValue, source: None, backtrace: Some(   0: backtrace::backtrace::libunwind::trace
...
```
____

## For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

PR #287, opened 2020-10-06
Fix query client consensus command

ancazamfir:
Closes: #286

## Description
Added new query parameter for the epoch value.
Running the query against the gaiad/ go relayer setup:
```
$ rrly -c simple_config.toml query client consensus ibc0 ibconeclient 0 28
    Finished dev [unoptimized + debuginfo] target(s) in 0.15s
     Running `target/debug/relayer -c simple_config.toml query client consensus ibc0 ibconeclient 0 28`
     Options QueryClientConsensusOptions { client_id: ClientId("ibconeclient"), consensus_epoch: 0, consensus_height: 28, height: 0, proof: true }
[relayer/src/chain/cosmos.rs:63] "Todo: implement proof verification." = "Todo: implement proof verification."
client consensus state query result:  Tendermint(ConsensusState { timestamp: Time(2020-10-06T11:06:11.147301Z), root: CommitmentRoot([2, 140, 56, 156, 164, 1, 214, 81, 231, 130, 237, 6, 5, 85, 55, 52, 79, 27, 43, 202, 19, 49, 199, 249, 140, 146, 23, 88, 44, 151, 104, 246]), next_validators_hash: Hash::Sha256(702F9DF2C25EAC1CE7C19DC03E7AA10E3FD9BF1E65EE2A9B96DAEE3DD2E65FC5) })
```

### DIFF
--- a/modules/src/ics24_host/path.rs
+++ b/modules/src/ics24_host/path.rs
@@ -11,7 +11,11 @@ pub const IBC_QUERY_PATH: &str = "store/ibc/key";
 pub enum Path {
     ClientType(ClientId),
     ClientState(ClientId),
-    ClientConsensusState(ClientId, u64, u64),
+    ClientConsensusState {
+        client_id: ClientId,
+        epoch: u64,
+        height: u64,
+    },
     ClientConnections(ClientId),
     Connections(ConnectionId),
     Ports(PortId),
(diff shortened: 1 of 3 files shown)


modules/src/ics24_host/path.rs
```diff
@@ -11,7 +11,7 @@ pub const IBC_QUERY_PATH: &str = "store/ibc/key";
 pub enum Path {
     ClientType(ClientId),
     ClientState(ClientId),
-    ClientConsensusState(ClientId, u64),
+    ClientConsensusState(ClientId, u64, u64),
```
romac, line 14: It may be easy to mix up the two `u64` fields, so let's perhaps give them names? What do you think?

```suggestion
    ClientConsensusState {
        client_id: ClientId,
        epoch: u64,
        height: u64,
    },
```

romac: Maybe we can also do the same for the other paths which contain a single, plain `u64`?

romac (approved):
Looks good! Left a suggestion but feel free to ignore it and merge this if you feel it's not worth the change.

romac (approved):

Issue #297, opened 2020-10-08
Cleaner context mocks for ICS18 and ICS26


adizere:
## Summary

The last bits as part of overhauling mock context #269: this issue addresses specifically the mock context for ICS18 and ICS26. Once we clean these up, we will also be able to remove the old context mocks (`MockConnectionContext`, `MockClientContext`)

@ancazamfir [raised two interesting comments](https://github.com/informalsystems/ibc-rs/pull/298#discussion_r503259245) while reviewing the prior iteration of work on mock contexts:

> Wondering if we should check for sanity for the context, even if it's mock, just to make sure the tests don't initialize with a bad context. For example check that the client of a connection is present. Could also add the connection to client here.
I am also wondering if we should allow for a vector of connections, i.e. call it with_connections().

TODOs:
- [x] Anca's two suggestions
- [x] modules/src/ics03_connection/handler.rs the context should not be mutable
____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

Issue #296, opened 2020-10-08
Cleaner context mocks for ICS2


adizere:
## Summary

Part of overhauling mock context #269: this issue addresses specifically the mock context for ICS2.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #298, opened 2020-10-08
Cleanup mock context for ICS3

adizere:
Closes: #295 

## Description

______

For contributor use:

- [X] Unit tests written
- [X] Added test to CI if applicable 

### DIFF
--- a/modules/src/mock_context.rs
+++ b/modules/src/mock_context.rs
@@ -26,14 +26,14 @@ pub struct MockContext {
     /// A list of `max_history_size` headers, ascending order by their height (latest is last).
     history: Vec<MockHeader>,
 
-    // All the connections in the store.
-    connections: HashMap<ConnectionId, ConnectionEnd>,
-
     /// The set of all clients, indexed by their id.
     clients: HashMap<ClientId, MockClientRecord>,
 
     // Association between client ids and connection ids.
     client_connections: HashMap<ClientId, ConnectionId>,
+
+    // All the connections in the store.
+    connections: HashMap<ConnectionId, ConnectionEnd>,
 }
 
 impl MockContext {
@@ -50,20 +50,6 @@ impl MockContext {
             connections: Default::default(),
             clients: Default::default(),
             client_connections: Default::default(),
-        }
-    }
-
-    /// Internal interface for associating a connection to this context.
-    pub fn with_connection(
-        self,
-        connection_id: ConnectionId,
-        connection_end: ConnectionEnd,
-    ) -> Self {
-        let mut connections = self.connections.clone();
-        connections.insert(connection_id, connection_end);
-        Self {
-            connections,
-            ..self
         }
     }
 
@@ -87,9 +73,23 @@ impl MockContext {
         Self { clients, ..self }
     }
 
+    /// Internal interface for associating a connection to this context.
+    pub fn with_connection(
+        self,
+        connection_id: ConnectionId,
+        connection_end: ConnectionEnd,
+    ) -> Self {
+        let mut connections = self.connections.clone();
+        connections.insert(connection_id, connection_end);
+        Self {
+            connections,
+            ..self
+        }
+    }
+
     /// Internal interface. Accessor for a header of the local (host) chain of this context.
     /// May return `None` if the header for the requested height does not exist.
-    fn fetch_host_header(&self, height: Height) -> Option<MockHeader> {
+    fn host_header(&self, height: Height) -> Option<MockHeader> {
         let l = height.value() as usize;
         let h = self.latest_height.value() as usize;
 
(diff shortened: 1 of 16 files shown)


romac:
Aside from the build failing, this looks good!

modules/src/mock_context.rs
```diff
@@ -0,0 +1,199 @@
...
+    // All the connections in the store.
+    connections: HashMap<ConnectionId, ConnectionEnd>,
+
+    /// The set of all clients, indexed by their id.
+    clients: HashMap<ClientId, MockClientRecord>,
+
+    // Association between client ids and connection ids.
+    client_connections: HashMap<ClientId, ConnectionId>,
+}
+
```
ancazamfir, line 38: Could we keep the clients, connections, channels, packets,.. in the struct, init sequence, etc?
```suggestion
pub struct MockContext {
    /// Maximum size of the history.
    max_history_size: usize,

    /// Highest height of the headers in the history.
    latest_height: Height,

    /// A list of `max_history_size` headers, ascending order by their height (latest is last).
    history: Vec<MockHeader>,

    /// The set of all clients, indexed by their id.
    clients: HashMap<ClientId, MockClientRecord>,
    
    // Association between client ids and connection ids.
    client_connections: HashMap<ClientId, ConnectionId>,
    
    // All the connections in the store.
    connections: HashMap<ConnectionId, ConnectionEnd>,
}

```

modules/src/mock_context.rs
```diff
@@ -0,0 +1,199 @@
...
+        connection_end: ConnectionEnd,
+    ) -> Self {
+        let mut connections = self.connections.clone();
+        connections.insert(connection_id, connection_end);
+        Self {
+            connections,
+            ..self
+        }
+    }
+
```
ancazamfir, line 69: Wondering if we should check for sanity for the context, even if it's mock, just to make sure the tests don't initialize with a bad context. For example check that the client of a connection is present. Could also add the connection to client here.
I am also wondering if we should allow for a vector of connections, i.e. call it `with_connections()`.

modules/src/mock_context.rs
```diff
@@ -0,0 +1,199 @@
...
+            .consensus_states
+            .insert(height, MockConsensusState(MockHeader(height)));
+        clients.insert(client_id.clone(), client_record);
+
+        Self { clients, ..self }
+    }
+
+    /// Internal interface. Accessor for a header of the local (host) chain of this context.
+    /// May return `None` if the header for the requested height does not exist.
+    fn fetch_host_header(&self, height: Height) -> Option<MockHeader> {
```
ancazamfir, line 92: Leaving this comment here but it applies generally. We have a number of places with `fetch_<item>()`, `get_<item>()` or just `<item>()`. I personally prefer the last :)

```suggestion
    fn host_header(&self, height: Height) -> Option<MockHeader> {
```

ancazamfir (commented):
Looks good! A few comments but we can also make changes later.

modules/src/mock_context.rs
adizere: Agree with the fact we should be consistent in naming this class of functions. The simple name `<item>` seems like the best option to me also. Will rename this and keep an eye out for other spots.

modules/src/mock_context.rs
adizere: > I am also wondering if we should allow for a vector of connections, i.e. call it with_connections().

I'm slowly expanding the mocks to support more and more capabilities. So it's quite likely that we'll have such a function `with_connections`, which will enable more complex test cases. I can do that in the next iteration of mock  refurbishing (i.e., for #296).

> Wondering if we should check for sanity for the context, even if it's mock, just to make sure the tests don't initialize with a bad context. For example check that the client of a connection is present. Could also add the connection to client here.

Hmm, not sure I understand. Maybe this would be a good use case, indeed, but as I understand the mock context, its purpose is to support testing of handlers (and not more than that at the moment). So I would not add too much logic, validation, or any kind of support, unless it is to exercise more paths in the codebase.

ancazamfir: > Hmm, not sure I understand. Maybe this would be a good use case, indeed, but as I understand the mock context, its purpose is to support testing of handlers (and not more than that at the moment). So I would not add too much logic, validation, or any kind of support, unless it is to exercise more paths in the codebase.

Exactly, but in tests we always want to start with a correct context. This becomes harder to setup correctly with a more complex initial state.

ancazamfir: > I'm slowly expanding the mocks to support more and more capabilities. So it's quite likely that we'll have such a function with_connections, ....

This is would be the same as `with_connection`, just pass a slice and do a loop.

adizere: Now I understand! Will track these two suggestions in the other iteration on this work (https://github.com/informalsystems/ibc-rs/issues/297#issue-717562826).

modules/src/mock_context.rs
```diff
@@ -0,0 +1,200 @@
...
+        MockContext {
+            max_history_size,
+            latest_height,
+            history: (0..n)
+                .rev()
+                .map(|i| MockHeader((latest_height.value() - i).try_into().unwrap()))
+                .collect(),
+            connections: Default::default(),
+            clients: Default::default(),
+            client_connections: Default::default(),
```
ancazamfir, line 53:
```suggestion
            clients: Default::default(),
            client_connections: Default::default(),
            connections: Default::default(),
```

ancazamfir (approved):
Looks good! Thanks Adi!

Issue #272, opened 2020-09-30
Adapt code to the new `Height` def, in particular the `epoch_number`


ancazamfir:
## Problem Definition
With the latest proto definitions all heights are of the followint type:
https://github.com/informalsystems/ibc-rs/blob/ec9dafb7af199cf4e3d83b60a4fb971009dc141f/proto/src/prost/ibc.client.rs#L68-L85

Currently the `epoch_number` is ignored and we deserialize into `0` on receive and use `0` when we serialize.

## Proposal
Add new domain type `Height` and define operators like `eq`, `lt`, etc. Update the code to use this new type.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

greg-szabo:
This is done in Tendermint-rs, you can use that. (We can add Ord and Eq traits, if needed.)

Issue #191, opened 2020-08-03
Ambiguity in modules regarding the type Height


adizere:
## Summary of Bug

There are currently two type declarations for `Height` in the modules:

1. A general definition at the top-level of modules:

https://github.com/informalsystems/ibc-rs/blob/65487863a68e6307d1b08c02f202a4ba1f2c6340/modules/src/lib.rs#L36

2. Then the `Height` being used in each specific module is different, for example:

https://github.com/informalsystems/ibc-rs/blob/c447e63692c663a6af325196e8f92cf9a32bf91d/modules/src/ics03_connection/events.rs#L11-L14

Both types are imported from tendermint (one from the `light` the other from `block`) and both boil down to a `u64`. Unless there is a specific reason to keep both, we should be consistent here. Even better, maybe we can have our own definition of a `Height` that is (in some sense) even more general than the definitions from tendermint.

## Steps to Reproduce

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #292, opened 2020-10-07
Tendermint protobuf dependency for IBC proto

greg-szabo:
Closes: #284 .

## Description
This PR fixes the issue of duplicated tendermint protobuf structs. The compiler will now invoke the tendermint protobufs structs instead of recreating them into the IBC protobuf crate.

~~This requires a rebuild of the IBC protobuf structs, which should be done in a different issue.~~
______

For contributor use:

- [ ] Unit tests written
- [ ] Added test to CI if applicable 
- [ ] Updated CHANGELOG_PENDING.md
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Updated relevant documentation (`docs/`) and code comments
- [ ] Re-reviewed `Files changed` in the Github PR explorer

PR #304, opened 2020-10-12
Add Height domain type 

ancazamfir:
Closes: #272 
Closes: https://github.com/informalsystems/ibc-rs/issues/191

______

For contributor use:

- [ ] Unit tests written


modules/src/ics03_connection/error.rs
```diff
@@ -24,6 +23,9 @@ pub enum Kind {
     #[error("consensus height claimed by the client on the other party is too advanced: {0}")]
     InvalidConsensusHeight(Height),
 
+    #[error("invalid raw height")]
```
adizere, line 26: I think this is not used anywhere.

modules/src/ics03_connection/msgs/conn_open_ack.rs
```diff
@@ -55,7 +56,7 @@ impl MsgConnectionOpenAck {
     /// value `Height(0)` if this field is not set.
     pub fn consensus_height(&self) -> Height {
         match self.proofs.consensus_proof() {
-            None => Height::from(0_u32),
+            None => zero_height(),
```
adizere, line 59: Just an aesthetic idea: This would look nicer in my opinion if we typed it as `Height::zero()`. The current method `zero_height()` is an improvement over the older one (`Height::from(0_u32)`) but is somehow odd, as if the method was local to this module..

modules/src/ics07_tendermint/consensus_state.rs
```diff
@@ -50,7 +50,6 @@ impl DomainType<RawConsensusState> for ConsensusState {}
 impl TryFrom<RawConsensusState> for ConsensusState {
     type Error = Error;
 
-    // TODO: Fix height conversion below (which ignores epoch number, hence it's incorrect).
     // Related: https://github.com/informalsystems/ibc-rs/issues/191.
```
adizere, line 53: I guess we can remove the issue mentioned in this comment.

modules/src/lib.rs
```diff
@@ -46,8 +46,8 @@ pub mod context_mock;
 #[cfg(test)]
 pub mod mock_client;
 
-/// Height of a block, same as in `tendermint` crate
-pub type Height = tendermint::block::Height;
+/// short use name for IBC Height domain type
```
adizere, line 49: Great!

adizere: I guess we should usually prefer:

```
use crate::Height
```
instead of

```
use crate::ics02_client::height::Height;
```
as often as possible.

modules/src/proofs.rs
```diff
@@ -18,20 +19,20 @@ impl Proofs {
         object_proof: CommitmentProof,
         client_proof: Option<CommitmentProof>,
         consensus_proof: Option<ConsensusProof>,
-        proof_height: u64,
+        proof_height: ibc_proto::ibc::client::Height,
```
adizere, line 22: Maybe:
```suggestion
        proof_height: RawHeight,
```

modules/src/ics02_client/height.rs
```diff
@@ -0,0 +1,135 @@
+use serde_derive::{Deserialize, Serialize};
+use std::convert::TryFrom;
+
+use tendermint_proto::DomainType;
+
+use crate::ics02_client::error::{Error, Kind};
+use ibc_proto::ibc::client::Height as RawHeight;
+
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct Height {
```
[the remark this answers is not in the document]
romac: I too was wondering where the `version_number` and `version_height` fields names came from.

modules/src/proofs.rs
romac: 6b18f9f

modules/src/lib.rs
romac: 6b18f9f

modules/src/ics07_tendermint/consensus_state.rs
romac: 6b18f9f

modules/src/ics03_connection/msgs/conn_open_ack.rs
romac: 44490c4

modules/src/ics03_connection/error.rs
romac: 6b18f9f

modules/src/ics02_client/height.rs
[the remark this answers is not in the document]
ancazamfir: This was changed in the latest protos. But need #292 done first.

modules/src/ics02_client/height.rs
```diff
@@ -0,0 +1,135 @@
...
+            f,
+            "epoch: {}, height: {}",
+            self.version_number, self.version_height
+        )
+    }
+}
+
+/// is_epoch_format() checks if a chain_id is in the format required for parsing epochs
+/// The chainID must be in the form: `{chainID}-{version}
+fn is_epoch_format(chain_id: String) -> bool {
```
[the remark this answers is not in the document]
ancazamfir: Not afaik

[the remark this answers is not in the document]
adizere: Then is this requirement we invented ourselves? In any case, if this is a requirement on the name of a chain, it probably should be captured and standardized in cosmos/ics.

Perhaps @cwgoes knows more?

adizere (approved):
Thanks Anca & Romain for the quick fixes and feedback. Looking forward to getting this in master!

modules/src/ics02_client/height.rs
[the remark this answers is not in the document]
ancazamfir: We do the same handling as in cosmos-SDK but I agree it should be standardized.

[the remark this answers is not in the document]
colin-axner: This is a requirement we initially came up with when implementing epochs for the tendermint light client on the SDK's IBC implementation. The issue is we need the epoch number to be included in the signed header. The solution is to put it in the chainID, but this requires some standardized parsing format so we can obtain the epoch number from the chainID. 

ref: [issue](https://github.com/cosmos/cosmos-sdk/issues/7422), [comment](https://github.com/cosmos/cosmos-sdk/pull/7280#discussion_r488538877)

We also [renamed epoch to version](https://github.com/cosmos/cosmos-sdk/issues/7324) as per a [suggestion](https://github.com/cosmos/cosmos-sdk/issues/7194#issuecomment-687844241)

romac:
Let's merge this after #292.

modules/src/ics04_channel/msgs.rs
```diff
@@ -468,7 +468,7 @@ impl MsgTimeout {
         packet: Packet,
         next_sequence_recv: Option<u64>,
         proof: CommitmentProof,
-        proof_height: Height,
+        proof_height: RawHeight,
```
adizere, line 471: Minor comment: it's a bit odd to see raw type side by side with domain type. I guess we could have just as well used `crate::Height` instead of `ibc_proto::ibc::core::client::v1::Height as RawHeight;` unless I'm missing something.

We can fix this in a later iteration, and perhaps also split the modules/src/ics04_channel/msgs.rs file like we did in ICS3.

adizere (approved):

modules/src/ics04_channel/msgs.rs
romac: Agreed, let’s use `crate::Height` instead.

romac (approved):

romac (approved):

modules/src/ics02_client/height.rs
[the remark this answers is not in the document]
cwgoes: Aye. Admittedly this is a bit of a hack, but it's a compromise to make future (safe) upgrades easier.

We can add these details to the ICS standards as well.

Issue #308, opened 2020-10-14
Client queries broke after last protobuf updates


ancazamfir:
## Steps to Reproduce
Run the Rust relayer client queries against Gaiad/ Go Relayer (stargate-3 or stargate-4). The output will include something like:
```
UnknownConsensusStateType("/ibc.lightclients.tendermint.v1.ConsensusState"))
```
This is because `type-url`s have changed
____

## For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

Issue #254, opened 2020-09-24
Extract the type_url explicit strings into constants


ancazamfir:
## Summary
With the migration to the new protobufs we need to keep track of the `type_url` used by `prost_types::Any`.
Currently these are used in clear, e.g.:
https://github.com/informalsystems/ibc-rs/blob/3e60e67b679ffab65b1a173b1af12411eb08d4c5/modules/src/ics02_client/client_def.rs#L128-L130

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
We should use the fn `type_url` from trait Msg to encapsulate all calls to type_url:

https://github.com/informalsystems/ibc-rs/blob/12cb1d6187e878eee0e40295fe0b804ddef0b846/modules/src/tx_msg.rs#L21-L23

Currently this is only implemented for a handful of type, e.g., MsgCreateAnyClient:

https://github.com/informalsystems/ibc-rs/blob/12cb1d6187e878eee0e40295fe0b804ddef0b846/modules/src/ics02_client/msgs/create_client.rs#L81-L83

ancazamfir:
This has been taken care over a number of PRs (see git blame :))

PR #309, opened 2020-10-14
Fix Client queries after last protobuf updates

ancazamfir:
Closes: #308

## Description
Fixes the `type-url`s to be consistent with latest protos which are used in stargate-4. 
Added constants for the client related URLs. These may change with fix for #254

______

For contributor use:

- [ ] Unit tests written

### DIFF
--- a/modules/src/ics02_client/client_def.rs
+++ b/modules/src/ics02_client/client_def.rs
@@ -18,6 +18,12 @@ use crate::Height;
 
 use tendermint_proto::{DomainType, Error, Kind};
 
+pub const TENDERMINT_CLIENT_STATE_TYPE_URL: &str = "/ibc.lightclients.tendermint.v1.ClientState";
+pub const TENDERMINT_CONSENSUS_STATE_TYPE_URL: &str =
+    "/ibc.lightclients.tendermint.v1.ConsensusState";
+pub const MOCK_CLIENT_STATE_TYPE_URL: &str = "/ibc.mock.ClientState";
+pub const MOCK_CONSENSUS_STATE_TYPE_URL: &str = "/ibc.mock.ConsensusState";
+
 #[cfg(test)]
 use {
     crate::mock_client::client_def::MockClient,
@@ -137,12 +143,12 @@ impl TryFrom<Any> for AnyClientState {
     // TODO Fix type urls: avoid having hardcoded values sprinkled around the whole codebase.
     fn try_from(raw: Any) -> Result<Self, Self::Error> {
         match raw.type_url.as_str() {
-            "/ibc.tendermint.ClientState" => Ok(AnyClientState::Tendermint(
+            TENDERMINT_CLIENT_STATE_TYPE_URL => Ok(AnyClientState::Tendermint(
                 TendermintClientState::decode_vec(&raw.value)?,
             )),
 
             #[cfg(test)]
-            "/ibc.mock.ClientState" => Ok(AnyClientState::Mock(MockClientState::decode_vec(
+            MOCK_CLIENT_STATE_TYPE_URL => Ok(AnyClientState::Mock(MockClientState::decode_vec(
                 &raw.value,
             )?)),
 
@@ -157,12 +163,12 @@ impl From<AnyClientState> for Any {
     fn from(value: AnyClientState) -> Self {
         match value {
             AnyClientState::Tendermint(value) => Any {
-                type_url: "/ibc.tendermint.ClientState".to_string(),
+                type_url: TENDERMINT_CLIENT_STATE_TYPE_URL.to_string(),
                 value: value.encode_vec().unwrap(),
             },
             #[cfg(test)]
             AnyClientState::Mock(value) => Any {
-                type_url: "/ibc.mock.ClientState".to_string(),
+                type_url: MOCK_CLIENT_STATE_TYPE_URL.to_string(),
                 value: value.encode_vec().unwrap(),
             },
         }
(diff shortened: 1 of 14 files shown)


ancazamfir:
Tested the client queries manually.

```
$ rrly -c simple_config.toml query client state ibc0 ibconeclient
    Finished dev [unoptimized + debuginfo] target(s) in 0.31s
     Running `target/debug/relayer -c simple_config.toml query client state ibc0 ibconeclient`
     Options QueryClientStateOptions { client_id: ClientId("ibconeclient"), height: 0, proof: true }
[relayer/src/chain/cosmos.rs:65] "Todo: implement proof verification." = "Todo: implement proof verification."
client state query result:  Tendermint(ClientState { chain_id: "ibc1", trusting_period: 1209600s, unbonding_period: 1814400s, max_clock_drift: 600s, frozen_height: Height { version_number: 0, version_height: 0 }, latest_height: Height { version_number: 0, version_height: 34 }, upgrade_path: "upgrade/upgradedClient", allow_update_after_expiry: false, allow_update_after_misbehaviour: false })

$ rrly -c simple_config.toml query client consensus ibc0 ibconeclient 0 34
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/relayer -c simple_config.toml query client consensus ibc0 ibconeclient 0 34`
     Options QueryClientConsensusOptions { client_id: ClientId("ibconeclient"), consensus_epoch: 0, consensus_height: 34, height: 0, proof: true }
[relayer/src/chain/cosmos.rs:65] "Todo: implement proof verification." = "Todo: implement proof verification."
client consensus state query result:  Tendermint(ConsensusState { timestamp: Time(2020-10-14T10:07:21.778102Z), root: CommitmentRoot([126, 130, 139, 102, 89, 141, 92, 76, 215, 119, 177, 64, 230, 228, 79, 125, 21, 58, 215, 160, 59, 167, 17, 196, 224, 107, 245, 134, 6, 183, 17, 140]), next_validators_hash: Hash::Sha256(44C5CB324AB86ADEE044B8CC0A569608A85AF549C0487405442843E2D88AF99A) })
```

modules/src/ics02_client/client_def.rs
```diff
@@ -18,6 +18,12 @@ use crate::Height;
 
 use tendermint_proto::{DomainType, Error, Kind};
 
+pub const TENDERMINT_CLIENT_STATE_TYPE_URL: &str = "/ibc.lightclients.tendermint.v1.ClientState";
+pub const TENDERMINT_CONSENSUS_STATE_TYPE_URL: &str =
+    "/ibc.lightclients.tendermint.v1.ConsensusState";
+pub const MOCK_CLIENT_STATE_TYPE_URL: &str = "/ibc.mock.ClientState";
+pub const MOCK_CONSENSUS_STATE_TYPE_URL: &str = "/ibc.mock.ConsensusState";
```
ancazamfir, line 25: Probably not the ideal way to define these, let me know what you think. Maybe we leave this as is and address it in #254?

modules/src/ics02_client/msgs.rs
```diff
@@ -92,7 +92,7 @@ impl TryFrom<RawMsgCreateClient> for MsgCreateAnyClient {
             .ok_or_else(|| Kind::DecodeMessage.context(error::Kind::InvalidRawClientState))?;
 
         let client_type = match raw_client_state.type_url.as_str() {
-            "/ibc.tendermint.ClientState" => Ok(ClientType::Tendermint),
+            "/ibc.lightclients.tendermint.v1.ClientState" => Ok(ClientType::Tendermint),
```
romac, line 95: Can we extract this string as a constant as part of this PR already?

romac (changes requested):

modules/src/ics02_client/msgs.rs
ancazamfir: Oops...forgot this one. Actually I had this on my mind for a while, do we still need the `client_type` in `MsgCreateAnyClient`? We do the same match when we decide on the enum type for `AnyClientState` which is included there. And looking at the code we only initialize this field.

romac: Good point! It seems we can indeed remove it and replace it with a function which does the match you suggest.

romac: By the way, we should actually enforce in the TryFrom instance that the `client_state` we are getting in the raw message matches the type of client state, even if we don't store, but for ensuring that the raw message we are getting is consistent.

adizere (commented):
Waiting for the last changes.

modules/src/ics02_client/context.rs
```diff
@@ -22,32 +22,25 @@ pub trait ClientKeeper {
     fn store_client_result(&mut self, handler_res: ClientResult) -> Result<(), Error> {
         match handler_res {
             CreateResult(res) => {
-                self.store_client_type(res.client_id.clone(), res.client_type)?;
```
adizere, line 25: Not an issue, just a remark: 
ICS024 [mentions explicitly](https://github.com/cosmos/ics/tree/master/spec/ics-024-host-requirements#path-space) a "path" in the provable store at key `"clients/{identifier}/clientType"`. It seems we will no longer have that path in our provable store. I guess we can still prove client type based on what is stored in client state (at key `"clients/{identifier}/clientState"`).

modules/src/ics02_client/error.rs
```diff
@@ -39,6 +39,12 @@ pub enum Kind {
     #[error("invalid raw client consensus state")]
     InvalidRawConsensusState,
 
+    #[error("mismatch  raw client consensus state")]
```
adizere, line 42: Minor: there's double spaces `mismatch  raw` in there.

modules/src/ics02_client/handler/create_client.rs
```diff
@@ -21,33 +19,24 @@ pub fn process(
...
-        return Err(Kind::ClientAlreadyExists(client_id).into());
+    if ctx.client_state(&msg.client_id()).is_some() {
+        return Err(Kind::ClientAlreadyExists(msg.client_id()).into());
     }
 
     output.log("success: no client state found");
 
-    if ctx.client_type(&client_id).is_some() {
-        return Err(Kind::ClientAlreadyExists(client_id).into());
+    if ctx.client_type(&msg.client_id()).is_some() {
```
adizere, line 28: We might consider removing this check, since it's redundant in our implementation. Though it doesn't hurt keeping it.

adizere (approved):
Just minor comments & remarks. Looks ready!

modules/src/ics02_client/context.rs
ancazamfir: oy! good point!!! added back.

modules/src/ics02_client/client_def.rs
```diff
@@ -256,7 +266,7 @@ impl From<AnyConsensusState> for Any {
 
 impl ConsensusState for AnyConsensusState {
     fn client_type(&self) -> ClientType {
-        todo!()
+        self.client_type()
```
romac, line 269: Isn't this a recursive call the the same method?

romac (commented):
Looks great! I just have one question inline above.

modules/src/ics02_client/client_def.rs
ancazamfir: I think it calls the `AnyConsensusState` impl at: https://github.com/informalsystems/ibc-rs/blob/9f661824d02290ea67fdf0fbcfa5c9752e681989/modules/src/ics02_client/client_def.rs#L136
Btw, we have the same pattern for `latest_height()`

romac: TIL that's actually not a recursive call: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d4a17a4674079b43f8c460d113a2706f

romac (approved):

PR #348, opened 2020-10-29
Merge light clients config in relayer config and add commands to add/remove light clients

romac:
## Description

This PR merge the light clients config (previously in `light_clients.toml`) in the main relayer config and add commands to add/remove light clients.

### `relayer light`

```
USAGE:
    relayer light <SUBCOMMAND>

DESCRIPTION:
    basic functionality for managing the lite clients

SUBCOMMANDS:
    add        add a light client peer for a given chain
    rm         remove a light client peer for a given chain
```

### `relayer light add`

```
USAGE:
    relayer light add <OPTIONS>

DESCRIPTION:
    add a light client peer for a given chain

POSITIONAL ARGUMENTS:
    chain_id                  identifier of the chain

FLAGS:
    -i, --peer-id PEER-ID     peer id for this client
    -p, --primary             whether this is the primary peer
    -x, --hash HASH           trusted header hash
    -a, --address ADDRESS     RPC network address
    -h, --height HEIGHT       trusted header height
```

### `relayer light rm`

```
USAGE:
    relayer light rm <OPTIONS>

DESCRIPTION:
    remove a light client peer for a given chain

POSITIONAL ARGUMENTS:
    chain_id                  identifier of the chain

FLAGS:
    -i, --peer-id PEER-ID     peer id for this client
    -f, --force               force removal of primary peer
```

## Caveat

The main downside of this PR is that running either of the commands above will remove any comments or formatting in the config file. This could be improved by using format-preserving TOML parser/serializer (see [`toml_edit`](https://docs.rs/toml_edit/0.2.0/toml_edit/)), but it would be a substantial amount of work to manually apply the config changes to the parsed TOML value with `toml_edit`. Definitely something to revisit in the future.

<del>
The main downside of this PR is that running either of the commands above will "scramble" the config file when writing the new configuration to it.

This stems from the fact I have not figured out a way to express our config structs so that [they can be serialized](https://github.com/alexcrichton/toml-rs/issues/142) to a TOML string directly via serde.

As a workaround, I am serializing the config struct to a TOML `Value` first and then serializing that to a string. Unfortunately, this does not respect the field order of the Rust structs and thus scrambles the original file (but preserves its semantics, thankfully).

I see a few potential ways out of that:

a) figure out how to express the config so that it can be directly serialized to a TOML string via Serde (no luck so far)
b) use a format-preserving TOML parser/serializer (see [`toml_edit`](https://docs.rs/toml_edit/0.2.0/toml_edit/))
c) move to another config format ([Dhall](https://dhall-lang.org), [RON](https://github.com/ron-rs/ron), ?)

### Pros
a) quick fix if we can figure it out
b) would preserve formatting and comments
c) no need to massage our config struct to fit into what's accepted by TOML

### Cons
a) comments and formatting would still be lost
b) it might take a substantial amount of work to manually apply the config changes to the parsed TOML value with `toml_edit`.
c) comments and formatting would still be lost

</del>

______

For contributor use:


ancazamfir:
My concern with these commands is the UX experience. The user has to do some work to retrieve the peer-id, hash, height, etc.
This could be done by the command itself:
```
USAGE:
    relayer-cli light add <OPTIONS>

DESCRIPTION:
    add a light client peer for a given chain

POSITIONAL ARGUMENTS:
    address                 RPC network address

FLAGS:
    -p, --primary             whether this is the primary peer
    -f, --force               allow overriding an existing peer
```
And obtain all the info from the `status()` rpc to the `address` specified, i.e. from `network`, `id`, `latest_block_hash`, `latest_block_height`.

relayer/src/config.rs
```diff
@@ -78,11 +89,28 @@ pub struct ChainConfig {
     pub client_ids: Vec<String>,
     #[serde(default = "default::gas")]
     pub gas: u64,
+    #[serde(default = "default::clock_drift", with = "humantime_serde")]
+    pub clock_drift: Duration,
     #[serde(default = "default::trusting_period", with = "humantime_serde")]
     pub trusting_period: Duration,
+    #[serde(default)]
+    pub trust_threshold: TrustThreshold,
```
ancazamfir, line 97: Should this be in the client config? Same question for the `trusting_period`? Different apps may use different clients of same chain if they have different security requirements.

romac:
Good point! Couple questions:

1. Is there any value in letting the user specify the height they want to trust?
2. What do you think of having the command print out the details it fetched from the node and ask for confirmation? Is that overkill?

ancazamfir:
> Good point! Couple questions:
> 
> 1. Is there any value in letting the user specify the height they want to trust?

I think the user should be able to specify any number of params, at the minimum the rpc `address`. If `height` is specified then it should be used. If hash is specified then height must also be. But to start with, if we can allow just the `address` it will go a long way, especially in the development phase, CI integration and initial release with lower focus on security.

> 2. What do you think of having the command print out the details it fetched from the node and ask for confirmation? Is that overkill?

That would be nice :)

romac:
>  But to start with, if we can allow just the address it will go a long way, especially in the development phase, CI integration and initial release with lower focus on security.

Done.

---

Here's the new usage of the `light add` command:

```
USAGE:
    relayer-cli light add <OPTIONS>

DESCRIPTION:
    add a light client peer for a given chain

POSITIONAL ARGUMENTS:
    address                   RPC network address

FLAGS:
    -c, --chain-id CHAIN-ID   identifier of the chain
    -p, --primary             whether this is the primary peer
    -f, --force               allow overriding an existing peer
    -y, --yes                 skip confirmation
```

I also adapted the options for the `light rm` command for consistency with the new `light add` options:

```
USAGE:
    relayer-cli light rm <OPTIONS>

DESCRIPTION:
    remove a light client peer for a given chain

POSITIONAL ARGUMENTS:
    peer_id                   peer id for this client

FLAGS:
    -c, --chain-id CHAIN-ID   identifier of the chain
    -f, --force               force removal of primary peer
```

ancazamfir (approved):
Looks great! Thanks Romain!

relayer/src/config.rs
romac: @ebuchman ^

romac: Tracked in #356

Issue #277, opened 2020-09-30
Relayer CLIs for client update message


adizere:
## Summary

Implement the command: 

- `relayer tx raw update-client <to-chain-B> <chain-A> <client-id-A>  -h <A-height>` 
  - updates the client of `chain-A` with id `client-id-A` installed on chain `to-chain-B` with the header/consensus state at height `A-height`
            - Note: if `-h` is not specified, or `0`, the latest header from local light client for `chain-A` should be used
  - detailed:
          - sends request to local relayer light client for `chain-A` to retrieve header `header_A` for `A-height` or latest
          - sends a `MsgUpdateClient` to `chain-B` with:
```
    client_id = client-id-A,
    header=header_A,
    signer: get_account(chain-B),
``` 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ancazamfir:
A few dependencies for this one: #47 (almost done), #336 and https://github.com/informalsystems/tendermint-rs/issues/634
Moving to v0.0.6

Issue #338, opened 2020-10-25
Rename parts to part_set_header


adizere:
## Crate

modules

## Summary of Bug

In the upcoming version of tendermint-rs (most likely RC2) the `parts` field in block_id [is renamed to](https://github.com/informalsystems/tendermint-rs/commit/7b888b33f4233699547b8bcadec0f818449efb27) `part_set_header`.
Some of our tests will fail once we start depending on this version.

Part of the fix will involve adapting a JSON file we use in our tests. See [here the diff](https://gist.github.com/adizere/de20784f3dbeb233a5f20138e53e0804) on how the fix should look like. 

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #219, opened 2020-09-03
Remove `signer` field from domain type messages


adizere:
> I think `AccAddress` was supposed to be the type for field signer in MsgConnectionOpenInit (and other messages). But seems we currently use `AccountId` (imported from tendermint) for this purpose.
> We should probably replace AccountId with AccAddress ?

> Yes, we need to make this change. Didn't do it here because it touches all messages. We don't do anything with this field in `Msg` in IBC modules. It is going to be handled up the stack once we have auth, accounts, etc. Can you open an issue for this?

_Originally posted by @ancazamfir in https://github.com/informalsystems/ibc-rs/pull/188#discussion_r482834495_

adizere:
Following-up on work in #355, @ancazamfir made the observation that modules have no need to access the signer in any way. The decisions then is to remove this field from the Domain types which have it, and let the relayer logic add the signer directly in the raw types.


## Original discussion

Yeah, this caused us some grief for nothing. I propose we remove the typed `signer` from IBC domain type structures as it is never really used there.  We should either remove it altogether or make it a `String`. But this should happen in a different issue, maybe together with @andynog's work on the `keystore`.
The only time this is written is when the relayer sends a message. The signer should be build from the `keystore`. When the relayer builds the domain type message it doesn't need the `signer`. The only place it is required after, is in the raw message created from the domain type message. The relayer could just add it the raw message at that point.
The place where the signer is read on the handlers side is in some future module that will do all the account, auth, etc. verification. This will happen before the IBC handlers are invoked and they make no use for the signer.

_Originally posted by @ancazamfir in https://github.com/informalsystems/ibc-rs/pull/355#discussion_r516610337_

adizere:
At the same time with removing the `signer` field, we should also introduce proper constructors with `new` for domain type messages, cf [this comments](https://github.com/informalsystems/ibc-rs/pull/388#discussion_r520414557).

PR #355, opened 2020-11-02
Implement client update and fix client create CLIs

ancazamfir:
Closes: #277 #338

## Description
Here is a summary of the changes:
- pointing to branch `greg/research-json` where the tendermint changes for stargate-4 are and tendermint domain types are implemented for the IBC dependencies (validator set, consensus params, etc)
- removed unneeded serde (de)serializers
- refactored a bit the conversion between accountId and String. We should change the signer type to String in handlers as any validation and use of this happens before the hanlers are invoked. Code will be simpler and we could remove the `address.rs`
- defined domain types for AnyHeader, MsgUpdateAnyClient, ics07 header,
- update tendermint client_state with stargate-4 required fields and impls
- implement get_sign_bytes() in Msg trait
- add update client CLI
- refactor the Tx CLIs to take typed input, remove Option from positional params.
- add client message builder to chain impl of cosmos
- add connection init message builder to chain trait
- some refactor of chain methods. We should sit down and review what goes in `Chain` vs `impl Chain for CosmosSDK` vs `impl CosmosSDK`. I guess these would move to the `ChainHandle` but the approach should be the same.
- some changes in tests and refactor of the different test utils.

TODO:
- update changelog
- split client `msgs.rs` file before merging. -> deferred to https://github.com/informalsystems/ibc-rs/issues/367

______

For contributor use:


modules/src/tx_msg.rs
```diff
@@ -1,6 +1,7 @@
+use prost_types::Any;
 use tendermint::account::Id as AccountId;
 
-pub trait Msg: Sized {
+pub trait Msg: Clone {
```
romac, line 4: This `Clone` bound might cause issue if we ever need to deal with `Box<dyn Msg>` values as traits with `Clone` bounds cannot be made in trait objects. Not a problem at the moment, just pointing this out in case that comes up later on.

relayer/src/chain.rs
```diff
@@ -1,60 +1,71 @@
+use std::convert::TryFrom;
 use std::time::Duration;
 
 use anomaly::fail;
 use prost_types::Any;
 use serde::{de::DeserializeOwned, Serialize};
 
+use tendermint_proto::DomainType;
+
+/// TODO - tendermint deps should not be here
```
romac, line 10:
```suggestion
// TODO - tendermint deps should not be here
```

relayer/src/tx/client.rs
```diff
@@ -1,125 +1,95 @@
...
-        ClientStatePath(opts.clone().dest_client_id),
-        tendermint::block::Height::from(0_u32),
-        false,
-    );
-
-    if response.is_ok() {
+    // Verify that the client has not been created already, i.e the destination chain does not
+    // have a state for this client.
+    let client_state =
+        dest_chain.query_client_state(&opts.dest_client_id, 0_u64.try_into().unwrap(), false);
```
romac, line 50:
```suggestion
        dest_chain.query_client_state(&opts.dest_client_id, 0_u32.into(), false);
```

relayer/src/tx/client.rs
```diff
@@ -1,125 +1,95 @@
...
-        Kind::CreateClient(opts.dest_client_id.clone(), "bad signer".into()).context(e)
-    })?;
-
-    // Build the domain type message
-    let new_msg = MsgCreateAnyClient::new(
+    let mut dest_chain = CosmosSDKChain::from_config(opts.clone().dest_chain_config)?;
+
+    // Get the latest trusted height from the client state on destination.
+    let trusted_height = dest_chain
+        .query_client_state(&opts.dest_client_id, 0_u64.try_into().unwrap(), false)?
```
romac, line 76:
```suggestion
        .query_client_state(&opts.dest_client_id, 0_u32.into(), false)?
```

modules/src/ics02_client/msgs.rs
```diff
@@ -126,56 +130,109 @@ impl From<MsgCreateAnyClient> for RawMsgCreateClient {
             client_id: ics_msg.client_id.to_string(),
             client_state: Some(ics_msg.client_state.into()),
             consensus_state: Some(ics_msg.consensus_state.into()),
-            signer: ics_msg.signer.to_string(),
+            signer: account_to_string(ics_msg.signer).unwrap(),
```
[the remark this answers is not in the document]
ancazamfir: Yeah, this caused us some grief for nothing. I propose we remove the typed `signer` from IBC domain type structures as it is never really used there.  We should either remove it altogether or make it a `String`. But this should happen in a different issue, maybe together with @andynog's work on the `keystore`.
The only time this is written is when the relayer sends a message. The signer should be build from the `keystore`. When the relayer builds the domain type message it doesn't need the `signer`. The only place it is required after, is in the raw message created from the domain type message. The relayer could just add it the raw message at that point.
The place where the signer is read on the handlers side is in some future module that will do all the account, auth, etc. verification. This will happen before the IBC handlers are invoked and they make no use for the signer.

relayer-cli/src/commands/tx/client.rs
```diff
@@ -102,3 +58,97 @@ impl Runnable for TxCreateClientCmd {
...
+            Ok(result) => result,
+        };
+        status_info!("Message", "{:?}", opts);
+
+        let res: Result<Vec<u8>, Error> =
+            update_client(opts).map_err(|e| Kind::Tx.context(e).into());
+
+        match res {
+            Ok(receipt) => status_info!("client updated, result: ", "{:?}", receipt),
+            Err(e) => status_info!("client update failed, error: ", "{}", e),
```
romac, line 108:
```suggestion
            Ok(receipt) => status_ok!("Success", "client updated: {:?}", receipt),
            Err(e) => status_err!("client update failed: {}", e),
```

relayer-cli/src/commands/tx/client.rs
```diff
@@ -102,3 +58,97 @@ impl Runnable for TxCreateClientCmd {
...
+        .ok_or_else(|| "missing destination chain configuration".to_string())?;
+
+    let src_chain_config = config
+        .chains
+        .iter()
+        .find(|c| c.id == src_chain_id)
+        .ok_or_else(|| "missing source chain configuration".to_string())?;
+
+    let signer_seed = std::fs::read_to_string(seed_file)
+        .map_err(|e| "invalid signer seed file".to_string() + &e.to_string())?;
```
romac, line 145:
```suggestion
        .map_err(|e| anomaly::Context::new("invalid signer seed file", Some(e.into()))?;
```

relayer-cli/src/commands/tx/connection.rs
```diff
@@ -1,128 +1,70 @@
...
-        let dest_connection_id = self
-            .dest_connection_id
-            .as_ref()
-            .map(|v| {
-                v.parse()
-                    .map_err(|_| "bad destination connection identifier".to_string())
-            })
-            .transpose()?;
+        let signer_seed = std::fs::read_to_string(&self.seed_file)
+            .map_err(|e| "invalid signer seed file".to_string() + &e.to_string())?;
```
romac, line 57:
```suggestion
        .map_err(|e| anomaly::Context::new("invalid signer seed file", Some(e.into()))?;
```

romac:
@ancazamfir Amazing job by the way :)

relayer/src/chain/cosmos.rs
```diff
@@ -209,31 +284,120 @@ impl Chain for CosmosSDKChain {
...
+        TrustThreshold {
+            numerator: 1,
+            denominator: 3,
+        }
     }
 
     fn unbonding_period(&self) -> Duration {
         // TODO - query chain
-        Duration::from_secs(24 * 7 * 3)
+        Duration::from_secs(24 * 7 * 3 * 3600)
```
adizere, line 297: As discussed, unclear if we should keep these in the chain trait. Just to keep in mind for future work here.

relayer/src/tx/connection.rs
```diff
@@ -1,72 +1,64 @@
...
-pub fn conn_init(opts: ConnectionOpenInitOptions) -> Result<Vec<u8>, Error> {
-    // Get the destination chain
-    let mut dest_chain = CosmosSDKChain::from_config(opts.clone().dest_chain_config)?;
+pub fn conn_init(opts: &ConnectionOpenInitOptions) -> Result<Vec<u8>, Error> {
+    // Get the source and destination chains
+    let src_chain = CosmosSDKChain::from_config(opts.src_chain_config.clone())?;
+    let mut dest_chain = CosmosSDKChain::from_config(opts.dest_chain_config.clone())?;
+
+    // Check that the destination chain will accept the message, i.e. it does not have the connection
+    dest_chain.check_connection_for_init(opts.dest_connection_id.clone())?;
```
adizere, line 40: This method is quite fine-grained to keep in the chain trait. It would make sense to rely on a more generic method – simply `query_connection` – to implement this check, but let's consider this in a follow-up PR!

adizere (commented):
Cool stuff!

Left a few ideas on how we can continue this work + do a bit of cleanup, but this is almost ready.

modules/src/address.rs
```diff
@@ -1 +1,23 @@
...
+
+use std::convert::TryFrom;
+
+use anomaly::{BoxError, Context};
+use bech32::FromBase32;
+use bech32::ToBase32;
+
+use tendermint::account::Id as AccountId;
+
+pub fn account_to_string(addr: AccountId) -> Result<String, BoxError> {
```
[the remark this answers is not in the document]
ancazamfir: Let's discuss this there. Not sure about the benefit of having a new type since we completely ignore this field in the IBC handlers (other than converting it back and forth).

modules/src/ics03_connection/connection.rs
```diff
@@ -118,7 +117,7 @@ impl ConnectionEnd {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Eq)]
```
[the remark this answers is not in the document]
ancazamfir: this was added by @Romain

modules/src/ics03_connection/msgs.rs
```diff
@@ -25,40 +25,20 @@ pub mod conn_open_try;
 pub const TYPE_MSG_CONNECTION_OPEN_CONFIRM: &str = "connection_open_confirm";
 
 /// Enumeration of all possible messages that the ICS3 protocol processes.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub enum ConnectionMsg {
     ConnectionOpenInit(MsgConnectionOpenInit),
     ConnectionOpenTry(Box<MsgConnectionOpenTry>),
-    ConnectionOpenAck(MsgConnectionOpenAck),
+    ConnectionOpenAck(Box<MsgConnectionOpenAck>),
```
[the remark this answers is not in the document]
ancazamfir: Yes, forgot to mention, `Ack` is also too large now because tendermint client state (included in the message) has new fields (e.g.consensus params, proofOps). And we don't allow `large_enum_variant` for `ConnectionMsg`. It would be better imo to allow it and to remove `Box` from `Ack` and `Try` as they make 50% of all variants. Let me know what you think.

modules/src/ics07_tendermint/client_state.rs
```diff
@@ -142,6 +155,11 @@ impl TryFrom<RawClientState> for ClientState {
             upgrade_path: raw.upgrade_path,
             allow_update_after_expiry: raw.allow_update_after_expiry,
             allow_update_after_misbehaviour: raw.allow_update_after_misbehaviour,
+            consensus_params: raw
+                .consensus_params
+                .ok_or_else(|| Kind::InvalidRawClientState.context("missing consensus parameters"))?
+                .try_into()
+                .map_err(|_| Kind::InvalidRawHeight)?,
```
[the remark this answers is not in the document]
ancazamfir: Good one, will fix it.

modules/src/ics24_host/identifier.rs
```diff
@@ -101,6 +101,12 @@ impl FromStr for ClientId {
     }
 }
 
+impl Default for ClientId {
```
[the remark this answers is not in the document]
ancazamfir: This is required by abscissa for `Options` and because the positional fields not being `Option` i think. Didn't look much into it to be honest, just reacted to the compiler complaint.

relayer/src/tx/connection.rs
ancazamfir: Yes it is. I made this change after looking into the `OpenTry` message. Checking if the connection state is compatible with a certain handshake step is more complex for some messages, especially with flexible connection id. I think we will revise these methods once we have all connection tx-es implemented.

modules/src/address.rs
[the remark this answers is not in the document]
adizere: Yes! Since handlers/modules ignore the signer, it is clearly best to remove this field altogether. We can do that as part of #219.

modules/src/ics03_connection/msgs.rs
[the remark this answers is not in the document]
adizere: Ah I suspected some field must have grown in the meantime, but I missed which one exactly. Thanks for the explanation!

I have no particular opinion about introducing a `allow(clippy::..)` here. In the interest of merging this soon, maybe we keep things as they are currently. In future, I think this deserves a discussion about clippy in general, if we want to have a repo-wide configuration where we adjust the clippy parameters (suggested in [another comment](https://github.com/informalsystems/ibc-rs/pull/355#discussion_r516654739)) so we can remove the allow/clippy annotations altogether.

modules/src/ics03_connection/connection.rs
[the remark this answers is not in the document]
romac: It does not matter too much, but if a datatype form a total order (as opposed to a partial one), it's usually good practice to add an `Eq` bound.

NB: Most types, with the notable exception of floats (`f32`, `f64`, etc.) form a total order.

adizere (approved):
Thanks Anca for the replies & updates. Looks good!

relayer/src/tx/connection.rs
ancazamfir: Ok i removed this for now and also took out the message builder from chain.

relayer/src/chain/cosmos.rs
ancazamfir: Moved to `impl CosmosSDK`

Issue #386, opened 2020-11-09
TLA+ specification for ICS-020


istoilkovska:
## Summary

Write a TLA+ specification of the Fungible Token Transfer Protocol in IBC ([ICS-020](https://github.com/cosmos/ics/tree/master/spec/ics-020-fungible-token-transfer)).

## Proposal

- Encode the token transfer application logic using TLA+. 
- Use the existing encoding of the basic packet flow in order to be able to send packet datagrams with ICS-020 application data in the data field of the IBC packets.
- Define basic functionalities of the bank application using TLA+.
____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #387, opened 2020-11-09
TLA+ specification of the IBC Fungible Token Transfer Protocol

istoilkovska:
Closes: #386

## Description

A TLA+ specification of the IBC Fungible Token Transfer Protocol ([ICS-020](https://github.com/cosmos/ics/tree/master/spec/ics-020-fungible-token-transfer)).

The specification has the following modules: 
  - `ICS20Environment.tla` (the main module, creates instances of 
`ICS20Chain.tla` and encodes the relayer logic)
  - `ICS20Chain.tla` (captures the chain logic relevant to ICS-020)
  - `ICS20Definitions.tla` (defines operators that are used across all modules)
  - `PacketHandlers.tla`
  - `FungibleTokenTransferHandlers.tla`
  - `Bank.tla`

The modules `PacketHandlers.tla` and `FungibleTokenTransferHandlers.tla` contain definition of operators that handle packet datagrams and encode the token transfer application logic, respectively. 
The module `Bank.tla` encodes operators defined by the bank application.

## Assumptions/current state
1. We assume that initially the channel handshake is open. 
2. Timeouts are not captured currently.
3. Properties and invariants are still to be expressed.
______

For contributor use:


josef-widder (approved):
Look great!

I guess for getting tools to do generate traces for testing and verification, we should abstract over the chainstore. (I dump some ideas here, they are not blocking this PR)

In principle for each chain, we just need to store a set of pending packets (or incoming data to createOutgoingPacket) together with the function they are called with.

After a createOutgoingPacket on chain A either
  - a packet for recvpacket is added to the set on chain B, 
  - or a packet for timeout is added at chain A

If a recvpacket happens at chain A, an acknowledgemant is added to chain B

For liveness every (packet, function) pair eventually is executed

the state on each chain is just the bank, and we write specifications over them.

@andrey-kuprianov and @istoilkovska : Would it make sense to generate traces in such an abstract model (as sketched), and then link it to the model in this PR to get more details into test scenarios (filling all the data we have abstracted away)? Let's discuss this synchronously.

docs/spec/fungible-token-transfer/Bank.tla
```diff
@@ -0,0 +1,89 @@
+-------------------------------- MODULE Bank --------------------------------
+
+EXTENDS Integers, FiniteSets
+
+\* subtract coins from account
+SubtractCoins(accounts, accountID, amount) ==
```
milosevic, line 6: I think we will need to add denoms also to the picture. So when you subtract you also need to specify denom as accounts are multi-denomed one.

docs/spec/fungible-token-transfer/Bank.tla
```diff
@@ -0,0 +1,89 @@
...
+    [accounts EXCEPT ![accountID] = accounts[accountID] - amount]
```
milosevic, line 7: What if amount is bigger than what is on account?

docs/spec/fungible-token-transfer/FungibleTokenTransferHandlers.tla
```diff
@@ -0,0 +1,190 @@
...
+ICS20OnPacketRecv(chainID, chain, accounts, packet) ==
+    \* get packet data and denomination
+    LET data == packet.data IN
+    LET denomination == data.denomination IN 
+    
+    LET escrowAccounts == chain.escrowAccounts IN
+    
+    \* receiving chain is source if 
+    \* the denomination is prefixed by srcChannelID
+    LET source == (Head(denomination) = packet.srcChannelID) IN
```
milosevic, line 75: Prefix should be portId + channelId?

docs/spec/fungible-token-transfer/ICS20Chain.tla
```diff
@@ -0,0 +1,172 @@
...
+
+(***************************************************************************
+ Token transfer operators
+ ***************************************************************************)
+\* Create a packet: Abstract away from ports, and timestamp. 
+\* Assume timeoutHeight is MaxHeight + 1
+CreatePacket(packetData) ==
+    AsPacket([
+        sequence |-> appPacketSeq,
+        timeoutHeight |-> MaxHeight + 1,
```
milosevic, line 32: We might want to pick timeout non-deterministically.

Issue #315, opened 2020-10-15
Implement the channel message domain types


ancazamfir:
## Summary
`ics04-channel` module is incomplete wrt to domain type implementations

## Proposal
[X] Add the `From/TryFrom` impls for all messages and supporting structures where missing
[X] Restructure the tests to initialize with raw messages and test (de)serialization
[X] Split the `msgs.rs` file in per message modules

Note: this does _not_ include packet-related messages (e.g., timout, ack). https://github.com/informalsystems/ibc-rs/issues/95 is for that.

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #391, opened 2020-11-10
ICS 4 Domain Types for channel handshakes

adizere:
Closes: #315 

## Description

______

For contributor use:


modules/src/ics04_channel/msgs/chan_open_try.rs
```diff
@@ -0,0 +1,322 @@
...
+        proof_init: CommitmentProof,
+        proofs_height: Height,
+        signer: AccountId,
+    ) -> Result<MsgChannelOpenTry, Error> {
+        let connection_hops: Result<Vec<_>, _> = connection_hops
+            .into_iter()
+            .map(|s| ConnectionId::from_str(s.as_str()))
+            .collect();
+
+        let version =
```
adizere, line 50: Note to self: this was changed in a separate PR: https://github.com/informalsystems/ibc-rs/pull/388/files#diff-a9dd86861adb60ba50e159c43913d6f20a5d406115a2ef4d17c7d0725054c7cc

ancazamfir:
Looking at this now and wondering why we need `impl DomainType<RawMyType> for <MyType>` for many of these. Just the TryFrom/From would be enough. The code compiles and runs fine if we remove all except the cases where `RawMyType` is not a prost message but wire bytes. For example the client state that is stuffed into an `Any` type as Vec<u8>, or merkle proofs.

ancazamfir (approved):
Looks good, thanks Adi! Let's merge this and we'll make more changes as we implement the handlers (making use of the deserializer/ `TryFrom` part) and the relayer (used the serializer/ `From` part).

Issue #371, opened 2020-11-06
Relayer CLI for MsgChannelOpenInit


adizere:
## Crate

primarily `cli`

## Summary

We should be able to send transactions from the relayer CLI for `MsgChannelOpenInit`:

https://github.com/informalsystems/ibc-rs/blob/12cb1d6187e878eee0e40295fe0b804ddef0b846/modules/src/ics04_channel/msgs.rs#L16-L21

## Proposal

The following CLIs should be implemented by the relayer:

`relayer tx raw chan-open-init <to-chain-A> <port-id> <channel-id>`

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #373, opened 2020-11-06
Relayer CLI for MsgChannelOpenAck


adizere:
## Crate

primarily `cli`

## Summary

We should be able to send transactions from the relayer CLI for `MsgChannelOpenTry`:

https://github.com/informalsystems/ibc-rs/blob/12cb1d6187e878eee0e40295fe0b804ddef0b846/modules/src/ics04_channel/msgs.rs#L158-L164

## Proposal

The following CLIs should be implemented by the relayer:

`relayer tx raw chan-open-ack <to-chain-A> <port-id> <channel-id>`

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ebuchman:
Believe this is closed by #403

PR #403, opened 2020-11-17
Channel open handshake message relaying/CLIs

ancazamfir:
Closes: #371 #372 #373 #374

## Description
- CLIs for channel open handshake messages
- building and relaying of channel open handshake messages

[will revert the Cargo.toml changes for tendermint dependencies after more testing]
______

For contributor use:


relayer-cli/src/commands/tx/channel.rs
```diff
@@ -0,0 +1,389 @@
...
+
+    #[options(
+        help = "json key file for the signer, must include mnemonic",
+        short = "k"
+    )]
+    seed_file: String,
+}
+
+impl TxRawChanConfirmCmd {
+    fn validate_options(&self, config: &Config) -> Result<ChannelOpenOptions, String> {
```
adizere, line 331: The three `validate_options` methods for try, ack, and confirm are identical. That's some potential to improve when we refactor this (unless you prefer to do it now). There is overlap across the `TxRaw..Cmd` structs also.

adizere (changes requested):
The present version seems to have compilation issues, following [recent changes](https://github.com/informalsystems/tendermint-rs/commit/bfcbf835ffe19164682605f69d389b05be684ab1#diff-5d10b574e97438514650fd6f44664299baf20536aab73554ca2754906d2a72d4) in tendermint-rs master (domaintype redefined to protobuf). So we should remember to revert cargo modifications, or adapt to these changes.

Otherwise looks good IMO!

relayer-cli/src/commands/tx/channel.rs
ancazamfir: Yes, it's the same as for connection CLIs. I am planning to look at all in a different PR.

adizere (approved):

Issue #363, opened 2020-11-05
Command to add a key to a configured chain in the relayer


andynog:
## Crate

relayer-cli 
relayer

## Summary

This feature will implement a command to the relayer that will allow users to add keys to a chain configured on the relayer. 

## Problem Definition

Right now to test a transaction command (e.g. `tx raw`) the user needs to specify a flag (-k) which is the key_seed.json file every time a transaction is submitted. When this feature implemented, this flag will not be needed anymore. The users would have to run a `keys add` command once and that will add the key to the relayer home folder.

## Proposal

- [X] Implement command to the relayer to add keys
- [X] Remove the key flag (-k) from the `tx raw` commands
- [X] The added key specified in the config should be used to sign transactions
- [X] Keys are stored in the relayer home folder (unencrypted)
- [X] Perform regression and integration test
- [X] *Bonus*: Added `keys list` command
- [X] Add instructions to add keys on README

## Assumptions

* This will not implement a keychain or keyring. 
* The keys to be imported should be generated with `gaiad` and exported (e.g. key_seed.json). 
* The key file will be stored in the relayer home folder exactly like they were exported.
* The key name should be specified in the relayer config file (.toml)
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

andynog:
Most of the logic to have this working is implemented. The only things missing is to add instructions on README and do some regression and integration tests.  If all works well will open a PR tomorrow for this.

andynog:
Tested conn-init command against a `cosmos-test-stargate` chain and using the new `keys add` command and it works. Will update the instructions on README with new way to submit transactions.

PR #408, opened 2020-11-19
Implemented 'keys add' and 'keys list' commmands

andynog:
Closes: #363 

## Description

This PR includes the implementation of the `keys add` and `keys list` commands on the relayer. Changes:

* Removed all the -k parameter from the `tx` commands. So not `key_seed.json` file needs to be specified for a `tx raw` command
* Refactored keybase logic to simplify methods, it has now the chain config so most parameters are used from the config
* Implemented a `keys add` command. Basically this adds the `key_seed.json` file on a folder `$HOME/.rrly`. The file will be saved with the `key_name` parameter specified in the config (e.g. testkey.json). 
* The key file is stored un-encrypted for now. No integration at this time with an OS keyring that would safely store the file. Once the keyring logic is properly implemented then this file can be properly stored in a secure way. For now, the purpose is just to simplify the logic for the `tx raw` commands.
* Added instructions on relayer-cli `README.md` file
* Also implemented a `keys list` command that can be used to list the key configured for a chain
_____

For contributor use:

- [X] Updated relevant documentation (`docs/`) and code comments.


relayer-cli/src/commands/query/channel.rs
```diff
@@ -121,7 +121,7 @@ mod tests {
     #[test]
     fn parse_channel_query_end_parameters() {
         let default_params = QueryChannelEndCmd {
-            chain_id: Some("ibc0".to_string().parse().unwrap()),
```
romac, line 124: Why is this needed?

relayer-cli/tests/fixtures/two_chains.toml
```diff
@@ -3,12 +3,13 @@ timeout = '10s'
 strategy = 'naive'
 
 [[chains]]
-id = 'ibc0'
+id = 'ibc-0'
```
romac, line 6: Is this change needed because the identifiers of the chains spawned by the Go relayer setup scripts have changed?

relayer/src/keyring/store.rs
```diff
@@ -127,38 +163,108 @@ impl KeyRingOperations for KeyRing {
...
+                                Kind::KeyStoreOperation.context("failed to get key entry")
+                            })?;
+                            Ok(key_entry)
+                        }
                         None => Err(Kind::InvalidKey.into()),
                     }
                 }
             }
+            KeyRing::TestKeyStore {
+                store: s,
```
romac, line 192: I would rather use the full names `store` and `chain_config` in the bound variables to improve readability a bit, here and in the other places.

relayer/src/keyring/store.rs
```diff
@@ -16,27 +17,45 @@ use k256::{
...
+pub const KEYSTORE_TEST_BACKEND: &str = "keyring-test";
+pub const KEYSTORE_FILE_EXTENSION: &str = "json";
+
 pub type Address = Vec<u8>;
 
 pub enum KeyRing {
-    MemoryKeyStore { store: BTreeMap<Address, KeyEntry> },
+    MemoryKeyStore {
+        store: BTreeMap<String, String>,
+        chain_config: ChainConfig,
```
romac, line 39: Do we really need to full chain config here? Can we instead just store the data that's actually needed from it (eg. the key name)?

romac (changes requested):
Looks good, aside from some minor remarks. Thanks :)

relayer-cli/tests/fixtures/two_chains.toml
andynog: Yes, they changed

relayer/src/keyring/store.rs
andynog: Actually I had it before with key name and chain id but it made calling the functions more complex and thought that this way was cleaner and also if we need more params in the config here in the future this would make it simpler to implement

relayer/src/keyring/store.rs
andynog: Ok, makes sense. I can change that.

andynog:
I noticed that with other PRs merged to master this PR got broken again, I will fix this one again so it can be merged

relayer-cli/tests/fixtures/two_chains.toml
ancazamfir: Hmm, i don't understand why this needs change. We are still on stargate-4. This change is coming in stargate-5.

ancazamfir (commented):
Found some issues while testing. I can only submit transactions on one chain (`ibc0` of stargate-4/ gaia setup with Go relayer). For the other one the account is not recognized. Not sure why, will try to debug.
Also not sure why the protos had changed compared with master.

andynog:
> Found some issues while testing. I can only submit transactions on one chain (`ibc0` of stargate-4/ gaia setup with Go relayer). For the other one the account is not recognized. Not sure why, will try to debug.
> Also not sure why the protos had changed compared with master.

I've changed to `ibc-0` because I was testing against the latest Go relayer release that now is on https://github.com/cosmos/relayer. For some reason their scripts to start two chains changed the chain-id to `ibc-0`. But if we are not using that to test then we can revert back to `ibc0`. 

Concerning the transactions, please make sure you add the right key for each chain. So you should have a key_seed.json for each chain, after you add them you should see something like this under your `$HOME/.rrly` folder

```
$tree $HOME/.rrly
$HOME/.rrly/
└── keys
    ├── ibc-0
    │   └── keyring-test
    │       └── testkey.json
    └── ibc-1
        └── keyring-test
            └── testkey.json
```

But each testkey.json should have different contents. You can verify by doing:

```
relayer -c relayer-cli/tests/fixtures/two_chains.toml keys list ibc-0
keys list result:  "chain: ibc-0 -> testkey (cosmos1hkza0857rpkfl4uvwptv87he5t7f9g5v5yel7r)"
```

```
relayer -c ./relayer-cli/tests/fixtures/two_chains.toml keys list ibc-1
keys list result:  "chain: ibc-1 -> testkey (cosmos1hud8teqlrzw6zstnz3vf2qns547enm26ul22fw)"
```

which should display different addresses. Then the transactions should be signed with the key for the chain (right now there's no way to use more than one key per chain). If you have the right setup and still not working then might be a bug in the logic.

relayer/src/keyring/store.rs
andynog: But I couldn't find a good pattern here. Ideally this is used as a property of the chain (`keybase`) so if the config could be "passed down" into the keybase then this `chain_config` property would not be needed.

relayer/src/keyring/store.rs
andynog: Pushed changes to fix this https://github.com/informalsystems/ibc-rs/pull/408/commits/f36389685e7ac7a992f74e82bd70f7c7322c0984

relayer-cli/tests/fixtures/two_chains.toml
andynog: do you want me to change this back to `ibc0` ?

relayer-cli/src/commands/query/channel.rs
andynog: I didn't implement this as part of my issue, maybe something merged from master? Maybe @ancazamfir knows ?

ancazamfir:
> Concerning the transactions, please make sure you add the right key for each chain. So you should have a key_seed.json for each chain, after you add them you should see something like this under your $HOME/.rrly folder

yes, I did that. All accounts and keys seem fine and I doubled checked they are the same as the ones from the Go relayer seed files. Does this work for you? Having both chains accepting transactions?

ancazamfir:
> I've changed to ibc-0 because I was testing against the latest Go relayer release that now is on https://github.com/cosmos/relayer. For some reason their scripts to start two chains changed the chain-id to ibc-0. But if we are not using that to test then we can revert back to ibc0.

Yes but for the latest (stargate-5) we need a number of changes in ibc-rs and tendermint-rs (I made changes for this on `anca/stargate-5` on both). With current we cannot do proper client, connection and channel handshakes. IIRC only ConnOpenInit works.

ancazamfir:
> > Concerning the transactions, please make sure you add the right key for each chain. So you should have a key_seed.json for each chain, after you add them you should see something like this under your $HOME/.rrly folder
> 
> yes, I did that. All accounts and keys seem fine and I doubled checked they are the same as the ones from the Go relayer seed files. Does this work for you? Having both chains accepting transactions?

I figured what went wrong. From previous testing I had some sections `[[chains.peers.light_clients]]` in the config file and for `ibc1` one of the client peers was configured with the port of `ibc0`. This caused the transaction to be sent to `ibc0`. I will still have to investigate why, as the rpc address was correctly configured and the light clients are not currently started with the CLIs.
The good news is that your changes work, I tested almost all CLIs. The problem is that master is now broken after the upgrade to tokio 0.3 (compiles but CLIs are broken)  so we should fix that before we merge your PR.

ancazamfir (approved):
Looks great! All commands work without the key seed file option, which is great! Thanks a lot Andy, let's merge this today.

romac (approved):

Issue #158, opened 2020-07-19
Local chain type


ebuchman:
We should add a new type of Chain type that is a local in-process chain. This chain will basically be our [implementation of the IBC handlers](https://github.com/informalsystems/ibc-rs/issues/119). It can also be used for in-process testing of the relayer.

A Chain is expected to be running a blockchain (generating at least light blocks ...) and to have some on-chain state that can be queried and updated by transactions. So we'll need:

- A way to produce a chain of real Tendermint light blocks (we should have all the tools for this now given the tendermint-rs model-based testing work to do this)
- An on chain kv-store. For now we should mock out proofs so we don't worry about checking them but soon we'll want to implement a basic merkle tree backing here, even a simple merkle tree that's recomputed every block. Eventually we'll use something like [merk](https://github.com/nomic-io/merk) or otherwise.
- A way to query the on chain kv-store (should roughly match the ABCI query interface)
- IBC handlers: ie. functions that take IBC msgs as input and executes them against the kv-store
- A way to send the msgs to the handlers (ie. "transactions")

So basically we need to mock out a Tendermint blockchain using light blocks and a simple state machine for the ibc handlers.

Note this wouldn't be using ABCI directly, since that would assume a running instance of Tendermint Core. So we don't need to hook up real ABCI queries or any kind of transaction stuff, just a simple interface for passing in the IBC msg types and for querying from the KV store. Ultimately though we'll want this application state/logic to be able to run on top of ABCI, so it's worth keeping in mind.

PR #422, opened 2020-11-25
Local chain implementation & basic tests

adizere:
Closes: #158 

______

For contributor use:


modules/src/ics02_client/msgs/create_client.rs
```diff
@@ -19,7 +19,7 @@ use crate::ics02_client::error::{Error, Kind};
 use crate::ics24_host::identifier::ClientId;
 use crate::tx_msg::Msg;
 
-const TYPE_MSG_CREATE_CLIENT: &str = "create_client";
+pub const TYPE_URL: &str = "/ibc.core.client.v1.MsgCreateClient";
```
adizere, line 22: FYI: Introduced these public constants in all messages to be able to use them in pattern `match`ing, which is useful in the ICS26 handler:

https://github.com/informalsystems/ibc-rs/blob/a82f53a1b58dc868f3e865d34313bf995e9d7291/modules/src/ics26_routing/handler.rs#L23-L26

relayer/src/chain/cosmos.rs
```diff
@@ -256,6 +298,74 @@ impl Chain for CosmosSDKChain {
...
+        // Get the key from key seed file
+        let key = self
+            .keybase()
+            .get_key()
+            .map_err(|e| Kind::KeyBase.context(e))?;
+
+        Ok(key)
+    }
+
+    fn build_client_state(&self, height: ICSHeight) -> Result<Self::ClientState, Error> {
```
adizere, line 326: These methods are unchanged, just moved around to appear in the same order here as in `trait Chain`.

romac (approved):
LGTM, great work!

modules/src/ics26_routing/handler.rs
```diff
@@ -1,21 +1,45 @@
...
+                let domain_msg = update_client::MsgUpdateAnyClient::decode_vec(&*any_msg.value)
+                    .map_err(|e| Kind::MalformedMessageBytes.context(e))?;
+                Ok(ICS2Msg(ClientMsg::UpdateClient(domain_msg)))
+            }
+            // TODO: ICS3 messages
+            _ => Err(Kind::UnknownMessageTypeURL(any_msg.type_url)),
+        }?;
+        dispatch(ctx, envelope)?;
+    }
+    Ok(())
```
ancazamfir, line 42: We probably shouldn't mutate context if one message processing fails. i.e. we need to apply these messages on a copy of the context and only change the context when the loop finishes.

modules/src/mock/context.rs
```diff
@@ -225,8 +224,9 @@ impl MockContext {
     }
 
     /// A datagram passes from the relayer to the IBC module (on host chain).
+    /// Alternative method to `ICS18Context::send` that does not exercise any serialization.
     /// Used in testing the ICS18 algorithms, hence this may return a ICS18Error.
-    fn recv(&mut self, msg: ICS26Envelope) -> Result<(), ICS18Error> {
+    pub fn recv(&mut self, msg: ICS26Envelope) -> Result<(), ICS18Error> {
```
ancazamfir, line 229: maybe change the name?

ancazamfir (changes requested):
Looks good! Just a couple of comments.

modules/src/mock/context.rs
adizere: Good idea
https://github.com/informalsystems/ibc-rs/pull/422/commits/c92fb0ad22eb3b52e70f66210189c50461732565

modules/src/ics26_routing/handler.rs
adizere: I'm not sure if this is ideal, I still want to play with it in the context of other tests, but this may work:

https://github.com/informalsystems/ibc-rs/pull/422/commits/c92fb0ad22eb3b52e70f66210189c50461732565

modules/src/mock/context.rs
adizere: Good idea
https://github.com/informalsystems/ibc-rs/pull/422/commits/c92fb0ad22eb3b52e70f66210189c50461732565

ancazamfir (approved):
Looks good, thanks Adi!

Issue #445, opened 2020-12-10
Implement ChanOpenInit handler


ancazamfir:
## Crate
Modules

## Summary
Implement the dispatch, process, keeper etc for all `ChanOpenInit`. See #316

## Problem Definition

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #449, opened 2020-12-14
Migration to recent stargate

ancazamfir:
Closes: #451
Closes: #443
Closes: #468
Closes: #379

## Description
Draft PR for migration to latest `tm` and `cosmos`.
Change summary:
- tendermint dep `romac/skip-verif` off `v0.17`

- added confio's ics23 crate
   - [ ] not fully integrated yet

-  [x] Height related fields `version_..` change to `revision_..` @adizere 

- `createClient, connOpenInit, connOpenTry, chanOpenInit, chanOpenTry` do not pass an ID anymore, the allocated ID is returned in the `TxResult`
  - minimal changes to compile the code
  - [X] config not changed
  - [X] relayer loop not changed
  - [ ] CLIs work because the IDs are ignored
  - [X] tests fail @adizere 
  - [X] handlers need changes @adizere 

- connection `Version` is typed (used to be `String`)
   - changed msg structs accordingly

- some raw types have changed
  - minimum changes so the code compiles

- proto-compiler 
   - changed the commit hash to one extracted from go.mod of Go relayer on branch `colin/329-handshake-refactor`
   - ~~[X] the tonic compiler overwrites the previous prost compiler results, effect is missing files for genesis, tx, etc.~~
   - ~~[X] with temp fix for ^ the protos need some manual fixes and also the genereated files (on inculde of tendermint doesn't work)~~
   - [ ] need to modify the proto files for tonic to compile
   - [ ] grpc client generated code is not formatted

- added some ack and timeout related code
  - [X] CLIs work, need to finish the relaying part 
______

For contributor use:

- [ ] Updated the __Unreleased__ section of [CHANGELOG.md](https://github.com/informalsystems/ibc-rs/blob/master/CHANGELOG.md) with the issue.
- [ ] If applicable: Unit tests written, added test to CI.
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Updated relevant documentation (`docs/`) and code comments.
- [ ] Re-reviewed `Files changed` in the Github PR explorer.

PR #452, opened 2020-12-17
Handler for channel open init

cezarad:
Closes: #445

## Description

1. Main modified files: 

- ics04_channel/handler/chan_open_init.rs contains the definition of the handler and three very basic tests. 

ChanOpenInit returns the capability associated with the port's identity and channel's identity in the spec and the go version. Since the ChannelId is no longer visible in the handler probably the keeper should store the capability of the channel then there is no need for the handler to return a capability. Current version just returns the port capability (in the channel result) which respects the prototype in the spec but not the semantics.  

- ics04_channel/context.rs defines the interface for the context under which channels exists 

- ibc-rs/modules/src/mock/context.rs extends the store of MockContext with the stores required by channels: A map to store channels and a Map for the sequence numbers used by packets. 
2. medium changes: 

- ics26_routing/handler.rs  added handler for ChanOpenInit and tests for it. 
- ics04_channel/msgs/chan_open_init.rs added an auxiliary function that creates a "wrong" RawMsgChannelOpenInit for the tests in ics26_routing/handler.rs 

 
3. small modifications 
 The only modification in ics03_connection/connection.rs is to create a connection useful for the test on ChanOpenInit

______

For contributor use:


modules/src/ics03_connection/connection.rs
```diff
@@ -66,6 +68,21 @@ impl ConnectionEnd {
         })
     }
 
+    pub fn test_channel_new(
```
romac, line 71: This method should be moved under the `tests` module in `ics04_channel/handler/chan_open_init.rs`.

romac: Reason for that is that we don't want to include test-only code in the public API of the project.
As an additional benefit, since the `tests` module in `ics04_channel/handler/chan_open_init.rs` is defined as 

```rust
#[cfg(test)]
mod tests {
  ...
}
```

it will only be compiled when building the tests, eg. via `cargo test`.

modules/src/ics04_channel/channel.rs
```diff
@@ -279,8 +280,12 @@ pub fn validate_version(version: String) -> Result<String, Error> {
 
 #[cfg(test)]
 pub mod test_util {
+
     use ibc_proto::ibc::core::channel::v1::Channel as RawChannel;
     use ibc_proto::ibc::core::channel::v1::Counterparty as RawCounterparty;
+    // use crate::ics03_connection::msgs::conn_open_init::MsgConnectionOpenInit;
```
romac, line 286: Can we remove these commented lines altogether?

modules/src/ics04_channel/channel.rs
```diff
@@ -296,11 +301,22 @@ pub mod test_util {
...
+        RawChannel {
+            state: 1,
+            ordering: 0,
+            counterparty: Some(get_dummy_raw_counterparty()),
+            connection_hops: vec!["noconnection".to_string()],
+            version: "ics20".to_string(), // The version is not validated.
         }
     }
 }
+//MsgConnectionOpenInit::try_from(get_dummy_msg_conn_open_init()).unwrap().connection_id()
```
romac, line 319: Same here.

modules/src/ics04_channel/context.rs
```diff
@@ -0,0 +1,104 @@
+//! ICS4 (channel) context. The two traits `ConnectionReader` and `ConnectionKeeper` define
+//! the interface that any host chain must implement to be able to process any `ChannelMsg`.
+//! TODO make "ADR 004: IBC protocol implementation" for more details.
+//!
+//use crate::ics02_client::client_def::{AnyClientState, AnyConsensusState};
```
romac, line 5: The commented-out `use` statements in this file should probably be removed if you don't plan on uncommenting them later on.

modules/src/ics04_channel/context.rs
```diff
@@ -0,0 +1,104 @@
...
+                )?;
+
+                self.store_nextsequence_ack(
+                    &(result.port_id.clone(), result.channel_id.clone()),
+                    1,
+                )?;
+            }
+            State::TryOpen => {
+                self.store_channel(&(result.port_id, result.channel_id), &result.channel_end)?;
+                // TODOIf this is the first time the handler processed this channel, associate the
```
romac, line 66:
```suggestion
                // TODO: If this is the first time the handler processed this channel, associate the
```

modules/src/ics04_channel/error.rs
```diff
@@ -3,8 +3,13 @@ use thiserror::Error;
 
 pub type Error = anomaly::Error<Kind>;
 
+use crate::ics24_host::identifier::{ConnectionId,ChannelId};
+
 #[derive(Clone, Debug, Error)]
 pub enum Kind {
+    #[error("not implemented yet")]
+    Niy,
```
romac, line 11: In general, instead of returning a `Niy` error when some function is not implemented yet, you can use the `todo!` macro, eg.

```rust
fn my_unimplemented_function(x: u32) {
  todo!()
}
```

I would thus suggest removing this variant from the `Kind` struct.

modules/src/ics04_channel/handler/chan_open_init.rs
```diff
@@ -0,0 +1,184 @@
...
+    msg: MsgChannelOpenInit,
+) -> HandlerResult<ChannelResult, Error> {
+    let mut output = HandlerOutput::builder();
+
+    let pc_id = (msg.port_id().clone(), msg.channel_id().clone());
+
+    if ctx.channel_end(&pc_id).is_some() {
+        return Err(Kind::ChannelExistsAlready(msg.channel_id().clone()).into());
+    }
+    //TODO what about the port capabilities  ?
```
romac, line 22:
```suggestion
    // TODO: What about the port capabilities?
```

modules/src/ics04_channel/handler/chan_open_init.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    // An IBC connection running on the local (host) chain should exist.
+    if ctx
+        .connection_state(&msg.channel().connection_hops()[0])
+        .is_none()
+    {
+        return Err(Kind::MissingConnection(msg.channel().connection_hops()[0].clone()).into());
+    }
+
+    //TODO Check Version non Empty but not necessary coherent
```
romac, line 36:
```suggestion
    // TODO: Check that `version` is non empty but not necessary coherent
```

modules/src/ics04_channel/handler/chan_open_init.rs
```diff
@@ -0,0 +1,184 @@
...
+    if msg.channel().version() == "" {
```
romac, line 37: This can be rewritten as:

```suggestion
    if msg.channel().version().is_empty() {
```

(As advised by [Clippy](https://github.com/rust-lang/rust-clippy#usage): https://rust-lang.github.io/rust-clippy/master/#comparison_to_empty)

modules/src/ics04_channel/msgs/chan_open_init.rs
```diff
@@ -23,6 +23,28 @@ pub struct MsgChannelOpenInit {
...
+    pub fn port_id(&self) -> &PortId {
+        &self.port_id
+    }
+
+    /// Getter: borrow the `counterparty` from this message.
+    pub fn channel(&self) -> &ChannelEnd {
+        &self.channel
+    }
+
+    // /// Setter for `connection_id`. Amenable to chaining, since it consumes the input message.
```
romac, line 42: Can you explain why this is commented out?

romac (changes requested):
Great stuff, @cezarad :)

I did a first pass over the PR and left some comments which are mostly stylistic, but it all looks very good already!

It looks the build is failing because of the formatting. You can reformat the whole project at once by running `cargo fmt` in the `ibc-rs` folder. [Instructions for setting it up](https://github.com/rust-lang/rustfmt#on-the-stable-toolchain) (assuming you have installed Rust via https://rustup.rs.

modules/src/ics04_channel/msgs/chan_open_init.rs
cezarad: uncertain if we need a DS for connection -> channel because of confusing spec ICS04 :). 
The last query in ICS04 takes connection as parameter and I wonder if the intend is to retrieve all channels open over a certain connection. Waited to discuss with you ... but now I see it as copy paste  mistake in ICS04 queryChannel should have been portId instead of connId. :)

modules/src/ics04_channel/error.rs
cezarad: Thanks. It is a better one also I used it only as an intermediate step and forgot to remove it.

modules/src/ics03_connection/connection.rs
cezarad: yeah, this is indeed ugly. It appeared when I gave up on style to have at least one test. Compiling failed me because fields have the same name in different structures. :) In this case State is a field name in both in connection and Channel.  I need help understanding how I can create a connection in channel without rust complaining that is expects State to belong to either connection or channel.

modules/src/ics04_channel/handler/chan_open_init.rs
cezarad: I didn't see any capabilities. Do we ignore them for now/all together ?

modules/src/ics04_channel/context.rs
```diff
@@ -0,0 +1,130 @@
+//! ICS4 (channel) context. The two traits `ConnectionReader` and `ConnectionKeeper` define
```
adizere, line 1:
```suggestion
//! ICS4 (channel) context. The two traits `ChannelReader ` and `ChannelKeeper` define
```

modules/src/ics04_channel/context.rs
```diff
@@ -0,0 +1,130 @@
...
+//! the interface that any host chain must implement to be able to process any `ChannelMsg`.
+//! TODO make "ADR 004: IBC protocol implementation" for more details.
```
adizere, line 3:
```suggestion
```

modules/src/ics04_channel/context.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    /// Returns the ConnectionState for the given identifier `connection_id`.
+    fn connection_state(&self, connection_id: &ConnectionId) -> Option<ConnectionEnd>;
+
+    fn connection_channels(&self, cid: &ConnectionId) -> Option<Vec<(PortId, ChannelId)>>;
+
+    fn channel_client_state(&self, port_channel_id: &(PortId, ChannelId))
+        -> Option<AnyClientState>;
+
+    fn channel_consensus_state(
```
adizere, line 28: I don't see this function used anywhere. I'm wondering what's the use for it.

modules/src/ics04_channel/context.rs
```diff
@@ -0,0 +1,130 @@
...
+                    &result.channel_end.connection_hops()[0].clone(),
+                    &(result.port_id.clone(), channel_id),
+                )?;
+            }
+            State::TryOpen => {
+                self.store_channel(
+                    &(result.port_id, result.channel_id.clone().unwrap()),
+                    &result.channel_end,
+                )?;
+                // TODO: If this is the first time the handler processed this channel, associate the
```
adizere, line 81: Is this TODO still relevant? I guess we should try to address it.

modules/src/ics04_channel/error.rs
```diff
@@ -44,8 +46,23 @@ pub enum Kind {
...
     MissingCounterparty,
 
+    #[error("no commong version")]
+    NoCommonVersion,
+
     #[error("missing channel end")]
     MissingChannel,
+
+    #[error("channel exists (was initialized) already: {0}")]
+    ChannelExistsAlready(ChannelId),
```
adizere, line 56: This error msg is never used.

modules/src/ics04_channel/error.rs
```diff
@@ -44,8 +46,23 @@ pub enum Kind {
...
+
+    #[error("given connection {0} for channel innexitant ")]
```
adizere, line 58:
```suggestion
    #[error("given connection hop {0} does not exist")]
```

modules/src/ics04_channel/handler/chan_open_init.rs
```diff
@@ -0,0 +1,196 @@
...
+            msg_conn_init.counterparty().clone(),
+            get_compatible_versions(),
+            msg_conn_init.delay_period,
+        );
+
+        let ccid = <ConnectionId as FromStr>::from_str("defaultConnection-0");
+        let cid = match ccid {
+            Ok(v) => v,
+            Err(_e) => ConnectionId::default(),
+        };
```
adizere, line 123: This might be easier, and maybe it achieves the same thing.

```suggestion
        let ccid = ConnectionId::default();
```

modules/src/mock/context.rs
```diff
@@ -59,11 +67,30 @@ pub struct MockContext {
     /// All the connections in the store.
     connections: HashMap<ConnectionId, ConnectionEnd>,
 
+    /// All the channels in the store. TODO Make new key PortId X ChanneId
+    channels: HashMap<(PortId, ChannelId), ChannelEnd>,
```
adizere, line 71: Since we use the `(PortId, ChannelId)` tuple so often, it may be a good idea to address this comment and add a struct that captures the tuple.

Something like either of the two:
```rust
// A tuple struct
struct OwnedChannelId(PortId, ChannelId);

// A struct with two fields
struct OwnedChannelId {
    port_id: PortId,
    channel_id: ChannelId,
}
```

adizere (changes requested):
Nice work!

I left a bunch of comments, please see inline.

A good practice you can experiment with in future: you can try to give your branches a specific name, to help identify them easily. For instance, instead of [`Cezara`](https://github.com/informalsystems/ibc-rs/tree/Cezara) in future you can give the branch the name/id of the issue it is addressing, e.g., [`adi/68_errors`](https://github.com/informalsystems/ibc-rs/tree/adi/68_errors) or [`romac/backward-verif`](https://github.com/informalsystems/ibc-rs/tree/romac/backward-verif)

modules/src/ics04_channel/context.rs
cezarad: I looked into https://github.com/informalsystems/ibc-rs/blob/master/proto/src/prost/ibc.core.channel.v1.rs
and there at line 497 there's a QueryChannelConsensusStateResponse so I put it (tentatively) in the interface for the query implementation.

modules/src/ics04_channel/context.rs
cezarad: That part I haven't touched. It has to do with OpenTry.

modules/src/ics04_channel/error.rs
cezarad: Yes, it's from before the channel Id disappeared from the message. It was the error checkin the channel id in the ics04 spec that now is out of date.

modules/src/ics04_channel/handler/chan_open_init.rs
cezarad: The default is not yet versioned. Connection default returns just defaultConnection. I think I should modify it but I didn't want to mess with stuff I didn't do so soon.

modules/src/mock/context.rs
cezarad: yes :) we had a chat on this and you suggest the pair since is faster: and it was :).

modules/src/ics04_channel/handler/chan_open_init.rs
adizere: That makes sense, thanks for the explanation!

Is the `default` ConnectionId going to make the tests fail (due to the missing version number)? If the missing version is not problematic, I'd suggest we still go with `let ccid = ConnectionId::default();`; otherwise, if the missing version poses problems, we can just keep the code here as you wrote it initially.

cezarad: ConnectionId::default() needs to stay "defaultConnection" cause next_connection_id adds the counter as a suffix.  
For example the test in ics26 fails if ConnectionId::default() is versioned, cause the keeper will add the version  to the default connection and for channel a connection needs to be present. 
The function  ics04_channel::channel::test_util::get_dummy_raw_channel_end which returns a channel end with a versioned connection "defaultConnection-0". Here in this test I kind of do what get_dummy_raw_channel_end does in a way that compiles locally.

modules/src/ics04_channel/handler/chan_open_init.rs
cezarad: Added a place holder for capabilities.

cezarad:
> Nice work!
> 
> I left a bunch of comments, please see inline.
> 
> A good practice you can experiment with in future: you can try to give your branches a specific name, to help identify them easily. For instance, instead of [`Cezara`](https://github.com/informalsystems/ibc-rs/tree/Cezara) in future you can give the branch the name/id of the issue it is addressing, e.g., [`adi/68_errors`](https://github.com/informalsystems/ibc-rs/tree/adi/68_errors) or [`romac/backward-verif`](https://github.com/informalsystems/ibc-rs/tree/romac/backward-verif)

cezarad:
Yes indeed. I realised now that I started the try ... it

modules/src/mock/context.rs
adizere: Thanks for reminding me, that makes sense. We'll fix this later on then!

modules/src/ics04_channel/handler/chan_open_init.rs
adizere: Yep, we're ignoring capabilities for now.

modules/src/ics04_channel/context.rs
adizere: Ah, cool.

As a general comment, it's usually a good idea to avoid making a PR too big, otherwise it is difficult to review and understand it. So it's preferable to only bring in changes which are necessary.

For this method specifically, I don't understand why would it need to be part of the `ChannelReader` trait. The trait captures all methods necessary for the [channel handlers](https://github.com/cosmos/ics/tree/master/spec/ics-004-channel-and-packet-semantics#sub-protocols), and I'm not convinced this method is needed there (for any handler). What do you think?

modules/src/ics04_channel/handler/chan_open_init.rs
adizere: Ok, I understand. This has to do with the [recent changes](#449) on how `ConnectionId::default()` is used. So let's ignore my comment then, and we can refactor this later.

modules/src/mock/context.rs
```diff
@@ -196,6 +230,22 @@ impl MockContext {
             ..self
         }
     }
+    pub fn with_connection_capability(
```
ancazamfir, line 233: Maybe? Or have separate `with...` funcs?
```suggestion
    pub fn with_connection_and_port_capability(
```

ancazamfir (commented):
Looks good!! Left a few comments. Could you merge master and fix the clippy errors? I will do another review with focus on the port capability and port module impl.

modules/src/ics04_channel/context.rs
adizere: > IMO this is the right place for the implementation

Hmm, Reader/Keeper captures dependencies needed to execute/handle transactions. These are entirely different from handling queries. If the IBC module will be called for handling queries, then it may make sense to involve the Keeper, and we can keep queries code here, but this was not clear to me.

.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
+{
```
adizere, line 1: Let's not keep this files in the repo. The same for `.vscode/tasks.json`.

modules/src/ics04_channel/handler/chan_open_init.rs
```diff
@@ -0,0 +1,201 @@
...
+
+        for test in tests {
+            let res = dispatch(&test.ctx, test.msg.clone());
+            // Additionally check the events and the output objects in the result.
+            match res {
+                Ok(proto_output) => {
+                    assert_eq!(
+                        test.want_pass,
+                        true,
+                        "conn_open_init: test passed but was supposed to fail for test: {}, \nparams {:?} {:?}",
```
adizere, line 171:
```suggestion
                        "chan_open_init: test passed but was supposed to fail for test: {}, \nparams {:?} {:?}",
```

modules/src/ics04_channel/handler/chan_open_init.rs
```diff
@@ -0,0 +1,201 @@
...
+                        test.name,
+                        test.msg.clone(),
+                        test.ctx.clone()
+                    );
+                    assert_ne!(proto_output.events.is_empty(), true); // Some events must exist.
+
+                    // The object in the output is a ConnectionEnd, should have init state.
```
adizere, line 178:
```suggestion
                    // The object in the output is a ChannelEnd, should have init state.
```

modules/src/ics04_channel/handler/chan_open_init.rs
```diff
@@ -0,0 +1,201 @@
...
+                    let res: ChannelResult = proto_output.result;
+                    //assert_eq!(res.channel_id, msg_chan_init.channel_id().clone());
```
adizere, line 180: Could we have a test checking the port_id? Something like:

```suggestion
                    assert_eq!(res.port_id, msg_chan_init.port_id);
```

(Watch out for compilation issues, a `clone` may be necessary to `msg_chan_init`.)

modules/src/mock/context.rs
```diff
@@ -268,10 +318,161 @@ impl MockContext {
...
+    fn store_channel(
+        &mut self,
+        port_channel_id: &(PortId, ChannelId),
+        channel_end: &ChannelEnd,
+    ) -> Result<(), ICS4Error> {
+        self.channels
+            .insert(port_channel_id.clone(), channel_end.clone());
+        Ok(())
+    }
+    fn store_nextsequence_send(
```
adizere, line 427: A question of style (feel free to ignore). I'd rename these methods to be consistent with the fields `store_next_sequence_send`, `store_next_sequence_recv`, etc.

```suggestion
    fn store_next_sequence_send(
```

adizere (commented):
Thanks for the patience with these reviewz! We're pretty much there.

.vscode/launch.json
romac: We should add `.vscode` to `.gitignore`

ancazamfir (approved):
Looks good! Thanks Cezara!

romac (approved):

adizere (approved):
lgtm!

Issue #470, opened 2020-12-23
Add ability to send transfer packets from the CLI for testing purposes


ancazamfir:
## Crate
relayer-cli and relayer

## Summary
Add a CLI to send transfer packets (ICS-20) for testing purposes.

## Problem Definition
Until migration to latest ICS specs and protos (stargate-5/ stargate-6 see #451) we have been using the Go relayer CLI. With the changes for on-chain ID allocation it is not possible at this point to use it to test the Rust relayer loop (`packet-recv` and `packet-ack` CLIs can be used) . In addition we want more capabilities, e.g. specify the number of packets to be sent, the timeout height or timeout stamp etc.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #495, opened 2021-01-04
send packet transfer CLI

ancazamfir:
Closes: #470

## Description
- Added the CLI. For example:
`tx raw packet-send ibc-0 ibc-1 transfer channel-0 9999 111 -n 3`
Initiates three `9999` `samoleans` ics-20 transfer messages on `ibc-0` to be eventually sent to `ibc-1`, on `channel-0` with timeout height offset of `111`.

- The CLI doesn't use the runtime as we need it to work concurrently with a running relayer (using instead the `CosmosSDKChain::bootstrap()`)

- The `ics20-fungible-token-transfer` is moved under the `application` module.
______

For contributor use:


modules/src/application/ics20_fungible_token_transfer/msgs/transfer.rs
```diff
@@ -0,0 +1,89 @@
...
+/// Message definition for the "packet receiving" datagram.
+///
+#[derive(Clone, Debug, PartialEq)]
+pub struct MsgTransfer {
+    /// the port on which the packet will be sent
+    pub source_port: PortId,
+    /// the channel by which the packet will be sent
+    pub source_channel: ChannelId,
+    /// the tokens to be transferred
+    pub token: Option<ibc_proto::cosmos::base::v1beta1::Coin>,
```
romac, line 26: At some point (perhaps in a later PR), we should define this `Coin` type in a module rather than rely on the protobuf definition directly.

relayer/src/transfer.rs
```diff
@@ -0,0 +1,53 @@
...
+        .get_signer()
+        .map_err(|e| Kind::KeyBase.context(e))?;
+
+    let sender = packet_src_chain
+        .get_signer()
+        .map_err(|e| Kind::KeyBase.context(e))?;
+
+    let latest_height = packet_dst_chain.query_latest_height()?;
+
+    // TODO - send multiple messages
```
adizere, line 36: This todo may be outdated.

relayer/src/transfer.rs
```diff
@@ -0,0 +1,53 @@
...
+            amount: opts.amount.to_string(),
+        }),
+        sender,
+        receiver,
+        timeout_height: latest_height.add(opts.height_offset),
+        timeout_timestamp: 0,
+    };
+
+    let raw_msg = msg.to_any::<RawMsgTransfer>();
+    let msgs = vec![raw_msg; opts.number_msgs];
```
adizere, line 51: Confirm that the CLI tx works.

I'm wondering if we shouldn't add some basic validation and bound checking on the `-n` command line parameter for this tx. If I run (notice the `-n 0`)

> rrly -c loop_config.toml tx raw packet-send ibc-1 ibc-0 transfer channel-1 9999 111 -n 0

The error that there is no message to be sent is indeed caught, though a bit late, after the call to `build_and_send_transfer_messages`.

If I run with `-n 1000`, it seems the relayer will attempt to send everything in batches of ~33 messages. It's unclear if this works as intended, as there are also timeout errors popping-up.

We may want to enforce that `1 <= n <= MAX_MSGS` where MAX_MSGS is ~10.

adizere (approved):
Aside from minor comments (all of which can be fixed later), looks good!

Issue #505, opened 2021-01-07
Query channels per connection


adizere:
## Crate

Mostly `relayer` and `relayer-cli`

## Summary

Given a connection id, we'll need a CLI for querying all the existing channels that are associated with that connection.

____

#### For Admin Use

- [X] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

Issue #500, opened 2021-01-05
Add --json flag to force JSON output in relayer CLI


romac:
## Crate

`relayer-cli`

## Summary

Add a `--json` flag to force JSON output in all the CLI commands.

### Why do we need this?

- @andynog expressed the need for JSON output as part of his work on the CI
- other users may want to use the CLI programmatically without relying on using regular expressions to parse the output of the commands

## Proposal

- Add a `--json` flag to all commands, which enables JSON output.
- When this flag is set, a command should not use the `status_ok!`, `status_info!` and `status_err!` macros but instead emit a single-line JSON-encoded message.
- The message _must_ include a `status` field which must be set to one of `"success"`, `"info"`, or `"error"`.
- The message _should_ include a `msg` field with a human-readable description of the result of the command.
- The message _may_ include other fields.

### Example:

```json
{"status": "success", "msg": "client created", "client_id": "tendermint-07"}
```

```json
{"status": "info", "msg": "no action taken"}
```

```json
{"status": "error", "msg": "client not found", "client_id": "tendermint-07"}
```

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

adizere:
Maybe also take into consideration the following problem highlighted elsewhere (in #469):

Sort-out error messages and make them consistent. For example, updating a non-existing client results in a panic, but querying for a non-existing client returns an error. The commands are:
> rrly -c loop_config.toml tx raw update-client ibc-0 ibc-1 07-tendermint-X
> rrly -c loop_config.toml query client state ibc-0 07-tendermint-X

romac:
Agreed!

On top of that, we should also return a proper UNIX error code on exit, via [`abscissa_core::application::fatal_error`](https://docs.rs/abscissa_core/0.5.2/abscissa_core/application/fn.fatal_error.html). See the [`light rm` command](https://github.com/informalsystems/ibc-rs/blob/76134e257113ab4fe883346a4d97f66848f392cc/relayer-cli/src/commands/light/rm.rs#L70-L78) for an example of how to do this.

adizere:
I'm thinking we should make `--json` always on and simply stop using status_ok!, status_info! and status_err! macros altogether.

ancazamfir:
[...]

PR #506, opened 2021-01-07
CLI Command to query connection channels

adizere:
Closes: #505 

## Description

Added the following command:

```bash
rrly -c loop_config.toml query connection channels [CHAIN-ID] [CONNECTION-ID]
```

Sample outputs:

> $ rrly -c loop_config.toml query connection channels ibc-0 connection-1
> connection query result:  []

> $  rrly -c loop_config.toml query connection channels ibc-0 connection-0
> connection query result:  [ChannelId("channel-0")]

Error-prone cases:

> $ rrly -c loop_config.toml query connection channels ibc-3 connection-0
> error: invalid options: missing chain configuration for the given chain id

> $ rrly -c loop_config.toml query connection channels ibc-3 connection-0
> connection query error query error: GRPC error: transport error: error trying to connect: tcp connect error: Connection refused (os error 61)

~~Also offers option for json output (initial experience towards https://github.com/informalsystems/ibc-rs/issues/500):~~ (this work will take place in a different PR)

______

For contributor use:

- [X] If applicable: Unit tests written, added test to CI.
- [X] Updated relevant documentation (`docs/`) and code comments.


adizere:
This should be ready for review. If you wish, @cezarad and @vitorenesduarte please also have a look over the PR and any thoughts or questions would be welcome!

relayer-cli/src/commands/query/connection.rs
```diff
@@ -98,11 +98,89 @@ impl Runnable for QueryConnectionEndCmd {
...
+        config: &Config,
+    ) -> Result<(ChainConfig, QueryConnectionChannelsOptions), String> {
+        let chain_id = self
+            .chain_id
+            .clone()
+            .ok_or_else(|| "no chain chain identifier provided".to_string())?;
+        let chain_config = config
+            .chains
+            .iter()
+            .find(|c| c.id == chain_id)
```
vitorenesduarte, line 130: This could use the `Config::find_chain` method introduced recently in #510.

relayer/src/chain/cosmos.rs
```diff
@@ -655,6 +651,34 @@ impl Chain for CosmosSDKChain {
...
+
+        let response = self
+            .block_on(client.connection_channels(request))?
+            .map_err(|e| Kind::Grpc.context(e))?
+            .into_inner();
+
+        let vec_ids = response
+            .channels
+            .iter()
+            .filter_map(|ic| ChannelId::from_str(ic.channel_id.as_str()).ok())
```
vitorenesduarte, line 677: If there are any errors here, they will be silently ignored. Maybe we can instead do something like this: https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect

ancazamfir: We can do this but the question is do we want to fail the entire operation if we cannot extract one  `ChannelId`?

relayer-cli/src/commands/query/connection.rs
adizere: Good point, thanks for bringing that up! 

ea129d390229e4e3fe0403c880174f33e38d1e5e

relayer/src/chain/cosmos.rs
adizere: Good catch.

Between the option of failing the operation because an identifier fails to parse, and ignoring the offending identifier(s), I would rather go with the latter.

One middle-road solution is to have warnings in case some identifiers fail to parse. We're not (yet) principled with how we use warnings, if at all, so I'm adding a TODO here to keep an eye out in the future. c51b814

ancazamfir (approved):
Looks good! Thanks Adi!

PR #511, opened 2021-01-11
Replace `ChannelConfig` in `Channel::new`

vitorenesduarte:
## Description

Instead of updating a partially complete `ChannelConfig`, `Channel::new` now constructs the `ChannelConfig` from scratch given its arguments. I found this API nicer to use.

All the other changes are for consistency:
- both `ConnectionConfigSide::new` and `ChannelConfigSide::new` now have owned values as arguments
- the order of the arguments on these methods follows the creation order: chain, client, connection, port, channel
- `ConnectionConfig::new` and `ChannelConfig::new` are deleted as they're no longer used

I also noticed that if I use `PortId::default()` as a port, I get the following error: `could not retrieve module from port-id: ports/defaultPort: capability not found`. Is this expected?

______

For contributor use:

### DIFF
--- a/relayer/src/connection.rs
+++ b/relayer/src/connection.rs
@@ -20,7 +20,6 @@ use ibc::tx_msg::Msg;
 use ibc::Height as ICSHeight;
 
 use crate::chain::handle::ChainHandle;
-use crate::config;
 use crate::error::{Error, Kind};
 use crate::foreign_client::{build_update_client, ForeignClient};
 use crate::relay::MAX_ITER;
@@ -44,20 +43,20 @@ pub struct Connection {
 #[derive(Clone, Debug)]
 pub struct ConnectionSideConfig {
     chain_id: ChainId,
-    connection_id: ConnectionId,
     client_id: ClientId,
+    connection_id: ConnectionId,
 }
 
 impl ConnectionSideConfig {
     pub fn new(
         chain_id: ChainId,
-        connection_id: ConnectionId,
         client_id: ClientId,
+        connection_id: ConnectionId,
     ) -> ConnectionSideConfig {
         Self {
             chain_id,
-            connection_id,
             client_id,
+            connection_id,
         }
     }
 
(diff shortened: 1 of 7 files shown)


relayer/src/connection.rs
```diff
@@ -44,37 +43,37 @@ pub struct Connection {
...
     }
 
-    pub fn set_connection_id(&mut self, id: &ConnectionId) {
-        self.connection_id = id.clone()
+    pub fn connection_id(&self) -> &ConnectionId {
+        &self.connection_id
+    }
+
+    pub(crate) fn set_connection_id(&mut self, id: &ConnectionId) {
+        self.connection_id = id.clone();
```
romac, line 76: Can we rather have this function take an owned value? I'd rather expose the `clone` to the call site to make it more explicit rather than doing it implicitly under to the hood (seems to be a general good practice in the Rust ecosystem, although we don't currently follow it everywhere).

```suggestion
    pub(crate) fn set_connection_id(&mut self, id: ConnectionId) {
        self.connection_id = id;
```

I know that this code was not added in your PR, but it would be nice to improve on this while we're at it :)

vitorenesduarte: Thanks, this is a good suggestion. Just pushed a commit addressing this.

ancazamfir:
> I also noticed that if I use PortId::default() as a port, I get the following error: could not retrieve module from port-id: ports/defaultPort: capability not found. Is this expected?

Could you give more details on the use that shows this error?

vitorenesduarte:
> > I also noticed that if I use PortId::default() as a port, I get the following error: could not retrieve module from port-id: ports/defaultPort: capability not found. Is this expected?
> 
> Could you give more details on the use that shows this error?

The error should show up with something like this:

```rust
    let connection_0_1: Connection = ...;
    let ordering = Order::default();
    let port_0 = PortId::default();
    let port_1 = PortId::default();
    let channel_0_1 = Channel::new(connection_0_1, ordering, port_0, port_1)?;
```

The full example that triggered the error is here: https://github.com/vitorenesduarte/ibc-rs/blob/reproduce_bug_rust/bbt/src/main.rs#L74-L80 (the branch is not polished; I can extract the example from there in case it's useful)

I just confirmed, and the error can also be triggered using the CLI by replacing `transfer` with `defaultPort` in the following:
```
rrly -c loop_config.toml tx raw chan-init ibc-0 ibc-1 connection-0 transfer transfer defaultChannel defaultChannel
```

ancazamfir:
>  just confirmed, and the error can also be triggered using the CLI by replacing transfer with defaultPort in the following...

Yes, that is a chain error and it is expected as there is no module on chain binding to that `defaultPort`. The only supported one is `transfer`. The relayer does not have currently a way to query valid `PortId` -s so if the ID is valid it just passes it in the IBC message.

ancazamfir (approved):
Looks good, thanks Vitor!

Issue #514, opened 2021-01-12
Poor performance for Tx-es and queries


ancazamfir:
## Crate
relayer

## Summary of Bug
Tx-es and queries take a long time to complete. For example getting the account information takes ~1sec:
```
Jan 07 14:05:31.681  INFO relayer::chain::cosmos:  => getting account info
Jan 07 14:05:32.581  INFO relayer::chain::cosmos:  => got account info
```
Sending a Tx takes ~2sec:
```
Jan 07 14:05:32.663  INFO relayer::chain::cosmos:  => sending tx
Jan 07 14:05:34.728  INFO relayer::chain::cosmos:  => received tx response
```
The problem is here:
https://github.com/informalsystems/ibc-rs/blob/9c5d4f60732db62626cd7163f7e3b87f9f3c289a/relayer/src/chain/cosmos.rs#L132-L134

## Version
master

## Steps to Reproduce
Run relayer loop or CLIs an notice the slow processing times.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ancazamfir:
After replacing some of the `self.block_on` with util's `block_on`:
https://github.com/informalsystems/ibc-rs/blob/5c9739c536ee253baa5bba2b4f2b20968b3489b7/relayer/src/util/block_on.rs#L6-L13

in `anca/relayer_profiling` branch, the performance improves dramatically:

Old:
```
Jan 07 14:05:31.681  INFO relayer::chain::cosmos:  => getting account info
Jan 07 14:05:32.581  INFO relayer::chain::cosmos:  => got account info
Jan 07 14:05:32.663  INFO relayer::chain::cosmos:  => sending tx
Jan 07 14:05:34.728  INFO relayer::chain::cosmos:  => received tx response
```
New:
```
Jan 07 14:14:30.121  INFO relayer::chain::cosmos:  => getting account info
Jan 07 14:14:30.125  INFO relayer::chain::cosmos:  => got account info
Jan 07 14:14:30.205  INFO relayer::chain::cosmos:  => sending tx
Jan 07 14:14:31.035  INFO relayer::chain::cosmos:  => received tx response
```

[...]

Issue #94, opened 2020-06-10
Investigate and clean the validate_basic method


Shivani912:
The method `validate_basic` appears in almost all domain type definitions, e.g., `Counterparty`, `ConnectionEnd`, `MsgConnectionOpenConfirm`, etc. In most of the implementations, this method returns an `Ok()` result. It's unclear if we should keep it around.

Investigate:

- What was the original intended purpose of `validate_basic`?
- Can we remove it altogether, or do some types still need it? Intuitively, conversion from the Raw type into the domain type (see below an example) handles all validation. Domain types are a rather new features, and perhaps they make the method `validate_basic` unnecessary.

https://github.com/informalsystems/ibc-rs/blob/2b5029b50e1cfd4181e2dfcf8d1b76c7b9d833e6/modules/src/ics03_connection/msgs/conn_open_init.rs#L84-L99

PR #515, opened 2021-01-12
Relayer fixes after profiling

ancazamfir:
Closes: #514
Closes: #94

## Description
Main changes are:
- replaced `CosmosSDK` 's `block_on()` with util's `block_on()`
- added check in `build_update_client()` to wait for source chain to reach the `target_height`
- re-organized packet CLIs around `PacketMsgCollector` that is populated by the queries, with focus on correct query height versus proof and client update height.

Miscellaneous/ Cleanup: 
- added ack and timeout event extraction from Tx results
- removed `Msg` 's `validate_basic()`

______

For contributor use:

### DIFF
--- a/relayer-cli/src/commands/query/packet.rs
+++ b/relayer-cli/src/commands/query/packet.rs
@@ -83,8 +83,8 @@ impl Runnable for QueryPacketCommitmentsCmd {
             Ok(cs) => status_info!(
                 "Result for packet commitments query at height",
                 "{:?} {:#?}",
-                cs.0,
-                cs.1
+                cs.1,
+                cs.0
             ),
             Err(e) => status_info!("Error encountered on packet commitments query:", "{}", e),
         }
--- a/relayer/src/chain/cosmos.rs
+++ b/relayer/src/chain/cosmos.rs
@@ -67,6 +67,7 @@ use crate::event::monitor::{EventBatch, EventMonitor};
 use crate::keyring::store::{KeyEntry, KeyRing, KeyRingOperations, StoreBackend};
 use crate::light_client::tendermint::LightClient as TMLightClient;
 use crate::light_client::LightClient;
+use crate::util::block_on;
 use ibc::ics04_channel::packet::Sequence;
 
 // TODO size this properly
@@ -97,9 +98,7 @@ impl CosmosSDKChain {
         let request =
             tonic::Request::new(ibc_proto::cosmos::staking::v1beta1::QueryParamsRequest {});
 
-        let response = self
-            .block_on(client.params(request))?
-            .map_err(|e| Kind::Grpc.context(e))?;
+        let response = block_on(client.params(request)).map_err(|e| Kind::Grpc.context(e))?;
 
         let res = response
             .into_inner()
@@ -122,8 +121,7 @@ impl CosmosSDKChain {
     /// Query the consensus parameters via an RPC query
     /// Specific to the SDK and used only for Tendermint client create
     pub fn query_consensus_params(&self) -> Result<Params, Error> {
-        Ok(self
-            .block_on(self.rpc_client().genesis())?
+        Ok(block_on(self.rpc_client().genesis())
             .map_err(|e| Kind::Rpc.context(e))?
             .consensus_params)
     }
@@ -160,9 +158,8 @@ impl CosmosSDKChain {
             value: pk_buf,
         };
 
-        let acct_response = self
-            .block_on(query_account(self, key.account))?
-            .map_err(|e| Kind::Grpc.context(e))?;
+        let acct_response =
+            block_on(query_account(self, key.account)).map_err(|e| Kind::Grpc.context(e))?;
 
         let single = Single { mode: 1 };
         let sum_single = Some(Sum::Single(single));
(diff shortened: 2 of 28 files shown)


relayer/src/link.rs
```diff
@@ -395,189 +393,255 @@ fn build_ack_from_recv_event(
...
+    dst_msgs: Vec<Any>,       // recv packets to be send to destination chain
+    dst_query_height: Height, // proof height for acks and timeout
+    src_msgs: Vec<Any>,       // acks and/or timeouts to be sent to source chain
+}
+
+impl PacketMsgCollector {
+    fn new(
+        packet_dst_chain: Box<dyn ChainHandle>,
+        packet_src_chain: Box<dyn ChainHandle>,
+        opts: &PacketEnvelope,
```
romac, line 412:
```suggestion
        opts: PacketEnvelope,
```

relayer/src/link.rs
```diff
@@ -395,189 +393,255 @@ fn build_ack_from_recv_event(
...
+    ) -> Self {
+        PacketMsgCollector {
+            packet_src_chain,
+            packet_dst_chain,
+            opts: opts.clone(),
```
romac, line 417:
```suggestion
            opts,
```

relayer/src/foreign_client.rs
```diff
@@ -188,6 +189,11 @@ pub fn build_update_client(
     dst_client_id: &ClientId,
     target_height: Height,
 ) -> Result<Vec<Any>, Error> {
+    // Wait for source chain to reach `target_height`
+    while src_chain.query_latest_height()? < target_height {
+        thread::sleep(Duration::from_millis(40))
```
romac, line 194: 40ms between queries sounds a bit too short to me, but perhaps those are quicker than I expect?

romac (commented):
Great stuff, this is a lot cleaner 👍

relayer/src/foreign_client.rs
ancazamfir: What would be a good value? Changed it to 100ms in the last commit. `query_latest_height()` takes 4msec on my machine with gaia nodes running locally. Also, since in this version (V0) nothing else is going on (we only have one blocking thread waiting here for source chain to advance) I don't think we take away time from anything in the relayer process.

romac (approved):

relayer-cli/src/commands/query/packet.rs
```diff
@@ -83,8 +83,8 @@ impl Runnable for QueryPacketCommitmentsCmd {
             Ok(cs) => status_info!(
                 "Result for packet commitments query at height",
                 "{:?} {:#?}",
-                cs.0,
```
adizere, line 86: FYI there will be conflicts here (due to recent changes in #518).

relayer/src/chain/cosmos.rs
```diff
@@ -97,9 +98,7 @@ impl CosmosSDKChain {
         let request =
             tonic::Request::new(ibc_proto::cosmos::staking::v1beta1::QueryParamsRequest {});
 
-        let response = self
-            .block_on(client.params(request))?
-            .map_err(|e| Kind::Grpc.context(e))?;
+        let response = block_on(client.params(request)).map_err(|e| Kind::Grpc.context(e))?;
```
adizere, line 101: I noticed the method `CosmosSDKChain::block_on` is still being used in a few spots. Should we completely remove it?

relayer/src/link.rs
```diff
@@ -395,197 +393,263 @@ fn build_ack_from_recv_event(
...
+    opts: PacketEnvelope,
+    recv_seqs: Vec<Sequence>,
+    ack_seqs: Vec<Sequence>,
+    src_query_height: Height, // proof height for recv packets
+    dst_msgs: Vec<Any>,       // recv packets to be send to destination chain
+    dst_query_height: Height, // proof height for acks and timeout
+    src_msgs: Vec<Any>,       // acks and/or timeouts to be sent to source chain
+}
+
+impl PacketMsgCollector {
```
adizere, line 408: This is a sort of a helper to do all logic related to sending packets?

relayer/src/chain/cosmos.rs
ancazamfir: We cannot remove it everywhere as depending on the caller stack we may end up with runtime-in-a-runtime error. Let's hope this will be cleaner when we migrate to tokio1.0

relayer/src/link.rs
ancazamfir: Yes. Before we had all the action functions returning the information as they were retrieving information. Now all is collected in this struct so we don't pass it around from return function values to input parameters to next ones.
Also, I was hoping that I will be able to eventually merge this and the Link struct but it's not as straightforward as it seems at first look. So we will look at this in subsequent PRs.

relayer-cli/src/commands/query/packet.rs
ancazamfir: Right. There is also a problem with the returned height. We cannot specify a height in the current gRPC request. So no matter what height we put in the CLI it will always return the query result from latest height. I will add this to @andynog 's PR with UX improvements.

relayer/src/link.rs
adizere: Cool,. Merging them together in some coherent struct would be nice to have, but the modifications here are already pretty good!

adizere (approved):

PR #518, opened 2021-01-12
JSON output for queries and txs

adizere:
Closes: #500 

## Description

List of outputs upon successful execution of all queries and txs:

#### client tx

> {"status":"success","result":[{"CreateClient":{"client_id":"07-tendermint-1","client_type":"Tendermint","consensus_height":{"revision_height":206,"revision_number":1},"height":"1"}}]}

> {"status":"success","result":[{"UpdateClient":{"client_id":"07-tendermint-1","client_type":"Tendermint","consensus_height":{"revision_height":346,"revision_number":0},"height":"1"}}]}

#### connection tx

> {"status":"success","result":[{"OpenInitConnection":{"client_id":"07-tendermint-1","connection_id":"connection-1","counterparty_client_id":"07-tendermint-1","counterparty_connection_id":null,"height":"1"}}]}

> {"status":"success","result":[{"OpenTryConnection":{"client_id":"07-tendermint-1","connection_id":"connection-2","counterparty_client_id":"07-tendermint-1","counterparty_connection_id":"connection-1","height":"1"}}]}

> {"status":"success","result":[{"OpenAckConnection":{"client_id":"07-tendermint-1","connection_id":"connection-2","counterparty_client_id":"07-tendermint-1","counterparty_connection_id":"connection-3","height":"1"}}]}

> {"status":"success","result":[{"OpenConfirmConnection":{"client_id":"07-tendermint-1","connection_id":"connection-3","counterparty_client_id":"07-tendermint-1","counterparty_connection_id":"connection-2","height":"1"}}]}

#### channel tx

> {"status":"success","result":[{"OpenInitChannel":{"channel_id":"channel-1","connection_id":"connection-3","counterparty_channel_id":null,"counterparty_port_id":"transfer","height":"1","port_id":"transfer"}}]}

> {"status":"success","result":[{"OpenTryChannel":{"channel_id":"channel-1","connection_id":"connection-2","counterparty_channel_id":"channel-1","counterparty_port_id":"transfer","height":"1","port_id":"transfer"}}]}

> {"status":"success","result":[{"OpenAckChannel":{"channel_id":"channel-1","connection_id":"connection-3","counterparty_channel_id":"channel-1","counterparty_port_id":"transfer","height":"1","port_id":"transfer"}}]}

> {"status":"success","result":[{"OpenConfirmChannel":{"channel_id":"channel-1","connection_id":"connection-2","counterparty_channel_id":"channel-1","counterparty_port_id":"transfer","height":"1","port_id":"transfer"}}]}

#### packet, transfer tx

> {"status":"success","result":[[{"SendPacketChannel":{"height":"1","packet":{"data":[123,34,97,109,111,117,110,116,34,58,34,49,48,48,48,48,34,44,34,100,101,110,111,109,34,58,34,115,97,109,111,108,101,97,110,115,34,44,34,114,101,99,101,105,118,101,114,34,58,34,99,111,115,109,111,115,49,106,107,54,118,115,99,108,119,117,97,122,101,116,117,57,107,102,118,118,117,112,50,51,50,122,51,52,104,101,52,116,101,110,104,121,115,99,107,34,44,34,115,101,110,100,101,114,34,58,34,99,111,115,109,111,115,49,50,110,101,119,48,97,120,110,97,112,109,114,117,116,110,122,103,52,104,117,115,120,122,117,103,110,122,119,114,122,101,57,100,50,48,51,53,56,34,125],"destination_channel":"channel-1","destination_port":"transfer","sequence":1,"source_channel":"channel-1","source_port":"transfer","timeout_height":{"revision_height":3798,"revision_number":1},"timeout_timestamp":0}}}]]}

__Note:__ TxRawPacketRecvCmd and TxRawPacketAckCmd: not clear how to test these

#### client queries

> {"status":"success","result":[{"AnyClientState":"Tendermint","allow_update_after_expiry":false,"allow_update_after_misbehaviour":false,"chain_id":"ibc-1","frozen_height":{"revision_height":0,"revision_number":0},"latest_height":{"revision_height":3536,"revision_number":1},"max_clock_drift":{"nanos":0,"secs":3},"trust_level":{"denominator":"3","numerator":"1"},"trusting_period":{"nanos":0,"secs":1209600},"unbonding_period":{"nanos":0,"secs":1814400},"upgrade_path":["upgrade","upgradedIBCState"]}]}

> {"status":"success","result":[{"AnyConsensusState":"Tendermint","next_validators_hash":"F340187B7E4860DE518F5B9F208EA243F9322BCE7D7D7693C4670B41E0057E22","root":[199,144,30,80,92,207,201,98,77,131,83,6,67,108,232,70,102,201,188,217,73,220,23,162,110,17,32,174,66,80,99,229],"timestamp":"2021-01-13T11:52:30.313428Z"}]}

> {"status":"success","result":[["connection-1","connection-2"]]}

#### connection queries:

{"status":"success","result":[{"ConnectionEnd":"ConnectionEnd","client_id":"07-tendermint-1","counterparty":{"client_id":"07-tendermint-1","connection_id":"connection-3","prefix":[105,98,99]},"delay_period":0,"state":"Open","versions":[{"features":["ORDER_ORDERED","ORDER_UNORDERED"],"identifier":"1"}]}]}

#### channel query: 

> {"status":"success","result":[{"ChannelEnd":"ChannelEnd","connection_hops":["connection-2"],"ordering":"Unordered","remote":{"channel_id":"channel-1","port_id":"transfer"},"state":"Open","version":"ics20-1"}]}

#### packet queries:

- for packet commitments:

> {"status":"success","result":[[1],{"revision_height":5641,"revision_number":0}]}

- for packet commitment:

> {"status":"success","result":[{"client_proof":null,"consensus_proof":null,"height":{"revision_height":1,"revision_number":0},"object_proof":[10,170,3,10,1....115]}]}

- for packet ack:

> {"status":"success","result":[[[],{"client_proof":null,"consensus_proof":null,"height":{"revision_height":1,"revision_number":0},"object_proof":[10,190,3,...45,107]}]]}

- for packet unreceived-packets:

> {"status":"success","result":[[1]]}

- for packet acks:

> {"status":"success","result":[[],{"revision_height":6993,"revision_number":0}]}

- for packet unreceived-acks:

> {"status":"success","result":[[]]}

### Added derive serde `Serialize` annotation

Did this for several types, e.g., `AnyClientState`, which was necessary to be able to format these types into the JSON output format.

### Better error output in some error corner cases

For example, any Query error is now accompanied with the precise location (the query type) that failed, so that the chain of errors is explicit in the error output. Also, some query errors were misreported as _converting_ problems when the query result was empty (the problem was that the result was empty, not in converting the types). The output diff is as follows for `rrly -c loop_config.toml tx raw update-client ibc-1 ibc-0`:

```diff
- {"status":"error","result":["tx error: Bad parameter: error converting message type into domain type: unknown client state type: "]}
+ {"status":"error","result":["tx error: Query error occurred (failed to finish query for client state): error converting message type into domain type: empty client state"]}
```

### Output simplified for 'query packet commitments' command:

We used to output the [`RawPacketState`](https://github.com/informalsystems/ibc-rs/blob/db2be6f72fef647933f48d80b29a0f9f12c804d4/proto/src/prost/ibc.core.channel.v1.rs#L95) (to string) upon command `query packet commitments`. There is no associated domain type for this, and is not possible to annotate this raw type with a serde::Serialize derive macro. So instead of outputting the whole RawPacketState, we now output only the packet sequence numbers.

Here is how the output changed:

```diff
- Result for packet commitments query at height [PacketState { port_id: "transfer", channel_id: "channel-1", sequence: 1, data: [42, 195, 42, 106, 38, 32, 206, 55, 183, 248, 21, 191, 96, 95, 250, 161, 10, 105, 89, 6, 91, 122, 57, 205, 55, 167, 29, 101, 55, 32, 55, 127] }] Height {
-    revision_number: 0,
-    revision_height: 5187,
- }
+ {"status":"success","result":[[1],{"Height":"Height","revision_height":5702,"revision_number":0}]}
```

## Cases that are _not_ covered with JSON output:

The following is non-exhaustive list of commands that _do not have JSON_ output.
We'll fix these problems as they become important:

- client update with invalid client id: 
>$ rrly -c loop_config.toml tx raw update-client ibc-1 ibc-0 ddd
> error: invalid argument to option `dst_client_id`: identifier ddd has invalid length 3 must be between 9-64 characters

- conn-try with invalid connection id confuses success with error -- note the missing src_connection_id parameter:

> $ rrly -c loop_config.toml tx raw conn-try ibc-1 ibc-0 07-tendermint-1 07-tendermint-1 connection-1
> {"status":"success","result":[{"ChainError":"deliver_tx reports error: log=Log(\"failed to execute message; message index: 1: connection handshake open try failed: failed connection state verification for client (07-tendermint-1): chained membership proof contains nonexistence proof at index 0. If this is unexpected, please ensure that proof was queried from the height that contained the value in store and was queried with the correct key.: invalid proof\")"}]}

- ~~[`query packet commitment`](https://github.com/informalsystems/ibc-rs/blob/4feec430f08c7b1504916242b76282de54839083/relayer-cli/src/commands/query/packet.rs#L135) command does not output JSON upon success. Reason: it's not clear how this command will be used and what are the requirements on the output. Is the object proof going to be consumed by users/CI @ancazamfir @andynog? The same goes for `query packet ack`~~ this was addressed

______

For contributor use:

### DIFF
--- a/modules/src/ics03_connection/connection.rs
+++ b/modules/src/ics03_connection/connection.rs
@@ -1,10 +1,12 @@
 use std::convert::{TryFrom, TryInto};
 use std::str::FromStr;
 
+use serde::Serialize;
+use tendermint_proto::Protobuf;
+
 use ibc_proto::ibc::core::connection::v1::{
     ConnectionEnd as RawConnectionEnd, Counterparty as RawCounterparty,
 };
-use tendermint_proto::Protobuf;
 
 use crate::ics03_connection::error::Kind;
 use crate::ics03_connection::version::Version;
@@ -12,7 +14,7 @@ use crate::ics23_commitment::commitment::CommitmentPrefix;
 use crate::ics24_host::error::ValidationError;
 use crate::ics24_host::identifier::{ClientId, ConnectionId};
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Serialize)]
 pub struct ConnectionEnd {
     state: State,
     client_id: ClientId,
@@ -148,7 +150,7 @@ impl ConnectionEnd {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize)]
 pub struct Counterparty {
     client_id: ClientId,
     connection_id: Option<ConnectionId>,
@@ -237,7 +239,7 @@ impl Counterparty {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize)]
 pub enum State {
     Uninitialized = 0,
     Init = 1,
--- a/modules/src/ics03_connection/connection.rs
+++ b/modules/src/ics03_connection/connection.rs
@@ -15,6 +15,7 @@ use crate::ics24_host::error::ValidationError;
 use crate::ics24_host::identifier::{ClientId, ConnectionId};
 
 #[derive(Clone, Debug, PartialEq, Serialize)]
+#[serde(tag = "ConnectionEnd")]
 pub struct ConnectionEnd {
     state: State,
     client_id: ClientId,
--- a/relayer-cli/src/conclude.rs
+++ b/relayer-cli/src/conclude.rs
@@ -3,6 +3,7 @@
 //! queries and transactions.
 //!
 //! ## Examples:
+//!
 //! - Exit from a query/tx with a string error:
 //!
 //! ```
@@ -12,7 +13,10 @@
 //! Output::with_error().with_result(json!(e)).exit();
 //! ```
 //!
-//! - Exit from a query/tx with an anomaly error:
+//! - Exit from a query/tx with an error of type `anomaly`:
+//! In the case where the error is a complex type such as anomaly (including backtraces), it is
+//! better to simplify the output and only write out the chain of error sources, which we can
+//! achieve with `format!("{}", e)`. The complete solution is as follows:
 //!
 //! ```
 //! use crate::relayer_cli::error::{Error, Kind};
@@ -22,6 +26,12 @@
 //! ```
 //!
 //! - Exit from a query/tx with success:
+//!
+//! ```
+//! Output::with_success().with_result(json!(cs)).exit();
+//! ```
+//!
+//! - Exit from a query/tx with success and multiple objects in the result:
 //!
 //! ```
 //! Output::with_success().with_result(json!(cs)).exit();
@@ -52,8 +62,8 @@ pub struct Output {
     pub status: Status,
 
     /// The result of a command, such as the output from a query or transaction.
-    /// An arbitrary strongly typed JSON object.
-    pub result: Option<serde_json::Value>,
+    /// This is a vector, possibly empty, of strongly typed JSON objects.
+    pub result: Vec<serde_json::Value>,
 }
 
 impl Output {
@@ -61,7 +71,7 @@ impl Output {
     pub fn new(status: Status) -> Self {
         Output {
             status,
-            result: None,
+            result: vec![],
         }
     }
 
@@ -75,10 +85,14 @@ impl Output {
         Output::new(Status::Error)
     }
 
-    /// Builder-style method for attaching a result to an output object.
+    /// Builder-style method for attaching a result to an output object. Can be called multiple
+    /// times, with each subsequent call appending the given `res` JSON object to the output.
     pub fn with_result(self, res: serde_json::Value) -> Self {
+        let mut new_res = self.result.clone();
+        new_res.push(res);
+
         Output {
-            result: Some(res),
+            result: new_res,
             ..self
         }
     }
--- a/relayer-cli/src/conclude.rs
+++ b/relayer-cli/src/conclude.rs
@@ -88,14 +88,9 @@ impl Output {
 
     /// Builder-style method for attaching a result to an output object. Can be called multiple
     /// times, with each subsequent call appending the given `res` JSON object to the output.
-    pub fn with_result(self, res: serde_json::Value) -> Self {
-        let mut new_res = self.result.clone();
-        new_res.push(res);
-
-        Output {
-            result: new_res,
-            ..self
-        }
+    pub fn with_result(mut self, res: serde_json::Value) -> Self {
+        self.result.push(res);
+        self
     }
 
     /// Exits from the process with the current output. Convenience wrapper over `exit_with`.
--- a/relayer-cli/src/conclude.rs
+++ b/relayer-cli/src/conclude.rs
@@ -6,9 +6,7 @@
 //!
 //! - Exit from a query/tx with a string error:
 //!
-//! ```
-//! use serde_json::json;
-//! use crate::conclude::Output;
+//! ```ignore
 //! let e = String::from("error message");
 //! Output::with_error().with_result(json!(e)).exit();
 //! // or as an alternative:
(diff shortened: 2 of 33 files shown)


modules/src/ics03_connection/connection.rs
```diff
@@ -1,18 +1,21 @@
 use std::convert::{TryFrom, TryInto};
 use std::str::FromStr;
 
+use serde::Serialize;
```
adizere, line 4: FYI: I often let IntelliJ organize the imports for us. The order makes sense to me:

- standard libs
- non-standard libs
- imports from the same project but different crates
- imports from the same crate

I will continue organizing the imports automatically like this with subsequent PRs in the files I touch.

ancazamfir:
> - [`query packet commitment`](https://github.com/informalsystems/ibc-rs/blob/4feec430f08c7b1504916242b76282de54839083/relayer-cli/src/commands/query/packet.rs#L135) command does not output JSON upon success. Reason: it's not clear how this command will be used and what are the requirements on the output. Is the object proof going to be consumed by users/CI @ancazamfir @andynog? The same goes for `query packet ack`

I added these for incremental testing during packet relay development. Used it also when I was debugging some issues with packet proofs at the time. Not sure if we should keep these as they are, remove them or make the output json. I don't believe CI will ever use these.

adizere:
> > * [`query packet commitment`](https://github.com/informalsystems/ibc-rs/blob/4feec430f08c7b1504916242b76282de54839083/relayer-cli/src/commands/query/packet.rs#L135) command does not output JSON upon success. Reason: it's not clear how this command will be used and what are the requirements on the output. Is the object proof going to be consumed by users/CI @ancazamfir @andynog? The same goes for `query packet ack`
> 
> I added these for incremental testing during packet relay development. 

Ah that makes sense, thanks for the explanation!

> Used it also when I was debugging some issues with packet proofs at the time. Not sure if we should keep these as they are, remove them or make the output json. I don't believe CI will ever use these.

Ok. Then I think it would be best if I will change the output to JSON nonetheless, for the sake of completeness. I'll try to include as many fields as possible in the output. If necessary, we'll adapt later on to the exact fields that are necessary.

relayer-cli/src/conclude.rs
```diff
@@ -0,0 +1,115 @@
...
+    /// times, with each subsequent call appending the given `res` JSON object to the output.
+    pub fn with_result(self, res: serde_json::Value) -> Self {
+        let mut new_res = self.result.clone();
+        new_res.push(res);
+
+        Output {
+            result: new_res,
+            ..self
+        }
+    }
```
romac, line 99: This can probably be rewritten as follows, to avoid cloning the `result` field:

```suggestion
    pub fn with_result(mut self, res: serde_json::Value) -> Self {
        self.result.push(res);
        self
    }
```

andynog (approved):
I think it looks good to me. I like the new way to use `Output` instead of `status_info` macro. I think a lot of changes are related for formatting the libs so it's OK. 

Also, the JSON outputs specified in the PR looks good too.

Thanks!

relayer-cli/src/conclude.rs
```diff
@@ -0,0 +1,110 @@
+//! Custom-made solution to output a JSON return message and ensure a return code
+//! from a CLI command. The main use-case for this module is to provide a consistent output for
+//! queries and transactions.
+//!
+//! ## Examples:
+//!
+//! - Exit from a query/tx with a string error:
+//!
+//! ```ignore
```
ancazamfir, line 9: why are these ignored?

relayer-cli/src/lib.rs
```diff
@@ -18,6 +18,7 @@
 
 pub mod application;
 pub mod commands;
+pub(crate) mod conclude;
```
ancazamfir, line 21: I actually wanted to ask this before, why don't we use the crate scope more often. I believe in this case for example all mods, except application, can be public at crate level.

relayer-cli/src/conclude.rs
adizere: Ignored because it's not really testing anything, it's just code snippets to help the developer use the module in the future. There's nothing to assert about this code: The command `Output::with_...exit()` writes to stdout and exits the program, but I don't know how to assert that the exit code from the program was `0` or `1`, nor how to assert that the program wrote anything to stdout.

If it is seems valuable, I can propose two options to make this un-ignored:

1. replace the `ignore` attribute with [`no_run`](https://doc.rust-lang.org/rustdoc/documentation-tests.html#attributes). This means that the code snippet is expected to compile (but not run).

2. remove the `ignore` attribute and try to have full-fledged doc tests that compile+run, but it's unclear if there will by any asserts, actual tests.

Let me know what you think.

ancazamfir (approved):
Looks good! Thanks Adi!

relayer-cli/src/lib.rs
adizere: Agree, for 0.1.0 this may be relevant. @romac any thoughts on this (I can add an issue if so) or is it low-priority?

relayer-cli/src/conclude.rs
ancazamfir: I like the `no_run` option but it's up to you.

relayer-cli/src/lib.rs
romac: I don't thin it matters too much for the `relayer-cli` crate, but yeah in general it wouldn't hurt either to make more modules crate scoped, to limit the surface area of our libs.

relayer-cli/src/conclude.rs
adizere: In the end, I went with attribute `compile_fail` and added an explanation on why the doctests will not compile (the reason is that the doctests need access to the `Output` struct, but this is crate-private and hence not accessible within doc tests).

PR #524, opened 2021-01-14
Make denomination configurable

vitorenesduarte:
## Description

This PR allows the denomination to be configurable.
It also removes what looks like an unnecessary conversion from/to JSON.

______

For contributor use:

### DIFF
--- a/relayer-cli/src/commands/tx/transfer.rs
+++ b/relayer-cli/src/commands/tx/transfer.rs
@@ -30,11 +30,17 @@ pub struct TxRawSendPacketCmd {
     #[options(free, help = "identifier of the source channel")]
     src_channel_id: ChannelId,
 
-    #[options(free, help = "amount of samoleans to send (e.g. `100000`)")]
+    #[options(
+        free,
+        help = "amount of coins (samoleans, by default) to send (e.g. `100000`)"
+    )]
     amount: u64,
 
     #[options(free, help = "timeout in number of blocks since current")]
     height_offset: u64,
+
+    #[options(help = "denomination of the coins to send", short = "d")]
+    denom: Option<String>,
 
     #[options(help = "number of messages to send", short = "n")]
     number_msgs: Option<usize>,
@@ -50,6 +56,7 @@ impl TxRawSendPacketCmd {
             .find_chain(&self.dest_chain_id.parse().unwrap())
             .ok_or_else(|| "missing destination chain configuration".to_string())?;
 
+        let denom = self.denom.clone().unwrap_or("samoleans".to_string());
         let number_msgs = self.number_msgs.unwrap_or(1);
         if number_msgs == 0 {
             return Err("number of messages should be bigger than zero".to_string());
@@ -60,6 +67,7 @@ impl TxRawSendPacketCmd {
             packet_src_port_id: self.src_port_id.clone(),
             packet_src_channel_id: self.src_channel_id.clone(),
             amount: self.amount,
+            denom,
             height_offset: self.height_offset,
             number_msgs,
         };
--- a/relayer-cli/src/commands/tx/transfer.rs
+++ b/relayer-cli/src/commands/tx/transfer.rs
@@ -56,7 +56,10 @@ impl TxRawSendPacketCmd {
             .find_chain(&self.dest_chain_id.parse().unwrap())
             .ok_or_else(|| "missing destination chain configuration".to_string())?;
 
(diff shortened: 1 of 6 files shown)


relayer-cli/src/commands/tx/transfer.rs
```diff
@@ -30,12 +30,18 @@ pub struct TxRawSendPacketCmd {
...
+        free,
+        help = "amount of coins (samoleans, by default) to send (e.g. `100000`)"
+    )]
     amount: u64,
 
     #[options(free, help = "timeout in number of blocks since current")]
     height_offset: u64,
 
+    #[options(help = "denomination of the coins to send", short = "d")]
+    denom: Option<String>,
```
romac, line 43: Can we introduce an `enum` for the supported denominations, even if it's only samoleans at the moment?

vitorenesduarte: My understanding is that denominations are dynamic and can take any value at runtime. For example, after the steps in https://github.com/informalsystems/ibc-rs/blob/master/relayer-cli/relayer_operation_instructions.md, to send the coins back to `ibc-0`, we would need to use something like `ibc/27A6394C3F9FF9C9DCF5DFFADF9BB5FE9A37C7E92B006199894CF1824DF9AC7C` as the denominator. Such denominator can be found with this query: https://github.com/informalsystems/ibc-rs/blob/master/one-chain#L137

vitorenesduarte: Interestingly, I tried to do that and the coins disappeared (both balances are 0), so I'm likely doing something wrong. If someone wants to give it a try, at [this commit](https://github.com/vitorenesduarte/ibc-rs/commit/cb24b8826687d2885bc34bd130093e4c3607647d), the following command should do it: `echo y | dev-env loop_config.toml ibc-0 ibc-1 && instructions.sh`

ancazamfir: > My understanding is that denominations are dynamic and can take any value at runtime

This is correct.

> Interestingly, I tried to do that and the coins disappeared (both balances are 0),..

I tried to transfer some `samoleans` from `ibc-0` to `ibc-1` and then `ibc/27A6394C3F9FF9C9DCF5DFFADF9BB5FE9A37C7E92B006199894CF1824DF9AC7C` tokens back to `ibc-0` from `ibc-1`. Things seem to work. Note that these `ibc/....` tokens sent to original chain show back into the `samoleans` denom balance for `ibc-0`.
Also make sure you complete the handshake (i.e. do the `tx raw packet-recv...` and `tx raw packet-ack...` CLIs)

I will try more tests later. But I think this is good to go.

ancazamfir (commented):
Looks good, thanks Vitor! Could you fix clippy errors and merge master.

relayer-cli/src/commands/tx/transfer.rs
ancazamfir: Just looked now at `instructions.sh` in the other commit...in lines 65-67 you have to flip `ibc-0` and `ibc-1`. It should look like this:

```
DENOM="ibc/27A6394C3F9FF9C9DCF5DFFADF9BB5FE9A37C7E92B006199894CF1824DF9AC7C"
rrly tx raw packet-send ibc-1 ibc-0 transfer channel-0 $AMOUNT $TIMEOUT -n 1 -d $DENOM
rrly tx raw packet-recv ibc-1 ibc-0 07-tendermint-0 07-tendermint-0 transfer transfer channel-0 channel-0
rrly tx raw packet-ack ibc-1 ibc-0 07-tendermint-0 07-tendermint-0 transfer transfer channel-0 channel-0
rrly query packet unreceived-packets ibc-0 ibc-1 transfer channel-0
rrly query packet unreceived-acks ibc-1 ibc-0 transfer channel-0
```

ancazamfir:
Could you please also update the instructions [here](https://github.com/informalsystems/ibc-rs/blob/master/relayer-cli/relayer_operation_instructions.md) with a new section on how to query balances and transfer tokens back and forth?
We should also update CHANGELOG.md, ideally there should be an issue opened for this.

relayer-cli/src/commands/tx/transfer.rs
vitorenesduarte: Thanks for pointing it out. It's working as expected now.

ancazamfir (approved):
LGTM!!! Thanks!

Issue #555, opened 2021-01-26
Problems with relayer::error display for CLIs


ancazamfir:
## Crate
relayer

## Summary of Bug
Not enough detail when the IBC message Tx fails

## Version
9d0a7b8f11a252e68e039a976910edd33107f074

## Steps to Reproduce
current output for `rrly -c loop_config.toml tx raw chan-init ibc-0 ibc-1 connection-0 transfer transfer defaultChannel defaultChannel -o ORDERED` is:
```{"status":"error","result":["tx error: failed"]}```

If using `{:?}` formatter here https://github.com/informalsystems/ibc-rs/blob/9d0a7b8f11a252e68e039a976910edd33107f074/relayer-cli/src/commands/tx/channel.rs#L89

the output of the same command is:
```
[...]

PR #559, opened 2021-01-26
Fix for relayer::error displays

adizere:
- Closes: #555
- Additionally unifies the use the `required` tag across all command options

## Description

______

For contributor use:

### DIFF
--- a/relayer-cli/src/commands/query/client.rs
+++ b/relayer-cli/src/commands/query/client.rs
@@ -19,6 +19,7 @@ use crate::conclude::Output;
 use crate::error::{Error, Kind};
 use crate::prelude::*;
 
+// TODO: refactor commands: why is chain_id an `Option`? simpler to give it `ChainId` type.
 /// Query client state command
 #[derive(Clone, Command, Debug, Options)]
 pub struct QueryClientStateCmd {
@@ -31,7 +32,10 @@ pub struct QueryClientStateCmd {
     #[options(help = "the chain height which this query should reflect", short = "h")]
     height: Option<u64>,
 
-    #[options(help = "whether proof is required", short = "p")]
+    #[options(
+        help = "whether proof is required; default: false (no proof)",
+        short = "p"
+    )]
     proof: Option<bool>,
 }
 
@@ -97,16 +101,24 @@ impl Runnable for QueryClientStateCmd {
 /// Query client consensus command
 #[derive(Clone, Command, Debug, Options)]
 pub struct QueryClientConsensusCmd {
-    #[options(free, help = "identifier of the chain to query")]
+    #[options(free, required, help = "identifier of the chain to query")]
     chain_id: Option<ChainId>,
 
-    #[options(free, help = "identifier of the client to query")]
+    #[options(free, required, help = "identifier of the client to query")]
     client_id: Option<String>,
 
-    #[options(free, help = "epoch of the client's consensus state to query")]
+    #[options(
+        free,
+        required,
+        help = "epoch of the client's consensus state to query"
+    )]
     consensus_epoch: Option<u64>,
 
-    #[options(free, help = "height of the client's consensus state to query")]
+    #[options(
+        free,
+        required,
+        help = "height of the client's consensus state to query"
+    )]
     consensus_height: Option<u64>,
 
     #[options(help = "the chain height which this query should reflect", short = "h")]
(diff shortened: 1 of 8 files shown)


relayer-cli/src/commands/query/client.rs
```diff
@@ -19,19 +19,25 @@ use crate::conclude::Output;
 use crate::error::{Error, Kind};
 use crate::prelude::*;
 
+// TODO: refactor commands: why is chain_id an `Option`? simpler to give it `ChainId` type.
+//  see the tx/client.rs or tx/packet.rs for simpler approaches.
```
romac, line 23: That's because `gumdrop` requires types used in options to have a `Default` instance: https://gist.github.com/romac/23db89d7c5971b0e591a11286fc259a6

While this makes sense for options which are not marked `required`, it's beyond me why the requirement subsists for `required` options.

As such, I am not even sure we need the `required` annotation since we end up checking for the `Option` to be defined anyway but I guess it does not hurt if the error message thrown by `gumdrop` is good enough.

relayer-cli/src/conclude.rs
```diff
@@ -26,6 +26,18 @@
 //! Output::error(format!("{}", e)).exit();
 //! ```
 //!
+//! #### Note:
+//! The resulting output that this approach generates is determined by the 'error' annotation given
+//! to the error object `Kind::Query`. If this error object comprises any positional arguments,
+//! e.g. as achieved by `Query(String, String)`, then it is important to cover these arguments
+//! in the `error` annotation, for instance:
+//! ```compile_fail
```
romac, line 34: `ignore` is a more appropriate annotation here I believe, `compile_fail` should be reserved for snippets of code which must fail to compile for the documentation to be correct.

```suggestion
//! ```ignore
```

relayer-cli/src/commands/query/client.rs
romac: Oh but we have a `Default` instance for `ChainId`. Then yeah we don't need the option and can instead just use `required`.

ancazamfir: We added `Default` -s for everything when we started to use types in the CLI options. I think we should consider moving to something else in V1 (maybe directly `gumdrop`) as we need more flexibility.

ancazamfir, line 23: You mean `client_id`? Yes, this file slipped through the cracks. We had everything `String` initially and then moved gradually to typed options for most CLIs.

adizere: > As such, I am not even sure we need the required annotation since we end up checking for the Option to be defined anyway but I guess it does not hurt if the error message thrown by gumdrop is good enough.

The error message from gumdrop is a bit.. underwhelming. But it helps still. It says only:
> error: missing required free argument

Was looking at ways to parametrize this, but did not find anything.

The way I understand it, with 'required' we now have two levels of checks:

1. gumdrop enforces some (any) value to be present
2. in our validate_options methods we parse and do additional checks on the field.


> We added Default -s for everything when we started to use types in the CLI options. I think we should consider moving to something else in V1 (maybe directly gumdrop) as we need more flexibility.

Sounds like a plan.

adizere: I meant the `chain_id` field in the command:


https://github.com/informalsystems/ibc-rs/blob/9d0a7b8f11a252e68e039a976910edd33107f074/relayer-cli/src/commands/query/client.rs#L24-L26

No reason to wrap it in an Option there. But also applies to the `client_id` which should not be an option either and should be type `ClientId`.

relayer-cli/src/conclude.rs
adizere: This was initially `ignore` but [changed](https://github.com/informalsystems/ibc-rs/pull/518#discussion_r557326691) to `compile_fail`.

You're right, I didn't catch that subtlety! The documentation here fails to compile not as a way to assert its correctness, but because the module is not public..

https://github.com/informalsystems/ibc-rs/blob/80d222bbc6f28e0f0461393dec9f66a22e2ad975/relayer-cli/src/lib.rs#L22

I'll change to `ignore` which seems like a better option in this case. Thanks!

adizere: 4d78ed0

relayer-cli/src/commands/query/client.rs
ancazamfir: changed in ecb53189a66496c714ebec00f1769132776e191f. For the rest I opened https://github.com/informalsystems/ibc-rs/issues/572.

ancazamfir (approved):
Looks good, thanks Adi!

relayer-cli/src/commands/query/client.rs
adizere: Opened an issue to continue discussion, if necessary, and track this

https://github.com/informalsystems/ibc-rs/issues/573

Issue #501, opened 2021-01-05
Cannot run relayer commands concurrently


romac:
## Crate

`relayer`, `relayer-cli`, `tendermint-light-client`

## Problem Definition

If one attempts to run two relayer commands concurrently (eg. `tx raw packet-send` while the relayer is running), they will be greeted with an error like:

```
error: client create failed: tx error: Light client error: IO error: could not acquire lock on "data/DEEB0AB3F2D7BEC0B8C9FF1532715635314F54D5/db": Os { code: 35, kind: WouldBlock, message: "Resource temporarily unavailable" }
```

This happens because each processes will instantiate a chain runtime which in turn will instantiate a light client, and both light clients will attempt to open the same `sled` database.

In turns out that `sled` does not allow multiple processes to have the database file open at once, hence the error above.

## Proposal

Implement a new instance of the `LightStore` trait backed by a database which supports concurrent use across processes (eg. SQLite, RocksDB).

Alternatively, standalone commands (ie. `tx` commands) could use an in-memory light store, thus reserving the on-disk on for the main relayer loop (`start` command). This has performance implication for those commands as they will start from an empty light store and should thus be evaluated carefully.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #562, opened 2021-01-27
Override the light client store with the in-memory one in all commands except the relayer loop

romac:
Closes: #501

## Description

Override the configured light store with the in-memory implementation in all standalone commands, thus reserving the on-disk store for the main relayer loop (`start` command). 

This allows running one-off commands concurrently the main relayer loop, at the expense of performance for those commands as they will start from an empty light store every time.

This is currently good enough for v0.1.0, but be re-evaluated for v0.2.0.

This was tested by launching the relayer loop and running multiple `tx raw update-client ibc-0 ibc-1 07-tendermint-0` commands concurrently.

______

For contributor use:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
   - Replace `ChannelConfig` in `Channel::new` ([#511])
   - Add `packet-send` CLI ([#470])
   - UX improvements for relayer txs ([#536, #540, #554])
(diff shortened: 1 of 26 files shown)


ancazamfir (approved):
Fantastic! Works beautifully and reproduced a bug I was expecting in the relayer loop retry part. Let's merge this, thanks Romain!

Issue #553, opened 2021-01-26
Query all connections CLI


ancazamfir:
## Crate
relayer, relayer-cli

## Summary
Add new `rrly -c config.toml query connections <chain-id>` to allow the user to query all connections on a given chain. This should return a list of `ConnectionId`-s that can be used to query individual connections.

## Proposal

____

#### For Admin Use

- [x] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #425, opened 2020-11-27
ibc-proto planning issue


greg-szabo:
## Crate
ibc-proto

## Summary
New requests came in to enrich the ibc-proto crate with cosmos-SDK related protobuf structs or rename it or separate it.

This issue tries to coordinate among these requests to find out what would be the best possible future for the ibc-proto crate.

## Problem Definition
ibc-proto only implements a subset of the Cosmos SDK protobuf definitions. Notably, the ones for IBC.

It would be nice to have one go-to protobuf crate for Cosmos SDK protobuf structs in Rust. But that might require its own repository (as well as better version control as other issues already point that out).

## Proposal
(Will be updated based on the discussion.)
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

greg-szabo:
How does this sound like:
a separate `cosmos-sdk-proto` crate that encompasses all the structs from cosmos/cosmos-sdk. Version control and recompilation options would be kept in the repository too. (proto-compiler or similar)

jkilpatr:
ibc-proto as it exists includes some Cosmos protos that are more generic, BaseAccount for example. Obviously it's not complete in that aspect.

I like the pattern in ibc-proto and tendermint-proto and I've talked with the ICF about moving them upstream and gotten the unimaginatively named https://github.com/cosmos/rust-monorepo repo for the purpose. 

Right now I think just moving the proto definitions there, making them more complete, and creating a definitive location for the maintenance of proto data would be a big win. We can work on having a standard client library and other such functionality later.

CharlyCst:
I was about to open an issue for a feature request, but it seems that it would be more appropriate to post it here:
I'm building a mock node for testing purpose and I rely on `ibc-proto` for proto definitions, however `proto-compiler` only builds client-related definitions (see [this line](https://github.com/informalsystems/ibc-rs/blob/2aa305bfc2079708e3d7cc77595d5587b4c67821/proto-compiler/src/cmd/compile.rs#L117)). For now I simply forked the crate but I would very much appreciate having all the definitions in a single place.

I'm not very familiar with `ibc-proto` and the needs of the cosmos ecosystem in general, but if someone is willing to provide some guidance I can help on moving thins forward.

adizere:
[...]

PR #565, opened 2021-01-27
Added CLI to query for all connections

adizere:
Closes: #553 

Example output:

```
$ rrly -c config.toml query connections ibc-1

{"status":"success","result":[["connection-0","connection-1"]]}
```
______

For contributor use:


modules/src/ics02_client/raw.rs
```diff
@@ -44,3 +44,15 @@ impl From<ConnectionIds> for RawClientConnections {
         }
     }
 }
+
+impl FromIterator<ConnectionId> for ConnectionIds {
```
ancazamfir, line 48: Haven't  noticed this file before. Why is it called `raw.rs` and part of the `ics02_client`? Its content relates to connections. It would be nice to move this to `ics03_connection`.

ancazamfir (approved):
Looks great! Just one small comment, will leave it up to you.

ancazamfir:
One more thing, could you optimize imports on the changed files?

modules/src/ics02_client/raw.rs
adizere: This file was added during one of the first queries, `query client connections` I think in #169.

The `RawClientConnections` connections and `ConnectionIds` are the main types in this file. It's not really clear to me at this point where it would be best to place these types. (ICS03 is better than ICS02, but still this file contains a raw/proto type, which is the weirdest part that should be solved first.) I propose we refactor this and move stuff around as necessary in https://github.com/informalsystems/ibc-rs/issues/425.

ancazamfir: `RawClientConnections` is defined locally, not part of any protofbuf. And it is not what the name suggests, it's just  a vector of `String` -s. I'm not sure we need it at all to be honest. If we keep it, ics03 is the best place. I will take care of cleaning it in the next PR.

Issue #577, opened 2021-01-28
Split the dev-env script in `setup_chains` and `init_clients`


ancazamfir:
## Crate
n/a

## Problem Definition
Currently the `dev-env` script starts two chains (one gaia node per chain) and then does the subjective initialization for the relayer light clients. We need to be able to do the two steps separately. In case the chains don't need to be started only the latter part is needed. We can still keep the `dev-env` and call the two from there.
Also, we should add a store parameter to the light client init script such that we can run multiple relayer instances (with different config files and light client sets) on the same machine.

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #542, opened 2021-01-19
Relayer documentation


andynog:
## Summary

Documentation for the Rust Relayer

## Problem Definition

Currently the documentation for setting up a relayer with two chains is outdated or missing. There's not current easy way to setup a relayer without some previous knowledge on how to do it or ask for assistance. 

## Proposal

Create initial documentation and instructions up to date for the relayer. 

Create docs using some framework e.g. mdbook or VuePress that can render the site automatically when docs are updated.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
- [ ] cleanup the sample .toml configurations: there's currently three of them under version control and it's not clear why and what are each used for: 
    - `relayer/tests/config/fixtures/relayer_conf_example.toml` and 
    - `relayer-cli/tests/fixtures/two_chains.toml` and 
    - `relayer/tests/config/fixtures/simple_config.toml`.

PR #589, opened 2021-01-29
Relayer documentation

romac:
Closes: #542
Closes: #577

TODOs:
- [x] Connection tx and queries (4.3.2 and 4.5.2) - @romac 
- [x] Packet tx and queries (4.3.4 and 4.5.3) - @ancazamfir 
- [x] Channel open tx and queries (4.3.3 and 4.5.3) - @romac
- [x] Channel close tx - @romac
- [x] Split `dev-env` scripts - #577 - @romac
- [x] #598 Update all README.md files in relayer-cli crate and main repo. - @adizere 
- [x] Section 4.6 (not entirely sure what should go here, maybe the listen command?),
- [x] Fill section 5.4 - @ancazamfir 
- [x] Section 6 - @adizere 
- [x] Section 7- remove
- [x] Feature list - @ancazamfir (in progress)
- [x] Clarify throughout the docs that only unordered channels are currently supported - @ancazamfir 
- [x] 4.1.4 needs cleanup - @ancazamfir 
- [x] 5.4.5 channel close  - @romac
- [x] new section 8 with everything that didn't fit in the other sections, need a home or removed -- @adizere 
______

For contributor use:


ebuchman:
Some rough notes from working through this as a group on Friday, some can probably be fixed here and some followed up in new issues:

local_chains.md:

- create the loop_config.toml file for me
- ./dev-env is in ./scripts ...

relay_packets.md
- output JSON should be pretty printed!
- 07-tendermint-1 is wrong

query client consensus
    - params are confusing (0 1393)
    - root shows [1,2,3] bytes not hex

query connection end
    - prefix is bytes - should be asccii ...

chan-open-init
    - dont need dst channel and src channel

all the commands are wrong because of the xxx-1

3.5 formatting is broken

packet relaying:
- packet data is bytes, should be at least hex, better if JSON ...

packet send:
- what address are we sending to ?
- src and dest are swapped here compared to others ...


query packet commitment 
- result is bytes should be hex
- how to query actual packet data from event system?

tx raw packet-recv
- all bytes again. should be hex or actual data

romac:
@ebuchman

> packet relaying:
> 
> * packet data is bytes, should be at least hex, better if JSON ...

Given that the `data` field is defined in the .proto file as an opaque byte array, I am not sure how we would proceed to display it a JSON. Is this field typically JSON or Protobuf-encoded? If so, then we could perhaps try to optimistically decode it as either and fall back to hex if that didn't work.

https://github.com/informalsystems/ibc-rs/blob/97f27e3475d159660ba0af6724732f071fd2616d/proto/src/prost/ibc.core.channel.v1.rs#L80-L82

ancazamfir:
> packet send:
what address are we sending to ?
src and dest are swapped here compared to others ...

I will clarify in 4.3.4..not there yet. This is a test CLI so we can actually have some packets to send. It should probably be renamed to `ics20-transfer` or something like that. This CLI creates a `MsgTransfer` packet that is stored on chain inside `send_packet` as its opaque `data`.  The address is, in the example, the relayer account on `ibc-1` and the transfer is from the relayer account on `ibc-0`.  You can use the CLI to send multiple packets and also specify the timeout offset. Again, all this is for testing purposes.

Regarding the decoding of packet data, the data is extracted from the Tx results, it is an opaque bytearray and we don't currently decode it. We would need to do it based on the port associated with it, e.g. if `transfer` then it would be a `MsgTransfer` packet.

ancazamfir:
> create the loop_config.toml file for me 

added `config_example.toml` in `ibc-rs`

> ./dev-env is in ./scripts ...

fixed

> query client consensus
- params are confusing (0 1393)

will clarify in section 4.3.1 (currently empty), i don't see it documented anywhere. Where did you get those numbers from?

> chan-open-init
- dont need dst channel and src channel

Most CLIs need cleanup, will do it in #594

> all the commands are wrong because of the xxx-1
fixed the docs

romac:
> * params are confusing (0 1393)
> 
> will clarify in section 4.3.1 (currently empty), i don't see it documented anywhere. Where did you get those numbers from?
> 
> >

These should perhaps be flags instead of positional arguments, eg.

```
$ hermes query client consensus ibc-1 07-tendermint-2 --consensus-epoch 0 --consensus-height 948
```

adizere (approved):

ancazamfir (approved):
Amazing work everyone!! This has been an ironman race, let's merge this!

PR #592, opened 2021-02-01
TLA spec bugs fixed in IBCCore and IBCTokenTransfer

istoilkovska:
## Description

@vitorenesduarte was checking the type invariant of the `IBCCore.tla` spec (for the ICS03 part) by running TLC, and found the following bug in the TLA spec: 
 - the relayer (`ibc-core/ICS18Relayer.tla`) creates connection datagrams even when the counterparty clients are not initialized on the chains. 
 
By checking the type invariant for the full `IBCCore.tla`, I've found bugs in the ICS04 part of the spec: 
  - in `WritePacketCommitment` in the file `ibc-core/ICS04PacketHandlers.tla`, the IF-THEN-ELSE conditions were improperly nested
  - in `SendPacket`in the file `ibc-core/Chain.tla`, sending packet was enabled even when `WritePacketCommitment` did not update the chain store, i.e., when it did not write a packet commitment. 

This PR fixes these spec bugs, both in the IBCCore and IBCTokenTransfer specs. Additionally, a type invariant for IBCTokenTransfer is added, and it holds for the parameter values defined in `fungible-token-transfer/IBCTokenTransfer.cfg`.

______

For contributor use:


docs/spec/tla/ibc-core/ICS04PacketHandlers.tla
```diff
@@ -164,19 +164,19 @@ WritePacketCommitment(chain, packet) ==
...
+              THEN [chain EXCEPT 
+                    !.packetCommitments =  
+                        chain.packetCommitments \union {[portID |-> packet.srcPortID,
+                                                         channelID |-> packet.srcChannelID,
+                                                         sequence |-> packet.sequence,
+                                                         timeoutHeight |-> packet.timeoutHeight]}
+                   ]
+              \* otherwise, do not update the chain store
+              ELSE chain
+    ELSE chain
```
vitorenesduarte, line 179: It's not clear to me why this change is required. The `TypeOK` invariant should not hold if the `channelEnd.order \not \in {"ORDERED", "UNORDERED"}`.

vitorenesduarte:
This look great, thanks @istoilkovska.

vitorenesduarte (approved):

docs/spec/tla/ibc-core/ICS04PacketHandlers.tla
istoilkovska: This change was added because the `ELSE IF channelEnd.order = "UNORDERED"` was wrongly aligned -- a packet commitment on an unordered channel was written even if the conditions on lines 143-153 did not hold. This in fact means that a packet is sent on an uninitialized or closed channel, which should not be the case. 

The line 178 can be the outcome of `WritePacketCommitment` in case the channel is ordered, but the sequence numbers are not correct. This is not a violation of the type invariant, but of the property which checks if packets are sent in order on ordered channels. We currently don't have any properties/invariants for the packet transmission in the IBC spec. We should address this in a separate PR.

vitorenesduarte: Oh, I see. Thanks for explaining.

Issue #546, opened 2021-01-21
Implement MBT for ICS02 Client handlers


andrey-kuprianov:
We need to initiate model-based testing for IBC-rs; IBC client handlers (ICS02) have been chosen for that.

The goal is to complement the current manual tests for client handlers with MBT-based ones. The model to be used for that can be found in Ilina's [modularize_client branch](https://github.com/informalsystems/ibc-rs/tree/ilina/modularize_client/docs/spec/client). This needs to be discussed also with Adi who is doing work on re-enabling the hand-written tests in #535.

andrey-kuprianov:
We have figured out that client handlers are too primitive to throw the heavy MBT machinery at them; e.g. the [CreateClient handler](https://github.com/informalsystems/ibc-rs/blob/9d0a7b8f11a252e68e039a976910edd33107f074/modules/src/ics02_client/handler/create_client.rs#L24-L52) is only incrementing one counter, so no real logic to test.

Considering that now the Channel handlers are coming to completeness, we are shifting focus to them. They have quite non-trivial implementation, that deserves thorough testing; see e.g. the [ChanOpenTry handler](https://github.com/informalsystems/ibc-rs/blob/9d0a7b8f11a252e68e039a976910edd33107f074/modules/src/ics04_channel/handler/chan_open_try.rs#L16-L158).

The corresponding [TLA+ model](https://github.com/informalsystems/ibc-rs/blob/9d0a7b8f11a252e68e039a976910edd33107f074/docs/spec/tla/ibc-core/ICS04ChannelHandlers.tla#L64-L122) is also non-trivial, so this seems to be a good match for MBT.

Issue #405, opened 2020-11-18
Clarify the Height and ChainId types (ref: "epochs")


adizere:
## Crate

`ibc`

## Summary

The `Height` type and the `ChainId` types are closely related: the `version_number` field in a height of a chain denotes the revision of that chain. The revision used to be called "epoch," then renamed to "version", but it seems that the name of this field will stabilize to "revision".

There's two things to improve/clarify on our side:

1. Our `ChainId` implementation is a bit messy and uses both the names of "version" and "epoch". 

https://github.com/informalsystems/ibc-rs/blob/45ae7b86844c906570f7b16792ef4806ed45ab54/modules/src/ics24_host/identifier.rs#L17-L25

2. The field of type `Height` might soon evolve so we probably need to rename `version_number` into `revision_number`.

https://github.com/informalsystems/ibc-rs/blob/b641f794cff8e5f21c308bd419fe4c29f028fda6/modules/src/ics02_client/height.rs#L9-L12

Dependencies:
- sdk: https://github.com/cosmos/cosmos-sdk/issues/7709
- spec: https://github.com/cosmos/ics/issues/481

## Proposal

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

thanethomson:
It'd make things a lot easier just to name this `epoch` everywhere. It has a pretty clear meaning and is quite distinct as a variable name.

Right now (from what I've managed to find so far) it's referred to as:
1. `revision_number`
2. `version`
3. `counter`

depending on where it's being referenced.

adizere:
I think using "epoch" as a field name in the `Height` type is not really an option, it would be more appropriate to align with the specs and the .proto and have the `Height` type consist of fields `revision_number` and `revision_height`. We did this alignment [in this PR](https://github.com/informalsystems/ibc-rs/pull/449) I think.

I'm not convinced using `epoch` in the `ChainId` type is a good idea. Does the term epoch originate from tendermint-go?

[...]

PR #601, opened 2021-02-02
Model-based tests for ICS02

vitorenesduarte:
Towards: #546

## Description

This PR is the first step towards #601. For ease of reviewing, we plan to introduce one ICS at a time, starting now with ICS02.

For some details on MBT and current limitations, please check [modules/tests/README.md](https://github.com/informalsystems/ibc-rs/blob/vitor/ics02_mbt/modules/tests/README.md) (which steals ideas from the [light-client one](https://github.com/informalsystems/tendermint-rs/blob/master/light-client/tests/support/model_based/README.md)).

______

For contributor use:


modules/Cargo.toml
```diff
@@ -37,6 +37,8 @@ dyn-clonable = "0.9.0"
 regex = "1"
 bech32 = "0.7.2"
 
+eyre = "0.6.5"
```
romac, line 40: This should be moved the `dev-dependencies` section as it is only used in tests and we want to keep the `modules` crate free of error handling libraries.

vitorenesduarte: Done in d57cf4119a9dbfe5260ef7c1afc212bc5c305458. Once `modelator.rs` lives in its own crate, we will depend only on that, so that dependency can go away.

modules/tests/model_based.rs
```diff
@@ -0,0 +1,266 @@
...
+        ics26_error
+            .source()
+            .expect("expected source in ICS26 error")
+            .downcast_ref::<anomaly::Error<K>>()
+            .expect("ICS26 source should be an handler error")
+            .kind()
+            .clone()
+    }
+
+    // TODO: this is sometimes called version/revision number but seems
```
adizere, line 95: Chain "epoch" was redefined to "revision" I think, we just didn't get a chance to make it consistent thoughout the codebase. The same goes for the epoch in heights. Ref: https://github.com/informalsystems/ibc-rs/issues/405

True this has nothing to do with `Version` type for channels or connections.

modules/tests/model_based.rs
```diff
@@ -0,0 +1,266 @@
...
+use std::collections::HashMap;
+use std::error::Error;
+use std::fmt::{Debug, Display};
+use step::{ActionOutcome, ActionType, Chain, Step};
+use tendermint::account::Id as AccountId;
+
+#[derive(Debug)]
+struct IBCTestExecutor {
+    // mapping from chain identifier to its context
+    contexts: HashMap<String, MockContext>,
```
adizere, line 29: I was wondering if it would have any positive effect if we used the domain type [`ChainId`](https://github.com/informalsystems/ibc-rs/blob/9ee48bd2a7c337277de3889df06d963b0fd005bf/modules/src/ics24_host/identifier.rs#L18) instead of String here. This would make the MBT suite more tedious here because of additional parsing/error catching (though serde could help there), but maybe it would align it better with the rest of the codebase.

modules/tests/modelator.rs
```diff
@@ -0,0 +1,49 @@
+use eyre::{eyre, Context, Result};
```
adizere, line 1: Could we give a bit more structure to this directory, e.g., put modelator.rs and step.rs inside a module of their own (which we can call `model_based`).

modules/tests/support/model_based/IBC.cfg
```diff
@@ -0,0 +1,12 @@
+CONSTANTS
+    ChainIds = {"chain-A", "chain-B"}
```
adizere, line 2: Would be a good idea to give chain identifiers the same format as in production, e.g., "chainA-0" and "chainB-0".

modules/tests/support/model_based/IBCTests.cfg
```diff
@@ -0,0 +1,8 @@
+CONSTANTS
```
adizere, line 1: Not clear what is the difference b/t this file and `modules/tests/support/model_based/IBC.cfg`.

modules/tests/support/model_based/README.md
```diff
@@ -0,0 +1,38 @@
+## Model-based tests for IBC modules
```
adizere, line 1: I would move this readme to a higher-level, inside `modules/tests/`. Here it's a bit buried and it's easy to overlook it.

adizere (changes requested):
Impressive work!

My main worry is that we're introducing new TLA+ specs here, instead of leveraging (part of) what we already had. I'm curious what @ancazamfir, @andrey-kuprianov, and @istoilkovska think about this: would it be possible, at least eventually, to reuse Ilina's TLA+ specs for MBT?

modules/tests/modelator.rs
```diff
@@ -0,0 +1,49 @@
...
+    for (i, step) in steps.into_iter().enumerate() {
+        // check the step
+        let ok = if i == 0 {
+            executor.initial_step(step.clone())
+        } else {
+            executor.next_step(step.clone())
+        };
+
+        if !ok {
+            return Err(eyre!(
```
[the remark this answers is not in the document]
romac: Agreed, it would be nice if we could get away with only `thiserror`.

modules/src/ics02_client/msgs/create_client.rs
```diff
@@ -22,9 +22,9 @@ pub const TYPE_URL: &str = "/ibc.core.client.v1.MsgCreateClient";
 /// A type of message that triggers the creation of a new on-chain (IBC) client.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct MsgCreateAnyClient {
-    client_state: AnyClientState,
-    consensus_state: AnyConsensusState,
-    signer: AccountId,
+    pub client_state: AnyClientState,
```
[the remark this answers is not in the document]
romac: We are quite inconsistent with this all over the codebase, and it's something we should uniformly sort out at some point.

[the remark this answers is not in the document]
vitorenesduarte: The convention seems to be to use `pub`, which explains the change here. I'll do a follow up PR where we remove all the getters from these structs.

modules/tests/support/model_based/IBC.cfg
vitorenesduarte: Could you point me to place where this format is defined? Can there be a `"chainA-0"` and a `"chainA-1"`?

modules/tests/support/model_based/README.md
vitorenesduarte: Done in 5e9dd634a3bc04c0f193ab3872b83dc00505d1d4.

modules/tests/model_based.rs
vitorenesduarte: Done in fb20ce0161ff8ad7009b68af9ff8978e508f4ee9.

modules/tests/model_based.rs
vitorenesduarte: Thanks. Renamed `epoch` to `revision` in 75db6bb9621e3a6db269100332e137d3a526a822.

modules/tests/modelator.rs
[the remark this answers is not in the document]
vitorenesduarte: Tried something in a10286398eee2ff375e727720b716d1293bcf12c. Please take a look @adizere @romac. I should also say that the plan is to eventually (hopefully soon) have the functionality in `moderator.rs` moved to its own crate.

modules/tests/modelator.rs
vitorenesduarte: Once `modelator.rs` goes way, this directory would only contain `step.rs`. Do you mind if we delay this and try to improve structure in future PRs?

modules/tests/modelator.rs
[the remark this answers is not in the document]
romac: Looks good

modules/tests/support/model_based/IBCTests.cfg
vitorenesduarte: Tried to improve the README explaining this in 2e5d6de82f69a14046b8555f95c706701b140415.

vitorenesduarte:
Thanks @adizere and @romac for the review! I just finished addressing all your comments.

modules/tests/model_based.rs
```diff
@@ -0,0 +1,269 @@
...
+                    consensus_state: Self::consensus_state(client_height),
+                    signer: Self::signer(),
+                }));
+                let result = ctx.deliver(msg);
+
+                // check the expected outcome: client create always succeeds
+                match step.action_outcome {
+                    ActionOutcome::ICS02CreateOK => {
+                        // the implementaion matches the model if no error occurs
+                        result.is_ok()
```
ancazamfir, line 190: should we also check that the new client is stored in the `ctx`, i.e. it can be retrieved?

modules/tests/model_based.rs
```diff
@@ -0,0 +1,269 @@
...
+                    header: Self::header(client_height),
+                    signer: Self::signer(),
+                }));
+                let result = ctx.deliver(msg);
+
+                // check the expected outcome
+                match step.action_outcome {
+                    ActionOutcome::ICS02UpdateOK => {
+                        // the implementaion matches the model if no error occurs
+                        result.is_ok()
```
ancazamfir, line 225: should we also check that the client in `ctx` has the updated consensus state?

modules/tests/model_based.rs
```diff
@@ -0,0 +1,269 @@
...
+                            ICS02ErrorKind::ClientNotFound(error_client_id)
+                            if error_client_id == Self::client_id(client_id)
+                        )
+                    }
+                    ActionOutcome::ICS02HeaderVerificationFailure => {
+                        let handler_error_kind =
+                            Self::extract_handler_error_kind::<ICS02ErrorKind>(result);
+                        // the implementaion matches the model if there's an
+                        // error matching the expected outcome
+                        handler_error_kind == ICS02ErrorKind::HeaderVerificationFailure
```
ancazamfir, line 243: should we also check that the client's consensus state has not been updated?

modules/tests/step.rs
```diff
@@ -0,0 +1,49 @@
...
+    #[serde(alias = "chainId")]
+    pub chain_id: Option<String>,
+
+    #[serde(alias = "clientId")]
+    pub client_id: Option<u64>,
+
+    #[serde(alias = "clientHeight")]
+    pub client_height: Option<u64>,
+}
+
```
ancazamfir, line 29: For create client we need to have a `client_state` and for update client we need a `header`. It is true that for the mock client these boil down to a `client_height` but if we want to extend these tests for other client types or make changes to the mock client (e.g. add a `chain-id` or client specific parameters to the mock state, or add a timestamp to the header/ consensus state)

ancazamfir (commented):
Looks good after a fist pass. Entered a few comments, will take another look.

modules/tests/modelator.rs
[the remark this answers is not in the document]
andrey-kuprianov: yes, minimising the dependencies is a good idea

modules/tests/model_based.rs
vitorenesduarte: This is a good idea! In the same way we check chain heights after each model action, we can extend it to also check the state of clients. I've opened https://github.com/informalsystems/ibc-rs/issues/629 to track this idea (which is not so easy to implement right now).

modules/tests/model_based.rs
vitorenesduarte: Same.

modules/tests/model_based.rs
vitorenesduarte: Same.

modules/tests/step.rs
vitorenesduarte: Currently the model does not distinguish between client state or consensus state. For each client, only the current height is kept. For this reason, there's no `"ICS02ConsensusStateNotFound"` error in the model, only `"ICS02ClientNotFound"` (as they cannot be distinguished).

Are you suggesting that the model should be extended to make the distinction? In general, we cannot hope to model everything, but maybe this is one of the situations where we should. At this point it is not obvious to me why the current abstraction is insufficient. It might become clear once there are other client types to be tested.

modules/src/ics02_client/msgs/create_client.rs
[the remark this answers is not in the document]
vitorenesduarte: Opened https://github.com/informalsystems/ibc-rs/issues/631.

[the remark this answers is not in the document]
adizere: thank you!

modules/tests/support/model_based/IBC.cfg
adizere: This format is just a convention at the moment, not sure it's formally defined anywhere.

> Can there be a "chainA-0" and a "chainA-1"?

This would be the same chain, in two subsequent epochs or revisions: revision `0` (due to suffix `-0`), and revision `1`. This also explains why a ibc::Height has two fields: a revision and a number.

For "chainA-0", the revision would be "0" at all height numbers of that chain.
For "chainA-1", the revision would be "1" at all height numbers of that chain...

modules/tests/step.rs
ancazamfir: > Are you suggesting that the model should be extended to make the distinction? 

Yes

> In general, we cannot hope to model everything, but maybe this is one of the situations where we should. At this point it is not obvious to me why the current abstraction is insufficient. It might become clear once there are other client types to be tested.

Understood we cannot model everything. But the create_client message includes a client state (`AnyClientState`) and a consensus state (`AnyConsensusState`) and I think we should pass those in the `action`.  update_client should take `AnyHeader`. 
It's not only for the different client types and MBT testing for clients. When we do MBT for connection we will need to pass `AnyConsensusState` (and not a client height) in some of the messages. And this would be something that we should have previously created a client with (or not, depending on the test).

vitorenesduarte: > Understood we cannot model everything. But the create_client message includes a client state (`AnyClientState`) and a consensus state (`AnyConsensusState`) and I think we should pass those in the `action`. update_client should take `AnyHeader`.

I can make this distinction in the model, but it's not clear to me what it brings to ICS02. For example, what should be a client state and a consensus state in the model? I can model both as a height, but it feels like this is the same as having a single height (which is what's modeled now).

> It's not only for the different client types and MBT testing for clients. When we do MBT for connection we will need to pass `AnyConsensusState` (and not a client height) in some of the messages. And this would be something that we should have previously created a client with (or not, depending on the test).

If the distinction will be needed for ICS03, can we address this concern in the PR with MBT for ICS03?

ancazamfir: We don't have to hold this PR necessarily.  I will leave it up to you. But I think we need to make the change so that we can easily extend the tests. And just to clarify this is not only needed for ICS03.

modules/tests/support/model_based/IBC.cfg
vitorenesduarte: That makes sense, thanks! Done in 5d6bf08c7fcfe4d9040c661f0f044ebf641ea31f.

modules/tests/step.rs
vitorenesduarte: After discussing with @ancazamfir, we have reached the conclusion that the model should keep, per client, the latest `N` heights, not just the latest one. The current model is a special case where `N = 1`. I've opened an issue to track this: https://github.com/informalsystems/ibc-rs/issues/633. Thanks @ancazamfir!

ancazamfir (approved):

ancazamfir:
Let's merge this one and address any pending comments in #633.

adizere (approved):
Good work Vitor!

Issue #582, opened 2021-01-29
Speed up CI by re-using relayer binary for E2E tests


romac:
We could likely significantly speed up the end-to-end test job by copying the relayer binary built as part of the `Rust` job into the image instead of rebuilding it.

___

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
The original CI infrastructure was done as part of https://github.com/informalsystems/ibc-rs/pull/544.
This was later extended with comprehensive tests in https://github.com/informalsystems/ibc-rs/pull/571.

The work in the original CI consists of three docker images:
- two of them were gaia chain images called "informaldev/ibc-0" and "informaldev/ibc-1" 
- the third image is for the relayer, and is dependent on the gaia images; this is built on-the-fly upon every commit to a pull-request

There are two basic differences between the original work and the new PR:
1. the relayer binary is now compiled as part of the Github action, and then this binary is made available (`cp`) to the relayer docker image:

https://github.com/informalsystems/ibc-rs/blob/81af0b0b0418c4a365617e3cbf6006db80a3c831/.github/workflows/e2e.yaml#L13-L17


 in the original work, the binary was compiled directly inside the `docker-compose` task:

https://github.com/informalsystems/ibc-rs/blob/148e196308a217542f2dc49bfe5881362bf1013d/ci/relayer.Dockerfile#L28

2. the `cargo build` command for obtaining the relayer binary no longer relies on `--all --release`, shortening the build cycle by ~5 minutes (almost halved)

PR #604, opened 2021-02-03
Speed-up CI end-to-end tests

adizere:
Close: #582

______

For contributor use:

- [X] If applicable: Unit tests written, added test to CI.


.github/workflows/e2e.yaml
```diff
@@ -2,13 +2,22 @@ name: End to End testing
...
     steps:
       - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+      - uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --workspace --all-targets
```
adizere, line 16: The binary is built in debug mode (without `--release`). This is what accounts primarily for shortening the workflow cycle.

.github/workflows/rust.yml
```diff
@@ -44,10 +44,6 @@ jobs:
         with:
           toolchain: stable
           override: true
-      - uses: actions-rs/cargo@v1
-        with:
-          command: build
```
adizere, line 49: This build was not necessary.

ci/relayer.Dockerfile
```diff
@@ -35,11 +6,11 @@ LABEL maintainer="hello@informal.systems"
 
 ARG RELEASE
 
-# Add jq and Python 3
-RUN apt-get update -y && apt-get install python3 jq -y
```
adizere, line 39: `jq` wasn't used in the relayer image.

.github/workflows/e2e.yaml
romac, line 16: While we are at it we can also remove `--all-targets` as we don't not need to build tests, etc.

```suggestion
          args: --workspace
```

andynog (approved):
The idea of building the relayer inside the Docker container was to avoid any side effects of building in one platform but since the Docker container is Debian and the CI uses Ubuntu, I think it's all good and the speed improvement are great

adizere:
> The idea of building the relayer inside the Docker container was to avoid any side effects of building in one platform but since the Docker container is Debian and the CI uses Ubuntu, I think it's all good and the speed improvement are great

Hmm this is something I overlooked!

In a future PR, maybe we run the relayer binary directly from Github actions? This would further simplify the workflow (and probably reduce the time).

PR #612, opened 2021-02-05
Add table of contents to query and tx reference pages

romac:
https://lib.rs/crates/mdbook-toc


andynog (approved):
Looks good to me. Tested the `toc` and seems to be working fine. Good work!

PR #613, opened 2021-02-05
Unambiguous packet query

vitorenesduarte:
Closes: #614

## Description

Before, a `query_txs` for a packet on a given chain would filter only by the packet source port, source channel and sequence.
With 2 chains, there can be at most a single packet matching such filter, and this fact was assumed in the code.

However, this assumption does not hold with more than 2 chains. To understand why, consider the following example with chains `ibc-0`, `ibc-1` and `ibc-2`:
- a channel handshake occurs between `ibc-0` and `ibc-1`; the channel identifiers are `channel-0` on both ends
- a channel handshake occurs between `ibc-1` and `ibc-2`; the channel identifier on `ibc-1` is `channel-1` and on `ibc-2` is `channel-0`

If now both `ibc-0` and `ibc-2` send packets to `ibc-1` and we do a packet query filtering only by the source, such query will return
packets from both `ibc-0` and `ibc-1` (as both have `channel-0` as the source channel in their packets).

In this commit, we start filtering also by the destination port and destination channel. This makes the packet query unambiguous.

Thanks @ancazamfir for the help debugging this issue. It was a fun one!

______

For contributor use:

### DIFF
--- a/modules/src/ics04_channel/channel.rs
+++ b/modules/src/ics04_channel/channel.rs
@@ -324,6 +324,8 @@ pub struct QueryPacketEventDataRequest {
     pub event_id: IBCEventType,
     pub source_channel_id: ChannelId,
     pub source_port_id: PortId,
+    pub destination_channel_id: ChannelId,
+    pub destination_port_id: PortId,
     pub sequences: Vec<Sequence>,
     pub height: Height,
 }
--- a/relayer/src/chain/cosmos.rs
+++ b/relayer/src/chain/cosmos.rs
@@ -826,6 +826,14 @@ fn packet_query(request: &QueryPacketEventDataRequest, seq: &Sequence) -> Result
         request.source_port_id.to_string(),
     )
     .and_eq(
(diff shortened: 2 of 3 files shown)


vitorenesduarte:
> With 2 chains, there can be at most a single packet matching such filter, and this fact was assumed in the code.

I left a TODO to make this assumption explicit. It will be addressed in a following PR.

ancazamfir (approved):
Looks good! Thanks again!

romac:
@vitorenesduarte Awesome, thanks! Could you please create an issue and open a PR to add a changelog entry for this, mentioning said issue?

romac:
^ This can be done as part of your follow-up PR.

vitorenesduarte:
> @vitorenesduarte Awesome, thanks! Could you please create an issue and open a PR to add a changelog entry for this, mentioning said issue?

Oh, right 😅 @romac Here it goes: https://github.com/informalsystems/ibc-rs/pull/617

PR #615, opened 2021-02-05
Fix panic in conn open try when no connection id is provided

vitorenesduarte:
Closes: #XXX

## Description

Currently, a conn open try can only succeed if some `previous_connection_id` is provided in the `MsgConnectionOpenTry`:
https://github.com/informalsystems/ibc-rs/blob/b1b9dac6132a2fd2a86fa1c6f0179f47db3e3454/modules/src/ics03_connection/msgs/conn_open_try.rs#L26-L35
However, to have a valid `previous_connection_id`, one has to have a successful conn open try. This circular dependency makes it so that it's never possible to have successful conn open try.

This issue was found during the MBT work (#601). I tried to add a regression test but didn't find an easy way to do it. The best candidate was the `routing_module_and_keepers` test in `ics26_routing/handler.rs`: https://github.com/informalsystems/ibc-rs/blob/b1b9dac6132a2fd2a86fa1c6f0179f47db3e3454/modules/src/ics26_routing/handler.rs#L152-L316

There's a single chain here, which I think means it's not possible to have a successful conn open try. I've also noticed that the tests here don't check that the returned error matches the expected error. This is done in the `IBCTestExecutor` introduced in #601. Maybe at some point it will be possible to reuse this test executor to write these simple unit tests.

______

For contributor use:

### DIFF
--- a/modules/src/ics03_connection/context.rs
+++ b/modules/src/ics03_connection/context.rs
@@ -72,14 +72,14 @@ pub trait ConnectionKeeper {
                 )?;
             }
             State::TryOpen => {
-                self.store_connection(
-                    &result.connection_id.clone().unwrap(),
-                    &result.connection_end,
-                )?;
+                let connection_id = result
+                    .connection_id
+                    .unwrap_or_else(|| self.next_connection_id());
+                self.store_connection(&connection_id, &result.connection_end)?;
                 // If this is the first time the handler processed this connection, associate the
                 // connection end to its client identifier.
                 self.store_connection_to_client(
-                    &result.connection_id.clone().unwrap(),
+                    &connection_id,
                     &result.connection_end.client_id(),
                 )?;
             }


ancazamfir (commented):
Looks good as a quick fix. But I think we should have a consistent approach with client and channel handling. I thought we decided when we re-wrote the client handler to return an id from `process()`. @adizere ? Looking at channels now, we do the same as in connection.

adizere:
> Looks good as a quick fix. But I think we should have a consistent approach with client and channel handling. I thought we decided when we re-wrote the client handler to return an id from `process()`. @adizere ? Looking at channels now, we do the same as in connection.

Currently only ICS02 performs identifier selection and returns that from `process`.

We should adapt both ICS03 (and ICS04) to return identifiers from `process`. The changes should not be difficult. I agree we need a consistent approach, and the approach in ICS02 should work.

ancazamfir:
> We should adapt both ICS03 (and ICS04) to return identifiers from process. The changes should not be difficult. I agree we need a consistent approach, and the approach in ICS02 should work.

Sounds good. I will leave it up to you @vitorenesduarte if you want to do it in this PR or we do it in a separate one. I opened #622, please mention in `Closes: #XXX` if you decide to fix it here.

vitorenesduarte:
> > We should adapt both ICS03 (and ICS04) to return identifiers from process. The changes should not be difficult. I agree we need a consistent approach, and the approach in ICS02 should work.
> 
> Sounds good. I will leave it up to you @vitorenesduarte if you want to do it in this PR or we do it in a separate one. I opened #622, please mention in `Closes: #XXX` if you decide to fix it here.

Let's do it in another PR.

adizere (approved):
Looks good, thank you.

romac:
@vitorenesduarte Great stuff! Can you please also create an issue, mention it in the body of this PR and update the CHANGELOG accordingly in a follow-up PR?

vitorenesduarte:
> @vitorenesduarte Great stuff! Can you please also create an issue, mention it in the body of this PR and update the CHANGELOG accordingly in a follow-up PR?

Thanks @romac. Done in https://github.com/informalsystems/ibc-rs/pull/627.

PR #617, opened 2021-02-06
Explicit assumptions in packet_from_tx_search_response 

vitorenesduarte:
Closes: #XXX

## Description

This PR makes the following two assumptions from `packet_from_tx_search_response` explicit:
- the number of returned transactions is at most 1
- if a transaction is returned, then there's exactly 1 transaction event matching the packet query

To make the second assumption explicit, it's required that we iterate all transactions events, which may come with a performance cost. To alleviate this, we're now only trying to parse channel events with `ChannelEvents::try_from_tx(&e)` instead of the more general `from_tx_response_event(e)`.

This is a follow-up to #613.

______

For contributor use:

### DIFF
--- a/relayer/src/chain/cosmos.rs
+++ b/relayer/src/chain/cosmos.rs
@@ -6,6 +6,7 @@ use std::{
 use anomaly::fail;
 use bitcoin::hashes::hex::ToHex;
 use crossbeam_channel as channel;
+use ibc::ics04_channel::events as ChannelEvents;
 use prost::Message;
 use prost_types::Any;
 use tendermint::abci::Path as TendermintABCIPath;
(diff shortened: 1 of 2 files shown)


relayer/src/chain/cosmos.rs
```diff
@@ -839,28 +840,37 @@ fn packet_query(request: &QueryPacketEventDataRequest, seq: &Sequence) -> Result
...
-            if e.type_str != request.event_id.as_str() {
-                continue;
-            }
+        let mut matching = Vec::new();
+        for e in r.tx_result.events {
+            assert_eq!(
+                e.type_str,
+                request.event_id.as_str(),
+                "packet_from_tx_search_response: unexpected event type"
+            );
```
ancazamfir: This doesn't seem right. It is true that there is one single transaction but it may include multiple events. I think what was there before was correct:
```suggestion
            if e.type_str != request.event_id.as_str() {
                continue;
            }
```

vitorenesduarte: Thanks! This was the reason why the tests were failing. Fixed in b0a9aa0a36126cbeec414f5ac659b997cf594594.

ancazamfir (approved):
Looks good, thanks Vitor!

Issue #560, opened 2021-01-27
Listen to and act on ChanCloseInit in the relayer loop


ancazamfir:
## Crate
relayer

## Summary
The relayer loop should listen to `ChanCloseInit` event, relay the `TimeoutOnClose`, `ChanCloseConfirm` and exit the loop.

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #623, opened 2021-02-08
Process ChanCloseInit in relayer loop (start command)

ancazamfir:
Closes: #560

## Description

- collect `CloseInitChannel` event (`collect_events()`) and handle it in `build_msg_from_event()`
- handle the race condition where client updates are sent on destination for heights higher than event height:
   - added a new function `adjust_events_height()` that checks:
       - if latest consensus height <= event_height (no change) or
       - if a client consensus of height `event_height + 1` exists (no adjustment) or
       - one at height `> event_height + 1` (change events height to latest client consensus height - 1)
   - the race condition can be reproduce this way:
       - start the realyer loop on an open channel (`hermes start ibc-0 ibc-1 -p transfer -c channel-0`)
       - from one terminal send tens of packets to `ibc-0` (`hermes tx raw ft-transfer ibc-0 ibc-1 transfer channel-0 9999 1000 -n 30`) 
       - from a different terminal send a `chan-close-init` to `ibc-1` (`hermes tx raw chan-close-init ibc-1 ibc-0 connection-0 transfer transfer -d channel-1 -s channel-0`)
       - before the `CloseInitChannel` event with some height `X` from `ibc-1` is processed, the `SendPacket` events processing results in `MsgTimeoutOnClose` packets to be sent to `ibc-0` prepended by a `MsgClientUpdate` at some height > `X`. This causes the subsequent `chan-close-init` event processing to fail.
    - there are probably many other ways to cause the race

Minor changes:
- added `timeout_on_close_packet` event decoder, with some awkwardness because `message.action` is `timeout_on_close_packet ` while the event name (prefix of tags) is `timeout_packet`. Needs more digging in cosmos-sdk, maybe ask them to make the changes.
- changed the channel event names (removed `Channel` suffix) 
 

______

For contributor use:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,53 +1,53 @@
 # Changelog
 
-[comment]: <> (## Unreleased Changes)
+## Unreleased Changes
 
 ### FEATURES
 
 - <high level summary>
 
-- [modules]
+- [ibc]
   -
 
-- [relayer]
+- [ibc-relayer]
   - Listen to channel close initialization event and perform the close handshake ([#560])
 
-- [relayer-cli]
+- [ibc-relayer-cli]
   -
 
 ### IMPROVEMENTS
 
 - <high level summary>
 
-- [modules]
+- [ibc]
   -
 
-- [relayer]
+- [ibc-relayer]
   -
 
-- [relayer-cli]
+- [ibc-relayer-cli]
   -
 
 ### BUG FIXES:
 
-- [modules]
+- [ibc]
   -
 
-- [relayer]
+- [ibc-relayer]
   -
 
-- [relayer-cli]
+- [ibc-relayer-cli]
   -
 
 ### BREAKING CHANGES:
 
-- [modules]
+- [ibc]
   -
 
-- [relayer]
+- [ibc-relayer]
   -
 
-- [relayer-cli]
+- [ibc-relayer-cli]
 
 [#593]: https://github.com/informalsystems/ibc-rs/issues/560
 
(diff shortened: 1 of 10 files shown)


romac (commented):
Looks good to me, as far as I can tell it does indeed implement the fix described above. Just pushed a commit with some minor stylistic changes.

I have only one remark aside from the one inlined below: as the event processing logic is getting more and more complex, it might eventually warrant a refactoring, perhaps even be rewritten as a state machine of some kind? Just some food for thought, this should be done in a separate PR anyway.

relayer/src/link.rs
```diff
@@ -165,22 +167,64 @@ impl RelayPath {
         event: &IBCEvent,
     ) -> Result<(Option<Any>, Option<Any>), LinkError> {
```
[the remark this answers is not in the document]
ancazamfir: I believe we will need quite a bit of restructuring when we fix https://github.com/informalsystems/ibc-rs/issues/640. Would also require some sort of packet sending thread to implement a delay period timer and send packets when the timer fires. Let's do these changes there, including the refactoring.

CHANGELOG.md
```diff
@@ -2,6 +2,55 @@
 
 [comment]: <> (## Unreleased Changes)
```
romac, line 3: This should be uncommented now that there is content for unreleased changes.

CHANGELOG.md
```diff
@@ -2,6 +2,55 @@
...
 
+### FEATURES
+
+- <high level summary>
+
+- [modules]
```
romac, line 9: We should use the actual crate names instead of the folder names, ie. ibc, ibc-relayer, ibc-relayer-cli

romac (changes requested):
Can we also fix or silence the clippy warnings, depending on whether we expect those functions to potentially returns errors at some point?

romac (approved):

Issue #549, opened 2021-01-25
Use ICS Height in IBC Events


ancazamfir:
## Crate
modules, relayer, relayer-cli

## Summary
Currently the IBC Events have the tendermint height. Use ICS Height instead.
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #638, opened 2021-02-11
Change event height to ICS height

ancazamfir:
Closes: #549

## Description

______

For contributor use:


adizere (approved):
Super, much cleaner now. Thanks Anca!

One idea that popped was whether we can get rid of the boilerplate around `IBCEvent::height()` and `set_height`. But I can't see a simple solution right away (and this is not the right time for that anyway).

Issue #672, opened 2021-02-16
Hermes failing to compile


Fraccaman:
## Crate
- ibc-relayer-cli

## Summary of Bug

Following the guide [here](https://hermes.informal.systems/installation.html#building-with-cargo-build), building `hermes` either from source or using cargo, leads to compile errors.
Here the error logs: [pastebin](https://pastebin.com/jvDapyBr)

## Version
- cargo: `cargo 1.50.0 (f04e7fab7 2021-02-04)`
- rustc: `rustc 1.50.0 (cb75ad5db 2021-02-10)`
- hermes: `v0.1.0`

## Steps to Reproduce

- `cargo install ibc-relayer-cli`

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
Hi, thank you for opening this issue.

Intuitively, the problem is that an indirect dependency (`funty`) of crate `ibc-relayer-cli` had a minor-version update recently, which [is causing build errors](https://github.com/informalsystems/ibc-rs/issues/665) in release `v0.1.0`. We have solved this problem, but did not release a new version since then.

The simplest way to mitigate this would be to build Hermes from source -- as documented in [here](https://hermes.informal.systems/installation.html#build-from-source), __but ignore the following step:__

```
git checkout v0.1.0
```

Instead, no need to check-out any specific release (just use `master`, the default version that is checked-out when you clone the repo).

Fraccaman:
Okay, thanks! On `master` is compiling without any error.

adizere:
Great!

We will update the documentation shortly to mention this problem & workaround explicitly.

PR #674, opened 2021-02-16
Document the v0.1.0 installation issue due to dependency problem

adizere:
Closes: #672

## Description

`cargo install ibc-relayer-cli` currently fails.

This PR modifies the guide to document this problem and provide a workaround.

______

For contributor use:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 
 - [ibc-relayer-cli]
   - Fix wrong acks sent with `tx raw packet-ack` in a 3-chain setup ([#614])
+  - Guide: documented the v0.1.0 cargo install problem and added workaround ([#672])
 
 ### BREAKING CHANGES
 
@@ -59,6 +60,7 @@
 [#626]: https://github.com/informalsystems/ibc-rs/issues/626
 [#643]: https://github.com/informalsystems/ibc-rs/issues/643
 [#665]: https://github.com/informalsystems/ibc-rs/issues/665
+[#672]: https://github.com/informalsystems/ibc-rs/issues/672
 
 ## v0.1.0
 *February 4, 2021*
--- a/guide/src/installation.md
+++ b/guide/src/installation.md
@@ -35,6 +35,20 @@ hermes version
 hermes 0.1.0
 ```
 
+> #### Update 16.02.2021: Regarding installation errors
+>
+> It is possible that the command `cargo install ibc-relayer-cli` will fail.
+> This is a [known issue][issue/672], caused by a recent update of a dependency.
+>
+> The simplest way to work around this problem is to build Hermes from source,
+> using the most up-to-date version. Simply follow the recipe in the section
+> below __but skip the step__ `git checkout v0.1.0`. In that way, you'll be
+> using the latest commit in the `master` branch. The rest of the steps remain
+> unchanged.
+
+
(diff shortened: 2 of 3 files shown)


romac (approved):

romac:
This issue with dependencies breaking compilation makes me think we should
a) pin all dependencies to an exact version
b) provide binaries for each release (mac + linux)

adizere:
> This issue with dependencies breaking compilation makes me think we should
> a) pin all dependencies to an exact version

I like this. Alternatively, I'm curious if committing the Cargo.lock file (for `ibc-relayer-cli` only), as proposed [here](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) could also solve this issue.

ancazamfir:
I am also not sure this is the right way to go for the issue. People who get the `v0.1.0` from `crates.io` will not see the doc update. I suggested earlier to release `v0.1.1` with a cargo patch for `funty`.

romac:
> > This issue with dependencies breaking compilation makes me think we should
> > a) pin all dependencies to an exact version
> 
> I like this. Alternatively, I'm curious if committing the Cargo.lock file (for `ibc-relayer-cli` only), as proposed [here](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) could also solve this issue.

Right, that would indeed solve the issue, provided one invokes `cargo install` with the [`--locked`](https://doc.rust-lang.org/cargo/commands/cargo-install.html#manifest-options) flag.

In any case, we should definitely commit it, for just `ibc-relayer-cli` if that's possible or the whole workspace if not (the file is ignored for libraries anyway).

adizere:
> I am also not sure this is the right way to go for the issue. People who get the `v0.1.0` from `crates.io` will not see the doc update. I suggested earlier to release `v0.1.1` with a cargo patch for `funty`.

Agree. Working on releasing 0.1.1 now.

Using `[patch.crates-io]` section to override the dep on funty does not seem to work. (The patch section is meant to replace the crates.io default URL with a different URL, a local copy of that repo, or a specific branch from a git repo; but we're still using the crates.io version of `funty` so I think the `patch` is not the way to go.)

romac:
> > I am also not sure this is the right way to go for the issue. People who get the `v0.1.0` from `crates.io` will not see the doc update. I suggested earlier to release `v0.1.1` with a cargo patch for `funty`.
> 
> Agree. Working on releasing 0.1.1 now.
> 
> Using `[patch.crates-io]` section to override the dep on funty does not seem to work. (The patch section is meant to replace the crates.io default URL with a different URL, a local copy of that repo, or a specific branch from a git repo; but we're still using the crates.io version of `funty` so I think the `patch` is not the way to go.)

Yeah, you instead need to pin it as a regular dependency and hope that Cargo will figure out a build plan that works for the pinned version. See the fix in `tendermint-rs`:

https://github.com/informalsystems/tendermint-rs/pull/817/files#diff-502a532f4ade695845390cb6c6933ccdc9a0895a2a96612320ad26dc62caa7d2R42-R43

romac:
Opened https://github.com/informalsystems/ibc-rs/issues/677 to track the binary release.

adizere:
> I suggested earlier to release `v0.1.1` ...

Tracking the release of `v0.1.1` in #680.

Issue #685, opened 2021-02-19
Overflow if `ctx.host_current_height().revision_height < ctx.host_chain_history_size()`


vitorenesduarte:
## Crate

modules

## Summary of Bug

If `ctx.host_current_height().revision_height < ctx.host_chain_history_size()`, the following will overflow:

https://github.com/informalsystems/ibc-rs/blob/ae36bba8db5404243e28566f2a7ce2dbef1640ae/modules/src/ics03_connection/handler/verify.rs#L200-L201

This issue was found during the MBT work (#546), after initializing the `MockContext` with `max_history_size = usize::MAX`.

## Version

ae36bba8db5404243e28566f2a7ce2dbef1640ae
____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #686, opened 2021-02-19
Fix overflow when `ctx.host_current_height().revision_height < ctx.host_chain_history_size()`

vitorenesduarte:
Closes: #685 

## Description

______

For contributor use:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,11 +32,10 @@
 
 [comment]: <> (  - [nothing yet])
 
-[comment]: <> (### BUG FIXES)
+### BUG FIXES
 
-[comment]: <> (- [ibc])
-
-[comment]: <> (  - [nothing yet])
+- [ibc]
+  - Overflow if `ctx.host_current_height().revision_height < ctx.host_chain_history_size()` ([#685])
 
 [comment]: <> (- [ibc-relayer])
 
@@ -59,6 +58,8 @@
 [comment]: <> (- [ibc-relayer-cli])
 
 [comment]: <> (  - [nothing yet])
+
+[#685]: https://github.com/informalsystems/ibc-rs/issues/685
 
 ## v0.1.1
 *February 17, 2021*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,7 @@ involving packet acks in a 3-chain setup. More details below.
 ### FEATURES
 - [ibc-relayer]
   - Listen to channel close initialization event and perform the close handshake ([#560])
+  - Updated to tendermint-rs `v0.18.1` ([#682], [#671])
 
 ### IMPROVEMENTS
 
@@ -115,7 +116,9 @@ involving packet acks in a 3-chain setup. More details below.
 [#626]: https://github.com/informalsystems/ibc-rs/issues/626
 [#643]: https://github.com/informalsystems/ibc-rs/issues/643
 [#665]: https://github.com/informalsystems/ibc-rs/issues/665
+[#671]: https://github.com/informalsystems/ibc-rs/pull/671
 [#672]: https://github.com/informalsystems/ibc-rs/issues/672
+[#682]: https://github.com/informalsystems/ibc-rs/issues/682
 
 ## v0.1.0
 *February 4, 2021*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,15 @@
   - Consistent identifier handling across ICS 02, 03 and 04 ([#622])
 
 - [ibc-relayer]
+  - [nothing yet]
+
+- [ibc-relayer-cli]
+  - [nothing yet]
 
 ### IMPROVEMENTS
 
 - [ibc]
-  - [nothing yet]
+  - Follow Rust guidelines naming conventions ([#689])
 
 - [ibc-relayer]
   - [nothing yet]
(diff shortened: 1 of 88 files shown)


modules/src/ics03_connection/handler/verify.rs
```diff
@@ -197,10 +197,7 @@ pub fn check_client_consensus_height(
         return Err(Kind::InvalidConsensusHeight(claimed_height, ctx.host_current_height()).into());
     }
 
-    let oldest_available_height =
-        ctx.host_current_height().revision_height - ctx.host_chain_history_size() as u64;
```
adizere, line 201: Can we remove method `host_chain_history_size`, which is no longer useful now that you introduced `host_oldest_height`?

CHANGELOG.md
```diff
@@ -32,11 +32,10 @@
...
 [comment]: <> (  - [nothing yet])
 
-[comment]: <> (### BUG FIXES)
-
-[comment]: <> (- [ibc])
+### BUG FIXES
 
-[comment]: <> (  - [nothing yet])
+- [ibc]
+  - Overflow if `ctx.host_current_height().revision_height < ctx.host_chain_history_size()` ([#685])
```
adizere, line 38: Nit: Would be nice to have a slightly higher level description in the changelog.

```suggestion
  - Fix overflow bug in ICS03 client consensus height verification method ([#685])
```

adizere (changes requested):
Nice find, Vitor! I had two minor remarks.

modules/src/ics03_connection/handler/verify.rs
vitorenesduarte: Done in c851c336c039286ef803cdae5e693d058cf0c342.

vitorenesduarte:
Thanks for the review @adizere. I've addressed your suggestions.

I'm thinking that the tests for connection open ack were relying on wrong behaviour since they started failing with this PR. I tried to fix them in 4da1d34c9d0505a360060c69957394b9bb55690a but there's one (now commented out) I didn't figure out how.

vitorenesduarte:
> I'm thinking that the tests for connection open ack were relying on wrong behaviour since they started failing with this PR. I tried to fix them in [4da1d34](https://github.com/informalsystems/ibc-rs/commit/4da1d34c9d0505a360060c69957394b9bb55690a) but there's one (now commented out) I didn't figure out how.

The test that was commented out is expecting a `MissingLocalConsensusState` to be thrown here: https://github.com/informalsystems/ibc-rs/blob/vitor/overflow/modules/src/ics03_connection/handler/verify.rs#L167-L170

However, before this, we're already checking if the height is stale here: https://github.com/informalsystems/ibc-rs/blob/vitor/overflow/modules/src/ics03_connection/handler/verify.rs#L200-L203

If the height is not stale, then the consensus state must exist, in which case the error `MissingLocalConsensusState` should instead be a panic. If you agree with this analysis, I'll replace the error with a panic in a follow-up PR.

adizere (approved):
> I'm thinking that the tests for connection open ack were relying on wrong behaviour since they started failing with this PR. I tried to fix them in 4da1d34 but there's one (now commented out) I didn't figure out how.

That may well be. Tried to consolidate tests in that module recently, but there's still some work to be done. Your improvements definitely help!

adizere:
> > I'm thinking that the tests for connection open ack were relying on wrong behaviour since they started failing with this PR. I tried to fix them in [4da1d34](https://github.com/informalsystems/ibc-rs/commit/4da1d34c9d0505a360060c69957394b9bb55690a) but there's one (now commented out) I didn't figure out how.
> 
> The test that was commented out is expecting a `MissingLocalConsensusState` to be thrown here: https://github.com/informalsystems/ibc-rs/blob/vitor/overflow/modules/src/ics03_connection/handler/verify.rs#L167-L170
> 
> However, before this, we're already checking if the height is stale here: https://github.com/informalsystems/ibc-rs/blob/vitor/overflow/modules/src/ics03_connection/handler/verify.rs#L200-L203
> 
> If the height is not stale, then the consensus state must exist, in which case the error `MissingLocalConsensusState` should instead be a panic. If you agree with this analysis, I'll replace the error with a panic in a follow-up PR.

I agree with your analysis. To summarize: in conn open try & ack handlers there are two verifications on the field `msg.proofs.consensus_proof().height` as follows:

1. first we verify for staleness with [`check_client_consensus_height`](https://github.com/informalsystems/ibc-rs/blob/vitor/overflow/modules/src/ics03_connection/handler/verify.rs#L200-L203), i.e., verify that this height is not older than the oldest height of the host (local) chain, nor newer than the current height of the host chain.
2. second we verify that the consensus state for this height exists in the local chain, as part of [`verify_consensus_proof`](https://github.com/informalsystems/ibc-rs/blob/vitor/overflow/modules/src/ics03_connection/handler/verify.rs#L167-L170).

I think the first verification is more of a fail-fast. While the second verification is a pre-check as part of the whole ensemble of proof verification. I agree there is some redundancy there. I would suggest to open an issue documenting this, but I wouldn't fix things with a panic (we have no panics in the modules yet, and not sure if it's a good idea to fix it like that).

vitorenesduarte:
I've updated #698. @adizere please confirm that it reflects the discussion here. 

Regarding `panic`s, there are already some `unwrap`s, which can also panic. My view is that both are okay as long as they are prefixed with some explanation why there are "safe".

Issue #599, opened 2021-02-02
Add missing support for ordered channels


ancazamfir:
## Crate
relayer, relayer-cli

## Summary
Complete and test `ordered` channels in relayer. While some support is present in the relayer for `ordered` channels, they are not fully supported. This is mainly because only `unordered` channels are currently supported in cosmos-sdk.

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #687, opened 2021-02-19
add ordered channel support

ancazamfir:
Closes: #599

## Description
The main changes are:
- add back option to specify the channel order in the `tx raw` commands
  - the changes to the `hermes channel create` and `hermes start` will be done separately as those CLI are currently heavily re-worked
- add the next receive sequence query
- send packets in order
- send a single timeout message per ordered channel (all the others are rejected as the channel is closed on-chain on the first one)
- ~~changes to the scripts to be able to run with latest cosmos (there are now two grpc ports, a new grpc-web port is required~~ (no need for these until we upgrade gaia/cosmos)

I tested this by using a custom SDK with an ICS-20 impl that uses ordered channels.

______

For contributor use:


scripts/setup-chains
```diff
@@ -70,8 +70,8 @@ CHAIN_1_RPC_ADDR="localhost:$CHAIN_1_RPC_PORT"
 CHAIN_0_SAMOLEANS=100000000000
 CHAIN_1_SAMOLEANS=100000000000
 
-"$ONE_CHAIN" gaiad "$CHAIN_0_ID" ./data $CHAIN_0_RPC_PORT 26656 6060 9090 $CHAIN_0_SAMOLEANS
-"$ONE_CHAIN" gaiad "$CHAIN_1_ID" ./data $CHAIN_1_RPC_PORT 26556 6061 9091 $CHAIN_1_SAMOLEANS
+"$ONE_CHAIN" gaiad "$CHAIN_0_ID" ./data $CHAIN_0_RPC_PORT 26656 6060 9090 9091 $CHAIN_0_SAMOLEANS
```
adizere, line 73: If the `gaiad` invocation changed, did the branch/release also change? If so, maybe we should update docs/code to reflect the new release.

adizere (changes requested):
Looks good, just had a minor comment/question.
Should we also update the changelog?

scripts/setup-chains
ancazamfir: Good point, l just undid those changes. I was working with a more recent gaia off a recent cosmos-sdk. We don't need these changes here. I think we need to update the docker images to use `gaia v4.0.4` and make the changes there.

ancazamfir: Opened https://github.com/informalsystems/ibc-rs/issues/702 for the gaia upgrade in CI

adizere (approved):

Issue #632, opened 2021-02-10
Implement misbehavior monitoring and evidence relaying


ancazamfir:
## Crate
relayer, relayer-cli, modules

## Summary
Add a new capability for the relayer to monitor IBC client updates, conflicting headers and submit proof of misbehavior in `MsgSubmitMisbehaviour` transaction.

## Proposal
(details will be added soon)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #691, opened 2021-02-19
Misbehavior monitoring and handling

ancazamfir:
Closes: #632

## Description
- Added header and deserialization in `UpdateClient` event and Tx result output for CLIs
- Extract header from `UpdateClient` transactions
- Get headers from `query_tx()`
- Misbehaviour event deserialization
- Added the `hermes misbehaviour` command for monitoring a given client on a chain
- Fix `tx raw create-client` to not allow same source and dest chains
- Removed src chain from  `tx raw update-client` as it is redundant
- Added target and trusted heights to `tx raw update-client` to allow testing with different updates.
- Added option to query all consensus states and also display only the heights.
- Added new command to query header used when a consensus state was created.
- Submit evidence supported for headers at same height but different BlockId-s and also for BFT time violation.
- Checks previous updates on hermes (re)start.
- Added `query_clients()` to the runtime.

______

For contributor use:


relayer-cli/src/commands/misbehaviour.rs
```diff
@@ -0,0 +1,106 @@
...
+                        client_id,
+                    );
+
+                    // 3 - decode light block from the update event
+                    // TODO - code below may change depending on the header representation in the new update event
+                    // Currently this doesn't work as `update.header` has a weird format
+                    if update.client_type() == ClientType::Tendermint {
+                        let chain_header = Any {
+                            type_url: TENDERMINT_HEADER_TYPE_URL.to_string(),
+                            value: Vec::from(update.header.as_bytes()),
```
colin-axner, line 84: this seems a little odd to me. The header bytes should contain the type_url? This is the definition in the [SDK](https://github.com/cosmos/cosmos-sdk/blob/a193522f7e54cd747d549bc4adb83618aac223af/codec/types/any.go#L77). I don't know rust, but it looks like you are creating an `Any` by passing in the type url using a constant and then trying to decode the encoded any into the value of the `Any`?

```
Any{
    type: tendermint,
    value: Any{
        type: tendermint,
        value: ibc.header{}
        }
  }
  ```

Comment (colin-axner) [reply relayer-cli/src/commands/misbehaviour.rs:84]: Here is how the header is being encoded in the [SDK](https://github.com/cosmos/cosmos-sdk/blob/a193522f7e54cd747d549bc4adb83618aac223af/codec/proto_codec.go#L171)

**Comment by ancazamfir** [comment]
> this seems a little odd to me. The header bytes should contain the type_url? This is the definition in the SDK. I don't know rust, but it looks like you are creating an Any by passing in the type url using a constant and then trying to decode the encoded any into the value of the Any?

This was with a version that you had where you were not using `Any`.

File: relayer/src/lib.rs [hunk 2]
```diff
@@ -8,7 +8,7 @@
     unused_qualifications,
     rust_2018_idioms
 )]
-// #![allow(dead_code, unreachable_code, unused_imports, unused_variables)]
+#![allow(clippy::large_enum_variant)]
```
adizere, line 11: I wonder if we shouldn't move this annotation to the site where it's necessary, namely here:

https://github.com/informalsystems/ibc-rs/blob/e3ccb5055306f15016403e5bd08a8e91393680b2/relayer/src/chain/handle.rs#L53

modules/src/ics07_tendermint/header.rs
```diff
@@ -11,26 +13,57 @@ use crate::ics02_client::header::AnyHeader;
...
+
+        match self_header_height.cmp(&&ibc_client_height) {
+            Ordering::Equal => {
+                // fork
+                self.signed_header.commit.block_id
+                    != ibc_client_header.signed_header.commit.block_id
+            }
+            Ordering::Greater => {
+                // BFT time violation
+                self.signed_header.header.time < ibc_client_header.signed_header.header.time
```
cezarad, line 50: Same here maybe ? The headers are compatible if the order on the heigh is the same as the order on the time. ```self.signed_header.header.time > ibc_client_header.signed_header.header.time```

modules/src/ics07_tendermint/header.rs
```diff
@@ -11,26 +13,57 @@ use crate::ics02_client::header::AnyHeader;
...
+            }
+            Ordering::Less => {
+                // BFT time violation
+                self.signed_header.header.time > ibc_client_header.signed_header.header.time
```
cezarad, line 54: Same here ?

relayer/src/foreign_client.rs
```diff
@@ -406,6 +476,231 @@ impl ForeignClient {
...
+    ) -> Result<Option<UpdateClient>, ForeignClientError> {
+        let request = QueryClientEventRequest {
+            height: Height::zero(),
+            event_id: IbcEventType::UpdateClient,
+            client_id: self.id.clone(),
+            consensus_height,
+        };
+
+        // Tx query fails if we don't wait a bit here (block is not ready/ indexed?)
+        thread::sleep(Duration::from_millis(100));
```
adizere, line 492: Perhaps add a retry loop here to make the query to `QueryTxRequest` more robust? The `100ms` may work in local tests but not sure how it will fare in actual networks.

adizere (commented):
This is impressive, great work Anca!

I left several suggestions, but should be treated as lower priority. I didn't get the chance to look closely at the important part in `ForeignClient::handle_misbehaviour`.

modules/src/ics07_tendermint/header.rs
```diff
@@ -11,26 +13,57 @@ use crate::ics02_client::header::AnyHeader;
...
+
+    pub fn compatible_with(&self, ibc_client_header: &Header) -> bool {
+        let ibc_client_height = ibc_client_header.signed_header.header.height;
+        let self_header_height = self.signed_header.header.height;
+
+        match self_header_height.cmp(&&ibc_client_height) {
+            Ordering::Equal => {
+                // fork
+                self.signed_header.commit.block_id
+                    != ibc_client_header.signed_header.commit.block_id
```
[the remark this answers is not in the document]
ancazamfir: Yeah, as I mentioned earlier I flipped the logic in this function to do some basic testing (i.e. submit headers with same hashes, and good timestamps) with gaia and SDK as it's simpler than constructing headers with different hashes that verify.  Just switched back now as I am done with testing.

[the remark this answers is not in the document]
ancazamfir: yup, left it like that for testing.

modules/src/ics07_tendermint/header.rs
ancazamfir: yes see ^

modules/src/ics02_client/client_consensus.rs
```diff
@@ -125,3 +165,11 @@ impl ConsensusState for AnyConsensusState {
         self
     }
 }
+
+#[derive(Clone, Debug)]
+pub struct QueryClientEventRequest {
```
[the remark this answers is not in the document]
ancazamfir: This is one type of query Tx. We also have a packet one that we keep in `ics04_channel`. And the global in `modules/src/query.rs`. The reason was that we would have a backend impl at some point for these. I think I will leave them where they are for now and we can discuss later.

relayer/src/foreign_client.rs
```diff
@@ -406,6 +476,232 @@ impl ForeignClient {
...
+                    misbehaviour,
+                    signer,
+                };
+
+                let events = self
+                    .dst_chain()
+                    .send_msgs(vec![msg.to_any()])
+                    .map_err(|e| {
+                        ForeignClientError::Misbehaviour(format!(
+                            "failed sending evidence to dst chain ({}) with err: {}",
```
romac, line 695:
```suggestion
                            "failed sending evidence to destination chain ({}), error: {}",
```

relayer/src/foreign_client.rs
```diff
@@ -406,6 +476,232 @@ impl ForeignClient {
...
+        update: Option<UpdateClient>,
+    ) -> Result<Option<IbcEvent>, ForeignClientError> {
+        match self.handle_misbehaviour(update)? {
+            None => Ok(None),
+            Some(misbehaviour) => {
+                error!("MISBEHAVIOUR detected {}, sending evidence", misbehaviour);
+
+                let signer = self.dst_chain().get_signer().map_err(|e| {
+                    ForeignClientError::Misbehaviour(format!(
+                        "failed getting signer for dst chain ({}) with error: {}",
```
romac, line 678:
```suggestion
                        "failed getting signer for destination chain ({}), error: {}",
```

relayer/src/foreign_client.rs
```diff
@@ -406,6 +476,232 @@ impl ForeignClient {
...
+    }
+
+    pub fn detect_misbehaviour_and_send_evidence(
+        &self,
+        update: Option<UpdateClient>,
+    ) -> Result<Option<IbcEvent>, ForeignClientError> {
+        match self.handle_misbehaviour(update)? {
+            None => Ok(None),
+            Some(misbehaviour) => {
+                error!("MISBEHAVIOUR detected {}, sending evidence", misbehaviour);
```
romac, line 674:
```suggestion
                error!("MISBEHAVIOUR DETECTED: {}, sending evidence...", misbehaviour);
```

relayer/src/foreign_client.rs
```diff
@@ -406,6 +476,232 @@ impl ForeignClient {
...
+                    .send_msgs(vec![msg.to_any()])
+                    .map_err(|e| {
+                        ForeignClientError::Misbehaviour(format!(
+                            "failed sending evidence to dst chain ({}) with err: {}",
+                            self.dst_chain.id(),
+                            e
+                        ))
+                    })?;
+
+                Ok(Some(events[0].clone()))
```
romac, line 701: Are we guaranteed to have at least one event here? If not, we should probably add a check otherwise this code will panic. (I hate how Rust usually makes potential panics very explicit (`unwrap` etc.) _except_ for indexing into vectors…)

relayer/src/light_client/tendermint.rs
```diff
@@ -56,6 +63,79 @@ impl super::LightClient<CosmosSdkChain> for LightClient {
 
         self.fetch_light_block(AtHeight::At(height))
     }
+
+    /// Given a client update event that includes the header used in a client update.
+    /// it looks for misbehaviour by fetching a header at same or latest height.
+    /// TODO - return also intermediate headers.
+    fn build_misbehaviour(
```
romac, line 70: Can we find a better name for this method and associated ones? To me the `build_` prefix implies that there is always a misbehavior to build whereas the actual behavior of the function is to figure out if there is a misbehavior by checking if the headers are compatible.

Perhaps

```suggestion
    fn check_misbehaviour(
```
?

relayer/src/light_client/tendermint.rs
```diff
@@ -56,6 +63,79 @@ impl super::LightClient<CosmosSdkChain> for LightClient {
...
+        &mut self,
+        client_state: &AnyClientState,
+        update: UpdateClient,
```
romac, line 73: For consistency with `verify`:

```suggestion
        update: UpdateClient,
        client_state: &AnyClientState,
```

relayer-cli/src/commands/misbehaviour.rs
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+}
+
+pub fn monitor_misbehaviour(
+    chain_id: &ChainId,
+    client_id: &ClientId,
+    config: &config::Reader<CliApp>,
+) -> Result<(), BoxError> {
+    let chain = spawn_chain_runtime(&config, chain_id)
```
romac, line 49: Note for follow-up: At some point later down the line we could use the `Registry` to avoid spawning a runtime multiple time (eg. if there are more than one client per chain).

relayer-cli/src/commands/misbehaviour.rs
```diff
@@ -0,0 +1,129 @@
...
+        .map_err(|e| {
+            format!(
+                "could not run misbehaviour detection for {}: {}",
+                client_id, e
+            )
+        })?;
+
+    if let Some(evidence_submission_result) = misbehaviour_detection_result {
+        info!(
+            "\nEvidence submission result {:?}",
```
romac, line 123:
```suggestion
            "Evidence submission result {:?}",
```

relayer-cli/src/commands/query/client.rs
```diff
@@ -102,12 +103,115 @@ impl Runnable for QueryClientConsensusCmd {
...
+                    client_id: self.client_id.to_string(),
+                    pagination: None,
+                });
+
+                match res {
+                    Ok(states) => {
+                        if self.heights_only {
+                            let heights: Vec<Height> = states
+                                .iter()
+                                .filter_map(|cs| Option::from(cs.height))
```
romac, line 139: Seems to be a no-op:

```suggestion
```

relayer-cli/src/commands/tx/client.rs
```diff
@@ -53,34 +53,64 @@ pub struct TxUpdateClientCmd {
...
-            dst_chain: chains.dst,
-            src_chain: chains.src,
+            dst_chain,
+            src_chain,
             id: self.dst_client_id.clone(),
         };
 
         let res: Result<IbcEvent, Error> = client
-            .build_update_client_and_send()
+            .build_update_client_and_send(height, trusted_height)
```
romac, line 113: See comment on `build_update_client_and_send`: why not pass the `Option`s here directly instead of mapping `None` to `Height::zero()`?

romac (changes requested):
Great work @ancazamfir! Pushed a couple commits with some minor changes (none semantics) and otherwise left a bunch of comments and suggestions inline.

relayer/src/foreign_client.rs
```diff
@@ -406,6 +476,232 @@ impl ForeignClient {
...
+    /// Ensure that consensus state times are monotonically increasing with height.
+    ///
+    /// Other notes:
+    /// - the algorithm builds misbehavior at each consensus height, starting with the
+    /// highest height assuming the previous one is trusted. It submits the first constructed
+    /// evidence (the one with the highest height)
+    /// - a lot of the logic here is derived from the behavior of the only implemented client
+    /// (ics07-tendermint) and might not be general enough.
+    ///
+    pub fn handle_misbehaviour(
```
romac, line 586: Looks good to me, as far as I can tell point 1 and 2 are indeed implemented here and within the light client `build_misbehaviour` method.

We could perhaps make it a bit more explicit where each of the attacks described above are checked, ie. add a reference to the attack number within the inline code comments.

relayer/src/foreign_client.rs
```diff
@@ -406,6 +476,232 @@ impl ForeignClient {
...
+
+    /// Retrieves all consensus states for this client and sorts them in reverse height
+    /// order. If consensus states are not pruned on chain, then last consensus state is the one
+    /// installed by the `CreateClient` operation.
+    fn consensus_states(&self) -> Result<Vec<AnyConsensusStateWithHeight>, ForeignClientError> {
+        let mut consensus_states = self
+            .dst_chain
+            .query_consensus_states(QueryConsensusStatesRequest {
+                client_id: self.id.to_string(),
+                pagination: None,
```
romac, line 526: Do we know what the semantics of setting `pagination` to `None` are? The Protobuf definition for `PageRequest` says (emphasis mine):

>     /// limit is the total number of results to be returned in the result page.
>     /// **If left empty it will default to a value to be set by each app.**

which leads me to think that we might not get all values at once. Perhaps we should always set it to something like `u64::MAX` to be sure we get them all.

https://github.com/informalsystems/ibc-rs/blob/a11c3c0847677280408a518dc353c89612c3054d/proto/src/prost/cosmos.base.query.v1beta1.rs#L20-L23

ancazamfir: Yes this is an issue, please see #811

relayer/src/foreign_client.rs
```diff
@@ -323,7 +355,33 @@ impl ForeignClient {
                 ))
             })?;
 
-        let trusted_height = client_state.latest_height();
+        // If not specified, set trusted state to the highest height smaller than target height.
+        // Otherwise ensure that a consensus state at trusted height exists on-chain.
+        let cs_heights = self.consensus_state_heights()?;
+        let trusted_height = if trusted_height == Height::zero() {
```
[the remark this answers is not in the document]
ancazamfir: Right, so I started with `trusted_height` as an option, then noticed that i would have to change the `target_height` also (to be consistent), then do this throughout the code. So I gave up.
The history here is coming from tendermint where `0` means the latest (when you do queries for example). We adopted that in ibc. Then `ibc::Height` was created and we extended that to `Height::zero()` meaning latest.
Now here i change that to "latest height that makes sense". 
I think we should have another PR to clean this up throughout the code base.

relayer/src/foreign_client.rs
```diff
@@ -406,6 +476,231 @@ impl ForeignClient {
...
+                    e
+                ))
+            })?;
+
+        if res.is_empty() {
+            return Ok(None);
+        };
+
+        assert_eq!(res.len(), 1);
+        let event = res[0].clone();
```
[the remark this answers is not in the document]
ancazamfir: nope

relayer-cli/src/commands/misbehaviour.rs
ancazamfir: Agreed. I was just about to add it but we decided to only monitor one client with the CLI for now.

relayer/src/foreign_client.rs
```diff
@@ -406,6 +476,231 @@ impl ForeignClient {
...
+                    "failed to query Tx-es for update client event {}",
+                    e
+                ))
+            })?;
+
+        if res.is_empty() {
+            return Ok(None);
+        };
+
+        assert_eq!(res.len(), 1);
```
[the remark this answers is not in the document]
ancazamfir: Great question! This was true when only one update client at a given height was allowed. In theory it should not be the case anymore. However, I just tried to do multiple updates at same height and I still see a single event. I changed the code for `query_txs()` to return them "all" and let the caller just pick one. 
Since the query is for a given client and consensus state height all the updates must have the same content so for misbehavior we just pick the first.

relayer/src/foreign_client.rs
```diff
@@ -406,6 +476,231 @@ impl ForeignClient {
...
+    /// highest height assuming the previous one is trusted. It submits the first constructed
+    /// evidence (the one with the highest height)
+    /// - a lot of the logic here is derived from the behavior of the only implemented client
+    /// (ics07-tendermint) and might not be general enough.
+    ///
+    pub fn handle_misbehaviour(
+        &self,
+        mut update: Option<UpdateClient>,
+    ) -> Result<Option<AnyMisbehaviour>, ForeignClientError> {
+        thread::sleep(Duration::from_millis(100));
```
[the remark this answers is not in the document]
ancazamfir: added retries

relayer/src/foreign_client.rs
ancazamfir: Good point! We should get some event the way we have the code now. But in the future we might prepend some client updates before the misbehaviour message. I changed the code to return all events.

relayer/src/foreign_client.rs
ancazamfir: I added numbers in `compatible_with`. Let me know if you think i should add in other places.

ancazamfir:
Thanks @romac, @cezarad and @adizere for the great comments!! I think I addressed most of them but please double check.

romac (approved):

Issue #695, opened 2021-02-22
handler for send packets 


cezarad:
## modules 
Relates to issue #362 

### Summary 
The handler for send_packets and its set-up 
The message send_transfer is processed from ICS26. 

#### For Admin Use

- [ x] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #706, opened 2021-02-24
Ics04 packets setup and send packet handler

cezarad:
Closes: #695

-  ics04 handler for send packets and tests for it 
-  added time to the mock header 
- channel reader and writer for packets 
-  mock implementations for the context reader and writer 
-  ics20 send transfer setup to call for ics04 send packets 
- filled in the deliver message for missing message types in ics26/handler

The message for send packets is processed via ICS26. Minimal application set-up relevant for send packets. 

______

For contributor use:


modules/src/ics02_client/client_def.rs
```diff
@@ -282,6 +284,23 @@ pub enum AnyConsensusState {
 }
 
 impl AnyConsensusState {
+    pub fn latest_timestamp(&self) -> u64 {
```
adizere, line 287: Two ideas:

1. Would make sense to rename this into `timestamp` as there is no earliest or latest timestamp in a single consensus state.

2. We should try to handle the panic in converting the integer here. One solution: use the value `0` as an invalid timestamp which we return if conversion failed; to catch the error in converting integer -> uint we can use try_from. Example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=56d9637ca51dafadc0f4bbf4b18f6227

modules/src/ics02_client/client_def.rs
```diff
@@ -282,6 +284,23 @@ pub enum AnyConsensusState {
...
+        match self {
+            Self::Tendermint(cs_state) => {
+                let date = <DateTime<Utc> as From<Time>>::from(cs_state.timestamp);
```
adizere, line 290: I think it's usually a good practice to avoid the syntax `<A as B>::method()`. This could make it easier to read:

```suggestion
                let date: DateTime<Utc> = cs_state.timestamp.into();
```

romac (changes requested):

modules/Cargo.toml
```diff
@@ -38,6 +38,7 @@ dyn-clonable = "0.9.0"
 regex = "1"
 bech32 = "0.8.0"
 subtle-encoding = "0.5"
+rust-crypto = "0.2.36"
```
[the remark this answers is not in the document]
cezarad: it's for the hashing some fields of a packet.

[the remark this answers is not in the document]
romac: Can you point me to where it's used?

modules/src/mock/context.rs
```diff
@@ -362,20 +382,30 @@ impl ChannelReader for MockContext {
...
-            None => Err(ICS4Error::from(ICS4Kind::NoPortCapability(port_id.clone()))),
+            None => Err(Ics4Error::from(Ics4Kind::NoPortCapability(port_id.clone()))),
         }
     }
 
+    fn get_next_sequence_send(&self, port_channel_id: &(PortId, ChannelId)) -> Option<&u64> {
+        self.next_sequence_send.get(port_channel_id)
+    }
+
+    fn hashing(&self, input: String) -> String {
```
adizere, line 403: @romac I think this is where the new `rust-crypto` dependency is being used. A hash/digest is necessary when we call `store_packet_commitment`.

That being said, since we only use this method in the mock for tests, it would make sense to make this crate a dep only for `[dev-dependencies]`.

@cezarad: I would give this method a better name (usually methods in the Reader and Keeper are actions/verbs):

```suggestion
    fn hash(&self, input: String) -> String {
```

modules/src/mock/context.rs
```diff
@@ -426,14 +456,27 @@ impl ChannelKeeper for MockContext {
...
         Ok(())
     }
 
     fn increase_channel_counter(&mut self) {
         self.channel_ids_counter += 1;
     }
+
+    fn store_packet_commitment(
+        &mut self,
+        key: &(PortId, ChannelId, Sequence),
```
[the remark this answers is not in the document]
cezarad: yes indeed. It will touch the context of both connection and channel.

modules/src/ics02_client/client_def.rs
cezarad: Done&Done.

modules/src/ics02_client/client_def.rs
```diff
@@ -282,6 +284,23 @@ pub enum AnyConsensusState {
...
+                let value = date.timestamp();
+                if value < 0 {
+                    panic!("Negative timestamp")
+                } else {
+                    value as u64
+                }
```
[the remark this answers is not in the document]
cezarad: I used tryfrom as suggested by @Adi above.

modules/src/ics04_channel/packet_handler/send_packet.rs
```diff
@@ -79,7 +79,9 @@ pub fn send_packet(ctx: &dyn ChannelReader, packet: Packet) -> HandlerResult<Pac
         .client_consensus_state(&client_id, latest_height)
         .ok_or_else(|| Kind::MissingClientConsensusState(client_id.clone(), latest_height))?;
 
-    let latest_timestamp = consensus_state.latest_timestamp();
+    let latest_timestamp = consensus_state
+        .timestamp()
+        .map_err(|e| Kind::NegativeChainTimestamp.context(e))?;
```
adizere, line 84: nice!

modules/src/ics04_channel/error.rs
```diff
@@ -129,6 +129,9 @@ pub enum Kind {
     #[error("Receiving chain block timestamp >= packet timeout timestamp")]
     LowPacketTimestamp,
 
+    #[error("Negativ chain timestamp: overflow ")]
+    NegativeChainTimestamp,
```
adizere, line 133: I think calling this "chain timestamp" may give rise to confusion. Also, it would be nice to know what is the exact value that provoked this error.. the original `i64` (?) value that caused the error.

```suggestion
    #[error("Invalid timestamp in consensus state:{0}; timestamp must be a positive value")]
     NegativeChainTimestamp(int64),
```

modules/Cargo.toml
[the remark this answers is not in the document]
adizere: Removed this dependency and replaced with `sha2`. This crate is also an optional dependency. 35335dac

adizere (approved):

modules/src/application/ics20_fungible_token_transfer/relay_application_logic.rs
```diff
@@ -0,0 +1,23 @@
...
+use crate::{ics04_channel::packet::PacketResult, ics26_routing::context::Ics26Context};
+
+use super::error::{Error, Kind};
+use super::msgs::transfer::MsgTransfer;
+use super::relay_application_logic;
+
+pub mod send_transfer;
+
+/// Entry point for processing a transfer message in ICS20 token transfer
+pub fn dispatch<Ctx>(ctx: &mut Ctx, msg: MsgTransfer) -> Result<HandlerOutput<PacketResult>, Error>
```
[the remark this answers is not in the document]
cezarad: We don't need it. In this PR there are two ways (redundant) to call send packet one from ICS26 and one from ICS20. The one form ICS20 (this one) exists in the go implementation. I can remove but if the Go code is the spec I wanted to make sure all in there is discussed before I change it.

modules/src/mock/client_state.rs
```diff
@@ -109,17 +110,24 @@ impl TryFrom<RawMockConsensusState> for MockConsensusState {
...
         Ok(Self(MockHeader::try_from(raw_header)?))
     }
 }
 
+impl MockConsensusState {
+    pub fn timestamp(&self) -> u64 {
+        (self.0).1
+    }
+}
+
```
ancazamfir, line 124: Can we move this to the `MockConsensusState` definition?

modules/src/ics04_channel/handler/send_packet.rs
```diff
@@ -0,0 +1,267 @@
...
+
+    if packet.sequence.is_zero() {
+        return Err(Kind::ZeroPacketSequence.into());
+    }
+    if packet.timeout_height.is_zero() && packet.timeout_timestamp == 0 {
+        return Err(Kind::ZeroPacketTimeout.into());
+    }
+    if packet.data.is_empty() {
+        return Err(Kind::ZeroPacketData.into());
+    }
```
ancazamfir, line 36: This should be done in the `impl TryFrom<RawPacket> for Packet {...`
It will be needed later in `impl TryFrom<RawMsgRecvPacket> for MsgRecvPacket {` and all other messages that include a `Packet`, like acks, timeouts, etc.

modules/src/ics04_channel/handler/send_packet.rs
```diff
@@ -0,0 +1,267 @@
...
+        .ok_or_else(|| Kind::MissingClientState(client_id.clone()))?;
+
+    // prevent accidental sends with clients that cannot be updated
+    if client_state.is_frozen() {
+        return Err(Kind::FrozenClient(connection_end.client_id().clone()).into());
+    }
+
+    // check if packet height timeouted on the receiving chain
+
+    let latest_height = client_state.latest_height();
```
ancazamfir, line 82:
```suggestion
    // check if packet height timed out on the receiving chain
    let latest_height = client_state.latest_height();
```

modules/src/ics04_channel/handler/send_packet.rs
```diff
@@ -0,0 +1,267 @@
...
+    let packet_height = packet.timeout_height;
+
+    if !packet.timeout_height.is_zero() && packet_height.cmp(&latest_height).eq(&Ordering::Less) {
```
ancazamfir, line 85:
```suggestion
    if !packet.timeout_height.is_zero() && packet_height <= latest_height {
```

modules/src/ics04_channel/handler/send_packet.rs
```diff
@@ -0,0 +1,267 @@
...
+        return Err(Kind::LowPacketHeight(latest_height, packet.timeout_height).into());
+    }
+
+    //check if packet timestamp timeouted on the receiving chain
```
ancazamfir, line 89: We don't really check the receiving chain. This (like the timeout height above) is a weak check, but if the latest client consensus state we have for the receiving chain is newer (higher timestamp) then for sure we have a timeout. 
```suggestion
    //check if the packet timeout timestamp is newer than the timestamp of the latest consensus of the receiving chain
```

modules/src/ics04_channel/handler/send_packet.rs
```diff
@@ -0,0 +1,267 @@
...
+        .client_consensus_state(&client_id, latest_height)
+        .ok_or_else(|| Kind::MissingClientConsensusState(client_id.clone(), latest_height))?;
+
+    let latest_timestamp = consensus_state
+        .timestamp()
+        .map_err(Kind::ErrorInvalidConsensusState)?;
+
+    let packet_timestamp = packet.timeout_timestamp;
+    if !packet.timeout_timestamp == 0 && packet_timestamp.cmp(&latest_timestamp).eq(&Ordering::Less)
+    {
```
ancazamfir, line 100:
```suggestion
    if !packet.timeout_timestamp == 0 && packet_timestamp <= latest_timestamp {
```

modules/src/ics04_channel/handler/send_packet.rs
```diff
@@ -0,0 +1,267 @@
...
+        return Err(Kind::LowPacketTimestamp.into());
+    }
+
+    // check sequence number
+    let next_seq_send = ctx
+        .get_next_sequence_send(&(packet.source_port.clone(), packet.source_channel.clone()))
+        .ok_or(Kind::MissingNextSendSeq)?;
+
+    if !packet.sequence.eq(&next_seq_send) {
```
ancazamfir, line 109:
```suggestion
    if packet.sequence != next_seq_send {
```

modules/src/ics04_channel/msgs.rs
```diff
@@ -1,12 +1,17 @@
...
 
+use acknowledgement::MsgAcknowledgement;
+
 use crate::ics04_channel::msgs::chan_close_confirm::MsgChannelCloseConfirm;
 use crate::ics04_channel::msgs::chan_close_init::MsgChannelCloseInit;
 use crate::ics04_channel::msgs::chan_open_ack::MsgChannelOpenAck;
 use crate::ics04_channel::msgs::chan_open_confirm::MsgChannelOpenConfirm;
 use crate::ics04_channel::msgs::chan_open_init::MsgChannelOpenInit;
 use crate::ics04_channel::msgs::chan_open_try::MsgChannelOpenTry;
+//use crate::application::ics20_fungible_token_transfer::msgs::transfer::MsgTransfer;
```
ancazamfir, line 12: remove?

modules/src/ics04_channel/msgs.rs
```diff
@@ -34,3 +39,12 @@ pub enum ChannelMsg {
...
     ChannelCloseConfirm(MsgChannelCloseConfirm),
 }
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum PacketMsg {
+    RecvPacket(MsgRecvPacket),
+    AckPacket(MsgAcknowledgement),
+    ToPacket(MsgTimeout),
+    ToClosePacket(MsgTimeoutOnClose),
+    //SendPacket(MsgTransfer)
```
ancazamfir, line 49: remove?

modules/src/mock/host.rs
```diff
@@ -47,7 +47,10 @@ impl HostBlock {
     /// Generates a new block at `height` for the given chain identifier and chain type.
     pub fn generate_block(chain_id: ChainId, chain_type: HostType, height: u64) -> HostBlock {
         match chain_type {
-            HostType::Mock => HostBlock::Mock(MockHeader(Height::new(chain_id.version(), height))),
+            //TODO: check timestamp
```
ancazamfir, line 50: What is this comment referring to?

modules/src/application/ics20_fungible_token_transfer/relay_application_logic.rs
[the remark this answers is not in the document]
ancazamfir: Yes, looks good with it removed. We can come back to it later if needed.

modules/src/application/ics20_fungible_token_transfer/error.rs
```diff
@@ -1,10 +1,41 @@
...
-#[derive(Clone, Debug, Error)]
-pub enum Kind {}
+#[derive(Clone, Debug, Error, PartialEq, Eq)]
+pub enum Kind {
+    #[error("unrecognized ICS-20 transfer message type URL {0}")]
+    UnknownMessageTypeUrl(String),
+
+    #[error("the message is malformed and cannot be decoded")]
+    MalformedMessageBytes,
+
```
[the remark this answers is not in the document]
adizere: I think @cezarad is using these errors in the packet receive (correct me if I'm wrong), so we left them here, and will cleanup the Kind type in ICS20 with that PR.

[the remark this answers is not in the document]
cezarad: I removed them. We will do ICS20 from scratch.

romac (approved):
Looks good!

Issue #561, opened 2021-01-27
Cleanup pending packets when (re)starting the relayer loop


ancazamfir:
## Crate
relayer

## Summary
When starting the relayer loop, after subscribing to IBC events but before starting processing them, any pending `send_packet` and `acknowledge` packets should be handled. This is the same as used in the `packet-recv` and `packet-ack` CLIs, it just needs proper integration with the relayer loop.

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #708, opened 2021-02-24
Add CI test for the relayer loop


ancazamfir:
## Crate

n/a

## Summary

A new python test should be added to `e2e` suite to test packet relayer loop. The test should ensure that the relayer loop will relay all pending packets on channels it is configured to relay.

## Proposal
Something along these lines:
1) create some unreceived acks
```
hermes tx raw ft-transfer ibc-0 ibc-1 transfer channel-0 10000 1000 -n 2
hermes tx raw ft-transfer ibc-1 ibc-0 transfer channel-0 10000 1000 -n 2
hermes tx raw packet-recv ibc-1 ibc-0 transfer channel-0
hermes tx raw packet-recv ibc-0 ibc-1 transfer channel-0
```

2. create some unreceived packets
```
hermes tx raw ft-transfer ibc-0 ibc-1 transfer channel-0 10000 1000 -n 3
hermes tx raw ft-transfer ibc-1 ibc-0 transfer channel-0 10000 1000 -n 4
```

3. verify the expected number of unreceived packets and acks on each channel end
```
hermes query packet unreceived-packets ibc-0 ibc-1 transfer channel-0
hermes query packet unreceived-acks ibc-0 ibc-1 transfer channel-0
hermes query packet unreceived-packets ibc-1 ibc-0 transfer channel-0
hermes query packet unreceived-acks ibc-1 ibc-0 transfer channel-0
```
4. start relaying on the channel - it should clear the unreceived packets
```hermes start ibc-0 ibc-1 -p transfer -c channel-0```

5. in parallel send some packets
```
hermes tx raw ft-transfer ibc-0 ibc-1 transfer channel-0 10000 1000 -n 3
hermes tx raw ft-transfer ibc-0 ibc-1 transfer channel-0 10000 1000 -n 4
```
6. wait a bit and make sure there are no pending packets (same commands as in 3. but empty vecs in results):
```
hermes query packet unreceived-packets ibc-0 ibc-1 transfer channel-0
hermes query packet unreceived-acks ibc-0 ibc-1 transfer channel-0
hermes query packet unreceived-packets ibc-1 ibc-0 transfer channel-0
hermes query packet unreceived-acks ibc-1 ibc-0 transfer channel-0
```

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #707, opened 2021-02-24
Add test for MsgTimeout in CI


ancazamfir:
## Crate
n/a
the test should be added to `./e2e/run.py`

## Summary
Add a test that sends a packet with low timeout height and ensures the packet is cleared (`MsgTimeout`) is received.

## Proposal

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #709, opened 2021-02-24
Relay pending packets in the relayer loop

ancazamfir:
Closes: #561

## Description

- added the new `relay_pending_packets()` that queries the chains for the unreceived packets and acks, builds and sends transactions. This function uses the exact methods used by the corresponding CLIs. It adds a retry around these.
- changed `relay_from_events()` to perform relaying of pending packets after events subscription but before the first event is processed. The height for queries and client updates must be the height of the first event minus one.
- noticed that the signer for msg timeout is wrong in master after my last commit and fixed that. We need to add a timeout test to the CI (#707)
- also opened #708 for relayer loop CI
______

For contributor use:

### DIFF
--- a/relayer/src/link.rs
+++ b/relayer/src/link.rs
@@ -912,10 +912,8 @@ impl Link {
                     clear_events_a = false;
                     Ok(())
                 }
-                Err(e) => match e {
-                    LinkError::OldPacketClearingPending => Ok(()),
-                    _ => Err(e),
-                },
+                Err(LinkError::OldPacketClearingPending) => Ok(()),
+                Err(e) => Err(e),
             }?;
 
             match self.b_to_a.relay_from_events(clear_events_b) {
@@ -923,10 +921,8 @@ impl Link {
                     clear_events_b = false;
                     Ok(())
                 }
-                Err(e) => match e {
-                    LinkError::OldPacketClearingPending => Ok(()),
-                    _ => Err(e),
-                },
+                Err(LinkError::OldPacketClearingPending) => Ok(()),
+                Err(e) => Err(e),
             }?;
 
             // TODO - select over the two subscriptions
--- a/relayer/src/link.rs
+++ b/relayer/src/link.rs
@@ -63,6 +63,12 @@ pub enum LinkError {
 
     #[error("exhausted max number of retries:")]
     RetryError,
+
+    #[error("old packets not cleared yet")]
+    OldPacketClearingPending,
+
+    #[error("clearing of old packets failed")]
+    OldPacketClearingFailed,
 }
 
 pub struct RelayPath {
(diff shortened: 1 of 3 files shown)


relayer/src/link.rs
```diff
@@ -856,13 +895,39 @@ impl Link {
...
+            match self.b_to_a.relay_from_events(clear_events_b) {
+                Ok(()) => {
+                    clear_events_b = false;
+                    Ok(())
+                }
+                Err(e) => match e {
+                    LinkError::OldPacketClearingPending => Ok(()),
+                    _ => Err(e),
+                },
+            }?;
```
romac, line 930: I find this slightly more readable as it decreases the indentation a bit, but feel free to ignore.

```suggestion
            match self.a_to_b.relay_from_events(clear_events_a) {
                Ok(()) => {
                    clear_events_a = false;
                    Ok(())
                }
                Err(LinkError::OldPacketClearingPending) => Ok(()),
                Err(e) => Err(e),
            }?;

            match self.b_to_a.relay_from_events(clear_events_b) {
                Ok(()) => {
                    clear_events_b = false;
                    Ok(())
                }
                Err(LinkError::OldPacketClearingPending) => Ok(()),
                Err(e) => Err(e),
            }?;
```

romac (approved):
Wow, that's much less code than I thought it would take :)

Left a suggestion inline, but feel free to ignore.

relayer/src/link.rs
```diff
@@ -391,10 +397,39 @@ impl RelayPath {
         self.timeout_msgs = vec![];
     }
 
-    fn relay_from_events(&mut self) -> Result<(), LinkError> {
+    fn relay_pending_packets(&mut self) -> Result<(), LinkError> {
+        println!("clearing old packets on {:#?}", self.channel);
+        for _i in 0..MAX_ITER {
```
romac, line 402: More general question, which does not be addressed in this PR: would it be sensible to introduce a delay or even implement some kind of exponential backoff strategy when retrying an operation which has failed (in this case, relaying the pending packets)?

ancazamfir: Yeh, we've been avoiding the whole error handling for some time. We have this pattern **everywhere**! We blindly retry in a loop for all errors. We need to sit down and see how to deal with each error. We have RPC, light client, on-chain and internal errors. The most complex ones are the on-chain errors that are a bit more subtle (leaving aside errors in the relayer code) and also they are not typed (we just get an error message and the reason is embedded in a string). The retry with exponential backoff may apply for some cases, in some cases we should not retry (e.g. a transaction has already been submitted by another relayer), in other cases we should rebuild stuff, etc. I think we should start a little doc and try to cover all errors and what actions we should take. 
Because the problem is so broad I would leave this handling here for later.

ancazamfir: Opened https://github.com/informalsystems/ibc-rs/issues/712 for error handling.

PR #719, opened 2021-03-01
Replace deprecated rust-crypto dep with RustCrypto org packages

adizere:
Closes: #352 

## Description

Our security audits are starting to fail:
https://github.com/informalsystems/ibc-rs/runs/2006092577?check_suite_focus=true

After a brief search, I choose to use crates from https://github.com/RustCrypto. Their crates seem to be maintained (repos had activity within the last few hours/days).

- ~~__Note that we're still depending on rust-crypto transitively via `bitcoin-wallet v1.1.0`.__ hence the security audit still fails.~~ h/t @romac for suggesting `tiny-bip39` to replace `bitcoin-wallet`.
- Also removed dependency `ics23` since we didn't use it and triggered audit failure. Maintainers of `ics23` are aware of the problem ([ics23/#37](https://github.com/confio/ics23/issues/37)).

______

For contributor use:


romac:
As [we only use](https://github.com/informalsystems/ibc-rs/blob/052703bf76543b93debc3a6b9ab50bca5dde6828/relayer/src/keyring/store.rs#L14) `bitcoin-wallet` for BIP30 mnemonics, I suggest we switch over to https://lib.rs/crates/tiny-bip39 which provides just the functionality we need.

romac (approved):

Issue #357, opened 2020-11-04
Relayer CLI for client upgrade message


adizere:
## Crate

relayer, relayer-cli, ibc

## Summary

The `MsgUpgradeClient` raw message is defined here:

https://github.com/informalsystems/ibc-rs/blob/9a930e62d715ad01d544f8c5e31af033b4649913/proto/src/prost/ibc.core.client.v1.rs#L84

We need:
- [x] a corresponding domain type message definition + implementation
- [x] clarify the exact cli syntax for this command and add support for it in `relayer-cli` and `relayer` crates:
    - `relayer tx raw upgrade-client <client-id> -h <upgrade-height>` 
    - done with ADR 006
- [X] the handlers will be handled separately https://github.com/informalsystems/ibc-rs/issues/722

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #722, opened 2021-03-03
Handler for upgrade client


adizere:
## Crate

`ibc`

## Summary

- [ ] handler implementation in ics02_client::handler::upgrade_client.rs
     - raise an issue with cosmos/ics to add canonical algorithm for handling this message? not clear if that's desirable since this is documented and handled mostly in the SDK: https://github.com/cosmos/cosmos-sdk/tree/master/docs/ibc/upgrades
     - probalby inspired from https://github.com/cosmos/cosmos-sdk/pull/7367 since there is no canonical algorithm in the cosmos/ics spec (yet)
- [ ] clarify the use of `upgrade_path` field in [Tendermint client state](https://github.com/informalsystems/ibc-rs/blob/db2be6f72fef647933f48d80b29a0f9f12c804d4/modules/src/ics07_tendermint/client_state.rs#L25) and include that in tests

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

PR #723, opened 2021-03-03
Hermes CLI for upgrading client

adizere:
Closes: #357
Closes: #734 

## Description

- added a new CLI
    - `hermes tx raw upgrade-client ibc-1 ibc-0 07-tendermint-1`
- testing instructions for the CLI can be found [in this guide](https://github.com/informalsystems/ibc-rs/blob/adi/357_upgrade/guide/src/upgrade_test.md)

Additional notes:
- added the corresponding proto file: `proto/src/prost/cosmos.upgrade.v1beta1.rs`
- also added the proto-compiler `Cargo.lock` file (note: this file account for over 1K new lines of code, so the PR is not as big as it seems)
    - also mentioned in proto-compiler that we should use the `locked` flag so that the binary is identical every time we re-generate these files;
    - this is important to avoid having aesthetic differences in the generated proto files (for example `pub log: std::string::String,` vs `pub log: ::prost::alloc::string::String,`)
- added some stubs for the upgrade client handler, but the rest will be handled in a separate issue (https://github.com/informalsystems/ibc-rs/issues/722)
- also [cleaned up](https://github.com/informalsystems/ibc-rs/pull/723/commits/12722c0b4e7bf6ff468eff979490ac105f54d5f2) some rust doc

______

For contributor use:


scripts/setup-chains
```diff
@@ -63,9 +63,7 @@ mkdir -p "$GAIA_DATA" && cd "$GAIA_DATA" && cd ../
 ONE_CHAIN="$(dirname "$0")/one-chain"
 
 CHAIN_0_RPC_PORT=26657
-CHAIN_0_RPC_ADDR="localhost:$CHAIN_0_RPC_PORT"
```
adizere, line 66: Was not used.

scripts/one-chain
```diff
@@ -133,5 +133,7 @@ $BINARY --home $CHAIN_DIR/$CHAIN_ID start --pruning=nothing --grpc.address="0.0.
 # Show validator's and user's balance
 sleep 3
 RPC_ADDR="tcp://localhost:$RPC_PORT"
+echo "Balances for validator '$VALIDATOR' @ '$RPC_ADDR'"
```
adizere, line 136: Made it more explicit as it wasn't clear from output which was user <> validator.

guide/src/upgrade_test.md
```diff
@@ -0,0 +1,176 @@
...
+ fi
+
+-GAIA_DATA="$(pwd)/data"
++GAIA_DATA="~/go/src/github.com/cosmos/relayer/data"
+
+ CHAIN_0_RPC_PORT=26657
+ CHAIN_0_RPC_ADDR="localhost:$CHAIN_0_RPC_PORT"
+```
+
+No setup the clients for Hermes to use:
```
romac, line 156:
```suggestion
Now setup the clients for Hermes to use:
```
```suggestion
No setup the clients for Hermes to use:
```

modules/src/ics02_client/events.rs
```diff
@@ -214,3 +218,19 @@ impl From<ClientMisbehavior> for IbcEvent {
         IbcEvent::ClientMisbehavior(v)
     }
 }
+
+/// Signals a recent upgrade of an on-chain client (IBC Client).
+#[derive(Debug, Deserialize, Serialize, Clone)]
```
romac, line 223:
```suggestion
#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Deserialize, Serialize)]
```

modules/src/ics02_client/events.rs
```diff
@@ -214,3 +218,19 @@ impl From<ClientMisbehavior> for IbcEvent {
...
+pub struct UpgradeClient(Attributes);
+
+impl UpgradeClient {
+    pub fn set_height(&mut self, height: Height) {
```
romac, line 227: Are there no other attributes we might want to expose?

modules/src/ics02_client/handler/upgrade_client.rs
```diff
@@ -0,0 +1,52 @@
...
+    let client_state = ctx
+        .client_state(&client_id)
+        .ok_or_else(|| Kind::ClientNotFound(client_id.clone()))?;
+
+    if client_state.is_frozen() {
+        return Err(Kind::ClientFrozen(client_id).into());
+    }
+
+    // Not implemented yet: https://github.com/informalsystems/ibc-rs/issues/722
+    todo!()
```
romac, line 37: Will this be addressed in this PR or a subsequent one?

modules/src/ics02_client/msgs/upgrade_client.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    fn route(&self) -> String {
+        crate::keys::ROUTER_KEY.to_string()
+    }
+
+    fn type_url(&self) -> String {
+        TYPE_URL.to_string()
+    }
+}
+
```
romac, line 37: We are missing a `impl Protobuf<RawMsgUpgradeClient> for MsgUpgradeAnyClient` and `TryFrom<RawMsgUpgradeClient> for MsgUpgradeAnyClient ` here.

modules/src/ics24_host/path.rs
```diff
@@ -41,6 +54,14 @@ pub enum Path {
         channel_id: ChannelId,
         sequence: Sequence,
     },
+    Upgrade(ClientUpgradePath),
+}
+
+/// Paths that are specific for client upgrades.
+#[derive(Clone, Debug)]
```
romac, line 61: Let's always add `PartialEq, Eq, PartialOrd, Ord, Hash` bounds whenever possible, as equality is often necessary (eg. in tests) and `*Ord` bounds allow storing those types in `BTreeMap` while the `Hash` bound allows storing them in `HashMap`.

scripts/init-clients
```diff
@@ -66,8 +66,8 @@ cargo run --bin hermes -- -c "$CONFIG_FILE" light rm -c "$CHAIN_1_ID" --all -y &
 
 # set the primary peers for clients on each chain
 echo "Adding primary peers to light client configuration..."
-cargo run --bin hermes -- -c "$CONFIG_FILE" light add $CHAIN_0_RPC_ADDR -c "$CHAIN_0_ID" -f -p -s "$GAIA_DATA/$CHAIN_0_ID/data" -y &>/dev/null
-cargo run --bin hermes -- -c "$CONFIG_FILE" light add $CHAIN_1_RPC_ADDR -c "$CHAIN_1_ID" -f -p -s "$GAIA_DATA/$CHAIN_1_ID/data" -y &>/dev/null
+cargo run --bin hermes -- -c "$CONFIG_FILE" light add $CHAIN_0_RPC_ADDR -c "$CHAIN_0_ID" -f -p -s "$GAIA_DATA/$CHAIN_0_ID/data" -y &>/dev/null || printf "\tError adding primary peer to %s\n" "$CHAIN_0_RPC_ADDR"
```
romac, line 69: 👍

romac (changes requested):

modules/src/ics02_client/handler/upgrade_client.rs
adizere: Definitely subsequent PR.

modules/src/ics02_client/msgs/upgrade_client.rs
adizere: I did not add it on purpose as it was not necessary. I find it difficult to understand which messages need which kind of de/serialization support, so I decided to stop adding these methods unless they're clearly needed.

But if you think it's better to just have the methods already (e.g., for consistency sake), I will add them.

modules/src/ics02_client/events.rs
adizere: I didn't notice anything else being necessary here at the moment.

modules/src/ics02_client/msgs/upgrade_client.rs
romac: I'd rather we add them everywhere, both for consistency, but also because `ibc` is library aimed at being used outside of our own `ibc-relayer` use case, and other users might have different needs. Of course, we could still wait until someone raises the issue to us, but in the case of messages, I think it wouldn't hurt adding those instances by default.

adizere: Good point! 77691624

adizere:
@ancazamfir @cezarad can you please have a look over this?

ancazamfir:
FYI, I added a new command `upgrade-chain` in 42b326a54375b69553fd0889ebca718af03d2c82 to help with the client upgrade testing. Also removed the patches and dependencies on the Go relayer. Updated the instructions.

relayer-cli/src/commands/tx/client.rs
```diff
@@ -101,3 +102,52 @@ impl Runnable for TxUpdateClientCmd {
...
+        help = "identifier of the chain which underwent upgrade (source chain)"
+    )]
+    src_chain_id: ChainId,
+
+    #[options(
+        free,
+        required,
+        help = "identifier of the client to be upgraded on destination chain"
+    )]
+    dst_client_id: ClientId,
```
ancazamfir, line 123: I think we should make `dst_client_id` optional and if not entered to upgrade all clients of `src_chain_id`  present on `dst_chain_id`. The way it is, with a lot of clients, the user needs to query all clients and issue the upgrade for each client. 
Also, we should have a `hermes client upgrade..` command in place of or in addition to this command.

romac: Do you want to do this in this PR or a follow-up one? If the latter, then we should open an issue to track this.

relayer/src/upgrade_chain.rs
```diff
@@ -0,0 +1,124 @@
...
+    let upgrade_height = dst_chain
+        .query_latest_height()
+        .unwrap()
+        .add(opts.height_offset);
+    let client_state = src_chain
+        .query_client_state(&opts.src_client_id, Height::zero())
+        .unwrap();
+    let mut upgraded_client_state = client_state.zero_custom_fields();
+    upgraded_client_state.latest_height = upgrade_height.increment();
+    upgraded_client_state.unbonding_period = Duration::from_secs(400 * 3600); // TODO add to options
```
romac, line 57: Do we want to address this in this PR or a follow-up one? If the latter we should track this in an issue.

romac (approved):
Awesome, pushed a couple commits for nitpicks and left a comment about a TODO, but otherwise it's all good!

relayer/src/upgrade_chain.rs
ancazamfir: We could do it here although I don't think is that important. We can probably remove the TODO.

relayer/src/upgrade_chain.rs
```diff
@@ -0,0 +1,120 @@
...
+        .query_client_state(&opts.src_client_id, Height::zero())
+        .unwrap();
+
+    let mut upgraded_client_state = ClientState::zero_custom_fields(client_state);
+    upgraded_client_state.latest_height = upgrade_height.increment();
+    upgraded_client_state.unbonding_period = Duration::from_secs(400 * 3600); // TODO add to options
+
+    let raw_client_state = AnyClientState::Tendermint(upgraded_client_state);
+    let plan = Plan {
+        name: "test".to_string(),
```
romac, line 61: Does this name matter or needs to be configured?

ancazamfir: No need to be configured. We can add later more options but I don't see a need right now.

relayer-cli/src/commands/tx/client.rs
ancazamfir: Follow up PR, thanks!

adizere: https://github.com/informalsystems/ibc-rs/issues/763

Issue #702, opened 2021-02-22
Update images, CI and scripts to use gaia v4.1.0


ancazamfir:
## Summary
We should update the docker images, CI and scripts to use gaia `v4.1.0`.
We should also have a process in place to track new gaia releases and do the required changes to our CI and scripts.
 

## Problem Definition
gaia v4.1.0 includes SDK v0.42.0 that, among other things, has the support required for misbehavior detection. I expect this  (or later version) would be used in the next upgrade.

## Proposal

____

#### For Admin Use

- [x] Not duplicate issue
- [ ] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

andynog:
we should use gaia 4.1.0 that was released today https://github.com/cosmos/gaia/releases/tag/v4.1.0

PR #735, opened 2021-03-10
Updating gaiad to version 4.1.0 and updating proto-compiler to use ibc-go

andynog:
Closes: #702 

## Description

Updated gaiad to the latest version 4.1.0 (released today March 10, 2021 which uses cosmos-sdk 0.42 but includes a security fix so all nodes upgraded to this version)
______

For contributor use:


andynog:
I think this is good to be merged @adizere and @ancazamfir.

ancazamfir:
Thanks! We also need to fix the scripts, `./scripts/dev-env ~/.hermes/config.toml ibc-0 ibc-1` currently fails. I realized that running `./e2e/run.py  -c ~/.hermes/config.toml` might be ok once the scripts are fixed as we currently don't decode in `master` the new `header` field in the update client result (this causes failures in my PR #691 where I do this)

ancazamfir:
We should also update the protos, now in https://github.com/cosmos/ibc-go/tree/main/proto/ibcgo

andynog:
> Thanks! We also need to fix the scripts, `./scripts/dev-env ~/.hermes/config.toml ibc-0 ibc-1` currently fails. I realized that running `./e2e/run.py -c ~/.hermes/config.toml` might be ok once the scripts are fixed as we currently don't decode in `master` the new `header` field in the update client result (this causes failures in my PR #691 where I do this)

I've tested the `dev-env` script with `gaiad v4.1.0` and it wokrs for me.

andynog:
Got logic to compile the protos from `ibc-go` for ibc protos and from `cosmos-sdk` for `cosmos` protos working. 

I could compile all crates with the new protos. Ran a `dev-env` test and it passed.

I think this is good to go

proto/src/prost/tendermint.abci.rs
```diff
@@ -0,0 +1 @@
...lt () ; let path = http :: uri :: PathAndQuery :: from_static ("/tendermint.abci.ABCIApplication/LoadSnapshotChunk") ; self . inner . unary (request . into_request () , path , codec) . await } pub async fn apply_snapshot_chunk (& mut self , request : impl tonic :: IntoRequest < :: tendermint_proto :: abci :: RequestApplySnapshotChunk > ,) -> Result < tonic :: Response < :: tendermint_proto :: abci :: ResponseApplySnapshotChunk > , tonic :: Status > { self . inner . ready () . await . map_err (| e | { tonic :: Status :: new (tonic :: Code :: Unknown , format ! ("Service was not ready: {}" , e . into ())) }) ? ; let codec = tonic :: codec :: ProstCodec :: default () ; let path = http :: uri :: PathAndQuery :: from_static ("/tendermint.abci.ABCIApplication/ApplySnapshotChunk") ; self . inner . unary (request . into_request () , path , codec) . await } } impl < T : Clone > Clone for AbciApplicationClient < T > { fn clone (& self) -> Self { Self { inner : self . inner . clone () , } } } impl < T > std :: fmt :: Debug for AbciApplicationClient < T > { fn fmt (& self , f : & mut std :: fmt :: Formatter < '_ >) -> std :: fmt :: Result { write ! (f , "AbciApplicationClient {{ ... }}") } } }
```
ancazamfir, line 1: Is there a way to not generate these `proto/src/prost/tendermint*` empty files?

ancazamfir (approved):

proto/src/prost/tendermint.abci.rs
adizere: It seems that the empty files related to tendermint are dependencies in other files, and that's why they get generated. I manually removed the empty ones from this commit. I left the abci.rs.

Issue #736, opened 2021-03-11
ics04 receive and ack packets 


cezarad:
Related to issue #362 

## Crate

modules 

## Summary

implementation of the packet handlers for 
- receive packet 
- write acknowledgement 
- acknowledge packet 
-
## Problem Definition
- missing handlers 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #737, opened 2021-03-11
Ics04 packet recv write_ack ack 

cezarad:
Closes: #736

## Description

 - implementation for receive packet, write acknowledgement and ack packet handlers 
 - unit test for the handlers 
 - update of the Mock accordingly 
 - unit tests in ICS26 for send packet and receive packet 

______

For contributor use:


modules/src/ics04_channel/context.rs
```diff
@@ -97,6 +113,47 @@ pub trait ChannelKeeper {
                     res.data,
                 )?;
             }
+            PacketResult::Recv(res) => {
+                match res.receipt {
+                    None => {
+                        //Ordered cchannel
```
romac, line 119:
```suggestion
                        // Ordered channel
```

modules/src/ics04_channel/context.rs
```diff
@@ -97,6 +113,47 @@ pub trait ChannelKeeper {
...
+                        self.store_next_sequence_recv(
+                            (res.port_id.clone(), res.channel_id.clone()),
+                            res.seq_number,
+                        )?
+                    }
+                    Some(r) => {
+                        //Unorderd channel
```
romac, line 126:
```suggestion
                        // Unordered channel
```

modules/src/ics04_channel/context.rs
```diff
@@ -37,9 +37,25 @@ pub trait ChannelReader {
 
     fn get_next_sequence_send(&self, port_channel_id: &(PortId, ChannelId)) -> Option<Sequence>;
 
+    fn get_next_sequence_recv(&self, port_channel_id: &(PortId, ChannelId)) -> Option<Sequence>;
```
romac, line 40: We don't have to do this in this PR, but I think that in general we should change the type of the key for the getters from:

```rust
key: &(A, B, C)
```

to

```rust
key: (&A, &B, &C)
```

to avoid having to `.clone()` the A, B and C at call site.

romac (changes requested):
Pushed some minor changes and left a couple comments and a suggestion.

modules/src/ics04_channel/context.rs
adizere: Or maybe add a new type that encapsulates this tuple? We wanted to this originally but kept postponing it.

adizere (approved):
Cool work Cezara! I just added a few commits with some improvements. I think the PR is ready.

modules/src/ics04_channel/context.rs
```diff
@@ -38,9 +38,25 @@ pub trait ChannelReader {
...
+    fn get_packet_acknowledgement(&self, key: &(PortId, ChannelId, Sequence)) -> Option<String>;
+
     /// A hashing function for packet commitments  
     fn hash(&self, value: String) -> String;
 
+    /// Returns the current height of the local chain.
+    fn host_height(&self) -> Height;
+
+    /// Returns the current timestamp of the local chain.
+    fn host_timestamp(&self) -> u64;
```
romac, line 58: Can we introduce a newtype wrapper for the timestamp, here and in other places where we currently use a `u64`? eg. a `pub struct Timestamp { seconds: u64 }` in the `host` module.

romac (approved):
Looks good, especially the tests :) I only have a little comment, but we can follow-up on this in a later PR if you want. Approved preemptively just in case.

modules/src/ics04_channel/context.rs
adizere: https://github.com/informalsystems/ibc-rs/issues/758

PR #742, opened 2021-03-12
Relay on all paths defined in the configuration

romac:
Closes: #748

## Description

If the start-multi command is ran without options, spawn a supervisor per connection defined in the config.

~~This PR is not ready yet because in a 3 chains setup, some events attached to an object that is not relevant to a worker are forwarded to it anyways.~~

~~To fix this, we must perform some queries to ensure that the object attached to an event is relevant to the two chains the supervisor is responsible for.~~

## TODO

- [x] Open issue to track progress on this PR
- [x] Ensure that the `Object` assigned to an event is relevant for the pair of chains the supervisor is responsible for.
- [x] Either avoid spawning workers for irrelevant events or let the worker exit gracefully if it's not relevant.
- [x] Think about how to deal with worker exiting because of internal error. Will be addressed in future PR that introduces [a proper architecture](https://tinyurl.com/6u6hbwh6).

______

For contributor use:


relayer-cli/src/registry.rs
```diff
@@ -0,0 +1,40 @@
...
+use crate::{cli_utils::spawn_chain_runtime, error::Error};
+
+/// Registry for keeping track of [`ChainHandle`]s indexed by a `ChainId`.
+///
+/// The purpose of this type is to avoid spawning multiple runtimes for a single `ChainId`.
+pub struct Registry<'a> {
+    config: &'a Config,
+    handles: HashMap<ChainId, Box<dyn ChainHandle>>,
+}
+
```
ancazamfir, line 17: nice!

adizere (commented):
Looks good. Should we update the changelog?

Also, it wasn't clear to me what is the expected behavior. Upon starting with no options, assuming a default configuration (with dev-env) is the expected behavior to just start the supervisor for both chains and then stop?

```
$ hermes start-multi
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/hermes start-multi`
{"timestamp":"Mar 16 10:27:00.559","level":"INFO","fields":{"message":"spawning supervisor for chains ibc-0 and ibc-1"},"target":"ibc_relayer_cli::commands::start_multi"}
{"timestamp":"Mar 16 10:27:00.772","level":"INFO","fields":{"message":"running listener","chain.id":"ibc-0"},"target":"ibc_relayer::event::monitor"}
{"status":"success","result":"ok"}
{"timestamp":"Mar 16 10:27:00.875","level":"INFO","fields":{"message":"running listener","chain.id":"ibc-1"},"target":"ibc_relayer::event::monitor"}
```

romac:
That's an oversight on my part, I forgot to commit the code which waits for the threads to finish.

There are also other issues with this PR which we discussed with @ancazamfir, so I am going to put it back in draft until I address them.

romac:
> Looks good. Should we update the changelog?

Done.

relayer/src/supervisor.rs
```diff
@@ -379,3 +435,42 @@ pub fn collect_events(batch: EventBatch) -> CollectedEvents {
...
+        src_channel_id = %src_channel_id,
+        src_port_id = %src_port_id,
+        "getting counterparty chain"
+    );
+
+    use ibc::ics02_client::client_state::ClientState;
+
+    let src_channel = src_chain.query_channel(src_port_id, src_channel_id, Height::zero())?;
+
+    // TODO: Check channel state?
```
adizere, line 456: Shall we fix the TODO or is this for later?

scripts/dev-env
```diff
@@ -23,17 +23,23 @@ if [ -z "$3" ]; then
   missing "CHAIN_1_ID"
 fi
 
-if [ "$#" -gt 3 ]; then
+# the 3rd chain can be omitted, won't be launched and configured
+# if [ -z "$4" ]; then
+#   missing "CHAIN_2_ID"
+# fi
```
adizere, line 29:
```suggestion
```

scripts/init-clients
```diff
@@ -25,16 +25,22 @@ if [ -z "$3" ]; then
   missing "CHAIN_1_ID"
 fi
 
-if [ "$#" -gt 3 ]; then
+# the 3rd chain can be omitted, won't be configured
+# if [ -z "$4" ]; then
+#   missing "CHAIN_2_ID"
+# fi
```
adizere, line 31:
```suggestion
```

scripts/setup-chains
```diff
@@ -21,13 +21,19 @@ if [ -z "$2" ]; then
   missing "CHAIN_1_ID"
 fi
 
-if [ "$#" -gt 2 ]; then
+# the 3rd chain can be omitted, won't be launched
+# if [ -z "$3" ]; then
+#   missing "CHAIN_2_ID"
+# fi
```
adizere, line 27:
```suggestion
```

adizere:
Using the following config h/t @romac 


```
[global]
timeout = '10s'
strategy = 'naive'
log_level = 'info'
[[chains]]
id = 'ibc0'
rpc_addr = 'tcp://localhost:26657'
grpc_addr = 'tcp://localhost:9090'
account_prefix = 'cosmos'
key_name = 'testkey'
store_prefix = 'ibc'
gas = 3000000
clock_drift = '5s'
trusting_period = '14days'
[chains.trust_threshold]
numerator = '1'
denominator = '3'
[[chains]]
id = 'ibc1'
rpc_addr = 'tcp://localhost:26557'
grpc_addr = 'tcp://localhost:9091'
account_prefix = 'cosmos'
key_name = 'testkey'
store_prefix = 'ibc'
gas = 3000000
clock_drift = '5s'
trusting_period = '14days'
[chains.trust_threshold]
numerator = '1'
denominator = '3'
[[chains]]
id = 'ibc2'
rpc_addr = 'tcp://localhost:26457'
grpc_addr = 'tcp://localhost:9092'
account_prefix = 'cosmos'
key_name = 'testkey'
store_prefix = 'ibc'
gas = 3000000
clock_drift = '5s'
trusting_period = '14days'
[chains.trust_threshold]
numerator = '1'
denominator = '3'
[[connections]]
a_chain = 'ibc0'
b_chain = 'ibc1'
[[connections.paths]]
a_port = 'transfer'
b_port = 'transfer'
[[connections]]
a_chain = 'ibc1'
b_chain = 'ibc2'
[[connections.paths]]
a_port = 'transfer'
b_port = 'transfer'
[[connections]]
a_chain = 'ibc2'
b_chain = 'ibc0'
[[connections.paths]]
a_port = 'transfer'
b_port = 'transfer'
```

adizere:
Not a problem, but got me confused (thinking that an error occurred), but upon running:

> ./scripts/dev-env ~/.hermes/multi_config.toml ibc0 ibc1 ibc2

There is in the output some error-level messages:

```
...
Removing light client peers from configuration...
error: hermes fatal error: no peers configured for chain: ibc0
error: hermes fatal error: no peers configured for chain: ibc1
error: hermes fatal error: no peers configured for chain: ibc2
...
```

relayer/src/supervisor.rs
romac: It's actually more of a question than a TODO, was hoping you or @ancazamfir could advise me as to what to do here.

relayer/src/supervisor.rs
```diff
@@ -379,3 +435,42 @@ pub fn collect_events(batch: EventBatch) -> CollectedEvents {
...
+
+    let src_connection_id = src_channel
+        .connection_hops()
+        .first()
+        .ok_or_else(|| format!("no connection hops for channel '{}'", src_channel_id))?;
+
+    let src_connection = src_chain.query_connection(&src_connection_id, Height::zero())?;
+
+    // TODO: Check connection state?
```
romac, line 465: @adizere @ancazamfir Same here, do we need to check the connection state to infer the corresponding object or does it not matter?

romac:
> Not a problem, but got me confused (thinking that an error occurred), but upon running:
> 
> > ./scripts/dev-env ~/.hermes/multi_config.toml ibc0 ibc1 ibc2
> 
> There is in the output some error-level messages:
> 
> ```
> ...
> Removing light client peers from configuration...
> error: hermes fatal error: no peers configured for chain: ibc0
> error: hermes fatal error: no peers configured for chain: ibc1
> error: hermes fatal error: no peers configured for chain: ibc2
> ...
> ```
> 
> Again, this is not exactly an error: there are no peers configured for the chain (and therefore they cannot be removed). But others may also find it confusing.

Yeah that's pretty bad UX, sorry about that. This should be a warning at most or even a different message

relayer/src/supervisor.rs
ancazamfir: If the channel doesn't exist, this query returns an `Uninitialized` channel with zero connection hops. So we will see an error when we try to get the connection further down.

relayer/src/supervisor.rs
ancazamfir: Just looked the query for non-existing connection returns a connection in `Uninitialized` state but with `07-tendermint-0` as client id (??). So I think for both channel and connection we should check that the state is not `Uninitialized` before continuing.

relayer/src/supervisor.rs
```diff
@@ -278,40 +303,70 @@ impl From<UnidirectionalChannelPath> for Object {
...
+            Self::UnidirectionalChannelPath(ref path) => &path.dst_chain_id,
+        }
+    }
+
     /// Build the object associated with the given [`SendPacket`] event.
-    pub fn for_send_packet(e: &SendPacket, chain_id: &ChainId) -> Self {
+    pub fn for_send_packet(e: &SendPacket, src_chain: &dyn ChainHandle) -> Self {
+        let dst_chain_id =
+            get_counterparty_chain(src_chain, &e.packet.source_channel, &e.packet.source_port)
+                .unwrap();
```
ancazamfir, line 322: I think we should fix these `.unwrap()` -s and just ignore any bogus events.

relayer/src/supervisor.rs
```diff
@@ -278,40 +303,70 @@ impl From<UnidirectionalChannelPath> for Object {
...
             src_port_id: e.packet.source_port.clone(),
         }
         .into()
     }
 
     /// Build the object associated with the given [`WriteAcknowledgement`] event.
-    pub fn for_write_ack(e: &WriteAcknowledgement, chain_id: &ChainId) -> Self {
+    pub fn for_write_ack(e: &WriteAcknowledgement, src_chain: &dyn ChainHandle) -> Self {
+        let dst_chain_id =
+            get_counterparty_chain(src_chain, &e.packet.source_channel, &e.packet.source_port)
```
ancazamfir, line 336: Testing fails with different channel ids and multiple chains. Here we should have:
```suggestion
            get_counterparty_chain(src_chain, &e.packet.destination_channel, &e.packet.destination_port)
```

romac:
> > Again, this is not exactly an error: there are no peers configured for the chain (and therefore they cannot be removed). But others may also find it confusing.
> 
> Yeah that's pretty bad UX, sorry about that. This should be a warning at most or even a different message

Tracked in Tracked in https://github.com/informalsystems/ibc-rs/issues/782

ancazamfir (approved):
Looks good! Thanks Romain!

Issue #752, opened 2021-03-19
Account number hardcoded


andynog:
## Crate

ibc-relayer

## Summary of Bug

Account number is hard-coded to '0' in the logic to sign a transaction

## Version

v0.1.1

## Steps to Reproduce

When using a key for an account that is not '0' the `tx raw` transaction will not work. For example testing with a chain where the key was for an account with number '1' the tx failed with unauthorized. 

Change the logic to use the account number retrieved from the chain.
____

## For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [X] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

romac:
@andynog Can you please open a PR for your fix if it's ready?

andynog:
> @andynog Can you please open a PR for your fix if it's ready?

Hi @romac this was fixed on PR #753, not sure why the issue was not closed

romac:
Oh right, will close it then :)

Issue #416, opened 2020-11-24
Support for chains that don't have "cosmos" as hrp for bech32 address


andynog:
## Crate

modules

## Summary of Bug

Currently, if one of the chains configure on the relayer doesn't have an address starting with "cosmos" , the transactions will fail with a 

```
conn init failed, error:  tx error: GRPC error: GRPC error: status: Unknown, message: "invalid Bech32 prefix; expected akash, got cosmos", details: [], metadata: MetadataMap { headers: {} }
```

I think this might have to do with this code since "cosmos" is hard-coded there https://github.com/informalsystems/ibc-rs/blob/db04e2468adc09551563b09cf137c872ba800df3/modules/src/address.rs#L11

## Steps to Reproduce

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
Closing in favor of #635.

romac:
Fixed in #753

PR #753, opened 2021-03-20
Fixes to make hermes work with non-gaiad custom chains (e.g. starport)

andynog:
Closes: 

* #416 
* #752 

## Description

This PR implements fixes to allow hermes relay packets between two non-gaiad based chains.

These changes were tested against two `starport v0.15.0` (supports `cosmos-sdk v0.42.1`). These starport chains have custom account prefix addresses (not `cosmos`) such as `chainb1pnv0elanhs8qcq2sst0jssak8magjpj3ggq45x`

This PR also fixes an issue where the account number is not `0`. For example the starport chains have multiple accounts (e.g. `validator` and `bob`). So using `bob` account key in the relayer would make the transactions fail because that account number is `1` in the key store. After the fix, it works now because the account number is dynamically fetched from the chain using grpc (same logic that gets the account sequence to submit transaction).
_____

For contributor use:

### DIFF
--- a/relayer/src/chain/cosmos.rs
+++ b/relayer/src/chain/cosmos.rs
@@ -159,6 +159,8 @@ impl CosmosSdkChain {
         let mut pk_buf = Vec::new();
         prost::Message::encode(&key.public_key.public_key.to_bytes(), &mut pk_buf).unwrap();
 
+        crate::time!("PK {:?}", hex::encode(key.public_key.public_key.to_bytes()));
+
         // Create a MsgSend proto Any message
         let pk_any = Any {
             type_url: "/cosmos.crypto.secp256k1.PubKey".to_string(),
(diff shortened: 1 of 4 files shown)


ancazamfir (approved):
lgtm, thanks Andy!

relayer/src/chain/cosmos.rs
```diff
@@ -181,7 +183,7 @@ impl CosmosSdkChain {
         // Gas Fee
         let coin = Coin {
             denom: "stake".to_string(),
-            amount: "1000".to_string(),
+            amount: "10000".to_string(),
```
romac, line 186: Is this supposed to stay hardcoded forever or do we eventually want to make this configurable/dynamic?

romac (approved):
Looks good, I just have one question but it does not block from merging.

relayer/src/chain/cosmos.rs
andynog: @romac I believe this should be dynamic, we can have a default value but should allow the value to be passed as a parameter, I had opened a separate issue for that https://github.com/informalsystems/ibc-rs/issues/754

Issue #770, opened 2021-03-26
Header trusted validators should be taken from trusted header's next validators


ancazamfir:
## Crate
relayer

## Summary of Bug
see
https://github.com/informalsystems/tendermint-rs/issues/831#issuecomment-808081715

## Steps to Reproduce
This was seen sporadically on `akashnet-2` when trying to update clients.

## Fix
```
@@ -1128,7 +1127,7 @@ impl Chain for CosmosSdkChain {
             trusted_height,
             signed_header: target_light_block.signed_header.clone(),
             validator_set: target_light_block.validators,
-            trusted_validator_set: trusted_light_block.validators,
+            trusted_validator_set: trusted_light_block.next_validators,
         })
     }
 }
```

This is because the consensus state created on-chain stores the `next_validator` hash:
```
	consensusState := &ConsensusState{
		Timestamp:          header.GetTime(),
		Root:               commitmenttypes.NewMerkleRoot(header.Header.GetAppHash()),
		NextValidatorsHash: header.Header.NextValidatorsHash,
	}
```

When this is later used as trusted consensus state agains a new Header the following is checked:
```
	tmTrustedValidators, err := tmtypes.ValidatorSetFromProto(header.TrustedValidators)
	if err != nil {
		return sdkerrors.Wrap(err, "trusted validator set in not tendermint validator set type")
	}

	// assert that trustedVals is NextValidators of last trusted header
	// to do this, we check that trustedVals.Hash() == consState.NextValidatorsHash
	tvalHash := tmTrustedValidators.Hash()
	if !bytes.Equal(consState.NextValidatorsHash, tvalHash) {
		return sdkerrors.Wrapf(
			ErrInvalidValidatorSet,
			"trusted validators %s, does not hash to latest trusted validators. Expected: %X, got: %X",
			header.TrustedValidators, consState.NextValidatorsHash, tvalHash,
		)
	}
```

So `header.trusted_validator_set` must come from the `trusted_header.next_validators`
____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #771, opened 2021-03-26
Use trusted block next validators when building header

ancazamfir:
Closes: #770

## Description

______

For contributor use:


romac (approved):

ancazamfir:
The first solution/ commit does not work as the `next_validators()` in the `LightBlock` doesn't have the proposer set and on-chain validation fails the update because of this. So I changed to pull the header at `trusted_height + 1` and pass the `trusted_height` in `build_header()`. Also had a version that decrements the `trusted_header.height()` inside `build_header()`. I thought that look just a bit uglier. Can't think of a better solution until we refactor the light client.

adizere (approved):
LGTM

romac (approved):

Issue #673, opened 2021-02-16
Re-evaluate use light client functionality in relayer


ancazamfir:
## Crate
relayer

## Summary
I have some doubts wrt the way we are using the tendermint light client in our relayer, trying to detail here.

## Problem Definition
We should be able to create IBC light clients on chain X for the same chain Y with different security parameters (aka options).
Currently the options are defined globally and relayer light clients are initialized before any IBC clients are created on chain. The configuration is also modified to keep the light client store information and having multiple relayer light clients per chain is hard to manage. 
There are also issues when we try to use IBC clients created by other relayers and that don't have a corresponding local light client instantiated.

## Proposal
At a high level:

- remove the light client sections from config. The config should only keep the chain info (full node rpc address and chain specific parameters) and 
- maybe keep some default light client options/ security params
- no need to have a light client store, the store is on chain
- no need for the supervisor I believe

- on `create-client`:
   - specify:
      - rpc address from config - primary, no need for witnesses here
      - options - hardcoded default or from config, allow height, hash, options to be passed in the CLI
   - relayer calls light client (lib) to fetch a light block from primary (similar to what we do at init today)
   - relayer stores this on chain via the `MsgCreateClient`

- on `update-client` the relayer:
   - gets the latest trusted light block from chain
   - gets the target height from context (e.g. latest if from CLI, height tight to proofs and events), same as today
   - calls light client with get_minimal_set(latest_trusted_block, target_height, options) and a set of light blocks is returned
   - updates the chain with the set of light blocks

- relayer implements misbehavior monitoring and evidence submission (in progress):
  - listens to update client events
  - gets the light block for the update height from chain
  - gets light block(s) from the full node(s) it is configured with (light client fetch light block) and 
  - verifies that they are the same
  - submits the the `MsgMisbehaviour` with the two blocks if needed
  - all this happens in a background process/ thread

summary of API needed from the light client:
 - `fetch_light_block(rpc_address, height)` that returns `LightBlock`
 - `get_minimal_set(rpc_address, trusted_light_block, target_height, options)` returns `Vec<LightBlock>`

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #783, opened 2021-03-29
Use a stateless light client 

romac:
See: #673

## Description

This PR introduces a stateless light client implementation, where we do not use a persistent store to hold the latest trusted block but rather supply the trusted block height at each use site.

The light client security parameters are now derived from the client state supplied to the light client at each request.

Moreover, this PR makes use of the light client instance directly instead of spawning a supervisor runtime. The latter is indeed  redundant since misbehavior monitoring and evidence submission are now the direct responsibility of the relayer (cf. https://github.com/informalsystems/ibc-rs/issues/632).

______

For contributor use:


ancazamfir (approved):
Looks great! Also tested with the hub to make sure, all good!

relayer/src/light_client/tendermint.rs
```diff
@@ -1,149 +1,128 @@
...
+
 use tendermint_light_client::{
-    builder::LightClientBuilder, builder::SupervisorBuilder, light_client, store, supervisor,
-    supervisor::Handle, supervisor::Supervisor, types::Height as TMHeight, types::LightBlock,
+    components::{self, io::AtHeight},
+    light_client::{LightClient as TmLightClient, Options as TmOptions},
+    operations,
+    state::State as LightClientState,
+    store::{memory::MemoryStore, LightStore},
+    types::Height as TMHeight,
```
ancazamfir, line 11: nit: why don't we get a clippy on `TMHeight`?

romac: Because it's a type alias (I think), will fix it in #793.

Issue #590, opened 2021-01-31
Help and usage show wrong executable name 'ibc-relayer-cli` 


andynog:
## Crate

ibc-relayer-cli

## Summary of Bug

When I compile the ibc-relayer-cli crate it generates an executable called `hermes` but if I run it, when the help message is displayed, `hermes` is not shown but `ibc-relayer-cli` which might be confusing to the end user.

```shell
$ ./target/release/hermes 
ibc-relayer-cli 0.1.0
Informal Systems <hello@informal.systems>

USAGE:
    ibc-relayer-cli <SUBCOMMAND>

SUBCOMMANDS:
    help       get usage information
    start      start the relayer (currently this refers to the v0 relayer)
    channel    channel functionality for managing channels
    listen     listen to IBC events
    config     manipulate the relayer configuration
    version    display version information
    query      query state from chain
    tx         create IBC transactions on configured chains
    light      basic functionality for managing the lite clients
    keys       manage keys in the relayer for each chain
```

## Version

v0.1.0

## Steps to Reproduce

Build and run
____

## For Admin Use

- [X] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
I looked into this and renaming the content of the help message seems difficult. This is because abscissa uses by default the package name (instead of binary name) to create all usage and help messages:

Some pointers:
- under the hood, when we run `./target/release/hermes` what we actually run is an `abscissa_core::EntryPoint`, parametrized with `CliCmd`:

https://github.com/informalsystems/ibc-rs/blob/9d0a7b8f11a252e68e039a976910edd33107f074/relayer-cli/src/application.rs#L60

- the `EntryPoint` [inherits its `name` field](https://github.com/iqlusioninc/abscissa/blob/82787669c61a371cfab634ba5520f698b6d1e4a2/core/src/command/entrypoint.rs#L62) (which is used in all usage/help messages) from `CliCmd::name()`

- `CliCmd` has its `name` setup by a derive: `#[derive(Command..` 

https://github.com/informalsystems/ibc-rs/blob/93c0e6d4342696523adbde4ea57f4e2c3965b6a2/relayer-cli/src/commands.rs#L35-L37

[...]

PR #747, opened 2021-03-17
Add packet delay feature

adizere:
Closes: #640

## Description

Primary feature added:

- `hermes create connection ibc-0 ibc-1 --delay 10` 
    - connection creation CLI now permits creation of new connections with the specified delay parameter
    - the delay is represented as `u64` almost everywhere (`relayer::Connection`, `ibc::ConnectionEnd`) except for `relayer::Channel` where it has type `Duration`
    - it is up to the IBC client implementation (e.g., ICS 07) to interpret this value; in SDK for 07 tendermint this is interpreted as nanoseconds;
    - the Hermes CLI parameter `--delay` interprets this value as __seconds__
- `hermes start ibc-0 ibc-1 -p transfer -c channel-2`
    - if the `-c` option to the start CLI is a channel building on a connection with a non-zero `delay_period`, then all packets relayed will respect this delay

There are two limitations, which I suggest we don't try to tackle in this issue but leave for later:
- adding a newtype wrapper for the `delay_period` and making this value general to any client implementation
    - this requires changes in most of ICS03 module

Other major changes included in this PR:

- Split link.rs into separate modules [this change is postponed until we finish reviewing the PR]
    - moved the `LinkError` type from `relayer/src/link.rs` into `relayer/src/link/error.rs`
    - moved the `RelayPath` type & impl from `relayer/src/link.rs` into `relayer/src/link/relay_path.rs`
- Added `OperationalData` struct as part of `RelayPath` to temporarily store batches of events which are pending processing 
- Refactored `RelayPath` struct and impl:
    - Removed the `src_height` and `dst_height` fields and several others
    - Added two fields `src_operational_data: Vec<OperationalData>` and `dst_operational_data: Vec<OperationalData>`.

## Notes on testing 

Creating a channel with zero packet delay:

- `hermes create channel ibc-0 ibc-1 --port-a transfer --port-b transfer`

Note the id of the newly created channel. Suppose this is `channel-0`.

Creating a channel with non-zero packet delay, say `10` seconds:

1. Create the connection `hermes create connection ibc-0 ibc-1 --delay 10`

Note the id of the newly created connection on chain ibc-0.  Suppose the id is `connection-2`.

2. Create a channel on top of the connection:

`hermes create channel ibc-0  --port-a transfer --port-b transfer --connection-a connection-2`

[...]

PR #796, opened 2021-04-06
Hermes: replace default JSON with human-readable output

adizere:
Closes: #805

Many thanks @romac for code snippets & helping with navigating the code around abscissa!

## Description

Primary change in this PR:

- Introduced the `-j`/`--json` flag to enable output in JSON format:

```
     Running `target/debug/hermes -j query channel end ibc-0 transfer channel-0`
{"timestamp":"Apr 07 21:48:04.410","level":"INFO","fields":{"message":"Using default configuration from: '.hermes/config.toml'"},"target":"ibc_relayer_cli::commands"}
{"timestamp":"Apr 07 21:48:04.415","level":"INFO","fields":{"message":"Options QueryChannelEndCmd { chain_id: ChainId { id: \"ibc-0\", version: 0 }, port_id: PortId(\"transfer\"), channel_id: ChannelId(\"channel-0\"), height: None }"},"target":"ibc_relayer_cli::commands::query::channel"}
{"status":"success","result":{"connection_hops":["connection-0"],"ordering":"Unordered","remote":{"channel_id":"channel-0","port_id":"transfer"},"state":"Open","version":"ics20-1"}}
```

versus pretty-printing (the default):

```
    Running `target/debug/hermes query channel end ibc-0 transfer channel-0`
Apr 07 21:48:31.510  INFO ibc_relayer_cli::commands: Using default configuration from: '.hermes/config.toml'
Apr 07 21:48:31.514  INFO ibc_relayer_cli::commands::query::channel: Options QueryChannelEndCmd { chain_id: ChainId { id: "ibc-0", version: 0 }, port_id: PortId("transfer"), channel_id: ChannelId("channel-0"), height: None }
Success
{
  "connection_hops": [
    "connection-0"
  ],
  "ordering": "Unordered",
  "remote": {
    "channel_id": "channel-0",
    "port_id": "transfer"
  },
  "state": "Open",
  "version": "ics20-1"
}
```

Notable code changes in this PR:

- override the default `abscissa::EntryPoint` with out custom `EntryPoint` in `relayer-cli/src/entry.rs` [per suggestion](https://github.com/informalsystems/ibc-rs/pull/796#issuecomment-814884114);
    - this allowed us to introduce the global application flag `-j` or `--json` to enable JSON only output (default is pretty printing)
    - also allowed us to remove the global `-v` verbose parameter
- introduced the `json()` method in `conclude.rs` to peek into the application state and infer if JSON output should be on or off
- we can dispense with `relayer-cli/build.rs`
    - this was a workaround introduced in https://github.com/informalsystems/ibc-rs/issues/590
    - the purpose was to ensure that Hermes CLI output shows `hermes` (the binary name) instead of `ibc-relayer-cli` (the crate name)

______

For contributor use:


romac:
imho whether or not we output JSON should be a global command-line flag (available on every commands) rather than a config option, otherwise if one wants to use the CLI programmatically (either in a script or in the e2e test suite) one has to ensure that the config file contains the proper value for `log_json`.

I wonder if we could do this by defining our own [`Entrypoint` struct](https://docs.rs/abscissa_core/0.5.2/src/abscissa_core/command/entrypoint.rs.html#11-33, adding the option there, and [telling the framework about it](https://github.com/informalsystems/ibc-rs/blob/6331c5cf01f35a41d0b8e052bec9496048d88f2d/relayer-cli/src/application.rs#L60)?

relayer-cli/Cargo.toml
```diff
@@ -13,7 +13,6 @@ authors = [
 
 description = """
     Implementation of `hermes`, an IBC Relayer developed in Rust.
-    This crate is a CLI wrapper over the `ibc-relayer` library.
```
romac, line 16: 👍

romac:
Looks great, thanks @adizere!

One last thing I wonder now that I am looking at an example: in the human-readable case (ie. when `--json` is not supplied), how about using the standard `dbg!`/`Debug` output instead of still outputting JSON? Since is the output meant to be human-readable, I find the `Debug` output to be more readable than JSON.

Just a thought, based on a personal preference. What do you think?

cezarad (commented):
The result of command and raw txs is different: E.g., create channel will print "Channel handshake finished for Channel{  ... } " in a different style than the "Success" message.

adizere:
Status of this PR as of commit a35a225:
- the last commit enforces filtering by modules, so that only log from crates `ibc_relayer` and `ibc_relayer_cli` are now printed
- with this, we can use `debug!` or `trace!` to print lower-level tracing information (without being overwhelmed by tendermint or hyper debug lines)
- this filter is enabled both for JSON output and pretty-printing

The feature of having debug and trace logging appeared to be useful in debugging recent work (https://github.com/informalsystems/ibc-rs/pull/747).

romac (approved):

Issue #801, opened 2021-04-06
Errors when re-compiling proto files


ancazamfir:
## Crate
ibc-proto

## Summary of Bug
`ibc-rs` doesn't build with the proto commits in `proto/src/prost`

## Steps to Reproduce
- Clone a new repo, branch `master`. 
- Get the refpoints from `COSMOS_.._COMMIT` files:
```
$ cat proto/src/prost/COSMOS_SDK_COMMIT
21814558eaa47b018018711e5fe16e0b16811fce
cat proto/src/prost/COSMOS_IBC_COMMIT
333c1f338b2a14a1928a6f8ab64c37123c0e97b6
```
- clone the proto repos:
```
$ cd proto-compiler
$ cargo run -- clone --out /tmp/cosmos --sdk-commit 21814558eaa47b018018711e5fe16e0b16811fce --ibc-go-commit 333c1f338b2a14a1928a6f8ab64c37123c0e97b6
```
- compile protos:
```cargo run -- compile --sdk /tmp/cosmos/sdk --ibc /tmp/cosmos/ibc --out ../proto/src/prost```

- notice the changes where there shouldn't be any:
```
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   ../proto/src/prost/cosmos.upgrade.v1beta1.rs
```

- try to build `ibc-rs`:
```
$ cd ..
$ cargo build
   Compiling ibc-relayer v0.1.1 (/Users/ancaz/rust/1ibc-rs/relayer)
error[E0432]: unresolved imports `ibc_proto::cosmos::upgrade::v1beta1::QueryCurrentPlanRequest`, `ibc_proto::cosmos::upgrade::v1beta1::QueryUpgradedConsensusStateRequest`
  --> relayer/src/chain/cosmos.rs:47:5
   |
47 |     QueryCurrentPlanRequest, QueryUpgradedConsensusStateRequest,
   |     ^^^^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `QueryUpgradedConsensusStateRequest` in `cosmos::upgrade::v1beta1`
   |     |
   |     no `QueryCurrentPlanRequest` in `cosmos::upgrade::v1beta1`

[...]

PR #802, opened 2021-04-06
Fix ibc-proto recompilation bug

adizere:
Closes: #801 

## Description

Added a safety check to prevent overwriting of cosmos-related files.

The problem is described here: https://github.com/informalsystems/ibc-rs/issues/801#issuecomment-814268643

Solution:
1. introduced separate directories for compiling the SDK files and the IBC-go files, respectively
2. when copying the generated Rust files into `../proto/...`, added a safety check to prevent overwriting of `cosmos.*` files
     - if two identical `cosmos.*` files are generated, only the original (from SDK) is retained
     - the following warning is emitted in this case:

> [warn ] Cosmos-related file exists already ../proto/src/prost/cosmos.upgrade.v1beta1.rs! Ignoring the one generated from IBC-go "/var/folders/sy/207jjj5d7j138ryjs32nbxk00000gn/T/ibc-proto-ibc-go.vsPlSpTwK1aO/cosmos.upgrade.v1beta1.rs"

______

For contributor use:


romac (approved):

romac:
Pushed a couple commits to fix an issue with the Git clone and a typo, looks good otherwise :)

Issue #806, opened 2021-04-08
Hermes token transfer CLI should permit custom destination address


adizere:
## Crate

`ibc-relayer` and `-cli`

## Summary

This is a sub-issue of #628 

> For ft-transfer, allow the supplied destination "address" to optionally be an arbitrary string. On the Agoric chain, where our bech32 prefix is "agoric", the transfer destinations are in their own namespace and look something like "board:1938586739".

## Acceptance Criteria

- the `ft-transfer` command permits an additional parameter that is optional, specifying which receiver address to use, e.g.:

`hermes tx raw ft-transfer ibc-0 ibc-1 transfer channel-2 9999 1000 -n 1 -r board:1938586739`

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

Issue #628, opened 2021-02-09
Agoric's dynamic IBC use case


michaelfig:
## Crate

ibc-relayer-cli

## Summary

Meta-issue for Agoric's required features.

## Problem Definition

https://github.com/Agoric/agoric-sdk implements IBC features for smart contracts running in our Javascript VM.  We've converged on a certain workflow, which we made the Go relayer support, and we would love it if Hermes also supported it.

We expect these things turn out to be common for other chains implementing dynamic IBC functionality, so if they are coherently implemented in Hermes, it will benefit everyone involved.

If there are disadvantages to this workflow, we would hope to engage Informal to adjust it to better fit your architecture.

## Proposal

We need the following features:

- [x] Establish a new IBC connection (not channel) between two chains.  Either have a script, or builtin functionality to keep retrying the process until it succeeds. -- #630
- [ ] Query the above-established connection configuration and report it in a consistent (both human- and machine-readable) way.  This is needed to give the JS contracts a way to name an established connection to another chain with just a connectionID. -- #796 
- [x] Establish a new channel given just a source chain, already-established connectionID on that chain, source and destination portIDs, channel mode (ordered/unordered), and source version (these could be used to build a fresh "path"). -- #715 
- [x] Start a streaming relayer, given only the path config (which could be copied from another relayer instance)
- [x] There should be no restrictions on the number of paths relayed via the same connection.  It may be worthwhile to make a "connection" its own first-class object that can support multiple "paths". -- #715 
- [ ] For `ft-transfer`, allow the supplied destination "address" to optionally be an arbitrary string.  On the Agoric chain, where our bech32 prefix is `"agoric"`, the transfer destinations are in their own namespace and look something like `"board:1938586739"`. - #806

I've only done a quick review of the docs to decide if they are implemented.  Please feel free to check off the ones that have been implemented.

Comments welcome!

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
>  Query the above-established connection configuration and report it in a consistent (both human- and machine-readable) way. This is needed to give the JS contracts a way to name an established connection to another chain with just a connectionID.

Can you provide more details on this? Wondering if this output fulfils your definition of both human- and machine-readable.

```
{
  "status": "success",
  "result": {
[...]

PR #807, opened 2021-04-08
Custom receiver address for `ft-transfer` CLI

adizere:
Closes: #806
Closes: #628 

## Description

- `hermes tx raw ft-transfer ibc-0 ibc-1 transfer channel-2 9999 1000 -n 1 -r board:1938586739`

>{"timestamp":"Apr 08 16:14:17.305","level":"INFO","fields":..... amount: 9999, denom: \"samoleans\", receiver: Some(\"board:1938586739\"), ...

> {"status":"success",....

______

For contributor use:


adizere:
@michaelfig when you have time could you have a look over this?

andynog:
Also @michaelfig could you please clarify the requirements. My understanding is that the receiver address on the destination chain can be any string (not a bech32 address) but the signer of the tx is a bech32 address. Is that correct?

michaelfig:
> Also @michaelfig could you please clarify the requirements. My understanding is that the receiver address on the destination chain can be any string (not a bech32 address) but the signer of the tx is a bech32 address. Is that correct?

That's correct!  Agoric will be enabling bech32 receivers for compatibility with wallets (like Keplr) that make assumptions of the chain's configuration, but we will also support differently-formatted names (like `board:...`) for more fine-grained control of the deposit destination.  It would be great if Hermes could support them both.

I will have a look at the PR now....

michaelfig (commented):
I can't read Rust very well, but allowing the receiver as an option looks right.

Not sure what the `custom_signer` is needed for.  Is it converting the receiver string to bech32?  If so, that's almost certainly not right.  The receiver (say `xyz`) option should be copied directly into the ics20-1 packet's (like: `"receiver": "xyz"`).  If that's what this code does, then I'm happy with it!

romac (approved):

PR #820, opened 2021-04-14
Added type annotations for Snowcat in IBCCore and IBCTokenTransfer TLA+ specs

istoilkovska:
## Description

The TLA+ specifications for the IBC core protocols and fungible token transfer are now enhanced with type annotations for [Apalache](https://apalache.informal.systems/docs/apalache/index.html). Running the new type checker [Snowcat](https://apalache.informal.systems/docs/apalache/typechecker-snowcat.html) on `IBCCore.tla` and `IBCTokenTransfer.tla` can be done by running: 

`apalache-mc typecheck IBCCore.tla`  and  `apalache-mc typecheck IBCTokenTransfer.tla`, respectively. 

In both cases, the type checker says that the types are great! :tada: 
______

For contributor use:


romac (approved):
These type annotations are 💯, thanks @istoilkovska!

I wonder if we may want to type check all the specs on the CI, as a sanity check until we actually check them on CI (about that, how long would you say it takes to verify all the specs in the repo?)

istoilkovska:
It took me a few hours of manual work to go from the old type annotations to the new ones, but once all specs are annotated correctly, running the type checker itself is pretty fast (e.g., 12 seconds for `IBCCore.tla` and 7 seconds for `IBCTokenTransfer.tla`).

romac:
Thanks a lot for going through all that work! How about actual verification with Apalache? Do all or some of the specs go through? And if so, does you have a sense of how much time it takes?

romac:
Will definitely add type checking to the CI then :)

istoilkovska:
I haven't checked the specs with Apalache yet, but I've created an issue on this and assigned @konnov.

konnov:
> It took me a few hours of manual work to go from the old type annotations to the new ones, but once all specs are annotated correctly, running the type checker itself is pretty fast (e.g., 12 seconds for `IBCCore.tla` and 7 seconds for `IBCTokenTransfer.tla`).

Wow. 12 seconds and 7 seconds is pretty slow for a type checker. We should look at the bottlenecks at some point :-)

konnov:
The type annotations are amazing. @istoilkovska, what is your experience in annotating the operators: Does it help to annotate them? It definitely improves my experience as a reader.

romac:
> Does it help to annotate them? It definitely improves my experience as a reader.

Same here!

istoilkovska:
> Does it help to annotate them? 

When adding the annotations here, I let myself be guided by the tool and only added annotations for operators that the tool asked for (as the manual advises). I understand that as a spec reader it definitely helps to have annotated operators. As a spec writer, I assume it would help to quickly refresh one's memory about what some old spec is doing.

Does it also help the type checker? Would it spend less time if the types of the currently non-annotated operators were explicitly given in advance by the spec writer?

konnov:
> Does it also help the type checker? Would it spend less time if the types of the currently non-annotated operators were explicitly given in advance by the spec writer?

I don't think the type checker would be any faster, if it has found a correct solution. It would only improve readability.

Issue #811, opened 2021-04-09
Some (g)RPC queries are missing proper pagination


ancazamfir:
## Crate
relayer

## Summary of Bug
all (g)RPC queries that return multiple items are missing proper pagination

## Steps to Reproduce
- Send > 100 packets: 
   `$ hermes tx raw ft-transfer ibc-0 ibc-1 transfer channel-0 9999 15 -n 500`
- Query packet commitments will return less than 500
  `hermes query packet unreceived-packets ibc-1 ibc-0 transfer channel-0`

Found this while I was debugging why `hermes start` doesn't clear all packets on start.

## Acceptance Criteria
All gRPC queries that return collections should be looked at. Here is a list (needs to be updated after we release `v0.2.0`:
- [x] `query_clients`
- [ ] `query_client_connections` - proto `QueryClientConnectionsRequest` doesn't have pagination
- [x] `query_connections`
- [x] `query_connection_channels`
- [x] `query_channels`
- [x] `query_packet_commitments`
- [x] ~~`query_unreceived_packets`- proto `QueryUnreceivedPacketsRequest` doesn't have pagination~~ - not needed as we ask for specific sequences
- [x] `query_packet_acknowledgements`
- [x] ~~`query_unreceived_acknowledgements` - proto `QueryUnreceivedAcksRequest` doesn't have pagination~~ - not needed as we ask for specific sequences
- [x] `query_consensus_states`

This is using tendermint rpc and should also be fixed:
- [ ] `query_txs`

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #827, opened 2021-04-15
Fix pagination for channels cli and reorg the query_txs()

ancazamfir:
Closes: #811

## Description
- fixed pagination for `hermes query channels`
- clarify that `query_txs()` only queries one tx at a time, reorg the code a bit. added comments

______

For contributor use:


relayer/src/chain/cosmos.rs
```diff
@@ -954,50 +955,70 @@ impl Chain for CosmosSdkChain {
...
                 for seq in &request.sequences {
-                    // query all Tx-es that include events related to packet with given port, channel and sequence
+                    // query first (and only) Tx that includes the event specified in the query request
                     let response = self
                         .block_on(self.rpc_client.tx_search(
                             packet_query(&request, *seq),
                             false,
                             1,
-                            1,
+                            1, // get onw Tx matching the query
```
romac, line 985:
```suggestion
                            1, // get only the first Tx matching the query
```

relayer/src/chain/cosmos.rs
```diff
@@ -954,50 +955,70 @@ impl Chain for CosmosSdkChain {
...
                             Order::Ascending,
                         ))
                         .unwrap(); // todo
```
romac, line 988: Can we fix this while we are it?

relayer/src/chain/cosmos.rs
```diff
@@ -1006,9 +1027,21 @@ impl Chain for CosmosSdkChain {
...
+
+                if let Some(event) = update_client_from_tx_search_response(
+                    self.id(),
+                    &request,
+                    response.txs[0].clone(),
+                ) {
+                    Ok(vec![event])
+                } else {
+                    Ok(vec![])
+                }
```
romac, line 1044:
```suggestion
                let event = update_client_from_tx_search_response(self.id(), &request, response.txs[0].clone());
                event.into_iter().collect()
```

romac (changes requested):

relayer/src/chain/cosmos.rs
```diff
@@ -954,61 +955,88 @@ impl Chain for CosmosSdkChain {
...
-
-                    let mut events =
-                        packet_from_tx_search_response(self.id(), &request, *seq, response)
-                            .map_or(vec![], |v| vec![v]);
-
-                    result.append(&mut events);
+                        .map_err(|e| Kind::Grpc.context(e))?;
+
+                    assert!(
+                        response.txs.len() <= 1,
```
romac, line 991: Shouldn't this be `response.txs.len() != 1`? Or alternatively, if we allow more than 1 response: `response.txs.len() < 1`?

romac (changes requested):

relayer/src/chain/cosmos.rs
ancazamfir: In `tx_search()` above we ask for a single Tx. This is because for a packet event of a specified type, path ans sequence number there can be at most one Tx that includes such an event. So only `response.txs.len()` 0 or 1 are valid.

romac: My bad!

relayer/src/chain/cosmos.rs
```diff
@@ -1364,12 +1346,11 @@ fn update_client_from_tx_search_response(
             IbcEvent::UpdateClient(update) => Some(update),
             _ => None,
         })
-        .filter(|update| {
+        .find(|update| {
```
romac, line 1349: I am not that this has the same semantics as the original code in case there is more than one event at this stage.

Initially, we were adding the events to a Vec and then returning the last one with `pop`, which we then changed to a call to `last` on the iterator, but now with `find`, the first element that matches the filter will be returned instead of the last.

relayer/src/chain/cosmos.rs
```diff
@@ -1259,109 +1287,70 @@ fn packet_from_tx_search_response(
...
-            }
-            let event = res.unwrap();
-            let packet = match &event {
+    response
+        .tx_result
+        .events
+        .into_iter()
+        .filter(|abci_event| abci_event.type_str == request.event_id.as_str())
+        .flat_map(|abci_event| ChannelEvents::try_from_tx(&abci_event))
+        .find(|event| {
```
romac, line 1303: Nitpick: we could split the call to `find` into a `flat_map` and a `find` (or `filter`+`last`, see my other comment) to make smaller logic units.

relayer/src/chain/cosmos.rs
ancazamfir: Yes it's true. I changed it because, currently, first or last are both ok. 
Here we are looking at multiple messages inside a single Tx. First one will be for the message that actually installed the consensus state. Subsequent client update messages for same height are noop-s wrt to chain state. For the same reason we get the first Tx. So this is just saying, from that first Tx (`response` here), get the first matching client update message event.

romac (approved):

Issue #794, opened 2021-04-01
Implement functionality to support ICS-27 (Interchain Accounts)


andynog:
## Crate

modules, relayer, relayer-cli

## Summary

We need to implement the ICS-27 spec needed support in order to relay packets to chains that will support Interchain Accounts (ics-27)

## Problem Definition

Interchain Accounts (ics-27) is coming to the hub in the near future and we need to ensure that Hermes is able to relay packets for `ordered channels` using the `ibcaccount` port.

References:

* Spec: https://github.com/cosmos/ibc/tree/master/spec/app/ics-027-interchain-accounts
* Implementation: https://github.com/cosmos/interchain-accounts

## Proposal

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #839, opened 2021-04-20
Support for ICS27

ancazamfir:
Closes: #794

## Description

- added module version and message actions for ICS27
- testing instructions at https://gist.github.com/ancazamfir/874543d88e4e336474c80dd46e61bd7d

______

For contributor use:


andynog:
OK, I'll try to test this tomorrow. Thanks !

ancazamfir:
@colin-axner is there a way to query the chain for the supported ports and their versions?

colin-axner:
> @colin-axner is there a way to query the chain for the supported ports and their versions?

There is no way to [query the currently binded ports](https://github.com/cosmos/ibc-go/issues/112). This is first on my todo list once ibc-go is ready for its release. The only way to query versions are:
- if the channel exists, query the version within the channel end
- if you want supported versions for a module, you'd need to use some sort of gRPC route for each individual module to get this [value](https://github.com/cosmos/ibc-go/blob/db6f316cb044aec21636c2e83f7926ba6c2edfe4/modules/apps/transfer/types/keys.go#L16)

To make querying for the version associated with a binded port scalable, we would need to modify the [module interface](https://github.com/cosmos/ibc-go/blob/db6f316cb044aec21636c2e83f7926ba6c2edfe4/modules/core/05-port/types/module.go#L13) which would be a more controversial change since the function added to get supported versions is unnecessary for IBC functionality and simply a client side improvement. 

I assume the ideal flow here is using a 05-port gRPC route to get the binded ports for a module (or the other way around) and using a module to get its supported version (all via 05-port routes)

If we didn't modify the module interface, the flow would likely be using 05-port gRPC to get the type of the module binded to the ports, and then switching on that type and performing a gRPC query corresponding to the type. This would require code changes for every added ICS application. 

On the go relayer application versions are user specified. Is there a use case for making these automatic? Or is this primarily needed when responding to a `OnChanInit` event since the versions on each side of the application could potentially be unequal?

ancazamfir:
Thanks for the detailed response @colin-axner!

> On the go relayer application versions are user specified. Is there a use case for making these automatic? Or is this primarily needed when responding to a OnChanInit event since the versions on each side of the application could potentially be unequal?

Both apply. How does a user find out the version to create that `.json` file for the relayer? It's seems like we assume well-known-version along with well-known-port identifiers. It's ok today with two apps but not if the number will scale.
For the second case the relayer sends the `Try` hoping for the best. On failure, since we currently do not distinguish between different errors, the relayer blindly retries a few times. We can discuss the difficulty of extracting SDK errors and lack of standard IBC errors in a different thread. In the meantime I got my answer for this PR and will use the user supplied version. Thanks again!

ics27_config_example.toml
```diff
@@ -0,0 +1,51 @@
+[global]
```
adizere, line 1: I would be in favor of not introducing another configuration file. I can see why it would be useful to have this ICS27-specific one around, however. We could improve things if we tidy up our config files and put all of them in a separate top-level folder.

adizere (approved):
Looks good!
Was able to test w/ the instructions from https://gist.github.com/ancazamfir/874543d88e4e336474c80dd46e61bd7d.

There was just this command 

>  $ icad tx bank send val2 $IBC_ACCOUNT 1000stake --chain-id test-2 --home ./data/test-2 --node tcp://localhost:26657

which generates on the relayer a warning
> Apr 28 18:28:12.362  WARN ibc_relayer::event::rpc: error while building event send_packet.packet_data

but I think this should be expected given that the tx is in the `bank` module. Do I understand correctly?

andynog:
Just tested the workflow in the gist file and had the same results as Adi. Seems to be working fine. Only one comment, in the gist mentioned that the `hermes.sh` file needs to have an executable flag `chmod +x [path]/hermes.sh` (at least I had to do it in my machine Linux based).

andynog (approved):
Looks good to me. Also performed the tests and it worked correctly.

PR #841, opened 2021-04-22
Use rust-cache GitHub Action helper to speed up CI

romac:
https://github.com/Swatinem/rust-cache


romac:
Won't be needed as we are working on versioning the guide itself.

PR #849, opened 2021-04-26
Only deploy the guide when pushing a new tag starting with `v`

romac:


adizere:
If I understand correctly, this avoids useless re-deploying of the guide. It's a good idea in principle. I just wonder: if we want to do a quick fix to the current guide (without doing a `v*` tag), then could we redeploy the guide? We would need to work around the changes introduced by this PR, right? Not sure if it's a problem, but something to keep in mind.

romac:
The idea is to only deploy the guide when we release a new version (ie. we push a tag), so that, until we have proper versioning, we can update the guide as we go along rather than at the end of a release cycle. It's true that this prevents us from pushing fixes to the current version of the guide, and I haven't found a good way of doing that yet, hence why this PR is just a suggestion.

romac:
Won't be needed as we are going to version the guide itself.

PR #850, opened 2021-04-26
Deny uses of `.unwrap()` in non-test code (first pass)

romac:
Closes: #871

- [x] Remove `unwrap` from crate `ibc` (@adizere) done ~50%, rest will be handled later, in #987 
- [x] Remove `unwrap` from crate `ibc_relayer` (@romac) deferred to later, #987


modules/src/application/ics20_fungible_token_transfer/msgs/transfer.rs
```diff
@@ -53,16 +53,25 @@ impl Msg for MsgTransfer {
...
+            source_channel: raw_msg.source_channel.parse().map_err(|e| {
+                Kind::InvalidProtoMsgTransfer(format!("invalid source channel: {}", e))
+            })?,
             token: raw_msg.token,
             sender: raw_msg.sender.into(),
             receiver: raw_msg.receiver.into(),
-            timeout_height: raw_msg.timeout_height.unwrap().try_into().unwrap(),
+            timeout_height: raw_msg
+                .timeout_height
+                .ok_or(Kind::InvalidProtoMsgTransfer(
```
romac, line 71: We should use `ok_or_else` here.

adizere (approved):

Issue #758, opened 2021-03-22
Introduce a newtype wrapper for the timestamp


adizere:
## Crate

`ibc`

## Summary

We could introduce a newtype wrapper for the timestamp, here and in other places where we currently use a `u64`, eg. a `pub struct Timestamp { seconds: u64 }` in the `host` module.

_Originally posted by @romac in https://github.com/informalsystems/ibc-rs/pull/737#discussion_r597109688_

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

soareschen:
To keep a better semantics of tracking the time, I will try to use define `TimeStamp` as a newtype over [`chrono::DateTime`](https://docs.rs/chrono/0.4.19/chrono/struct.DateTime.html).

PR #855, opened 2021-04-26
Use proper Timestamp type to track time

soareschen:
Closes: #758

## Description

This adds a new `Timestamp` newtype that wraps around `Option<DateTime<Utc>>` from the `chrono` trait. This provides a more precise representation of time, with `Option::None` used to denote the absence of a timestamp.

In this PR, we replace the uses of `u64` timestamps with `Timestamp`, using `Timestamp::from_nanoseconds` and assuming that all existing uses of `u64` timestamps all in nanoseconds precision. We do not directly expose conversion functions for other units, such as `seconds`, as they are not currently used in the IBC protocol. When needed, the helper methods can be added, or users can construct a `DateTime` directly and construct a `Timestamp` using `Timestamp::from_datetime`.

We also replace the timestamp comparison operators that check for expiry with `Timestamp::check_expiry`. The function returns an `Expiry` enum three possible results: `InvalidTimestamp` if either timestamps have `None` value, `Expired` if the first timestamp is strictly after the second timestamp, and `NotExpired` if otherwise. When doing the refactoring, we assume that the original use of the expiry comparison do not produce expired-related errors, if either of the timestamps have `0` value.

______

For contributor use:

---
This change is [Reviewable](https://reviewable.io/reviews/informalsystems/ibc-rs/855)

### DIFF
--- a/modules/src/application/ics20_fungible_token_transfer/error.rs
+++ b/modules/src/application/ics20_fungible_token_transfer/error.rs
@@ -23,6 +23,9 @@ pub enum Kind {
         "Destination channel not found in the counterparty of port_id {0} and channel_id {1} "
     )]
     DestinationChannelNotFound(PortId, ChannelId),
+
+    #[error("Invalid packet timeout timestamp value")]
+    InvalidPacketTimestamp,
 }
 
 impl Kind {
--- a/modules/src/application/ics20_fungible_token_transfer/error.rs
+++ b/modules/src/application/ics20_fungible_token_transfer/error.rs
@@ -13,19 +13,19 @@ pub enum Kind {
     #[error("error raised by message handler")]
     HandlerRaisedError,
 
-    #[error("Sending sequence number not found for port {0} and channel {1}")]
+    #[error("sending sequence number not found for port {0} and channel {1}")]
     SequenceSendNotFound(PortId, ChannelId),
 
-    #[error("Missing channel for port_id {0} and channel_id {1} ")]
+    #[error("missing channel for port_id {0} and channel_id {1} ")]
     ChannelNotFound(PortId, ChannelId),
 
     #[error(
-        "Destination channel not found in the counterparty of port_id {0} and channel_id {1} "
+        "destination channel not found in the counterparty of port_id {0} and channel_id {1} "
     )]
     DestinationChannelNotFound(PortId, ChannelId),
 
-    #[error("Invalid packet timeout timestamp value")]
-    InvalidPacketTimestamp,
+    #[error("invalid packet timeout timestamp value")]
+    InvalidPacketTimestamp(u64),
 }
 
 impl Kind {
--- a/modules/src/application/ics20_fungible_token_transfer/relay_application_logic/send_transfer.rs
+++ b/modules/src/application/ics20_fungible_token_transfer/relay_application_logic/send_transfer.rs
@@ -35,6 +35,9 @@ where
             Kind::SequenceSendNotFound(msg.source_port.clone(), msg.source_channel.clone())
         })?;
 
+    let timeout_timestamp = Timestamp::from_nanoseconds(msg.timeout_timestamp)
+        .map_err(|_| Kind::InvalidPacketTimestamp)?;
+
     //TODO: Application LOGIC.
 
     let packet = Packet {
@@ -45,7 +48,7 @@ where
         destination_channel: destination_channel.clone(),
         data: vec![0],
         timeout_height: msg.timeout_height,
-        timeout_timestamp: Timestamp::from_nanoseconds(msg.timeout_timestamp),
+        timeout_timestamp,
     };
 
     let handler_output =
--- a/modules/src/application/ics20_fungible_token_transfer/relay_application_logic/send_transfer.rs
+++ b/modules/src/application/ics20_fungible_token_transfer/relay_application_logic/send_transfer.rs
@@ -5,7 +5,7 @@ use crate::handler::HandlerOutput;
 use crate::ics04_channel::handler::send_packet::send_packet;
 use crate::ics04_channel::packet::Packet;
 use crate::ics04_channel::packet::PacketResult;
-use crate::ics24_host::timestamp::Timestamp;
+use crate::timestamp::Timestamp;
 
 pub(crate) fn send_transfer<Ctx>(
     ctx: &Ctx,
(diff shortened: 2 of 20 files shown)


modules/src/ics04_channel/handler/recv_packet.rs
```diff
@@ -78,7 +79,7 @@ pub fn process(ctx: &dyn ChannelReader, msg: MsgRecvPacket) -> HandlerResult<Pac
 
     // Check if packet timestamp is newer than the local host chain timestamp
     let latest_timestamp = ctx.host_timestamp();
-    if (packet.timeout_timestamp != 0) && (packet.timeout_timestamp <= latest_timestamp) {
+    if let Expiry::Expired = latest_timestamp.check_expiry(&packet.timeout_timestamp) {
```
soareschen, line 82: I am not too sure why there are different comparisons using `<=` or `>` in different places. I try to keep them consistent by always using `>` inside `check_expiry`. It is also not clear whether we need a strict equality here, considering this requires the exact same nanoseconds to satisfy.

cezarad: Where is the equality used ?

cezarad: The replacement with check_expiry are consistent. The >= or < are used interchanged, following the time flow intuition,   and there is an equivalent representation with just one, for which you picked a good name :).

soareschen: The original comparator is `packet.timeout_timestamp <= latest_timestamp`. If we strictly switch the position, it becomes `latest_timestamp >= packet.timeout_timestamp`. But in the new code I am doing `latest_timestamp > packet.timeout_timestamp`.

modules/src/timestamp.rs
```diff
@@ -0,0 +1,179 @@
...
+pub enum Expiry {
+    Expired,
+    NotExpired,
+    InvalidTimestamp,
+}
+
+impl Timestamp {
+    /// When used in IBC, all raw timestamps are represented as u64 Unix timestamp in nanoseconds.
+    ///
+    /// A value is 0 indicates that the timestamp is not set, and result in the underlying
```
ancazamfir, line 39:
```suggestion
    /// A value of 0 indicates that the timestamp is not set, and result in the underlying
```

modules/src/timestamp.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    /// Convert a `Timestamp` to an optional [`chrono::DateTime<Utc>`]
+    pub fn as_datetime(&self) -> Option<DateTime<Utc>> {
+        self.time
+    }
+
+    /// A timestamp is only valid if it is set, i.e. have non-zero raw value.
+    pub fn is_valid(&self) -> bool {
+        self.time.is_some()
+    }
```
ancazamfir, line 77: Should this be removed? Don't think it is used and also `None` is valid, e.g. for packets that are not timeout using the timestamp

modules/src/application/ics20_fungible_token_transfer/error.rs
```diff
@@ -23,6 +23,9 @@ pub enum Kind {
         "Destination channel not found in the counterparty of port_id {0} and channel_id {1} "
     )]
     DestinationChannelNotFound(PortId, ChannelId),
+
+    #[error("Invalid packet timeout timestamp value")]
```
romac, line 27: Could you please change this message and other ones above to start with a lowercase letter?

modules/src/application/ics20_fungible_token_transfer/relay_application_logic/send_transfer.rs
```diff
@@ -34,6 +35,9 @@ where
             Kind::SequenceSendNotFound(msg.source_port.clone(), msg.source_channel.clone())
         })?;
 
+    let timeout_timestamp = Timestamp::from_nanoseconds(msg.timeout_timestamp)
+        .map_err(|_| Kind::InvalidPacketTimestamp)?;
```
romac, line 39: Can we store the invalid timestamp (`u64`) in the `InvalidPacketTimestamp` variant and show it in the error message? This might help debug some issues.

modules/src/ics02_client/error.rs
```diff
@@ -83,6 +80,9 @@ pub enum Kind {
     #[error("invalid proof for the upgraded consensus state")]
     InvalidUpgradeConsensusStateProof(Ics23Kind),
 
+    #[error("Invalid packet timeout timestamp value")]
```
romac, line 83: Can we start the message with a lowercase letter here?

modules/src/ics04_channel/error.rs
```diff
@@ -126,11 +127,14 @@ pub enum Kind {
     PacketTimeoutHeightNotReached(Height, Height),
 
     #[error("Packet timeout timestamp {0} > chain timestamp {1}")]
-    PacketTimeoutTimestampNotReached(u64, u64),
+    PacketTimeoutTimestampNotReached(Timestamp, Timestamp),
 
     #[error("Receiving chain block timestamp >= packet timeout timestamp")]
     LowPacketTimestamp,
 
+    #[error("Invalid packet timeout timestamp value")]
```
romac, line 135: Can we start the message with a lowercase letter here?

modules/src/timestamp.rs
```diff
@@ -0,0 +1,179 @@
...
+/// A newtype wrapper over `Option<DateTime<Utc>>` to keep track of
+/// IBC packet timeout.
+///
+/// We use an explicit `Option` type to distinguish this when converting between
+/// a `u64` value and a raw timestamp. In protocol buffer, the timestamp is
+/// represented as a `u64` Unix timestamp in nanoseconds, with 0 representing the absence
+/// of timestamp.
+#[derive(PartialEq, Eq, Copy, Clone, Debug, Deserialize, Serialize, Hash)]
+pub struct Timestamp {
+    time: Option<DateTime<Utc>>,
```
romac, line 19: Rather than wrap an `Option<DateTime<Utc>>`, how about returning an `Option<Timestamp>` from `from_nanoseconds`? ie.

```rust
pub fn from_nanoseconds(nanoseconds: u64) -> Result<Option<Timestamp>, TryFromIntError> {
``

romac (changes requested):
See comments above.

modules/src/timestamp.rs
soareschen: I did considered that. However that cause issue because we can't derive or implement additional traits on `Option`. So the other alternative is to define a separate `OptionalTimestamp` and `Timestamp` newtypes. But looking around the usage, I think all current use of timestamp need to consider the null value.

romac: Alright, makes sense!

romac (approved):

Issue #847, opened 2021-04-26
Add coin-type for keys restore


allthatjazzleo:
## Crate

 relayer-cli
 relayer

## Summary

Go Relayer support `--coin-type` for `keys restore` command
Currently, it is hardcoded in ibc-rs

## Problem Definition

The key cannot be properly restore for chain with coin-type other than cosmos `118`

## Proposal

add optional flag for `hermes keys restore` to specify coin-type
add optional field in key file for `hermes keys add -f` 

## Acceptance Criteria

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #848, opened 2021-04-26
fix hardcoded coin-type

allthatjazzleo:
Closes: #847

## Description
The key cannot be properly restored for chain with coin-type other than cosmos 118

______

For contributor use:

- [ ] Updated the __Unreleased__ section of [CHANGELOG.md](https://github.com/informalsystems/ibc-rs/blob/master/CHANGELOG.md) with the issue.
- [ ] If applicable: Unit tests written, added test to CI.
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Updated relevant documentation (`docs/`) and code comments.
- [ ] Re-reviewed `Files changed` in the Github PR explorer.

PR #856, opened 2021-04-27
Add `--coin-type` option to `keys restore` command

romac:
Original PR by @allthatjazzleo: #848 

Closes: #847

## Description
The key cannot be properly restored for chain with coin-type other than cosmos 118

______

For contributor use:

### DIFF
--- a/relayer/src/keyring.rs
+++ b/relayer/src/keyring.rs
@@ -342,8 +342,10 @@ fn private_key_from_mnemonic(
     let seed = Seed::new(&mnemonic, "");
 
     // Get Private Key from seed and standard derivation path
-    let hd_path =
-        StandardHDPath::try_from(format!("m/44'/{}'/0'/0/0", coin_type.num()).as_str()).unwrap();
+    let hd_path_format = format!("m/44'/{}'/0'/0/0", coin_type.num());
+    let hd_path = StandardHDPath::try_from(hd_path_format.as_str())
+        .map_err(|_| Kind::InvalidHdPath(hd_path_format))?;
+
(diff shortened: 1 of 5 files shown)


relayer/src/keyring.rs
```diff
@@ -282,14 +318,18 @@ impl KeyRing {
...
+) -> Result<ExtendedPrivKey, Error> {
     let mnemonic = Mnemonic::from_phrase(mnemonic_words, Language::English)
         .map_err(|e| Kind::InvalidMnemonic.context(e))?;
 
     let seed = Seed::new(&mnemonic, "");
 
     // Get Private Key from seed and standard derivation path
-    let hd_path = StandardHDPath::try_from("m/44'/118'/0'/0/0").unwrap();
+    let hd_path =
+        StandardHDPath::try_from(format!("m/44'/{}'/0'/0/0", coin_type.num()).as_str()).unwrap();
```
andynog, line 332: It's better to hand the error here, instead of using unwrap. Since this value is based on user input it opens up for errors.

andynog (commented):
In the future we might need also to allow them to provide the full derivation path. Some addresses are tied to not only a 118 cointype but could also be in path that doesn't end with `0` for example could be `m/44'/118'/0'/0/0`. But for now I think this should be OK

romac:
@andynog Should we also allow specifying the coin type on the command line when adding a key from a keyfile? At the moment, if the keyfile contains a `coin_type` property, it will be picked up otherwise `118` is used.

andynog (approved):
Looks good to me

andynog:
I think this is OK for now, in the future from a CLI perspective I rather do the full HD path, that's what most libs do

ancazamfir (approved):

PR #866, opened 2021-04-28
Add e2e test for relayer loop

romac:
Closes: #708

## Description

Add an e2e test for the relayer loop following the instructions in #708.

______

For contributor use:


e2e/run.py
```diff
@@ -2,17 +2,128 @@
...
+
+    # hermes tx raw ft-transfer ibc-0 ibc-1 transfer channel-0 10000 1000 -n 2
+    packet.packet_send(c, src=IBC_1, dst=IBC_0, src_port=TRANSFER,
+                       src_channel=CHANNEL, amount=10000, height_offset=1000, number_msgs=2)
+
+    # hermes tx raw ft-transfer ibc-1 ibc-0 transfer channel-0 10000 1000 -n 2
+    packet.packet_send(c, src=IBC_0, dst=IBC_1, src_port=TRANSFER,
+                       src_channel=CHANNEL, amount=10000, height_offset=1000, number_msgs=2)
+
+    sleep(5.0)
```
brapse, line 34: Should we create a loop here which queries the chain to confirm that all the packets were sent and avoid the race condition here?

brapse (approved):
🚢

e2e/pyrightconfig.json
```diff
@@ -0,0 +1,2 @@
+{
```
adizere, line 1: Just flagging this in case it was added by mistake.

adizere (approved):

e2e/pyrightconfig.json
romac: It's not a mistake, was actually needed to get `pyright` to work properly on my machine, so figure I would leave it here as it may help others as well.

PR #879, opened 2021-04-30
Solution: use the last digit from chain-id

leejw51crypto:
Closes: #XXX

## Description

______

For contributor use:


modules/src/ics24_host/identifier.rs
```diff
@@ -59,7 +59,7 @@ impl ChainId {
         }
 
         let split: Vec<_> = chain_id.split('-').collect();
-        split[1].parse().unwrap_or(0)
+        split.last().expect("get revision number from chain_id").parse().unwrap_or(0)
```
adizere, line 62: Would it be possible to have a small test for this code? A doc-test for the method `chain_version` could be ideal, just to show the expected behavior.

For example (this code is not tested):

```rust
    /// ```
    /// use ibc::ics24_host::identifier::ChainId;
    ///
    /// assert_eq!(ChainId::chain_version("chain-a-0"), 0);
    /// assert_eq!(ChainId::chain_version("ibc-10"), 10);
    /// assert_eq!(ChainId::chain_version("cosmos-hub---97"), 97);
    /// ```
    pub fn chain_version(chain_id: &str) -> u64 {
        ...
    }
```

adizere (changes requested):

I left one suggestion for improvement. Please take the suggestion into account if you have time (otherwise, no problem, one of the repo owners will try to).

Issue #862, opened 2021-04-28
Single supervisor for multi path relayer


ancazamfir:
## Crate
relayer

## Summary
Use a single supervisor per relayer instance instead of one per connection.

Also simplifies the config content (remove connections & paths sections) (bundle with https://github.com/informalsystems/ibc-rs/issues/867).
Also should simplify a REST API integration.

## Problem Definition
This was the original intent to start with. Also, it will avoid multiple subscriptions and propagation for identical events.
I think it will also be easier when interacting with the rest server in the future, maintaining configuration, caching state, etc.

## Acceptance Criteria

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #872, opened 2021-04-29
Add client workers for active channels, clear packets on start

ancazamfir:
Closes: #786 
Closes: #784

## Description
tbd

______

For contributor use:

- [x] Updated the __Unreleased__ section of [CHANGELOG.md](https://github.com/informalsystems/ibc-rs/blob/master/CHANGELOG.md) with the issue.
- [ ] If applicable: Unit tests written, added test to CI.
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Updated relevant documentation (`docs/`) and code comments.
- [ ] Re-reviewed `Files changed` in the Github PR explorer.

Issue #712, opened 2021-02-25
Relayer error handling specification


ancazamfir:
## Crate
docs

## Summary
Currently the relayer performs retries for any error that occurs while building messages and transactions. 

## Problem Definition
There are RPC, light client, on-chain, internal errors, etc. The most complex ones are the on-chain errors that are a bit more subtle (leaving aside errors in the relayer code) and also they are not typed (we just get an error message and the reason is embedded in a string). The retry may apply for some cases, in others we should not retry (e.g. a transaction has already been submitted by another relayer), and yet in other cases we should rebuild stuff, etc.

## Proposal
Analyze the code, document all errors and the actions to be taken in each case. 
Consider also the exponential backoff as per @romac's  https://github.com/informalsystems/ibc-rs/pull/709#discussion_r582852413, where applicable.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ebuchman:
> also they are not typed (we just get an error message and the reason is embedded in a string)

May be worth trying to get these typed upstream in ibc

soareschen:
I tried to do a full trace of an example error message to see how it is propagated. There is a complicated path from the Hermes relayer to the IBC module, and there are many potential errors that can occur in each step. 

On the Go side, the error messages are formatted using helpers such as `sdkerrors.Wrapf` that are converted into strings. It would be challenging to modify the Go code to use more structured logging, so we can only do pattern matching on the string to find the type of error that are of our interest.

Here is an example error message:

> Error: tx error: failed with underlying cause: tx response error: deliver_tx reports error: log=Log("failed to execute message; message index: 1: connection handshake open try failed: failed connection state verification for client (07-tendermint-4): chained membership proof failed to verify membership of value: 0A0F... in subroot C9FE...at index 0. Please ensure the path and value are both correct.: invalid proof")

And here is the manual stack trace:

- `tx error` - [ibc-rs/relayer-cli/src/error.rs:33](https://github.com/informalsystems/ibc-rs/blob/9b3d1a344f1a0e179ef42b48cffba18e593243ee/relayer-cli/src/error.rs#L33)
[...]

PR #881, opened 2021-04-30
Use a single supervisor to subscribe to all configured chains

romac:
Closes: #862 

## Description

- [ ] update guide for config.toml content
- [x] apply any outstanding comments that Romain left in [this PR review](https://github.com/informalsystems/ibc-rs/pull/872#pullrequestreview-651063120)
- [x] test w/ 3 chains
- [x] test `start` should still work (using the `connections` config)
- [x] test with/out packet delay feature

______

For contributor use:


relayer/src/supervisor.rs
```diff
@@ -195,70 +205,128 @@ impl Supervisor {
...
+                return Ok(());
+            }
+            Err(e) => {
+                // Propagate errors.
+                return Err(format!(
+                    "unable to spawn workers for channel/chain pair '{}'/'{}' due to error: {:?}",
+                    channel.channel_id,
+                    chain.id(),
+                    e
+                )
```
ancazamfir, line 265: I think we should `error!` here or in the caller and continue best effort to create workers and relay with those. The way it is the supervisor exits.

romac: Yeah we should handle this error in the caller.

adizere: The only possible error that we propagate here is currently `Error::QueryFailed`. This can have one of three causes:

1. `query_channel` failed, or
2. `query_connection` failed, or
3. `query_client_state` failed.

I think all these failure cases should be handled with a retry inside method `channel_connection_client` (or deeper in the stack). And `channel_connection_client` should return `Error::QueryFailed` only if the number of retries was exhausted.

Until we have the retry-based fix (https://github.com/informalsystems/ibc-rs/issues/712), I've added code to handle the propagated error in the caller -- 2373a61a. Let me know if this is good. It should be safe for the supervisor to continue (because returning an error to the supervisor/caller means no single worker was spawned for the given channel, so there won't be some partially incomplete state left behind).

ancazamfir: looks good. in the future, even with retries, we probably don't want the supervisor to exit on the failures ^

ancazamfir (approved):
Yes!!! Very cool! Thanks @romac and @adizere!!

PR #899, opened 2021-05-05
Don't handle optimistic sends, don't exit if chain doesn't include header in updates

ancazamfir:
Closes: #XXX

## Description
Some leftover work from #872
- exit client worker only if expired or frozen
- don't handle optimistic sends
- start client worker when first channel is opened and not on update client event. The worker still receives the update client events and checks for misbehaviour

______

For contributor use:

### DIFF
--- a/relayer/src/foreign_client.rs
+++ b/relayer/src/foreign_client.rs
@@ -50,6 +50,9 @@ pub enum ForeignClientError {
     #[error("failed while finding client {0}: expected chain_id in client state: {1}; actual chain_id: {2}")]
     ClientFind(ClientId, ChainId, ChainId),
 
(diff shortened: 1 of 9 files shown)


relayer/src/foreign_client.rs
```diff
@@ -868,25 +874,39 @@ impl ForeignClient {
...
+                } else {
+                    warn!("[{}] misbehaviour checking result {:?}", self, e);
+                    MisbehaviourResults::ValidClient
+                }
             }
         }
     }
 }
 
+#[derive(Clone, Debug)]
```
romac, line 902: 💯

romac (approved):

romac (approved):

Issue #871, opened 2021-04-29
Deny uses of `unwrap` in non-test code (first pass)


adizere:
## Crate

all

## Acceptance Criteria

Enable `#![cfg_attr(not(test), deny(clippy::unwrap_used))]`

____

## For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #903, opened 2021-05-06
Make supervisor more resilient to node going down

romac:
Follow-up to #895
See also #871

## Description

The supervisor should now be more resilient to a node going down temporarily.
Instead of sitting there waiting for events via the subscription, the supervisor is
now notified that something went wrong, while the event monitor will attempt to
reconnect for a limited time (max retries with a delay between attempts).

Errors yielded by the client and packet workers are now caught at the top-level run loop of the worker,
and printed to the console rather than causing the worker to exits. This is quite brittle still
and will need more work and thought put into for the next milestone.

## Tested with

1. setup the chains, clients, connections, and channels:

```
❯ ./scripts/dev-env ibc-0 ibc-1 ibc-2
❯ hermes create channel ibc-0 ibc-1 --port-a transfer --port-b transfer -o unordered
❯ hermes create channel ibc-1 ibc-2 --port-a transfer --port-b transfer -o unordered
```

2. in a new console, with the debug log level in the config:

```
❯ hermes start-multi
```

3. send some packets if you want:

```
❯ hermes tx raw ft-transfer ibc-1 ibc-0 transfer channel-0 9999 1000 -n 5
...
```

4. kill one of the node, eg. `ibc-1`:

```
❯ ps aux | rg gaiad | rg ibc-1 | awk '{ print $2 }' | xargs -I{} kill -9 {}
```

5. watch the output of `start-multi`, it should show some errors about the WebSocket connection being down and perhaps some RPC queries failing but keep going and retrying to connect to the WebSocket.

6. start the nodes again (will automatically kill the remaining ones):

```
❯ ./scripts/dev-env ibc-0 ibc-1 ibc-2
❯ hermes create channel ibc-0 ibc-1 --port-a transfer --port-b transfer -o unordered
❯ hermes create channel ibc-1 ibc-2 --port-a transfer --port-b transfer -o unordered
```

7. after a while it should be able to reconnect and you can send some packets

```
❯ hermes tx raw ft-transfer ibc-1 ibc-0 transfer channel-0 9999 1000 -n 5
```

______

For contributor use:

### DIFF
--- a/relayer/src/chain/handle.rs
+++ b/relayer/src/chain/handle.rs
@@ -34,13 +34,16 @@ use ibc_proto::ibc::core::client::v1::QueryConsensusStatesRequest;
 use ibc_proto::ibc::core::commitment::v1::MerkleProof;
 pub use prod::ProdChainHandle;
 
-use crate::connection::ConnectionMsgType;
-use crate::keyring::KeyEntry;
-use crate::{error::Error, event::monitor::EventBatch};
+use crate::{
+    connection::ConnectionMsgType,
+    error::Error,
+    event::monitor::{Error as EventMonitorError, EventBatch},
+    keyring::KeyEntry,
+};
 
 mod prod;
 
-pub type Subscription = channel::Receiver<Arc<EventBatch>>;
+pub type Subscription = channel::Receiver<Arc<Result<EventBatch, EventMonitorError>>>;
 
 pub type ReplyTo<T> = channel::Sender<Result<T, Error>>;
 pub type Reply<T> = channel::Receiver<Result<T, Error>>;
--- a/relayer/src/chain/handle.rs
+++ b/relayer/src/chain/handle.rs
@@ -37,13 +37,13 @@ pub use prod::ProdChainHandle;
 use crate::{
     connection::ConnectionMsgType,
     error::Error,
-    event::monitor::{Error as EventMonitorError, EventBatch},
+    event::monitor::{EventBatch, Result as MonitorResult},
     keyring::KeyEntry,
 };
 
 mod prod;
 
-pub type Subscription = channel::Receiver<Arc<Result<EventBatch, EventMonitorError>>>;
+pub type Subscription = channel::Receiver<Arc<MonitorResult<EventBatch>>>;
 
 pub type ReplyTo<T> = channel::Sender<Result<T, Error>>;
 pub type Reply<T> = channel::Receiver<Result<T, Error>>;
--- a/relayer/src/event/monitor.rs
+++ b/relayer/src/event/monitor.rs
@@ -4,6 +4,7 @@ use crossbeam_channel as channel;
 use futures::stream::StreamExt;
 use futures::{stream::select_all, Stream};
 use itertools::Itertools;
+use retry::delay::Exponential;
 use thiserror::Error;
 use tokio::task::JoinHandle;
 use tokio::{runtime::Runtime as TokioRuntime, sync::mpsc};
@@ -18,11 +19,11 @@ use tendermint_rpc::{
 
 use ibc::{events::IbcEvent, ics02_client::height::Height, ics24_host::identifier::ChainId};
 
+use crate::util::retry::Clamped;
+
 const MAX_RETRIES: usize = 1000;
 const MAX_RETRY_DELAY: Duration = Duration::from_secs(5 * 60);
 const INITIAL_RETRY_DELAY: Duration = Duration::from_secs(2);
-const DEFAULT_RETRY_STRATEGY: RetryStrategy =
-    RetryStrategy::new(INITIAL_RETRY_DELAY, MAX_RETRY_DELAY, MAX_RETRIES);
 
 #[derive(Debug, Clone, Error)]
 pub enum Error {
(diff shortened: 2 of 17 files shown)


relayer/src/event/monitor.rs
```diff
@@ -81,37 +105,51 @@ pub struct EventMonitor {
     rt: Arc<TokioRuntime>,
 }
 
+async fn run_driver(
+    driver: WebSocketClientDriver,
+    tx: mpsc::UnboundedSender<tendermint_rpc::Error>,
+) {
+    if let Err(e) = driver.run().await {
+        if tx.send(e).is_err() {
+            println!("failed to relay driver error to event monitor");
```
adizere, line 114: I don't see where is the EventMonitor consumes (or handles) the errors which this method `send`s here.

adizere: Also, is it better to use `error!` (which prints to stderr) here instead of `println!`?

relayer/src/chain/handle.rs
```diff
@@ -34,13 +34,16 @@ use ibc_proto::ibc::core::client::v1::QueryConsensusStatesRequest;
...
+    connection::ConnectionMsgType,
+    error::Error,
+    event::monitor::{Error as EventMonitorError, EventBatch},
+    keyring::KeyEntry,
+};
 
 mod prod;
 
-pub type Subscription = channel::Receiver<Arc<EventBatch>>;
+pub type Subscription = channel::Receiver<Arc<Result<EventBatch, EventMonitorError>>>;
```
adizere, line 46: Would it be good to make `Arc<Result<EventBatch, EventMonitorError>>` a type of its own? Even better might be to have `Result<EventBatch, EventMonitorError>` a type of its own, since this has more potential for reuse. Something like `put type MonitorResult = Result<EventBatch, EventMonitorError>`. (Feel free to ignore the comment if it is more idiomatic to retain the `Result` as it is.)

I found the modifications to the monitor very difficult to understand, mostly because of the complex interactions between Monitor, Runtime, and Handler.
There are already quite a few types at the interface with the event monitor. I noticed the following:

- in file handle.rs:
    - `pub type Subscription = channel::Receiver<Arc<Result<EventBatch, EventMonitorError>>>;`
- in file monitor.rs:
    - `type SubscriptionResult = RpcResult<RpcEvent>;`
    - `type SubscriptionStream = dyn Stream<Item = SubscriptionResult> + Send + Sync + Unpin;`
    - `pub type Result<T> = std::result::Result<T, Error>;`
    - `pub type EventReceiver = channel::Receiver<Result<EventBatch>>;`


We can ignore `SubscriptionResult` and `SubscriptionStream` since these seem internal to the monitor.
If I understand correctly, an oversimplification of the flow of events flow is like this:

```

| Monitor |
     |---- Result<EventBatch, EventMonitorError> ---> | Runtime | 
                                                           |
                                                           |
                                   Arc<Result<EventBatch, EventMonitorError> 
                                                           |
                                                           |----> | Handler |

```

An ADR would be nice.

relayer/src/event/monitor.rs
romac: > Also, is it better to use error! (which prints to stderr) here instead of println!?
 
Right, that's a leftover from my spike which didn't use tracing.

romac: > I don't see where is the EventMonitor consumes (or handles) the errors which this method sends here.

https://github.com/informalsystems/ibc-rs/blob/31c992a46e40f11e0dfb6044f5520fdd0ca4bb3d/relayer/src/event/monitor.rs#L272

relayer/src/chain/handle.rs
romac: This should have read:

```rust
pub type Subscription = channel::Receiver<Arc<Result<EventBatch>>;
```

which is a bit more digestible imho. But yes we could go further and make an alias for `Result<EventBatch>`.

At the same time, all these aliases make me think that there maybe something wrong with the design, and I would like to refactor the whole thing anyway.

> Something like put type MonitorResult = Result<EventBatch, EventMonitorError>. 

I think internally it's customary to stick with `Result` and only rename it when imported in another module, but we could still do that.

> flow of events

At the moment there is still an event bus between the runtime and the supervisor, but we could maybe get rid of it (and the `Arc` wrapper) once we only have a single supervisor. We might need those events in other places so it's not clear if we can actually remove the event bus. To be investigated.

ancazamfir:
I still see an error when starting with 3 chains in config but only two gaia processes. The output is:
```
May 06 15:22:30.864  INFO ibc_relayer_cli::commands: Using default configuration from: '.hermes/config.toml'
May 06 15:22:30.879 DEBUG ibc_relayer::event::monitor: subscribing to query: tm.event = 'Tx'
May 06 15:22:30.880 DEBUG ibc_relayer::event::monitor: subscribing to query: tm.event = 'NewBlock'
May 06 15:22:30.881 DEBUG ibc_relayer::event::monitor: subscribed to all queries
May 06 15:22:30.881  INFO ibc_relayer::event::monitor: starting event monitor chain.id=ibc-0
May 06 15:22:30.881 TRACE ibc_relayer::registry: spawned chain runtime for chain identifier ibc-0
May 06 15:22:30.885 DEBUG ibc_relayer::event::monitor: subscribing to query: tm.event = 'Tx'
May 06 15:22:30.886 DEBUG ibc_relayer::event::monitor: subscribing to query: tm.event = 'NewBlock'
May 06 15:22:30.887 DEBUG ibc_relayer::event::monitor: subscribed to all queries
May 06 15:22:30.887 TRACE ibc_relayer::registry: spawned chain runtime for chain identifier ibc-1
May 06 15:22:30.887  INFO ibc_relayer::event::monitor: starting event monitor chain.id=ibc-1
Error: RPC error to endpoint http://127.0.0.1:26457/: error trying to connect: tcp connect error: Connection refused (os error 61) (code: 0)
```
I believe the error comes from here (see `?`):
https://github.com/informalsystems/ibc-rs/blob/ecbdb07a8bbed69e9cd9f88778a4b50edf0f5567/relayer/src/supervisor.rs#L332-L334

We should ignore the error...maybe create subscription as part of `spawn_workers()` below since we iter the chains there anyway.

relayer/src/event/monitor.rs
romac: d75bc22d

romac:
@ancazamfir Should be fixed in https://github.com/informalsystems/ibc-rs/pull/903/commits/493191b7c549871ad46f87b4006820d8b48008a4.

relayer/src/supervisor.rs
```diff
@@ -330,21 +335,42 @@ impl Supervisor {
...
-                }
-                Err(e) => {
-                    dbg!(e);
+                    let result = batch
+                        .unwrap_or_clone()
+                        .map_err(Into::into)
+                        .and_then(|batch| self.process_batch(chain.clone(), batch));
+
+                    match result {
+                        Ok(()) => trace!("[{}] batch processing complete", chain.id()),
```
ancazamfir, line 369: can we move this trace and we can figure it out later, we only want it for ibc events and not for new blocks. This is printed every block and even with 2 chains it makes the logs very verbose. I know it's tracing but still.

romac: Done

ancazamfir:
Another nice to have is this: If there are 3 chains in config and only two chains/ gaiad nodes are up then I should see the same behavior regardless on how this state was reached. Right now:
- when hermes starts with 2 nodes up and 3 chains, it tries to init 3rd chain and then it relays only for the two chains. It doesn't retry to reconnect with the third chain.
- when hermes starts with 3 nodes and then one is killed, it relays over the two chains but keeps retrying to reconnect with the third.

ancazamfir:
Also the `trying to reconnect to WebSocket end...` and `error when reconnecting` are too often maybe especially because they are warn/ error and happen every 5 sec (exponential backoff would be nice here) or maybe give up after a while.

relayer/src/chain/handle.rs
adizere: Thanks for thorough answers. ⬆️ 
Agree with everything, then let's keep the types as they are (except for the low-hanging fruit of `Subscription`), and seems like the monitor is a good target for refactoring sometime soon. Would like to get a chunk of that!

adizere (approved):
I think we should merge this.

Will open a follow-up PR to address [Anca's comment](https://github.com/informalsystems/ibc-rs/pull/903#issuecomment-833566686) regarding the backoff for the worker (`error when reconnecting`).

ancazamfir (approved):

Issue #930, opened 2021-05-12
Evaluate use of Tx simulate to determine the fees


ancazamfir:
## Crate
relayer

## Summary
Big Tx-es (e.g. including 30 msgs which is the default maximum) are sometimes rejected with `out of gas` error. 

## Problem Definition
We currently use the `gas` and `fee_amount` in the config file when sending a Tx. The Tx is sometimes rejected with `out of gas` error. Current workaround is to increase the value in the config file.
Instead we should evaluate the gas/ fee parameters at the moment we send the Tx.

## Proposal
@andynog made this analysis earlier:
The REST API has a `/cosmos/tx/v1beta1/simulate` endpoint that you can submit a tx and it returns the gas used, you can craft a 'fake' transaction e.g. MsgSend but it does enforce a few validation (address needs to exist on chain, account sequence number) and it will return something like:
```
{
  "gas_info": {
    "gas_wanted": "0",
    "gas_used": "50682"
  },
  "result": {
    "data": "CgYKBHNlbmQ=",
```
and this just simulates and doesn't execute the transaction. Anyways, this is probably the best way found so far to find how much gas a tx will use.

## Acceptance Criteria

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #945, opened 2021-05-14
Determine the Tx gas that would be required before sending the Tx

ancazamfir:
Closes: #930

## Description
- add gRPC query to simulate Tx
- if `gas_used` is higher than the config `gas` then do not sent the Tx

______

For contributor use:


ancazamfir:
Incorporated in #1091

PR #952, opened 2021-05-18
Split workers into their own module and struct

romac:
## Description

This PR splits out each worker into their own module and struct, rather than having them as methods of a single `Worker` struct.

- A [`Worker`](https://github.com/informalsystems/ibc-rs/blob/32a417786f7b26365d7a173cf52249a51de163d7/relayer/src/worker.rs#L30-L34)  enum now describes the different types of workers.
- Introduce a [`Client`](https://github.com/informalsystems/ibc-rs/blob/romac/worker-refactor/relayer/src/worker/client.rs) worker struct and a [`UniChanPath`](https://github.com/informalsystems/ibc-rs/blob/romac/worker-refactor/relayer/src/worker/uni_chan_path.rs) worker struct.
- Splits out [`WorkerCmd`](https://github.com/informalsystems/ibc-rs/blob/romac/worker-refactor/relayer/src/worker/cmd.rs) and  [`WorkerHandle`](https://github.com/informalsystems/ibc-rs/blob/romac/worker-refactor/relayer/src/worker/handle.rs) into their own module.
- Introduces a [`WorkerMap`](https://github.com/informalsystems/ibc-rs/blob/romac/worker-refactor/relayer/src/worker/map.rs) to encapsulate the lifecycle management of the workers.

This PR otherwise does not introduce any functional change to the relayer.


adizere (approved):
Very nice!

Issue #843, opened 2021-04-23
Hermes management, API specification and implementation (bootstrap)


ancazamfir:
## Crate
relayer

## Summary
Define the architecture for hermes operation management and implement it. Specify the initial APIs and implement a demo sample.

## Problem Definition
Current hermes operation is defined via the configuration file loaded at startup, cannot be changed programmatically or reloaded dynamically.

## Proposal
- [ ] Write a specification for the required APIs, including:
    - query, add, delete chain configuration
    - query, add, delete(or disable) relaying policy elements (e.g. “relay between chains A and B”, “relay on connection X for chain A”, etc.). This is evolving but at minimum specify the ones currently supported.
    - all queries currently supported by the CLIs: keys, clients, connections, channels, packets
    - all operations supported by the CLIs (mainly create/ update client, connections and channels)

- [ ] Decide on a client/ server architecture, be it REST or other, based on the above.
- [ ] Server implementation and the APIs for query, create and update client.

## Acceptance Criteria
Ability to interact with hermes via the agreed APIs and architecture to query, create, update clients.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

andynog:
I can work with @romac and others to evaluate from the technical side what's the best framework/crate to use for a REST API if we're going with that.

PR #955, opened 2021-05-19
RESTful API

greg-szabo:
Closes: #843

## Description

First stab at a RESTful API with basic functionality:

Endpoints:

- `curl -s localhost:8080/state -X GET | jq .`
- `curl -s localhost:8080/chain -X GET | jq .`
- `curl -s localhost:8080/chain/ibc-0 -X GET | jq .`
- `curl -s localhost:8080/ -X GET | jq .`
- **Not implemented yet**: `curl -s localhost:8080/chain -X POST ... `
    - postponed to later PR, as the need arises (see last todo below)

TODOs:

- [x] config inline doc for the rest address
    - if `rest_addr = '127.0.0.1:8080'` then the REST API starts; if `rest_addr = ''` or option not set in config file, the REST interface will not start
- [x] changelog
- [x] unit testing for `rouille`
- [x] ~~guide update~~ (We do this when we release new Hermes version)
- [x] `curl -s localhost:8080/reload -X POST` to instruct Hermes to reload its configuration file
    - ~~unclear if a useful requirement, to discuss with Cephalopod and Gravity Dex~~
    - Based on discussion with Gravity DeX team, it is unclear which of the two endpoints are preferable `/reload` vs. `/chain -X POST` we will handle that in a separate PR later

### OUTDATED: original PR description:

Features:
* ibc-relayer-rest: rouille synchronous web server in the new ibc-relayer-rest crate
* ibc-relayer: Messages and processing for the relayer library in the rest.rs module
* ibc-relayer-cli: Implemented the above two in the start-multi command
* New `[global]` config parameter: `rest_addr = '0.0.0.0:9999'` (mandatory)
* `/` endpoint returns the relayer library version. (Just so we know the server works.)
* `GET /chain` endpoint lists the names of all chain IDs in the configuration
* `GET /chain/{id}` endpoint returns the configuration details of the specified chain ID
* `POST /chain` adds a new chain to the configuration. Returns the chain configuration on success.
* This last endpoint doesn't actually add the chain to the configuration, but there is a placeholder "TODO" where that logic goes since Romain is already working on that.
* Some debug log was introduced in the relayer library to see when a request arrives.

Todo:
* Testing: Research how to unit test the rouille web server.
* Documentation: everything...

Edit: POST example in bash:
```bash
curl -s localhost:9999/chain -X POST -H "Content-Type: application/json" -d '{"id":"hello-1","rpc_addr":"http://localhost:26657","websocket_addr":"http://localhost:26657/ws","grpc_addr":"http://localhost:9090","account_prefix":"cosmos1","key_name":"wallet","store_prefix":"yolo","fee_denom":"stake","gas":1}'
```
______

For contributor use:


adizere (commented):
Cool stuff, thanks Greg!

I'll try to merge master in this branch directly and fix the conflicts, then add the code for the `POST /chain` endpoint.

greg-szabo:
This looks good to me. I'd like to run it locally and do some tests but I won't have the time to do that before the Tuesday IBC meeting. If it's merged by then, I'll do tests afterwards.

romac:
>  - config inline doc for the rest address
>     * if rest_addr = '127.0.0.1:8080' then the REST API starts; if rest_addr = '' or option not set in config file, the REST interface will not start

How about we instead add a section similar to the `telemetry` section?

```toml
[rest-server]
enabled = true
host = '127.0.0.1'
port = 9321
```

adizere:
> How about we instead add a section similar to the `telemetry` section?

This makes sense. It's a bit verbose but probably the best option!

relayer-rest/src/server.rs
```diff
@@ -0,0 +1,52 @@
...
+                trace!("[rest/server] GET /");
+                let result = assemble_version_info(&sender);
+                rouille::Response::json(&result)
+            },
+
+            (GET) (/chain) => {
+                // TODO(Soares): Add a `into_detail` to consume the error and obtain
+                //   the underlying detail, so that we avoid doing `e.0`
+                trace!("[rest/server] GET /chain");
+                rouille::Response::json(&all_chain_ids(&sender).map_err(|e| e.0))
```
adizere, line 36: @soareschen is this feasible with flex-error? We'd like to be able to do something as follows:

```suggestion
                trace!("[rest/server] GET /chain");
                rouille::Response::json(&all_chain_ids(&sender).map_err(|e| e.into_detail()))
```

The `into_detail` method should convert the error into the underlying `RestApiError` type (which derives `Serialize`) so we can output it in JSON format.

relayer/src/supervisor.rs
```diff
@@ -69,7 +70,11 @@ pub struct Supervisor {
 
 impl Supervisor {
     /// Create a [`Supervisor`] which will listen for events on all the chains in the [`Config`].
-    pub fn new(config: RwArc<Config>, telemetry: Telemetry) -> (Self, Sender<SupervisorCmd>) {
+    pub fn new(
+        config: RwArc<Config>,
+        rest_receiver: Option<rest::Receiver>,
```
romac, line 75:
```suggestion
        rest_rx: Option<rest::Receiver>,
```

relayer/src/supervisor.rs
```diff
@@ -61,6 +61,7 @@ pub struct Supervisor {
 
     cmd_rx: Receiver<SupervisorCmd>,
     worker_msg_rx: Receiver<WorkerMsg>,
+    rest_request_rx: Option<rest::Receiver>,
```
romac, line 64:
```suggestion
    rest_rx: Option<rest::Receiver>,
```

romac (commented):
Couple minor nitpicks wrt naming.

I would also suggest we rename the crate to `ibc-relayer-rpc` as the crate implements a way of doing RPC for Hermes, and the fact that we use REST is merely an implementation detail I would say. What do you think?

relayer-rest/src/server.rs
soareschen: Yes, we can include that in the macro generation: similar to the `detail()` method, which returns a reference: https://github.com/informalsystems/flex-error/blob/master/flex-error/src/macros.rs#L515-L517

adizere: Sounds good, we'll keep this in mind for later. Thanks Soares!

greg-szabo:
Regarding REST vs RPC:
Funny you mention that, I was torn between naming for a while. RESTful APIs cater really well to databases with their CRUD approach and to the DevOps community with their statelessness.
RPC APIs are fairly broad in definition and can mean anything, including stateful websockets. If we plan on adding gRPC or websockets, sure, this is an RPC API.

But I wanted to tinker with the RESTful paradigm to see if it can apply to our use-case. So far it worked but that might just be an accident.

relayer/src/config.rs
```diff
@@ -110,6 +110,8 @@ pub struct Config {
     #[serde(default)]
     pub global: GlobalConfig,
     #[serde(default)]
+    pub rest: RESTConfig,
```
adizere, line 113: Maybe we we can follow Rust conventions and call it `RestConfig` instead of `RESTConfig`,

config.toml
```diff
@@ -47,6 +44,22 @@ host = '127.0.0.1'
...
+
+# Specify the IPv4/6 host over which the built-in HTTP server will serve the restful
+# API requests. Default: 127.0.0.1
+host = '127.0.0.1'
+
+# Specify the port over which the built-in HTTP server will serve the restful API
+# requests. Default: 3000
+port = 3000
+
+
```
adizere, line 62: Two questions:

1. would it be more consistent to move the whole `rest` section above the `telemetry` section? Partly because of the port (3000) which is before the telemetry port (3001), partly because that's the order chosen in the `pub struct Config`.

2. How will we document the REST API in https://hermes.informal.systems/rest.html? Is there an automated way of doing that? (Note we do not want to create the page https://hermes.informal.systems/rest.html yet, we'll create that once we release next ver. of Hermes. Just asking to understand the complexity of documenting the REST API.).

romac: 1. 👍 
2. I don't think `rouille` provides a way of doing that, so I would suggest doing it by hand for now and if the API either changes to often or its surface increases too much we can revisit this.

relayer/src/config.rs
romac: https://github.com/informalsystems/ibc-rs/pull/955/commits/555469eff1086e0541fb8550f9cad6ec0e4534f7

config.toml
romac: 1. https://github.com/informalsystems/ibc-rs/pull/955/commits/5bafaf302febb6f172242be2b5e6b8beb3c4ec4e

romac:
This is ready to merge imho.

adizere (approved):
Agree, LGTM!

PR #979, opened 2021-05-23
update guide for connection with existing channel

ebuchman:
## Description

- update `guide/README` Pre-requisites to include `mdbook-toc`
- Add `New connection over existing clients` section to `Connection` guide

Maybe we should also update the `CLI` help menu to indicate that the second argument is optional ? It's probably best if the CLI help is as self-describing as possible to reduce dependency on (and maintenance burden of) commands documentation.

Tbh I'm not sure this optional second chain id is worth it. While it's cool, it doesn't save much, and given the need to explain it, it's not clear that it's worth the savings. Not a big deal though for now since the documentation now explains it pretty well :)

______

For contributor use:

### DIFF
--- a/guide/README.md
+++ b/guide/README.md
@@ -27,10 +27,11 @@ Install `mdBook` using [`cargo`](https://doc.rust-lang.org/cargo/):
 cargo install mdbook
 ```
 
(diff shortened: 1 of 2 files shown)


ancazamfir (approved):
Looks great, thanks for catching this @ebuchman!

Issue #989, opened 2021-05-25
Refactor retry logic in relayer to use retry crate


soareschen:
## Crate

`relayer`

## Summary

This is a spin off from #712 as the scope for the original issue is too large. This issue focus only on refactoring the ad hoc retry logic that make use of `MAX_RETRIES` and for loop, and refactor the code to use the `retry` crate and the Fibonacci retry strategy.

The refactoring will take place in multiple PRs, as there are quite a few separate places that require the refactoring.

This issue does _not_ cover the classification of errors to determine whether it is retryable or not. This is tracked by the original issue #712.

## TODO

- [x] `ibc_relayer::channel`
  - [x] https://github.com/informalsystems/ibc-rs/blob/e6e23b0f34259b9e51f52583fc2ca718d5d369c6/relayer/src/channel.rs#L341
  - [x] https://github.com/informalsystems/ibc-rs/blob/e6e23b0f34259b9e51f52583fc2ca718d5d369c6/relayer/src/channel.rs#L368
  - [x] https://github.com/informalsystems/ibc-rs/blob/e6e23b0f34259b9e51f52583fc2ca718d5d369c6/relayer/src/channel.rs#L392
- [ ] `ibc_relayer::connection`
  - [ ] https://github.com/informalsystems/ibc-rs/blob/1f498e3cd40b6e3cc5602f4aa5ffd2560581edf3/relayer/src/connection.rs#L256
  - [ ] https://github.com/informalsystems/ibc-rs/blob/1f498e3cd40b6e3cc5602f4aa5ffd2560581edf3/relayer/src/connection.rs#L273
  - [ ] https://github.com/informalsystems/ibc-rs/blob/1f498e3cd40b6e3cc5602f4aa5ffd2560581edf3/relayer/src/connection.rs#L289
- [ ] `ibc_relayer::link`
  - [ ] https://github.com/informalsystems/ibc-rs/blob/a1e95e1443a97f5db44bf3fd3fd53a2164e587b1/relayer/src/link.rs#L398
  - [ ] https://github.com/informalsystems/ibc-rs/blob/a1e95e1443a97f5db44bf3fd3fd53a2164e587b1/relayer/src/link.rs#L596
  - [ ] https://github.com/informalsystems/ibc-rs/blob/a1e95e1443a97f5db44bf3fd3fd53a2164e587b1/relayer/src/link.rs#L815
[...]

PR #990, opened 2021-05-25
Refactor retry logic in relayer::channel::handshake

soareschen:
Partially Fixes:  #989

## Description

This PR does a single refactoring of the retry logic in `relayer::channel::handshake`. The PR do not handle the logic of classifying errors to determine whether the channel handshake should be retried for the following reasons:

- The `ChannelError` type do not provide enough static information to analyze whether an error during the handshake is recoverable. To properly classify the error, we will need to redesign `ChannelError` to provide more details.

- The original retry logic is pretty complicated, and adding more changes will make the PR more difficult to review. Therefore it is better to do the retry classification in a separate PR.

______

For contributor use:

### DIFF
--- a/relayer/src/channel.rs
+++ b/relayer/src/channel.rs
@@ -4,7 +4,7 @@ use prost_types::Any;
 use serde::Serialize;
 use std::time::Duration;
 use thiserror::Error;
-use tracing::{info, debug, error};
+use tracing::{debug, error, info};
 
 use ibc::events::IbcEvent;
 use ibc::ics04_channel::channel::{ChannelEnd, Counterparty, IdentifiedChannelEnd, Order, State};
--- a/relayer/src/channel.rs
+++ b/relayer/src/channel.rs
@@ -4,7 +4,7 @@ use prost_types::Any;
 use serde::Serialize;
 use std::time::Duration;
 use thiserror::Error;
-use tracing::{debug, error};
+use tracing::{info, debug, error};
 
 use ibc::events::IbcEvent;
 use ibc::ics04_channel::channel::{ChannelEnd, Counterparty, IdentifiedChannelEnd, Order, State};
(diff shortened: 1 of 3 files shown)


relayer/src/channel.rs
```diff
@@ -24,7 +24,22 @@ use crate::connection::Connection;
...
+
+mod retry_strategy {
+    use crate::util::retry::clamp_total;
+    use retry::delay::Fibonacci;
+    use std::time::Duration;
+
+    // Default parameters for the retrying mechanism
+    const MAX_DELAY: Duration = Duration::from_secs(60); // 1 minute
+    const MAX_TOTAL_DELAY: Duration = Duration::from_secs(10 * 60); // 10 minutes
+    const INITIAL_DELAY: Duration = Duration::from_secs(1); // 1 second
```
soareschen, line 37: I am making a copy of `retry_strategy` from `monitor.rs`, as I suspect we may need to change the delay parameters to a smaller duration?

romac: Yes for now it's better to have each module define its own retry strategy until we figure out exactly what are the best parameters. At some we will maybe want to update those dynamically depending on the error rate, block time, etc.

relayer/src/util/retry.rs
```diff
@@ -2,9 +2,33 @@ use std::time::Duration;
...
+impl<T, E> ToRetryResult<T, E> for Result<T, E> {
+    fn err_to_retry_result(self, cont: impl FnOnce(E) -> RetryResult<T, E>) -> RetryResult<T, E> {
+        match self {
+            Ok(res) => RetryResult::Ok(res),
+            Err(err) => cont(err),
+        }
+    }
+}
+
+pub fn count_retry<E>(err: RetryError<E>) -> u64 {
```
romac, line 21: Reads a little bit better imho

```suggestion
pub fn retry_count<E>(err: RetryError<E>) -> u64 {
```

relayer/src/util/retry.rs
```diff
@@ -2,9 +2,33 @@ use std::time::Duration;
 
 pub use retry::{
     delay::{Fibonacci, Fixed},
-    retry_with_index, OperationResult as RetryResult,
+    retry_with_index, Error as RetryError, OperationResult as RetryResult,
 };
 
+pub trait ToRetryResult<T, E> {
```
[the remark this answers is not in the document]
soareschen: Yeah, I suppose we can add it back later when we implement the RetryableError logic. Right now since all errors are being retried, this is not really needed.

relayer/src/util/retry.rs
soareschen: Done.

adizere:
Hmm, what is the plan with this PR? I see 69 files changed.

soareschen:
Hmm I think something messed up while I tried to rebase.

soareschen:
Fixed now. It looks like a glitch on GitHub side when the site when down just now.

romac (approved):

romac:
Looks good, thanks @soareschen!

romac:
I updated #989 to track all locations which still need to be refactored.

Issue #992, opened 2021-05-25
Include chain-id in `hermes query clients` output 


ebuchman:
## Crate

relayer-cli

## Summary

`hermes query clients cosmoshub-4` outputs a list of all clients registered on the chain, but it's just the client ids, which isn't that helpful, especially since they're lexicographically ordered:

```
Success: [
    ClientId(
        "07-tendermint-0",
    ),
    ClientId(
        "07-tendermint-1",
    ),
    ClientId(
        "07-tendermint-10",
    ),
    ClientId(
        "07-tendermint-100",
    ),
    ClientId(
        "07-tendermint-101",
    ),
```

Of course the ordering is because these are strings, and we don't read the suffix number. It might be good to read the suffix number so we can order properly.

But more important would be to include the ChainID information, otherwise we don't really get anything valuable here since we don't know what each client corresponds to, but if we output the ChainID along with the ClientID it would at least be easier to find clients for particular chains.

## Acceptance Criteria

Output of `hermes query clients` is ordered numerically and includes ChainID along with ClientID
____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #995, opened 2021-05-26
Include chain-id in hermes query clients output

adizere:
Closes: #992

## Description

Three improvements for `hermes query clients`:

1. For each client id that is found, the command also outputs the chain identifier.

```
$ hermes query clients ibc-1
Success: [
    ChainClientTuple {
        client_id: ClientId(
            "07-tendermint-0",
        ),
        target_chain_id: "ibc-0",
    },
    ChainClientTuple {
        client_id: ClientId(
            "07-tendermint-1",
        ),
        target_chain_id: "ibc-0",
    },
]
```

The option `-o, --omit-chain-ids` can disable this behavior and print only the client identifiers.

```
$ hermes query clients ibc-1 -o
Success: [
    ClientId(
        "07-tendermint-0",
    ),
    ClientId(
        "07-tendermint-1",
    ),
]
```

2. The output of CLI `hermes query clients` is now sorted by the suffix of client ids

```
 $ hermes query clients ibc-0 -o
Success: [
    ClientId(
        "07-tendermint-0",
    ),
    ClientId(
        "07-tendermint-1",
    ),
    ClientId(
        "07-tendermint-2",
    ),
    ClientId(
        "07-tendermint-3",
    ),
    ClientId(
        "07-tendermint-4",
    ),
    ClientId(
        "07-tendermint-5",
    ),
    ClientId(
        "07-tendermint-6",
    ),
    ClientId(
        "07-tendermint-7",
    ),
    ClientId(
        "07-tendermint-8",
    ),
    ClientId(
        "07-tendermint-9",
    ),
    ClientId(
        "07-tendermint-10",
    ),
    ClientId(
        "07-tendermint-11",
    ),
]
```

3. The command can filter clients by their source chain id using the option `-s, --src-chain-id`

```bash
hermes query clients ibc-0 -s network5
May 26 17:11:12.313  INFO ibc_relayer_cli::commands::query::clients: printing identifiers of all clients hosted on chain ibc-0 which target chain network5
Success: [
    ClientId(
        "07-tendermint-6",
    ),
]
```

______

For contributor use:


romac:
Can we use a struct here instead of a tuple, to give names to the data?

```
$ hermes query clients network5 -i
 Success: [
    (
        ClientId(
            "07-tendermint-0",
        ),
        "network3",
    ),
]
```

ancazamfir:
A few comments:
- output typed ChainId
- should allow `hermes query clients ibc-0 [--src-chain ibc-1] [--include-chain-ids]`
- maybe make the `-i` default and add `-s` like in summary or some other better name for IDs only

relayer-cli/src/commands/query/clients.rs
```diff
@@ -49,10 +57,30 @@ impl Runnable for QueryAllClientsCmd {
...
-                    clients.into_iter().map(|cs| cs.client_id).collect();
-                Output::success(client_ids).exit()
+            Ok(mut clients) => {
+                // Sort by client identifier
+                clients.sort_by(|a, b| {
+                    a.client_id
+                        .suffix()
+                        .unwrap_or(0) // Fallback to `0` if client id is malformed
+                        .cmp(&b.client_id.suffix().unwrap_or(0))
+                });
```
ancazamfir, line 67: Ideally the sorting should be done in the `CosmosSdk` -s `query_client()`.

adizere:
@romac @ancazamfir thank you for the review! Most comments should be addressed.

ancazamfir:
> @romac @ancazamfir thank you for the review! Most comments should be addressed.

thanks, could you please update the description?

adizere:
> > @romac @ancazamfir thank you for the review! Most comments should be addressed.
> 
> thanks, could you please update the description?

Done!

ancazamfir:
Not sure if we should introduce `target_chain_id` here, I would keep it `chain_id` to be consistent with the state query.
So the two commands side-by-side would look like this (also maybe rename `ClientChain`, also it's not a tuple):
```
$ hermes query clients ibc-0
Success: [
    ClientChain {
        client_id: ClientId(
            "07-tendermint-0",
        ),
        chain_id: ChainId {
            id: "ibc-1",
            version: 1,
        },
    },
]
$ hermes query client state ibc-0 07-tendermint-0
Success: ClientState {
    chain_id: ChainId {
        id: "ibc-1",
        version: 1,
    },
    trust_level: TrustThresholdFraction {
        numerator: 1,
        denominator: 3,
    },
    trusting_period: 1209600s,
    unbonding_period: 1814400s,
    max_clock_drift: 5s,
    frozen_height: Height {
        revision: 0,
        height: 0,
    },
    latest_height: Height {
        revision: 1,
        height: 468,
    },
    upgrade_path: [
        "upgrade",
        "upgradedIBCState",
    ],
    allow_update: AllowUpdate {
        after_expiry: false,
        after_misbehaviour: false,
    },
}
```

modules/src/ics24_host/identifier.rs
```diff
@@ -179,6 +179,26 @@ impl ClientId {
...
+    /// let cl_b = ClientId::from_str("nosuffix_client").unwrap();
+    /// assert_eq!(cl_b.suffix(), None);
+    /// ```
+    pub fn suffix(&self) -> Option<u64> {
+        self.as_str()
+            .split('-')
+            .last()
+            .map(|e| e.parse::<u64>().ok())
+            .flatten()
+    }
```
ancazamfir, line 201: Maybe move this to `cosmos.rs` as query helper. Not all chains may have the same generation pattern as cosmos.

relayer-cli/src/commands/query/clients.rs
```diff
@@ -14,8 +16,27 @@ use crate::prelude::*;
...
+    #[options(
+        help = "filter for clients which target a specific chain id",
+        meta = "ID"
+    )]
+    src_chain_id: Option<ChainId>,
+
+    #[options(
+        help = "omit printing the chain identifier which each client targets; \
+                providing a `-s`/`--src-chain-id` will enable this option",
+        default = "false"
```
ancazamfir, line 31: Not sure if we can shorten this a bit. `hermes help query clients` is hard to read.

relayer-cli/src/commands/query/clients.rs
```diff
@@ -50,9 +71,47 @@ impl Runnable for QueryAllClientsCmd {
...
             Ok(clients) => {
-                let client_ids: Vec<ClientId> =
-                    clients.into_iter().map(|cs| cs.client_id).collect();
-                Output::success(client_ids).exit()
+                match self.src_chain_id.clone() {
+                    None => {
+                        match self.omit_chain_ids {
+                            true => {
+                                // Omit chain identifiers
+                                info!(
```
ancazamfir, line 79:
```suggestion
                                debug!(
```

relayer-cli/src/commands/query/clients.rs
```diff
@@ -50,9 +71,47 @@ impl Runnable for QueryAllClientsCmd {
...
+                                    "printing identifiers of all clients hosted on chain {}",
+                                    self.chain_id
+                                );
+                                let out: Vec<ClientId> =
+                                    clients.into_iter().map(|cs| cs.client_id).collect();
+                                Output::success(out).exit()
+                            }
+                            false => {
+                                // Include chain identifiers
+                                info!("printing identifiers (and target chain identifiers) of all clients hosted on chain {}", self.chain_id);
```
ancazamfir, line 89:
```suggestion
                                debug!("printing identifiers (and target chain identifiers) of all clients hosted on chain {}", self.chain_id);
```

relayer-cli/src/commands/query/clients.rs
```diff
@@ -50,9 +71,47 @@ impl Runnable for QueryAllClientsCmd {
...
+                                        client_id: cs.client_id,
+                                        target_chain_id: cs.client_state.chain_id().to_string(),
+                                    })
+                                    .collect();
+                                Output::success(out).exit()
+                            }
+                        };
+                    }
+                    Some(source_chain_id) => {
+                        info!(
```
ancazamfir, line 102:
```suggestion
                        debug!(
```

ancazamfir (changes requested):
Looks good! A few comments around help, debugs, etc.

modules/src/ics24_host/identifier.rs
adizere: Similar to the `ClientId::suffix()` method introduce in this PR, the method `ClientId::new()` also assumes the generation pattern with prefix/suffix for client ids, separated by a dash.

https://github.com/informalsystems/ibc-rs/blob/7c0418e3f6adf270a2a91f19b39a0141234c7545/modules/src/ics24_host/identifier.rs#L144-L159

I agree that this pattern is probably Cosmos-specific. But not sure we should separate this functionality in this PR. If you think this is the right time, it would be an easy refactor and I'll do it.

relayer-cli/src/commands/query/clients.rs
adizere: I adjusted the help messages a bit to make them more succint.

34c341111abb665df4a6b304adc9b04e51f0ee34

modules/src/ics24_host/identifier.rs
ancazamfir: The `new()` method should only be used by the on-chain handlers and it is ok to have it there as they are for a specific chain implementation, but not necessarily cosmos; ics24 probably not the best place for that, maybe the corresponding keeper, e.g. `ClientKeeper` for client IDs, `ConnectionKeeper` for connection IDs, etc. 
`suffix()` on the other hand is only for `CosmosSdk` chains.

relayer-cli/src/commands/query/clients.rs
ancazamfir: looks good!

modules/src/ics24_host/identifier.rs
adizere: > The new() method should only be used by the on-chain handlers and it is ok to have it there as that is for a specific chain implementation, but not necessarily cosmos

That's not how I interpreted it. The `new` method is `pub` so there's no reason why the relayer should not be using it. But I think I now understand your point. Thanks Anca for the explanation!

I refactored the problematic `ClientId::suffix()` into `cosmos.rs`. ad9ddf3e

ancazamfir (approved):
Looks good, thanks Adi!

Issue #998, opened 2021-05-26
High startup time for hermes


ancazamfir:
## Crate
relayer

## Summary of Bug
When hermes is started with two chains, one being cosmoshub, it takes 2-3 minutes to filter through all the hub channels and determine out it needs to handle maybe only one.
See also #994

## Steps to Reproduce
Start hermes with one channel between cosmos hub and other chain. It takes around 3 minutes for the two path relay workers to be spawned.

## Acceptance Criteria
Reduce startup time to a few seconds.

____

## For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

Issue #999, opened 2021-05-26
Hermes feature: New query channel client CLI


ancazamfir:
## Crate
relayer-cli

## Summary
Add following query to check the channel's client state:
`hermes query channel client <chain> <port-id> <channel-id`

## Acceptance Criteria

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

adizere:
For context: This was originally a requirement from Gravity Dex, but it seems no longer required by them. Adding it to backlog.

PR #1000, opened 2021-05-26
Optimize time for hermes start/ worker spawn

ancazamfir:
Closes: #998

## Description
Changed from iterating through and querying each channel to:
 - query all clients (one RPC)
 - filter clients that are not for configured chains
 - for connections for that client (one RPC)
 - query channels for that connection (one RPC)
- this is more efficient as after one RPC we know all channels obtained subsequently are very likely of interest (unless they are not-open/ inactive channels)

Also:
- added missing queries to the handle
- added a query to get the channel client state, in the end i did not use it but left it there and opened issue to add a CLI for this (#999)

- tested locally and also with cosmoshub-4 where there are ~125 channels. Startup time is reduced from 2-3mins for starting workers for a single pair of chains to 20 sec.
- as chains are added to hermes startup time will definitely increase but with useful work associated with it (spawning workers)

______

For contributor use:


adizere (approved):
Looks good!

PR #1006, opened 2021-05-27
Remove now unused `[connections]` section from the config definition

romac:


adizere (commented):
Could we also cleanup the files:

https://github.com/informalsystems/ibc-rs/blob/7c0418e3f6adf270a2a91f19b39a0141234c7545/relayer/tests/config/fixtures/relayer_conf_example.toml#L44

and

https://github.com/informalsystems/ibc-rs/blob/26ef2c6fcc2527998be1a206292ad870191b42c0/relayer/tests/config/fixtures/simple_config.toml#L40

In fact, I think we can even delete `fixtures/simple_config.toml` because it's not clear where it's used (but please double check).

romac:
> In fact, I think we can even delete `fixtures/simple_config.toml` because it's not clear where it's used (but please double check).

It's indeed not used. We have a file named the same at `ci/simple_config.toml` that's used as part of CI, so it looks safe to me to delete the one in `fixtures`.

adizere (approved):
Thanks Romain!

Issue #997, opened 2021-05-26
Hermes client update fails if client was recently created


adizere:
## Crate

`ibc-relayer` mostly

## Summary of Bug

Running `hermes create client` and immediately updating that client with `hermes update client` will fail.

```
./target/debug/hermes create client ibc-1 ibc-0 ; ./target/debug/hermes update client ibc-1 07-tendermint-6
May 26 18:46:52.144  INFO ibc_relayer_cli::commands: using default configuration from '/Users/adi/.hermes/config.toml'
Success: CreateClient(
    CreateClient(
        Attributes {
            height: Height {
                revision: 1,
                height: 6605,
            },
            client_id: ClientId(
                "07-tendermint-6",
            ),
            client_type: Tendermint,
            consensus_height: Height {
                revision: 0,
                height: 6652,
            },
        },
    ),
)
May 26 18:46:54.264  INFO ibc_relayer_cli::commands: using default configuration from '/Users/adi/.hermes/config.toml'
Error: tx error: error raised while updating client: chain ibc-1 is missing trusted state smaller than target height 0-6652
```

## Version

041ef642f7080e1faea60fc1325abde5afcf5c0a

## Steps to Reproduce

Compile hermes, then make sure you know the counter for the client that will be created (in this case, it's `07-tendermint-6`) and run:

```
./target/debug/hermes create client ibc-1 ibc-0 ; ./target/debug/hermes update client ibc-1 07-tendermint-6
```

## Acceptance Criteria

- ~~an appropriate error should be returned if the client cannot be updated; or if the update should succeed, then the bug should be fixed~~
- ~~after a closer look, it seems that the semantics of the CLI should be as follows: after running `hermes update client X`, client `X` should have the latest header installed; these semantics are already fulfilled by the current implementation of update client, yet the method returns `error`, so the correct return value from the method should be `success`~~
- update: make the `update client` CLI call blocking until the chain advances sufficiently so that method can effectively submit a newer consensus state to the target client
- also fix: "this probably silently fails sometime in our CI as we retry in `cmd.py`" -- comment by @ancazamfir

____

## For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

ancazamfir:
[...]

Issue #773, opened 2021-03-26
Client operations cf. ADR 006


adizere:
## Crate

`ibc-relayer-cli`

## Proposal

Toward improving user experience and simplifying the code-base, we'd like to retire some of the `tx raw` commands.
For client operations, these two CLIs are wrappers over tx raw commands:

- `hermes create client`
```rust
    /// Subcommand for creating a `client`
    #[options(help = "Create a new IBC client")]
    Client(TxCreateClientCmd),
```

- `hermes update client`

```rust
    /// Subcommand for updating a `client`
    #[options(help = "Update an IBC client")]
    Client(TxUpdateClientCmd),
```

The two commands `create client` and `update client` should not build on `TxUpdateClientCmd` and `TxCreateClientCmd`.

### Tasks

- [x] ~~to remove the code for `TxUpdateClientCmd` and `TxCreateClientCmd` and move most of that functionality in module `create.rs` and `update.rs`.~~
- [x] ~~additionally, we want to adopt the syntax laid-out in ADR 006, and start using the terminology "host chain" and "target chain" instead of "source" and "destination" chains when talking about IBC client.~~
- [x] Current hermes create client CLI creates a client using the per chain configuration. We should provide a way for the user to specify the security parameters for the client (either described via CLI options or maybe a .json or other type of file including the parameters)
    - tracked separately in #836 

Additionally, for the `update` method:
- [x] Since we know the client's chain once it is created, maybe make this one `hermes update client ibc-0 07-tendermint-0` ( instead of `hermes update client ibc-0 ibc-1 07-tendermint-0`). Can do in a different PR also. (_Originally posted by @ancazamfir in https://github.com/informalsystems/ibc-rs/pull/774#discussion_r602420456_)

____

#### For Admin Use

- [x] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

PR #1011, opened 2021-05-28
Refactoring and improve update client semantics

soareschen:
Closes: #997

## Description

______

For contributor use:

### DIFF
--- a/modules/src/ics02_client/height.rs
+++ b/modules/src/ics02_client/height.rs
@@ -1,5 +1,6 @@
 use std::cmp::Ordering;
 use std::convert::{Infallible, TryFrom};
+use std::fmt::{Debug, Display};
 use std::str::FromStr;
 
 use serde_derive::{Deserialize, Serialize};
@@ -16,6 +17,46 @@ pub struct Height {
 
     /// The height of a block
     pub revision_height: u64,
+}
+
+// A hack to indicate that a height is non-zero.
+// TODO: Make all height non-zero and use Option<Height> for
+//   optional height parameter:
+//   https://github.com/informalsystems/ibc-rs/issues/1009
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
+pub struct NonZeroHeight(Height);
+
+impl NonZeroHeight {
+    pub fn new(height: Height) -> Option<Self> {
+        if height == Height::zero() {
+            None
+        } else {
+            Some(NonZeroHeight(height))
+        }
+    }
+
+    /// Unsafe creation of a NonZeroHeight that may be zero.
+    /// This unfortunate counterintuitive function is due to the code base
+    /// right now using zero height to indicate for both the absense of height
+    /// and also zero height.
+    ///
+    /// Use this function to create zero height that bypass all checks that
+    /// behave differently on zero height by treating them as absense of height.
+    ///
+    /// This function can also be used for unfailable creation of
+    /// non-zero heights without the use of panics. The invariants
+    /// then have to be guaranteed by the caller.
+    ///
+    /// This function is unsafe, because we may be accidentally be converting
+    /// zero heights that are meant to indicate absense on height to actual
+    /// zero height and cause unexpected behavior.
+    pub fn unsafe_new(height: Height) -> Self {
+        NonZeroHeight(height)
+    }
+
+    pub fn height(self) -> Height {
+        self.0
+    }
 }
 
 impl Height {
@@ -123,7 +164,7 @@ impl From<Height> for RawHeight {
     }
 }
 
-impl std::fmt::Debug for Height {
+impl Debug for Height {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
         f.debug_struct("Height")
             .field("revision", &self.revision_number)
@@ -132,10 +173,15 @@ impl std::fmt::Debug for Height {
     }
 }
 
-/// Custom debug output to omit the packet data
-impl std::fmt::Display for Height {
+impl Display for Height {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
         write!(f, "{}-{}", self.revision_number, self.revision_height)
+    }
+}
+
+impl Display for NonZeroHeight {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        write!(f, "{}", self.0)
     }
 }
 
--- a/relayer-cli/src/commands/tx/client.rs
+++ b/relayer-cli/src/commands/tx/client.rs
@@ -61,11 +61,15 @@ pub struct TxUpdateClientCmd {
     )]
     dst_client_id: ClientId,
 
-    #[options(help = "the target height of the client update", short = "h")]
-    target_height: Option<u64>,
+    #[options(required, help = "the target height of the client update", short = "h")]
+    target_height: u64,
 
-    #[options(help = "the trusted height of the client update", short = "t")]
-    trusted_height: Option<u64>,
+    #[options(
+        required,
+        help = "the trusted height of the client update",
+        short = "t"
+    )]
+    trusted_height: u64,
 }
 
 impl TxUpdateClientCmd {
@@ -93,36 +97,11 @@ impl TxUpdateClientCmd {
         let client = ForeignClient::find(src_chain, dst_chain, &self.dst_client_id)
             .map_err(|e| format!("failed to find foreign client: {}", e))?;
 
-        let m_target_height = self
-            .target_height
-            .and_then(|h| NonZeroHeight::new(ibc::Height::new(src_chain_id, h)));
+        let target_height =
+            NonZeroHeight::unsafe_new(ibc::Height::new(src_chain_id, self.target_height));
 
-        let m_trusted_height = self.trusted_height.map(|h| {
-            // Allow trusted height parameter to be zero if force provided
-            NonZeroHeight::unsafe_new(ibc::Height::new(src_chain_id, h))
-        });
-
-        let target_height = match m_target_height {
-            Some(height) => {
-                client.wait_src_chain_reach_height(height).map_err(|e| {
-                    format!(
-                        "failed to wait for client to reach height {}: {}",
-                        height, e
-                    )
-                })?;
-                height
-            }
-            None => client
-                .latest_src_chain_height()
-                .map_err(|e| e.to_string())?,
-        };
-
-        let trusted_height = match m_trusted_height {
-            Some(h) => h,
-            None => client
-                .latest_consensus_state_height()
-                .map_err(|e| e.to_string())?,
-        };
+        let trusted_height =
+            NonZeroHeight::unsafe_new(ibc::Height::new(src_chain_id, self.trusted_height));
 
         client
             .build_update_client_and_send(target_height, trusted_height)
--- a/relayer-cli/src/commands/tx/client.rs
+++ b/relayer-cli/src/commands/tx/client.rs
@@ -3,6 +3,7 @@ use abscissa_core::{Command, Options, Runnable};
 use ibc::events::IbcEvent;
 use ibc::ics02_client::client_state::ClientState;
 use ibc::ics24_host::identifier::{ChainId, ClientId};
+use ibc::NonZeroHeight;
 use ibc_relayer::foreign_client::ForeignClient;
 
 use crate::application::app_config;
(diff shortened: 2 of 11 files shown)


modules/src/ics02_client/height.rs
```diff
@@ -133,12 +155,19 @@ impl std::fmt::Debug for Height {
...
 
 /// Custom debug output to omit the packet data
-impl std::fmt::Display for Height {
+impl Display for Height {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
         write!(f, "{}-{}", self.revision_number, self.revision_height)
     }
 }
 
+/// Custom debug output to omit the packet data
```
adizere, line 164: I guess this comment (and on line 157 also) is a copy-paste.

relayer/src/chain/cosmos.rs
```diff
@@ -1133,6 +1134,11 @@ impl Chain for CosmosSdkChain {
         let connection_end = ConnectionEnd::decode_vec(&res.value)
             .map_err(|e| Kind::Query("proven connection".into()).context(e))?;
 
+        debug!(
+            "Gotten proven connection respons for connection {} at height {}: {:#?}",
```
adizere, line 1138: We usually try to make all log output start with a a lowercase.

```suggestion
            "got proven connection response for connection {} at height {}: {:#?}",
```

relayer-cli/src/commands/tx/client.rs
```diff
@@ -67,50 +68,62 @@ pub struct TxUpdateClientCmd {
...
-            None => ibc::Height::zero(),
+        let m_target_height = self
+            .target_height
+            .and_then(|h| NonZeroHeight::new(ibc::Height::new(src_chain_id, h)));
+
+        let m_trusted_height = self
+            .trusted_height
+            .and_then(|h| NonZeroHeight::new(ibc::Height::new(src_chain_id, h)));
+
+        let target_height = match m_target_height {
```
adizere, line 104: These checks make the command quite intelligent, which is nice. But it might be better if the user is allowed to exercise the corner-cases of `update client CLIENT-ID --target-height 0` for example, especially for `tx raw`. To do that, it could be a good idea to leave `TxUpdateClientCmd` untouched (modulo the nice refactor with `execute`) and create a new command e.g., `UpdateClientCmd` that is similar to `TxUpdateClientCmd ` but which has more checks in place to cover corner-cases.

Having a new `UpdateClientCmd` command will also address (part of) #773 .

relayer/src/foreign_client.rs
```diff
@@ -398,31 +402,21 @@ impl ForeignClient {
...
     /// If the client already stores consensus states for this height, returns an empty vector.
     pub fn build_update_client_with_trusted(
         &self,
-        target_height: Height,
-        trusted_height: Height,
+        target_height: NonZeroHeight,
+        trusted_height: NonZeroHeight,
     ) -> Result<Vec<Any>, ForeignClientError> {
-        // Wait for source chain to reach `target_height`
-        while self.src_chain().query_latest_height().map_err(|e| {
```
adizere, line 414: This is probably still necessary, in particular for packet relaying algorithms.

adizere (changes requested):
I like your proposal for the `NonZeroHeight` type, which could help us remove the clumsy `if h == Height::zero()` checks around the codebase. It would be good if you can briefly argue for the rationale behind this new type in the PR description.

I know it's still a draft, so you're probably fixing these problems already, but please mind:
- [ ] clippy
- [ ] failing tests, probably because of the removed code in particular in ForeignClient
- [ ] fix for issue #997 which is driving this refactor (I updated the acceptance criteria in the issue)

relayer/src/foreign_client.rs
soareschen: I have added a `wait_src_chain_reach_height()` function to wait for the source chain to reach the target height. So far the only places that I know need waiting is in `RelayPath` and the command line.

relayer-cli/src/commands/tx/client.rs
soareschen: I have added a `NonZeroHeight::unsafe_new()` function for creation of heights that are really meant to be zero. Btw note that previously even if zero height is allowed in the command line, it still really means the absence of height and the relayer will still fetch the current height.

I have now added the `UpdateClientCmd` and have the two behaviors slightly different. In `UpdateClientCmd`, only the target height can optionally be provided, and the command will wait for the target height to reach before updating. For `TxUpdateClientCmd`, both trusted height and target height must be provided, and there is no sanity checking in the command. So if invalid heights are provided, they will go into the implementation code and return error from there.

relayer/src/chain/cosmos.rs
soareschen: Done

modules/src/ics02_client/height.rs
soareschen: Removed.

soareschen:
The PR is now ready for review. Here is a more reliable way to reproduce #997:

- Modify the `timeout_commit` and `timeout_propose` fields in the [one-chain](https://github.com/informalsystems/ibc-rs/blob/a1e95e1443a97f5db44bf3fd3fd53a2164e587b1/scripts/one-chain#L124-L134) script to a longer duration, such as 10s.
- Recreate the chains using `./scripts/dev-env`.
- Run commands such as`./target/debug/hermes create client ibc-0 ibc-1 ; ./target/debug/hermes update client ibc-0 07-tendermint-{N}`, with `{N}` replaced with expected new client ID.
- The command should fail on master, but pass on this PR.

relayer-cli/src/commands/update.rs
```diff
@@ -1,8 +1,94 @@
...
+
+    #[options(
+        free,
+        required,
+        help = "identifier of the client to be updated on destination chain"
+    )]
+    dst_client_id: ClientId,
+
+    #[options(help = "the target height of the client update", short = "h")]
+    target_height: Option<u64>,
```
adizere, line 27: Just FYI: we may need to update the guide ([this section](https://hermes.informal.systems/commands/path-setup/clients.html#update-client)) with this change.

relayer/src/foreign_client.rs
```diff
@@ -432,34 +463,6 @@ impl ForeignClient {
                 ))
             })?;
 
-        // If not specified, set trusted state to the highest height smaller than target height.
-        // Otherwise ensure that a consensus state at trusted height exists on-chain.
-        let cs_heights = self.consensus_state_heights()?;
-        let trusted_height = if trusted_height == Height::zero() {
```
adizere, line 438: Since we don't have this check in place anymore, this means that any caller of `build_update_client_with_trusted` has to make sure that the client has a consensus state at `trusted_height` installed. Do I understand correctly?

relayer/src/foreign_client.rs
```diff
@@ -540,9 +546,7 @@ impl ForeignClient {
             ))
         })?;
 
-        assert!(!events.is_empty());
-
-        Ok(events.pop().unwrap())
```
adizere, line 545: Nice refactor around this method.

adizere (commented):
Confirm that the fix for #997 works, thanks Soares!

```
> ./target/debug/hermes create client ibc-0 ibc-1 ; ./target/debug/hermes update client ibc-0 07-tendermint-6
Success: CreateClient(
    CreateClient(
        Attributes {
            height: Height {
                revision: 0,
                height: 522,
            },
            client_id: ClientId(
                "07-tendermint-6",
            ),
            client_type: Tendermint,
            consensus_height: Height {
                revision: 1,
                height: 521,
            },
        },
    ),
)
Jun 02 12:46:29.257  INFO ibc_relayer_cli::commands: using default configuration from '/Users/adi/.hermes/config.toml'
Jun 02 12:46:29.291  WARN ibc_relayer::foreign_client: [ibc-1 -> ibc-0:07-tendermint-6] skipping update: trusted height (1-521) >= chain target height (1-521)
Jun 02 12:46:29.291  INFO ibc_relayer::foreign_client: Client 07-tendermint-6 is already up-to-date with chain ibc-1@1-521
Success: None
```

The PR is difficult to review, however. There's multiple changes (in particular around the `ForeignClient::build*` methods) and could not understand their impact fully. The difficulty is mostly due to the complex ways in which these methods can be called and used (not necessarily because of this PR). So I'm wondering if it is possible to break the PR in two parts: one that strictly (and minimally) fixes 997, and another part with your refactoring.

relayer-cli/src/commands/update.rs
ancazamfir: definitely needs guide update. Also, if we want to change this as compared to `tx raw` we should discuss to understand why and how. At first look I would propose `hermes client update <chain-id> <cliend-id>` (i.e. drop the target_height)

ancazamfir (changes requested):

ancazamfir:
Running locally `e2e/run.py` i get:
```2021-06-02 12:37:20 [   DEBUG] cargo run --bin hermes -- -c /Users/ancaz/.hermes/config.toml --json update client ibc-0 07-tendermint-0
2021-06-02 12:37:20 [   DEBUG] {"result":null,"status":"success"}
Traceback (most recent call last):
  File "/Users/ancaz/rust/ibc-rs/./e2e/run.py", line 229, in <module>
    main()
  File "/Users/ancaz/rust/ibc-rs/./e2e/run.py", line 213, in main
    ibc0_client_id, ibc0_conn_id, ibc0_chan_id, ibc1_client_id, ibc1_conn_id, ibc1_chan_id = raw(config, ibc0 , ibc1, port_id)
  File "/Users/ancaz/rust/ibc-rs/./e2e/run.py", line 130, in raw
    ibc0_client_id = client.create_update_query_client(c, ibc0, ibc1)
  File "/Users/ancaz/rust/ibc-rs/e2e/e2e/client.py", line 128, in create_update_query_client
    update_client(c, dst, client.client_id)
  File "/Users/ancaz/rust/ibc-rs/e2e/e2e/client.py", line 111, in update_client
    res = cmd.run(c).success()
  File "/Users/ancaz/rust/ibc-rs/e2e/e2e/cmd.py", line 34, in success
    data = self.cmd.process(result)
  File "/Users/ancaz/rust/ibc-rs/e2e/e2e/client.py", line 49, in process
    return from_dict(ClientUpdated, result['UpdateClient']['common'])
KeyError: 'UpdateClient'
```
probably because the return value is `None` and update client event cannot be decoded.

(`one-chain` script changed with 10 seconds timeouts)

relayer/src/foreign_client.rs
soareschen: Correct. In pretty much all places, the trusted height is already obtained from the consensus state height, so the check is unnecessary. With the new `update client` command, I make it such that the trusted height is also automatically fetchd, so there is no need for checking. The only place left unchecked is in the `tx raw update-client` command, but since it is a raw command I think it is good to bypass checking so that we can test the behavior on the actual chain.

relayer-cli/src/commands/tx/client.rs
soareschen: This is just my suggestion, because otherwise I am not really sure what is the point of having two commands with separate implementations if their behavior are the same. Or do you have any recommendation of how the two commands should be differentiated?

soareschen:
> probably because the return value is None and update client event cannot be decoded.

Hmm makes sense. If the trusted height is already at target height, then `None` is returned, so the e2e script will need to handle that. I will fix the e2e script then.

soareschen:
> So I'm wondering if it is possible to break the PR in two parts: one that strictly (and minimally) fixes 997, and another part with your refactoring.

I suppose if we want to fix #997 naively, we can just replace the comparator operator in the original `build_update_client_and_send` from `<` to `<=`:

```diff
diff --git a/relayer/src/foreign_client.rs b/relayer/src/foreign_client.rs
index dad8993d..1385fe4b 100644
--- a/relayer/src/foreign_client.rs
+++ b/relayer/src/foreign_client.rs
@@ -439,7 +439,7 @@ impl ForeignClient {
             // Get highest height smaller than target height
             cs_heights
                 .into_iter()
-                .find(|h| h < &target_height)
+                .find(|h| h <= &target_height)
                 .ok_or_else(|| {
                     ForeignClientError::ClientUpdate(format!(
                         "chain {} is missing trusted state smaller than target height {}",
```

I agree that the original code is pretty difficult to understand. And at least without familiarizing myself with what the code actually does, I wouldn't be able to be certain that the fix really solves the issue. Right now with the correct understanding after the refactoring, I think the simple fix could work.

The refactored code should preserve the original semantics while keeping it simpler. But since the original code is so complex, it certainly requires some reading to make sure that it really doesn't break anything.

soareschen:
> I suppose if we want to fix #997 naively, we can just replace the comparator operator in the original build_update_client_and_send from < to <=:

Hmm maybe it's not that straightforward. Somewhere upstream the original code would still convert empty result to error.

```
Jun 02 11:19:29.047  WARN ibc_relayer::foreign_client: [ibc-1 -> ibc-0:07-tendermint-3] skipping update: trusted height (1-14) >= chain target height (1-14)
Error: tx error: error raised while updating client: Client 07-tendermint-3 is already up-to-date with chain ibc-1@1-14
```

romac:
> So I'm wondering if it is possible to break the PR in two parts: one that strictly (and minimally) fixes 997, and another part with your refactoring.

That would be great, and would perhaps allow us to merge the fix for 997 before the release and take some time to properly review the refactoring afterwards.

soareschen:
I have opened #1036 for the quick fix.

adizere:
This work will be subsided by later work that will replace `Height::zero` with `Option<Height>`.

Heights can have different meanings:
- some specific height
- none specified, which can mean an error (because a specific height is necessary)
- none specified, which can mean the _latest_ one

@soareschen:
- Create a discussion to capture this refactor, describe the rationale.

Issue #821, opened 2021-04-15
Add connection worker to the reactive relayer


ancazamfir:
## Crate
relayer

## Summary
Implement connection worker:
- [x] On start scan all configured chains for in-progress connections
- [x] Create worker per connection
- [x] Send last event to the worker to complete open handshake 

See #764

## Acceptance Criteria

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1019, opened 2021-05-31
Connection worker

cezarad:
worker for connection 

Closes: #821 

## Description

- added a worker for connection
- added query_connections to the chain handle and runtime
- allow entering the destination connection id in ```hermes tx raw conn-try... ``` and make dst_connection_id and option (used to be ``` ConnectionId::default()``` when not specified but this is a valid connection id.
- added tests to e2e

To setup for the tests:

- change strategy in config to `all` (relay from all events, including connection handshake ones)
- execute start the chains using ```./scripts/dev-env ~/.hermes/config.toml ibc-0 ibc-1 ``` and then
- create clients on ibc-0 and ibc-1 with ```hermes create client ibc-0 ibc-1 ``` and ``` hermes create client ibc-1 ibc-0 ```
 
case A:
1. start the relayer ```hermes start```
2. send ```hermes tx raw conn-init ibc-0 ibc-1 07-tendermint-0 07-tendermint-1```

Case B:
1. send ```hermes tx raw conn-init ibc-0 ibc-1 07-tendermint-0 07-tendermint-1```
2. start the relayer ```hermes start```

Case C:
1. send ```hermes tx raw conn-init ibc-0 ibc-1 07-tendermint-0 07-tendermint-1```
2. send ```hermes tx raw conn-try ibc-1 ibc-0 07-tendermint-1 07-tendermint-0 -s connection-x``` , with connection-x being the ID of the connection returned in step 1 above.
start the relayer ``` hermes start```

Case D:
1. send ```hermes tx raw conn-init ibc-0 ibc-1 07-tendermint-0 07-tendermint-1```
2. send```hermes tx raw conn-try ibc-1 ibc-0 07-tendermint-1 07-tendermint-0 -s connection-x```,
3. send ```hermes tx raw conn-ack ibc-0 ibc-1 07-tendermint-0 07-tendermint-1 -d  connection-x -s connection-y``` , with connection-x being the ID of the connection returned in step 1 and connection-y the ID of the connection returned in step 2 above
4. start the relayer ```hermes start```

After the last step, wait a couple of seconds for hermes to finish the handshake and verify the connection state on both chains using hermes query connection end ibc-0 connection-x hermes query connection end ibc-0 transfer connection-y. They should both be in open state

TODO: 
- [x] double-check check the e2e tests
- [X] guide update (done in the release PR #1098)

______

For contributor use:


ancazamfir (changes requested):

ancazamfir (approved):
Looks great @cezarad ! Thanks!!!

relayer/src/supervisor.rs
```diff
@@ -92,6 +92,37 @@ impl Supervisor {
...
+                    }
+                }
+
+                IbcEvent::OpenAckConnection(..) => {
+                    if !self.handshake_enabled() {
+                        continue;
+                    }
+
+                    if let Ok(connection_object) = Object::connection_from_conn_open_events(
+                        event.clone().connection_attributes().unwrap(),
```
romac, line 115: Is this `unwrap` guaranteed to succeed?

Also, the `clone` might not be required:

```suggestion
                        event.connection_attributes().unwrap(),
```

relayer/src/chain/counterparty.rs
```diff
@@ -3,17 +3,103 @@ use tracing::{error, trace};
...
+    let req = QueryClientConnectionsRequest {
+        client_id: counterparty_client_id.to_string(),
+    };
+
+    let counterparty_connections =
+        counterparty_chain
+            .query_client_connections(req)
+            .map_err(|e| {
+                Error::QueryFailed(format!(
+                    "counterparty_chain query_client {} _connections failed {}",
```
romac, line 57:
```suggestion
                    "counterparty::query_client_connections({}) failed {}",
```

relayer/src/supervisor.rs
ancazamfir: > Is this unwrap guaranteed to succeed?

Yes, but we should probably follow the same pattern as for channel and ignore if not present (shouldn't happen but it's nicer)

e2e/e2e/connection.py
```diff
@@ -250,3 +254,89 @@ def query_connection_end(c: Config, chain_id: ChainId, conn_id: ConnectionId) ->
...
+# =============================================================================
+
+def verify_state(c: Config,
+    ibc1: ChainId, ibc0: ChainId,
+    ibc1_conn_id: ConnectionId):
+
+    strategy = toml.load(c.config_file)['global']['strategy']
+    # verify connection state on both chains, should be 'Open' for 'all' strategy, 'Init' otherwise
+
+    if strategy == 'all':
```
adizere, line 269: Just to clarify: the intended use-case for this is for development-time, when running the e2e tests locally. The strategy is `packets` in the ci config, so these new tests are not expected to run in the CI, right?

relayer/src/worker/connection.rs
```diff
@@ -0,0 +1,127 @@
...
+                    }
+                };
+
+                if let Err(retries) = result {
+                    warn!("connection worker failed after {} retries", retries);
+
+                    // Resume handshake on next iteration.
+                    resume_handshake = true;
+                } else {
+                    resume_handshake = false;
```
soareschen, line 112: From what I can see, in the smooth case that no unexpected errors happen, then once `resume_handshake` is set to false, then it is not going to be set to true again ever. So does this means that the branch for `WorkerCmd::NewBlock` is going to always be skipped? Otherwise, are we using some subtle error variants to trigger the `Err` branch to set `resume_handshake` to true again?

relayer/src/worker/connection.rs
```diff
@@ -0,0 +1,127 @@
...
+
+                        debug!("connection worker starts processing {:#?}", last_event);
+
+                        match last_event {
+                            Some(event) => {
+                                let mut handshake_connection = RelayConnection::restore_from_event(
+                                    a_chain.clone(),
+                                    b_chain.clone(),
+                                    event.clone(),
+                                )?;
```
soareschen, line 67: Doing a `?` here short circuits the entire function and returns from the loop prematurely.

relayer/src/worker/connection.rs
```diff
@@ -0,0 +1,127 @@
...
+
+                        let height = current_height.decrement()?;
+
+                        let (mut handshake_connection, state) =
+                            RelayConnection::restore_from_state(
+                                a_chain.clone(),
+                                b_chain.clone(),
+                                self.connection.clone(),
+                                height,
+                            )?;
```
soareschen, line 98: Doing a `?` here short circuits the entire function and returns from the loop prematurely.

relayer/src/worker/connection.rs
```diff
@@ -0,0 +1,127 @@
...
+    /// Run the event loop for events associated with a [`Connection`].
+    pub(crate) fn run(self) -> Result<(), BoxError> {
+        let a_chain = self.chains.a.clone();
+        let b_chain = self.chains.b.clone();
+
+        // Flag that indicates if the worker should actively resume handshake.
+        // Set on start or when event based handshake fails.
+        let mut resume_handshake = true;
+
+        loop {
```
soareschen, line 49: What happens after the full connection handshake has been established? Is the loop supposed to continue forever?

relayer/src/worker/connection.rs
```diff
@@ -0,0 +1,127 @@
...
+            thread::sleep(Duration::from_millis(200));
+
+            if let Ok(cmd) = self.cmd_rx.try_recv() {
```
soareschen, line 52: If `try_recv()` returns `TryRecvError::Disconnected`, the sender side of the channel has been dropped and the loop is going to keep failing forever.

relayer/src/worker/connection.rs
```diff
@@ -0,0 +1,127 @@
...
+        Self {
+            connection,
+            chains,
+            cmd_rx,
+            telemetry,
+        }
+    }
+
+    /// Run the event loop for events associated with a [`Connection`].
+    pub(crate) fn run(self) -> Result<(), BoxError> {
```
soareschen, line 41: Can you add some documentation to explain a bit what is the expected execution flow of the function overall? My understanding is as follows:

- Wait for incoming block events for the connection state on the source chain `a_chain` to be updated.
- Match the new state machine and update to the next connection state.
  - Only cares about the direction from `a_chain` to `b_chain`, i.e. `(State::Init, State::TryOpen)` is not handled.
- If the connection state has reached `(State::Open, State::Open)`, loop forever (should have been return right?)
- If any error happens, retry until the connection succeed? 
   - The retry is in two stages: retry a number of times for each block event update, and retry again in the next block update.

relayer/src/worker/connection.rs
soareschen: I always find complicated control flow inside loops difficult to follow. I understand you are just following the convention before. It would be great if we can use functions to structure the control flows instead. That would make it much easier to understand the control flow. For example something like follows:

```rust
    pub(crate) fn run(self) -> Result<(), BoxError> {
        enum Action {
            Continue,
            End,
        }

        fn do_run(worker: &ConnectionWorker) -> Result<Action, BoxError> { ... }

        loop {
            match do_run(&self) {
                Ok(Action::End) => return Ok(()),
                Err(e) => {
                    error!("error: {}. retrying");
                    // or return Err if unrecoverable
                }
                _ => {}
            }
        }
  }
```

soareschen (changes requested):

e2e/e2e/connection.py
romac: The e2e test suite does work with `strategy = 'all'`, I've changed it in CI to reflect that.

relayer/src/worker/connection.rs
ancazamfir: > So does this means that the branch for WorkerCmd::NewBlock is going to always be skipped? 

It's going to be executed if `step_event()` or `step_state()` have not advanced the handshake after a number of tries.

relayer/src/worker/connection.rs
ancazamfir: correct.

relayer/src/worker/connection.rs
ancazamfir: correct.

relayer/src/worker/connection.rs
ancazamfir: > What happens after the full connection handshake has been established? Is the loop supposed to continue forever?

no, see #1115

ancazamfir: > I always find complicated control flow inside loops difficult to follow. I understand you are just following the convention before. It would be great if we can use functions to structure the control flows instead. 
...

Please open an issue for this and we will address in a separate PR.

relayer/src/worker/connection.rs
ancazamfir: yes this is true. All workers have the same pattern. The supervisor should not disconnect unless it has crashed and in this case not much can be done. But we could indeed check this error and exit. I think we discussed with @romac to have a way for the supervisor to gracefully close the worker, not sure this is done. I think the worker has a way to gracefully exit so the supervisor can clean its worker info.

PR #1012, opened 2021-05-28
Fix Hermes help to provide a useful message

adizere:
Closes: #1003 

## Description

When a command is invoked incorrectly, Hermes prints an error diagnostic followed by a generic help message, [for example](https://github.com/informalsystems/ibc-rs/issues/1003#issuecomment-850181883):

```
$ hermes create client
error: missing required free argument               <<<<<<---- diagnostic

hermes 0.3.2
Informal Systems <hello@informal.systems>                 <<--- 
Hermes is an IBC Relayer written in Rust.                 <<---- generic help message
                                                          <<--- 
FLAGS:                                                    <<--- 
    -c, --config CONFIG       path to configuration file  <<--- 
    -h, --help                print help message          <<--- 
    -j, --json                enable JSON output          <<--- 

```

The problem is that the `-h` mentioned in the generic help message does not work (#134, #1003).

With this PR, the generic help message is changed. The `-h` flag was removed, and the message is as follows:

```
$ hermes create client
error: missing required free argument

hermes 0.3.2
Informal Systems <hello@informal.systems>
For help, run `hermes [<FLAGS>] help <COMMAND>`, for example: `hermes help create client`

FLAGS:
    -c, --config CONFIG       path to configuration file
    -j, --json                enable JSON output
```

This help message should only appear when the user invoked a command with the wrong parameters. The other commands (e.g., `hermes help`) will not contain this new message:

```
$ hermes help
   hermes 0.3.2
Informal Systems <hello@informal.systems>
Hermes is an IBC Relayer written in Rust.

USAGE:
    hermes <SUBCOMMAND>

SUBCOMMANDS:
    help       Get usage information
    keys       Manage keys in the relayer for each chain
    create     Create objects (client, co
...
```

______

For contributor use:

- [x] Updated the __Unreleased__ section of [CHANGELOG.md](https://github.com/informalsystems/ibc-rs/blob/master/CHANGELOG.md) with the issue.
- [ ] If applicable: Unit tests written, added test to CI.
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Updated relevant documentation (`docs/`) and code comments.
- [ ] Re-reviewed `Files changed` in the Github PR explorer.

PR #985, opened 2021-05-25
Add telemetry and Prometheus endpoint

brapse:
Closes: #868

## Description

Adds telemetry to Hermes and exposes an Prometheus endpoint.

## Testing

* Configure telemetry exporter in config.toml by adding a `[telemetry]` section:
```
[telemetry]
enabled = true
port = 3001
```

1. Start dev environment, for example with 3 chains in a config.toml:
```
./scripts/dev-env ~/.hermes/config.toml ibc-0 ibc-1
```

2. Run Hermes:
```
$ cargo run -- start
```

In a separate terminal, watch the telemetry (please ensure the port matches the one specified in the `telemetry.port` config parameter
```
$ watch -n 1 curl -s localhost:3001/metrics
```

Expect the output to look like
```
# HELP hermes_chains_num Number of chains the relay is connecting to
# TYPE hermes_chains_num counter
hermes_chains_num{hermes="all"} 2
# HELP hermes_channels_num Number of channels the relay is connecting to
# TYPE hermes_channels_num counter
hermes_channels_num{hermes="all"} 2
# HELP hermes_tx_msg_ibc_acknowledge_packet Total number of IBC packets acknowledged
# TYPE hermes_tx_msg_ibc_acknowledge_packet counter
hermes_tx_msg_ibc_acknowledge_packet{hermes="all"} 2
```

Run the e2e test trigger other metrics....
______

For contributor use:

- [ ] Updated the __Unreleased__ section of [CHANGELOG.md](https://github.com/informalsystems/ibc-rs/blob/master/CHANGELOG.md) with the issue.
- [ ] If applicable: Unit tests written, added test to CI.
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Updated relevant documentation (`docs/`) and code comments.
- [ ] Re-reviewed `Files changed` in the Github PR explorer.

PR #1028, opened 2021-06-01
Guide update & release 0.4.0

adizere:
Closes: #1030

## Description

PRs that demand guide update, according to our tagging with [label:"needs guide update"](https://github.com/informalsystems/ibc-rs/pulls?q=is%3Apr+is%3Aclosed+label%3A%22needs+guide+update%22):
- [X] #1012 (307dd19, @adizere)
- [x] #995 (@adizere)
- [x] #971 (@romac)
- [x] #985 (@romac)
- [x] #1033 (@romac)
- [x] #924 (@ancazamfir)
- [x] double-check this list against _all_ PRs since last release (maybe we did not add the correct tag and we missed some) (@romac)
- [x] Add compatibility matrix to the guide (@ancazamfir)
- [x] Update changelog


proto-compiler/Cargo.lock
```diff
@@ -2,9 +2,9 @@
 # It is not intended for manual editing.
 [[package]]
 name = "anyhow"
-version = "1.0.38"
+version = "1.0.40"
```
adizere, line 5: I am wondering if we should keep the modifications to this file.

Ideally, we would change `proto-compiler/Cargo.lock` only when we re-build the `proto-compiler` binary, and we should rebuild this binary only when we use it to regenerate the .proto file (https://github.com/informalsystems/ibc-rs/tree/master/proto/src). Keeping `proto-compiler` binary aligned with the .proto files allows to regenerate those files in a way that produces consistent results (same Rust output, given same input .proto).

However, [some](https://github.com/informalsystems/ibc-rs/pull/1016) PRs changed the file `proto-compiler/Cargo.lock` anyway since the last time we used it to generate Rust proto files.. and maybe it's just easier to go with the flow and allow changes to this (since we have its history anyway, we should be able to obtain consistent output)

ancazamfir:
the guide build started to fail in CI since master merge 10af4ec8da73319262e4a76bc6a8d93f04b82541

proto-compiler/Cargo.lock
romac: As discussed offline, let's keep those in for this release.

romac (dismissed):

romac (approved):

ancazamfir (approved):

Issue #1032, opened 2021-06-01
Cannot reach telemetry endpoint from another machine in the same network


andynog:
## Crate

telemetry

## Summary of Bug

Testing the telemetry server, I could not reach the metric endpoint from another machine in the same network. This is important because Prometheus will probably be running on another machine.

## Version

e4a654385fc8e4e18aa39c68325ad0903f023d7f

## Steps to Reproduce

Run the relayer on one machine in the network with `hermes start`. Try to access the telemetry endpoint on another machine in the network e.g. if the hermes machine IP is 192.168.1.10, trying to access the endpoint http://192.168.1.10:3001/metrics won't work. 

The reason for that is because the telemetry server (rouille) is hardcoded to `localhost`. Setting it to `0.0.0.0` fixes the issue. 

## Acceptance Criteria

Be able to access the telemetry endpoint from a different computer in the same subnet other than the relayer. 

____

## For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

PR #1033, opened 2021-06-01
Add `host` parameter to `telemetry` config

andynog:
Closes: #1032 

## Description

Add `host` parameter to `telemetry` config section, to allow access from a different machine in the network (eg. by using `0.0.0.0` as the host).

_____

For contributor use:

### DIFF
--- a/telemetry/src/server.rs
+++ b/telemetry/src/server.rs
@@ -1,7 +1,7 @@
-use std::{net::ToSocketAddrs, sync::Arc};
+use std::{error::Error, net::ToSocketAddrs, sync::Arc};
 
 use prometheus::{Encoder, TextEncoder};
-use rouille::Request;
+use rouille::{Request, Response, Server};
 
 use crate::state::TelemetryState;
 
@@ -20,8 +20,11 @@ impl Route {
     }
 }
 
-pub fn run(address: impl ToSocketAddrs, telemetry_state: Arc<TelemetryState>) {
-    rouille::start_server(address, move |request| {
+pub fn listen(
+    address: impl ToSocketAddrs,
+    telemetry_state: Arc<TelemetryState>,
+) -> Result<Server<impl Fn(&Request) -> Response>, Box<dyn Error + Send + Sync>> {
+    let server = Server::new(address, move |request| {
         match Route::from_request(request) {
             // The prometheus endpoint
             Route::Metrics => {
@@ -37,5 +40,7 @@ pub fn run(address: impl ToSocketAddrs, telemetry_state: Arc<TelemetryState>) {
             // Return an empty response with a 404 status code.
             Route::Other => rouille::Response::empty_404(),
         }
-    })
+    })?;
+
+    Ok(server)
 }
--- a/telemetry/src/server.rs
+++ b/telemetry/src/server.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::{net::ToSocketAddrs, sync::Arc};
 
 use prometheus::{Encoder, TextEncoder};
 use rouille::Request;
@@ -20,8 +20,8 @@ impl Route {
     }
 }
 
-pub fn run(telemetry_state: Arc<TelemetryState>, port: u16) {
-    rouille::start_server(("0.0.0.0", port), move |request| {
+pub fn run(address: impl ToSocketAddrs, telemetry_state: Arc<TelemetryState>) {
+    rouille::start_server(address, move |request| {
         match Route::from_request(request) {
             // The prometheus endpoint
             Route::Metrics => {
--- a/telemetry/src/server.rs
+++ b/telemetry/src/server.rs
@@ -21,7 +21,7 @@ impl Route {
 }
 
 pub fn run(telemetry_state: Arc<TelemetryState>, port: u16) {
-    rouille::start_server(("localhost", port), move |request| {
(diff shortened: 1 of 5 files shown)


telemetry/src/server.rs
```diff
@@ -21,7 +21,7 @@ impl Route {
 }
 
 pub fn run(telemetry_state: Arc<TelemetryState>, port: u16) {
-    rouille::start_server(("localhost", port), move |request| {
+    rouille::start_server(("0.0.0.0", port), move |request| {
```
mircea-c, line 24: This really should be a config option someplace. Hardcoding either of these values is bad news when running in production.

andynog: OK, thanks @mircea-c, we do have a config that you can only specify the port. 

```
[telemetry]
enabled = true
port = 3001
```
 should we add a host parameter ? or replace the port with an `url` parameter where you specify a full URL ?

mircea-c: Issues caused by configuration are easy to fix. Hardcoding this to accept requests from every network interface on the machine can become a security issue. 

Telemetry should be considered a separate service from the relayer (which also shouldn't just listen to all interfaces by default) and have separate networking / access configuration. The operator needs to be allowed to decide what network interface / port the telemetry will be exposed on.

andynog: OK, that's fair. The telemetry might run from a server with multiple NICs

do you have a preference for having just one parameter like 
```
[telemetry]
enabled = true
address = '127.0.0.1:3001'
```
or
```
[telemetry]
enabled = true
port = 3001
host = '127.0.0.1'
```

mircea-c: I would prefer one parameter like `address`, but that's just me :slightly_smiling_face:
Access is fine as is, I wouldn't expect anything else.

romac: I would rather split the port and host into two options, as it makes clear that the port is required, otherwise it's too easy to just leave it out.

romac: I took the liberty of doing this in https://github.com/informalsystems/ibc-rs/pull/1033/commits/428c45a0ea805fa804495366c96c9111c61d087e.

andynog: Lgtm! Thanks!

andynog: @romac, testing locally, one use case is that if something is already running on the specified port (e.g. `3001`), when you start the relayer with `Hermes start` you get this message. Is this OK or should we not start it ?

```
./target/debug/hermes -c ./config.toml start
The application panicked (crashed).
Message:  Failed to start server: Os { code: 48, kind: AddrInUse, message: "Address already in use" }
Location: $HOME/.cargo/registry/src/github.com-1ecc6299db9ec823/rouille-3.1.1/src/lib.rs:221
```

romac: I think it's fine to crash, as otherwise one may not realize that the port was already in use and end up running Hermes without telemetry, effectively. We could perhaps show a nice error message and gracefully abort rather than panic?

romac: Oh and we also need a changelog entry for this.

andynog: I'd say just showing a friendly message saying telemetry not enable because port is in use should be fine. Not showing `The application panicked` should be better

ancazamfir: > Oh and we also need a changelog entry for this.

Isn't telemetry added in this release? I would just keep one entry for adding the feature and maybe list all issues that created and fixed things.

ancazamfir: re: "Address already in use" ...let's exit gracefully. The user should then either disable telemetry or try with a different address/port.

andynog (commented):
lgtm, the only suggestion is that if a more meaningful message could be added when exiting. The message "Address in use" does not have a context and for an operator is not helpful because it doesn't say the telemetry server failed to start because it's trying to use a port which is already allocated. If there's a way to at least say `telemetry server` or something like that I think that would be helpful than just bubbling up the error from the web server

```
Error: Hermes failed to start, last error: Address already in use (os error 48)
```

romac:
It says right before exiting with this message:

> Jun 02 22:32:48.130 ERROR ibc_relayer_cli::commands::start: telemetry service failed to start: Address already in use (os error 48)

which is not perfect but at least provides some more context. Is that enough or would you want it to also mention this in the final error? Maybe we could introduce an enum to capture the server error and then print it accordingly?

andynog:
> It says right before exiting with this message:
> 
> > Jun 02 22:32:48.130 ERROR ibc_relayer_cli::commands::start: telemetry service failed to start: Address already in use (os error 48)
> 
> which is not perfect but at least provides some more context. Is that enough or would you want it to also mention this in the final error? Maybe we could introduce an enum to capture the server error and then print it accordingly?



> It says right before exiting with this message:
> 
> > Jun 02 22:32:48.130 ERROR ibc_relayer_cli::commands::start: telemetry service failed to start: Address already in use (os error 48)
> 
> which is not perfect but at least provides some more context. Is that enough or would you want it to also mention this in the final error? Maybe we could introduce an enum to capture the server error and then print it accordingly?

Sorry, my bad. Indeed it says. I thought that ERROR line was only shown depending on the `log_level` parameter in the config but that's not the case, it will show in any case. So that's all good. 

Thanks!

adizere (approved):

Issue #1001, opened 2021-05-26
ChainEndpoint trait refactor (first phase)


adizere:
## Crate

`ibc`

## Summary

We would like to make the methods in the `Chain` trait more consistent. Refactoring this trait will probably take multiple rounds of work. This issue covers a first phase, which comprises two elements that we'd like to visit:

1. Fix some of the returned values for `query_*` methods, to make them consistent in the way in which they report non-existent values versus how they report a runtime/connection error.
2. Ensure consistency across methods with respect to their input arguments.

Both of these are described in more detail below.

## Problem Definition

The [`Chain` trait](https://github.com/informalsystems/ibc-rs/blob/041ef642f7080e1faea60fc1325abde5afcf5c0a/relayer/src/chain.rs#L68) is a major interface in the `ibc-relayer` library. The intention behind this trait is to capture all the dependencies which Hermes has towards any kind of chain (be it Cosmos SDK or something else). As it is a major connecting point between relaying logic and chain logic, this trait gets changed relatively often, and its design drifted to become less consistent over time.

Here are some examples of inconsistencies.

### Return values for query methods

The [`query_connection`](https://github.com/informalsystems/ibc-rs/blob/041ef642f7080e1faea60fc1325abde5afcf5c0a/relayer/src/chain.rs#L170) method returns a domain type that is possibly in state `Uninitialized`, and this signifies that the Connection object does not exist on-chain. We currently have to do this domain type interpretation to report to the user correctly that an object does not exist, e.g.:
https://github.com/informalsystems/ibc-rs/blob/26ef2c6fcc2527998be1a206292ad870191b42c0/relayer-cli/src/commands/query/connection.rs#L52-L65

An alternative to returning an `Uninitialized` domain type is to return a value of type `Result<Option<ConnectionEnd>, Error`. 
A returned `Ok(None)` indicates that the query suceedeed but nothing was found on-chain. It's not clear if this alternative design fits all queries, however, so the other queries methods should be visited in light of this proposal, and if the design is sound then we should modify the queries correspondingly.

### Consistency of input arguments

[...]

PR #861, opened 2021-04-28
Return PortId information in query_channels

soareschen:
Closes: #840

## Description

______

For contributor use:

- [ ] Updated the __Unreleased__ section of [CHANGELOG.md](https://github.com/informalsystems/ibc-rs/blob/master/CHANGELOG.md) with the issue.
- [ ] If applicable: Unit tests written, added test to CI.
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Updated relevant documentation (`docs/`) and code comments.
- [ ] Re-reviewed `Files changed` in the Github PR explorer.

Issue #745, opened 2021-03-16
Introduce `(PortId, ChannelId)` domain type


adizere:
## Crate

`ibc`

## Summary

Since we started working on ICS04 handlers(https://github.com/informalsystems/ibc-rs/pull/452) we noticed that there is a recurring need to encapsulate the tuple `(PortId, ChannelId)` as a type. This tuple is needed in error types:

https://github.com/informalsystems/ibc-rs/blob/d6250014e8d8fcddde651f9e3ceb6be5c93cbeae/modules/src/application/ics20_fungible_token_transfer/error.rs#L19-L20

or as parameters to structures:

https://github.com/informalsystems/ibc-rs/blob/71d9baa6a2eed5a9286f9fe88f1cfacbd51e19d8/modules/src/mock/context.rs#L86-L86

or as parameters to methods:

https://github.com/informalsystems/ibc-rs/blob/71d9baa6a2eed5a9286f9fe88f1cfacbd51e19d8/modules/src/mock/context.rs#L279-L282

So there's ample reasons to introduce a new type for capturing the two.

## Proposal

The new type should probably reside in ICS04. Potential names:

```rust
pub struct PortChannel(PortId, ChannelId);
pub struct BoundChannel(PortId, ChannelId);
pub struct AuthorizedChannel(PortId, ChannelId);
```

____

#### For Admin Use

- [X] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

romac:
How about `IbcSocket`? Much like a TCP socket consists of an IP address and a port number.

adizere:
Blocked on https://github.com/cosmos/ibc/issues/583

mzabaluev:
We can invent the type locally (already have as `PortChannelId`?) and use it in any internal or domain/idiomatic wrapper interfaces without the need for protobuf to be modified. In fact, a message type just to group the two fields consumes more space in protobuf wire encoding than the same two fields encoded inline.

> Potential names:

`Socket`?

> `BoundChannel(PortId, ChannelId)`

This looks too "stateful": there may not be a bound channel for this pair of IDs, as there isn't in your example of an error where this type could be used.

romac:
I kinda like _socket_ but i am afraid it's already too overloaded.

I agree that `BoundChannel` is too stateful.

plafer:
[...]

PR #1052, opened 2021-06-07
Refactor (PortId, ChannelId) to use PortChannelId

soareschen:
Partially Closes: #1001

## Description

This PR replaces the use of `(PortId, ChannelId)` to use the `PortChannelId` struct.

______

For contributor use:

### DIFF
--- a/modules/src/application/ics20_fungible_token_transfer/relay_application_logic/send_transfer.rs
+++ b/modules/src/application/ics20_fungible_token_transfer/relay_application_logic/send_transfer.rs
@@ -16,11 +16,9 @@ where
 {
     let port_channel_id = PortChannelId::new(msg.source_port.clone(), msg.source_channel.clone());
 
-    let source_channel_end = ctx
-        .channel_end(&port_channel_id)
-        .ok_or_else(|| {
-            Kind::ChannelNotFound(msg.source_port.clone(), msg.source_channel.clone())
-        })?;
+    let source_channel_end = ctx.channel_end(&port_channel_id).ok_or_else(|| {
+        Kind::ChannelNotFound(msg.source_port.clone(), msg.source_channel.clone())
+    })?;
 
     let destination_port = source_channel_end.counterparty().port_id().clone();
     let destination_channel = source_channel_end
@@ -31,11 +29,9 @@ where
         })?;
 
     // get the next sequence
-    let sequence = ctx
-        .get_next_sequence_send(&port_channel_id)
-        .ok_or_else(|| {
-            Kind::SequenceSendNotFound(msg.source_port.clone(), msg.source_channel.clone())
-        })?;
+    let sequence = ctx.get_next_sequence_send(port_channel_id).ok_or_else(|| {
+        Kind::SequenceSendNotFound(msg.source_port.clone(), msg.source_channel.clone())
+    })?;
 
     //TODO: Application LOGIC.
 
--- a/modules/src/application/ics20_fungible_token_transfer/relay_application_logic/send_transfer.rs
+++ b/modules/src/application/ics20_fungible_token_transfer/relay_application_logic/send_transfer.rs
@@ -5,6 +5,7 @@ use crate::handler::HandlerOutput;
 use crate::ics04_channel::handler::send_packet::send_packet;
 use crate::ics04_channel::packet::Packet;
 use crate::ics04_channel::packet::PacketResult;
+use crate::ics24_host::identifier::PortChannelId;
 
 pub(crate) fn send_transfer<Ctx>(
     ctx: &Ctx,
@@ -13,8 +14,10 @@ pub(crate) fn send_transfer<Ctx>(
 where
     Ctx: Ics20Context,
 {
+    let port_channel_id = PortChannelId::new(msg.source_port.clone(), msg.source_channel.clone());
+
     let source_channel_end = ctx
-        .channel_end(&(msg.source_port.clone(), msg.source_channel.clone()))
+        .channel_end(&port_channel_id)
         .ok_or_else(|| {
             Kind::ChannelNotFound(msg.source_port.clone(), msg.source_channel.clone())
         })?;
@@ -29,7 +32,7 @@ where
 
     // get the next sequence
     let sequence = ctx
-        .get_next_sequence_send(&(msg.source_port.clone(), msg.source_channel.clone()))
+        .get_next_sequence_send(&port_channel_id)
         .ok_or_else(|| {
             Kind::SequenceSendNotFound(msg.source_port.clone(), msg.source_channel.clone())
         })?;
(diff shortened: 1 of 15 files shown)


romac:
@adizere Can you describe the reasoning behind the suggested change that this PR addresses?

> Some improvements can be done for channel-related methods, specifically when both a ChannelId and a PortId are provided as input (for example here), and instead of these two types a single PortChannelId should be used.

I am not sure I see what using a struct over a tuple brings us here?

---

The problem I have with using `PortChannelId` instead of `(PortId, ChannelId)` is that in both cases we often need to clone the `PortId` and `ChannelId` when only a reference would be enough, eg. calling

```rust
fn get_something(key: &(PortId, ChannelId)) -> Foo {}
```

when one has owned `PortId` and `ChannelId` requires doing:

```rust
get_something(&(port_id.clone(), channel_id.clone())
```

Same with

```rust
fn get_something(key: &PortChannelId) -> Foo {}

get_something(& PortChannelId::new(port_id.clone(), channel_id.clone()))
```

If we stick with tuples (which I believe are fine for the use case of bundling two values to be used as a lookup key), we could instead have this API:

```rust
fn get_something(key: (&PortId, &ChannelId)) -> Foo {}

get_something((&port_id, &channel_id))
```

and avoid unnecessary clones.

Admittedly, we could introduce a `PortChannelIdRef` struct which holds references to its components, but again I am not sure what it really gets us over a tuple:

```rust
struct PortChannelIdRef<'a> {
  port_id: &'a PortId,
  channel_id: &'a ChannelId
}
```

I guess the question of *struct vs tuple* and *owned data vs references* are orthogonal, but I feel like using tuples+references is the simpler solution here. Happy to hear otherwise though.

@soareschen What are your thoughts on this? My apologies for nitpicking your PR when the culprit here is me not catching this and raising this discussion when the issue was opened.

romac (changes requested):
https://github.com/informalsystems/ibc-rs/pull/1052#issuecomment-855816230

soareschen:
I have actually tried to compare the two approaches, and I don't find significant advantage between either of them. Either way we are doing cloning in quite a number of places already, and I think it is premature to try to optimize the clones away.

The main issue I see is in the final implementation of the use of the port channel ID, which are in structures such as `HashMap<(PortId, ChannelId, Sequence), Receipt>`. So even if we pass by reference in the function calls, eventually clones have to be made in order to construct the tuple as a value and pass a single reference to `HashMap`. As a result, it is always either the caller having to do the clone by constructing `(PortId, ChannelId, Sequence)` and pass by reference, or the callee having to do the clone by accepting `&PortId, &ChannelId, &Sequence` and then construct the tuple `(PortId, ChannelId, Sequence)`.

We probably don't need to worry about the cost of cloning for now, since they are mainly consist of short string identifiers. It should be unlikely we will hit performance bottleneck on that part for the time being.

soareschen:
Btw there will be less need for cloning once we refactor structs like `Packet` to use `PortChannelId` directly instead of having separate `PortId` and `ChannelId` fields. Do you think we should also refactor that in this PR to make it slightly more efficient? Or I can do it in a follow up PR.

romac:
> The main issue I see is in the final implementation of the use of the port channel ID, which are in structures such as HashMap<(PortId, ChannelId, Sequence), Receipt>. So even if we pass by reference in the function calls, eventually clones have to be made in order to construct the tuple as a value and pass a single reference to HashMap.

Right, I hadn't noticed this. Nevermind then :)

> We probably don't need to worry about the cost of cloning for now, since they are mainly consist of short string identifiers. It should be unlikely we will hit performance bottleneck on that part for the time being.

I am not very concerned either, as you say we already clone a lot of stuff. Still, given that the `ibc` crate will eventually be `no-std + alloc` compatible, we might want to give it some special attention when it comes to allocating memory, as one might eventually want to use it in memory constrained environments, or in environments where memory allocations are particularly costly.

I still don't really see what we gain by grouping the port and channel id in a single (awkwardly named*) struct where a tuple, but will not oppose it as my other points are moot.

\* the fact that there is no obvious good name for this struct (at least to me) kinda tells me that it's not needed, but I am happy to proven wrong if anyone finds a good name. It's not required for this PR to be merged though.

romac:
Actually, after looking at the code again, I wonder why we don't just do 

```rust
fn get_next_sequence_send(&self, port_id: &PortId, channel_id: &ChannelId) -> Option<Sequence>
```

instead of using either a tuple (like we do currently) or a struct (like this PR suggests)?

This is more idiomatic, and while it won't avoid the problem with the `HashMap` that you mention, the latter is an implementation detail which should not interfere with the abstract API.

Sorry for being a pain about this, and changing my mind, but as the `ibc` crate will likely end up being the most widely used crate in the ibc-rs project, I would like to make sure we expose nice, idiomatic, easy to use APIs.

adizere:
The idea of the type `PortChannel` replacing the two types `PortId + ChannelId` came up some time ago, when Cezara first implemented the ICS04 handlers and we noticed that theres some potential for capturing these two types and reducing the number of parameters. Here's some background: #745. Based on the discussion in that issue, I recently opened an issue in cosmos/ibc (https://github.com/cosmos/ibc/issues/583) to implement the change upstream. The upstream change is pending, and not clear if it's going to happen at all, so maybe that will help inform us here if we should do it in the first place.

The rationale is to simplify some interfaces & calls. It would reduce the number of parameters, and would make it explicit that there is a special relation between a `Port` and a `Channel` identifier.

The idea of the new type is not set in stone, of course. @romac thanks for insisting on this, since it might actually prove that the original definitions from the spec ([core/ics-004](https://github.com/cosmos/ibc/tree/master/spec/core/ics-004-channel-and-packet-semantics#store-paths))
```rust
fn get_next_sequence_send(&self, port_id: &PortId, channel_id: &ChannelId) -> Option<Sequence>
```

... are actually the best way to do it.


> Btw there will be less need for cloning once we refactor structs like Packet to use PortChannelId directly instead of having separate PortId and ChannelId fields. Do you think we should also refactor that in this PR to make it slightly more efficient? 

I agree with you, Soares. But let's first try reach a consensus whether this new type is valuable at all. And if so, let's modify the `Packet` (and others) separately.

soareschen:
No problem, we can discuss further on how to better design the interface. I think with the PR in place we are better informed what the changes would look like.

I think an advantage of having a single type like `PortChannelId` is that it makes it more unlikely to accidentally mix up the use of different `portId` / `channelId` pairs. Other than that, it was introduced in #861 because we need it as a newtype for custom derivation the `Serialize` and `Deserialize` traits, and we can't implement that with the tuple type.

Otherwise, I am fine with passing them as two separate parameters as well.

romac:
Thanks both, you are both making compelling arguments. I like the `Socket` name mentioned in the `cosmos/ibc` issue. Having a good name for the compound data structure makes it more compelling to me, as does the argument for it making it harder to mix up different port and channel combinations. I vote for this change, but would suggest we wait until it gets upstream as well before merging this. What do you think?

adizere:
Closing this with note:

- wait for ibc-go to adopt `Socket` terminology
- redo the PR with the new name

Issue #1009, opened 2021-05-27
Option type should be used with non-zero Height


soareschen:
## Crate

`ibc-rs`

## Summary

Currently the `Height` type uses a special `Height::zero()` value to denote the absence of height value. This causes ad hoc checks of whether a height is zero in various places. We should instead use `Option<Height>` on places where the height value may be optional, and require non-zero value in the `Height` constructor.

## Acceptance Criteria

- Optional height parameters should use `Option<Height>` instead of `Height`.
- Checking of `height == Height::zero()` is replaced with Option matching.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

plafer:
My only potential concern with this solution is that our `Height` type would no longer be a partially ordered set over its entire domain, which is required in ICS 2, assuming that `None` serves as the zero-element. `Option<Height>` would correspond to `Height` in the spec, assuming a proper implementation for `cmp()`. 

However, looking at the code, the zero height is often used as a special value, sometimes even explicitly mentioned in [comments](https://github.com/informalsystems/ibc-rs/blob/6b536c7cee33bb46395f35419e25a8da913e6785/modules/src/core/ics03_connection/msgs/conn_open_try.rs#L41). This is confusing to me as I don't see any mention in the spec that the zero height should be treated differently.

What am I missing here? I definitely agree that `Option<Height>` + pattern matching should be used for treating the special case. However, I'm concerned that we might be introducing a bug w.r.t a "real" `height == 0` state, since I don't understand how the zero height is used in the implementation, and how that relates back to the spec.

hu55a1n1:
Note that `Height::zero()` can also mean `latest` height, especially in the context of Tendermint RPC -> https://docs.tendermint.com/master/rpc/#/Info/block

romac:
We should perhaps encode this across the whole codebase as 

```rust
enum Height {
  Latest,
  Specific(u64)
}
```

and then only convert `Height::Latest` to `0` or `None` on a case-by-case basis, depending on what the underlying API expects to mean "latest height".

plafer:
It seems like a bunch of events use the zero height as well:

[...]

PR #1059, opened 2021-06-07
Small refactor of `ForeignClient::build_update_client*` methods

romac:
## Description

- Improve legibility of loop condition when waiting for source chain to reach `target_height`
- Use `Option<Height>::None` instead of `Height::zero()` in ForeignClient::build_update_client*` methods

______

For contributor use:

### DIFF
--- a/relayer/src/foreign_client.rs
+++ b/relayer/src/foreign_client.rs
@@ -400,7 +400,7 @@ impl ForeignClient {
         &self,
         target_height: Height,
     ) -> Result<Vec<Any>, ForeignClientError> {
-        self.build_update_client_with_trusted(target_height, Height::zero())
+        self.build_update_client_with_trusted(target_height, None)
     }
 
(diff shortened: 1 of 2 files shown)


relayer/src/foreign_client.rs
```diff
@@ -400,24 +400,27 @@ impl ForeignClient {
...
+
         // Wait for source chain to reach `target_height`
-        while self.src_chain().query_latest_height().map_err(|e| {
-            ForeignClientError::ClientUpdate(format!(
-                "failed fetching src chain latest height with error: {}",
-                e
-            ))
-        })? < target_height
-        {
+        while latest_height()? < target_height {
```
romac, line 423: @ancazamfir Do we really want to wait here in a loop? Shouldn't this be handled via retries at the worker level?

romac:
Let's address this (#1009) all at once in another PR.

PR #1072, opened 2021-06-10
Propagate LightClient errors except IO as recoverable error all the way up to update client command

### LINKED ISSUE #712: Relayer error handling specification

soareschen:
Partial fix for #712

## Description

This is another attempt to fix PR #712 with a much more narrowed focus. The goal is:

- To classify the a single class of error, `tendermint_light_client::errors::ErrorKind`.
  - The variant `ErrorKind::Io` is recoverable.
  - All other variants such as `ErrorKind::Store`, `ErrorKind::NoPrimary`, etc are non recoverable.
 
- The retry logic is implemented at the top level in one of the command line.
  - I have implemented the retry logic in the `update client` command, as I forgot if we agreed on another command.
  - The command line have a new `-r` flag to activate the retry logic

For the retry logic to propagate, I have to modify outer error types such as `relayer::error::Kind` and `foreign_client::ForeignClientError` so that they can return the appropriate boolean when queries whether the error is retryable.

While the retry logic is working, the naive implementation of the `is_retryable` method is unfortunately not robust, as all other errors are currently being classified as being retryable. For example, the `ics02_client::error::Kind::ClientNotFound` error is by default also retryable. So when `update client` is now called with invalid client ID, the command line just naively keep retrying:

```
$ ./target/debug/hermes update client ibc-0 07-tendermint-1
Jun 10 10:06:06.202  INFO ibc_relayer_cli::commands: using default configuration from '~/.hermes/config.toml'
Jun 10 10:06:06.215 DEBUG ibc_relayer::event::monitor: starting event monitor chain.id=ibc-0
Error: relayer error: Query error occurred (failed to query for client state): error converting message type into domain type: the client state was not found
ubuntu@development:~/informal/ibc-rs$ ./target/debug/hermes update client ibc-0 07-tendermint-1 -r
Jun 10 10:06:15.709  INFO ibc_relayer_cli::commands: using default configuration from '~/.hermes/config.toml'
Jun 10 10:06:15.720 DEBUG ibc_relayer::event::monitor: starting event monitor chain.id=ibc-0
Jun 10 10:06:18.021 ERROR ibc_relayer_cli::commands::tx::client: Encounter recoverable error: relayer error: Query error occurred (failed to query for client state): error converting message type into domain type: the client state was not found. retrying update
Jun 10 10:06:19.035 DEBUG ibc_relayer::event::monitor: starting event monitor chain.id=ibc-0
Jun 10 10:06:19.042 ERROR ibc_relayer_cli::commands::tx::client: Encounter recoverable error: relayer error: Query error occurred (failed to query for client state): error converting message type into domain type: the client state was not found. retrying update
Jun 10 10:06:20.052 DEBUG ibc_relayer::event::monitor: starting event monitor chain.id=ibc-0
Jun 10 10:06:20.057 ERROR ibc_relayer_cli::commands::tx::client: Encounter recoverable error: relayer error: Query error occurred (failed to query for client state): error converting message type into domain type: the client state was not found. retrying update
...
``` 

The purpose of this PR is to demonstrate the issue of naively classifying only some errors as being retryable. As we can see in the code changes, even a limited error propagation like this is pretty non-trivial. It will also take significant effort to propagate other errors like `ics02_client::error::Kind` upward. And if we continue down the rabbit hole, we will end up with the same situation as in #950.

______

For contributor use:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1464,6 +1464,7 @@ dependencies = [
  "once_cell",
  "prost",
  "prost-types",
+ "retry",
  "serde",
  "serde_derive",
  "serde_json",
--- a/relayer-cli/Cargo.toml
+++ b/relayer-cli/Cargo.toml
@@ -46,6 +46,7 @@ crossbeam-channel = "0.5.1"
 subtle-encoding = "0.5"
 dirs-next = "2.0.0"
 itertools = "0.10.1"
+retry = { version = "1.2.1", default-features = false }
 
 [dependencies.tendermint-proto]
 version = "=0.19.0"
--- a/relayer-cli/src/cli_utils.rs
+++ b/relayer-cli/src/cli_utils.rs
@@ -39,15 +39,16 @@ pub fn spawn_chain_runtime(
     config: &Config,
     chain_id: &ChainId,
 ) -> Result<Box<dyn ChainHandle>, Error> {
-    let chain_config = config
-        .find_chain(chain_id)
-        .cloned()
-        .ok_or_else(|| format!("missing chain for id ({}) in configuration file", chain_id))
-        .map_err(|e| Kind::Config.context(e))?;
+    let chain_config = config.find_chain(chain_id).cloned().ok_or_else(|| {
+        Kind::Config.context(format!(
+            "missing chain for id ({}) in configuration file",
+            chain_id
+        ))
+    })?;
 
     let rt = Arc::new(TokioRuntime::new().unwrap());
     let chain_res = ChainRuntime::<CosmosSdkChain>::spawn(chain_config, rt)
-        .map_err(|e| Kind::Runtime.context(e));
+        .map_err(|e| Kind::Relayer(e.kind().clone()).context(e));
 
     let handle = chain_res.map(|(handle, _)| handle)?;
 
--- a/relayer-cli/src/commands/tx/client.rs
+++ b/relayer-cli/src/commands/tx/client.rs
@@ -4,6 +4,7 @@ use ibc::events::IbcEvent;
 use ibc::ics02_client::client_state::ClientState;
 use ibc::ics24_host::identifier::{ChainId, ClientId};
 use ibc_relayer::foreign_client::ForeignClient;
+use ibc_relayer::util::retry::retry_recoverable_with_index;
 
 use crate::application::app_config;
 use crate::cli_utils::{spawn_chain_runtime, ChainHandlePair};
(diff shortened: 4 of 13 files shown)


adizere (commented):
I have two observations, and a question.

1.

If I understand correctly, we are taking the approach that an error carries the information of whether there should be a retry or not for that error. But typically, it is the caller who gets to decide if an error is retryable or not (it's not the error itself who decides). This insight is relevant because I can imagine scenarios where a method `do_x` may generate a specific error, and the caller is the only one to have the right context and decide if it should retry `do_x` or not (of course, depending also on the error, but not only).

It looks like the reason why the `RetryableError` trait makes sense for Hermes (instead of letting the caller match on the error and always decide) is because we want Hermes to retry in a prevalent amount of cases. In other words, Hermes will comprise a _lot_ of `retry { } ` blocks, and matching on errors is too tedious/repetitive. Is this intuition correct?

2.

Suppose that we take the `RetryableError` approach, as in this experimental PR. We can think about two opposite strategies: errors are by default retryable, or are non-retryable by default. If we want to make the relayer as "stubborn" as possible, we choose to retry by default. If we'd prefer a "safety over liveness" mindset, then non-retryable by default. The latter might be better, but it's not obvious.

Whichever way we go forward, what about encode the outcome of these discussions in an ADR, which could also include some general coding practices w.r.t. errors, e.g., ["errors should be logged when they are handled"](https://www.lpalmieri.com/posts/error-handling-rust/).

soareschen:
> If I understand correctly, we are taking the approach that an error carries the information of whether there should be a retry or not for that error. But typically, it is the caller who gets to decide if an error is retryable or not (it's not the error itself who decides).

Yes, ideally the error case behavior should be handled by the caller. However this usually requires the error type to be simple enough so that the caller is not overloaded with too much information about how to handle the error. In the case of ibc-rs, currently we have probably hundreds of sub errors with no clear classification of what each of these errors really mean. So it may be more efficient to have a single source of truth to describe the property of the errors, at least for now. Note that the classification is really meant only for use by the relayer process, and we probably want to keep that logic out from the `ibc-rs` crate so that external users are not affected by this.

When there are many kinds of sub errors, a possible improvement might be to adopt error code schemes like the HTTP status code. With something like that, we can for example treat errors 4XX as being non-retryable, 5XX being retryable, etc.

> If we want to make the relayer as "stubborn" as possible, we choose to retry by default. If we'd prefer a "safety over liveness" mindset, then non-retryable by default. The latter might be better, but it's not obvious.

Yes, it is not very clear which strategy is better. Perhaps an alternative is to do a single big merge once we have managed to classify the errors. #988 is about half way doing that, and with all errors refactored, we should be able to have a better idea of how the errors are really being propagated. Nevertheless, the current behavior of the relayer is to retry on all errors, so we are kind of default on the stubborn case.

soareschen:
I have implemented a specific workaround to mark all errors from `ibc::ics02_client::error::Error` as being non-retryable, with some caveats:

- The only place that currently correctly classify errors coming from ICS02 is in `query_client_state`. All other existing code in the relayer still classify the errors as general errors such as `Kind::Grpc`. This means that most of the ICS02 errors will still be retried, because the relayer still see them as different errors.

- We are only skipping retry on specifically ICS02 errors. There are still many errors out there that should be non-retryable, such as connection/channel not found errors. The scope for properly identifying all possible errors is too large, and we will leave it to #988 to properly classify the remaining errors.

I have also done manual testing to show that the errors are classified correctly. The first case is when querying on non-existing client:

```bash
$ RUST_LOG=debug ./target/debug/hermes update client ibc-0 07-tendermint-not-found -r
Jun 15 08:26:03.691  INFO ibc_relayer_cli::commands: using default configuration from '/home/ubuntu/.hermes/config.toml'
Jun 15 08:26:03.705 DEBUG ibc_relayer::event::monitor: starting event monitor chain.id=ibc-0
Jun 15 08:26:05.792  WARN ibc_relayer::util::retry: non-recoverable error happened: relayer error: ICS 02 error: the client state was not found: the client state was not found. skipping retry
Error: relayer error: ICS 02 error: the client state was not found: the client state was not found
```

The second case is retrying when gaiad is killed:

```bash
$ killall gaiad
$ RUST_LOG=debug ./target/debug/hermes update client ibc-0 07-tendermint-0 -r
Jun 15 08:26:34.996  INFO ibc_relayer_cli::commands: using default configuration from '/home/ubuntu/.hermes/config.toml'
Jun 15 08:26:36.555  WARN ibc_relayer::util::retry: operation run_update_client failed with recoverable error: relayer error: RPC error to endpoint http://127.0.0.1:26657/: error trying to connect: tcp connect error: Connection refused (os error 111) (code: 0). retrying at count 2
Jun 15 08:26:37.561  WARN ibc_relayer::util::retry: operation run_update_client failed with recoverable error: relayer error: RPC error to endpoint http://127.0.0.1:26657/: error trying to connect: tcp connect error: Connection refused (os error 111) (code: 0). retrying at count 3
Jun 15 08:26:38.568  WARN ibc_relayer::util::retry: operation run_update_client failed with recoverable error: relayer error: RPC error to endpoint http://127.0.0.1:26657/: error trying to connect: tcp connect error: Connection refused (os error 111) (code: 0). retrying at count 4
...
```

And finally is to not retry on light client errors. For that I replace the `trusting_period` field in the Hermes config from `14days` to `30s`, and then try to update the client after 30 seconds:

```bash
$ sed -i "s|trusting_period = '14days'|trusting_period = '30s'|" config.toml
$ ./scripts/dev-env config.toml ibc-0 ibc-1
$ sleep 30
$ RUST_LOG=debug ./target/debug/hermes -c config.toml update client ibc-0 07-tendermint-0 -r
Jun 15 08:16:56.712  INFO ibc_relayer_cli::commands: using default configuration from '/home/ubuntu/.hermes/config.toml'
Jun 15 08:16:56.725 DEBUG ibc_relayer::event::monitor: starting event monitor chain.id=ibc-0
Jun 15 08:16:56.736 DEBUG ibc_relayer::event::monitor: starting event monitor chain.id=ibc-1
Jun 15 08:16:58.866  WARN ibc_relayer::util::retry: non-recoverable error happened: foreign client error: error raised while updating client: failed building header with error: Light client error for RPC address ibc-1: trusted state outside of trusting period. skipping retry
Error: foreign client error: error raised while updating client: failed building header with error: Light client error for RPC address ibc-1: trusted state outside of trusting period
```

I strongly feel that we need to improve the current practices of ad hoc manual testing and replace them with some automated unit testing. And by that I do not mean doing end-to-end tests that replicate the above cases. Rather, all these should be tested in unit tests without having to spawn `gaiad` chains.

adizere:
Postponing these changes until there is more robust testing in place (unit tests + more flexible E2E tests). Original issue will remain open #712.

Issue #1068, opened 2021-06-09
Tutorial commands give error "cannot read key file at..."


okwme:
## Crate
hermes 0.4.0

## Summary of Bug
Following instructions in the tutorial. We're on the [identifiers](https://hermes.informal.systems/tutorials/local-chains/identifiers.html) section. Running the following command:
```
hermes tx raw create-client ibc-1 ibc-0
```
gives the following error:
```
Jun 09 12:50:55.979  INFO ibc_relayer_cli::commands: using default configuration from '/Users/billy/.hermes/config.toml'
Error: tx error: error raised while creating client: failed while fetching the destination chain (ibc-1) signer: Keybase error: key store error: cannot read key file at '/Users/billy/.hermes/keys/ibc-1/keyring-test/testkey.json'
```
here's the tree of my `~/.hermes` directory:
```
> tree ~/.hermes
/Users/billy/.hermes
├── config.toml
└── keys
    ├── ibc-0
    │   └── keyring-test
    │       ├── testkey.json
    │       └── user2.json
    └── ibc-1
        └── keyring-test
            ├── testkey.json
            └── user2.json
```
And here's the tree of my `ibc-rs/data` directory:
```
> tree data -L 2
data
├── ibc-0
│   ├── config
│   ├── data
│   ├── keyring-test
│   ├── user2_seed.json
│   ├── user_seed.json
│   └── validator_seed.json
├── ibc-0.log
├── ibc-1
│   ├── config
│   ├── data
│   ├── keyring-test
│   ├── user2_seed.json
│   ├── user_seed.json
│   └── validator_seed.json
└── ibc-1.log
```
They look the same as the tutorial suggests, except the presence of the user-2 inside of the `~/.hermes` directory.

## Version
hermes 0.4.0

## Steps to Reproduce
Follow the tutorial. Both @jessicadosseh and I are having this error.

## Acceptance Criteria
Updates to the tutorial with instructions that succeed.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

andynog:
Helping them troubleshoot the error

adizere:
Confirm that @greg-szabo was encountering the same problem on a testnet with a Crypto.org Chain app.

The problem was manifesting as follows:

[...]

PR #1076, opened 2021-06-11
Update tutorial to ensure users check out the current release

romac:
Closes: #1068 

## Description

Update the *Two local chains* tutorial to ensure users check out the latest release of the repository, rather than work out of `master`. See the linked issue for a description of the issues encountered when one uses the `dev-env` script from `master` with Hermes 0.4.0.


adizere (approved):
Looks good!

Should we also update the changelog?

romac:
> Should we also update the changelog?

Since the guide is not versioned, I don't think adding a changelog entry for it makes sense at the moment.

PR #1078, opened 2021-06-11
Use flex-error for ics20 and ics07 

DaviRain-Su:
______

For contributor use:


DaviRain-Su:
@soareschen I modified some of the remaining ones, and I'll add the no_std feature later.

modules/src/lib.rs
```diff
@@ -26,6 +27,7 @@
 //! - ICS 26: Routing
 //! - Applications:
 //!    - ICS 20: Fungible Token Transfer
+extern crate no_std_compat as std;
```
soareschen, line 30: Can we use `core` and `alloc` directly without going through `no_std_compat`? The library has not been updated for a while, and very unfortunately the author seems to have [passed away](https://www.redox-os.org/news/open-source-mental-health/).

modules/src/tx_msg.rs
```diff
@@ -1,9 +1,13 @@
...
+use std::{
+    vec::Vec,
+    string::String,
+};
 
 use crate::ics24_host::error::ValidationError;
 
 pub trait Msg: Clone {
-    type ValidationError: std::error::Error;
+    type ValidationError: flex_error::StdErr;
```
soareschen, line 10: `flex-error` itself does not really define any error type. It looks to me that the error type here should be `crate::ics24_host::error::ValidationError`.

modules/src/ics24_host/error.rs
```diff
@@ -1,31 +1,33 @@
-use anomaly::{BoxError, Context};
-use thiserror::Error;
+use displaydoc::Display;
+use std::string::String;
 
-pub type ValidationError = anomaly::Error<ValidationKind>;
+pub type ValidationError = ValidationKind;
 
-#[derive(Clone, Debug, Error, PartialEq, Eq)]
+impl flex_error::StdErr for ValidationKind {}
```
soareschen, line 6: This doesn't really make sense.

modules/src/ics24_host/error.rs
```diff
@@ -1,31 +1,33 @@
-use anomaly::{BoxError, Context};
-use thiserror::Error;
+use displaydoc::Display;
+use std::string::String;
 
-pub type ValidationError = anomaly::Error<ValidationKind>;
+pub type ValidationError = ValidationKind;
```
soareschen, line 4: Can we also use `define_error!` to redefine the `ValidationError` here?

soareschen (changes requested):
I have given some feedback. Also the CI tests are failing and I can't build the code.

modules/src/ics24_host/error.rs
DaviRain-Su: `impl ValidationKind {
    pub fn contains_separator(id: String) -> Self {
        Self::ContainsSeparator { id }
    }

    pub fn invalid_length(id: String, length: usize, min: usize, max: usize) -> Self {
        Self::InvalidLength {
            id,
            length,
            min,
            max,
        }
    }

    pub fn invalid_character(id: String) -> Self {
        Self::InvalidCharacter { id }
    }

    pub fn empty() -> Self {
        Self::Empty
    }

    pub fn chain_id_invalid_format(id: String) -> Self {
        Self::ChainIdInvalidFormat { id }
    }

    pub fn kind(&self) -> &Self {
        &self
    }

}
`
If you define the error handling structure by `define_error!`, but how do you implement some functions for this structure? It seems that `flex-error` does not give some hints.

modules/src/ics24_host/error.rs
DaviRain-Su: Here because of the use of `displaydoc`.

modules/src/tx_msg.rs
DaviRain-Su: Here you need an Error triat, here I included a self-defined Error Trait in `flex-error`, if this is in the case of no-std, there is also an Error trait containing std. What you said here is directly a `crate::ics24_host::error:: ValidationError` to qualify is it possible?

modules/src/ics24_host/error.rs
soareschen: The name of the detail struct has a `Detail` postfix. So if you have `define_error!{ ValidationError; ... }`, then you have a `enum ValidationErrorDetail` defined, and you can define `impl ValidationErrorDetail { ... }` for example.

However in this case the helper functions seems to be just on constructing the errors, which we can replace entirely with the constructor functions defined by `flex-error`. So for example, instead of calling `ValidationKind::contains_separator(id)`, we can instead just use `error::contains_separator_error(id)`.

modules/src/lib.rs
DaviRain-Su: I think it is necessary to have a library that can switch between std and no_std. If you think switching to `no-std-compat` doesn't work, what do you think about switching to `sp-std`?

DaviRain-Su:
`
error[E0277]: the trait bound `flex_error::report::ErrorReport<ibc::ics23_commitment::error::ErrorDetail, anyhow::Error>: Clone` is not satisfied
   --> relayer/src/error.rs:159:20
    |
159 |     Ics023(#[from] ibc::ics23_commitment::error::Error),
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Clone` is not implemented for `flex_error::report::ErrorReport<ibc::ics23_commitment::error::ErrorDetail, anyhow::Error>`
    |
    = note: required by `std::clone::Clone::clone`
    = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
`
but It's already implemented.

`define_error! {
    #[derive(Clone)]
    Error {
        InvalidRawMerkleProof
        [ DisplayError<DecodeError> ]
        |_| { "invalid raw merkle proof" },

        CommitmentProofDecodingFailed
        [ DisplayError<DecodeError> ]
        |_| { "failed to decode commitment proof" },
    }
}
`
How to deal with problems? Can you help me?

modules/src/ics24_host/error.rs
DaviRain-Su: En, I understand.

soareschen:
We should change the definition to `Ics023(ibc::ics23_commitment::error::ErrorDetail)`. The compile error can propagate pretty wildly. Feel free to push your draft changes so that I can see what to be fixed.

DaviRain-Su:
@soareschen I made some fixes.

soareschen:
The changes look ok. However I feel that the scope of this PR is quite broad, and it will be more challenging for others to review this before merging to master. I suggest we split the changes into multiple PRs as follow:

- This PR should only have changes related to updating error definitions to use flex-error.
- The following should be done in a follow up PR:
  - The patches to other libraries `tendermint`, `tendermint-rpc` in Cargo.toml.
  - The feature flags for `std` and activation of `no_std`.
  - The use of dummy error in place of `std::error::Error` in `no_std`.

Basically we should aim for getting #988 merged first, and the submit a follow up PR, perhaps the original #890, to fix up the remaining changes to support `no_std`.

Feel free to also use different branches to track the changes relative to my `soares/flex-error` branch. I will avoid rebasing on the branch so that we can work on the parallel changes together.

DaviRain-Su:
> The changes look ok. However I feel that the scope of this PR is quite broad, and it will be more challenging for others to review this before merging to master. I suggest we split the changes into multiple PRs as follow:
> 
> * This PR should only have changes related to updating error definitions to use flex-error.
> * The following should be done in a follow up PR:
>   
>   * The patches to other libraries `tendermint`, `tendermint-rpc` in Cargo.toml.
>   * The feature flags for `std` and activation of `no_std`.
>   * The use of dummy error in place of `std::error::Error` in `no_std`.
> 
> Basically we should aim for getting #988 merged first, and the submit a follow up PR, perhaps the original #890, to fix up the remaining changes to support `no_std`.
> 
> Feel free to also use different branches to track the changes relative to my `soares/flex-error` branch. I will avoid rebasing on the branch so that we can work on the parallel changes together.

Hmmmmm, understand what you are saying about these difficult issues, so next, I'll revert this issue to std only. What do you think about fixing the support for no_std later on after starting over from you? Or in other cases, here we hope to support the no_std feature as soon as possible.

soareschen:
The remaining work can be save in another branch so that we do not lost the progress. I'd suggest putting up another PR as soon as this is merged. Right now some of the tests are also failing, which I think are caused by the no_std changes.

For supporting no_std as soon as possible, perhaps @romac and @ancazamfir can help with planning for the changes?

DaviRain-Su:
@soareschen  I backed off the modifications to the std feature only modifications, you check if there is anything else to modify. After that I hope to merge quickly.

soareschen:
Thanks for the fixes @ DaviRain-Su! I have merged this into my branch at #988. Next I will get rid of the `Box<dyn std::error::Error>` altogether so that we will track the concrete error types directly.

PR #1090, opened 2021-06-15
Simplify channel handshake logic to not involve retry

soareschen:
## Description

This is a follow up on #1066. I have refactored the logic in `do_chan_open_ack_confirm_step` so that it completes the full handshake without having to rely on the error retry loop.

The issue of using the same error retry loop for intentional retry is that the delay set in the error retry also affects the intentional retry. For the case of `do_chan_open_finalize_with_retry`, when `do_chan_open_ack_confirm_step` returns `PartialOpenHandshake`, the `retry_with_index` function will only try the next handshake step after the delay set in the fibonacci delay timer. This can significantly slow down the relayer if we do too many of such retries.

A better strategy is to explicitly write down the expected state transition, especially given that our state machine is pretty small. This also helps improve the readability, as readers do not need to understand the state machine defined else where in order to understand how the code works.

______

For contributor use:


adizere (approved):
LGTM!

Issue #1094, opened 2021-06-16
Add git commit hash to CLI help & error output


hu55a1n1:
## Crate

`ibc-relayer-cli`

## Summary

Print the git commit hash in the description. This makes it easier to identify the exact source that was used to create the CLI binary.

## Proposal
Either we print a valid semver that contains the short/long commit hash, e.g. `1.2.3-unstable+develop.a68295` or `1.1.2+a682956dccae752aa24597a0f5cd939f93614509` -> 
```bash
$ hermes
hermes 0.4.0-unstable+feature-branch.a68295
Informal Systems <hello@informal.systems>
Hermes is an IBC Relayer written in Rust.

# ...
```
Or, we print the commit hash as part of the description -> 
```bash
$ hermes
hermes 0.4.0
Informal Systems <hello@informal.systems>
v0.4.0 [a682956dccae752aa24597a0f5cd939f93614509] built on 2021-06-16T21:30:10
Hermes is an IBC Relayer written in Rust.

# ...
```

## Acceptance Criteria

The git commit hash must appear in the `usage`, `help` (including subcommand `help`s), `version` and error outputs.
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1101, opened 2021-06-17
Add CLI git hash 

hu55a1n1:
Closes: #1094

## Description

Export a env var (i.e. `HERMES_VERSION`) from build.rs and use it as the CLI's version string. 
The newer version string is a valid semver string of the form -> `<pkg-version>+<git-branch>-<git commit hash>`. There's also an optional suffix `-dirty` which is added if the git repo is dirty. e.g. -> `0.4.0+cli-git-hash-760270a8-dirty`

______

For contributor use:


romac:
Very nice, thanks @hu55a1n1!

One request: is there a way we can ensure that binaries built on CI and uploaded to the GitHub release show just the version without the `+branch-commit[-dirty]` suffix?

hu55a1n1:
> One request: is there a way we can ensure that binaries built on CI and uploaded to the GitHub release show just the version without the `+branch-commit[-dirty]` suffix?

Thanks for reviewing the PR @romac! 
I have been thinking about how to achieve that. Do you know if there's an env variable that the CI sets or some other way that I can use to figure out if the build script is running on the CI?

Otherwise, is it an option to change the regex in `acceptance.rs` to version meta? 
```rust
fn start_no_args() {
    // ...
    let mut runner = RUNNER.clone();
    let mut cmd = runner.capture_stdout().run();
    cmd.stdout().expect_regex(
        format!(
            "^[^ ]*{} {}", // <-------------------------- remove the trailing `$`?
            env!("CARGO_PKG_NAME"),
            env!("CARGO_PKG_VERSION")
        )
        .as_str(),
    );
    // ...
}
```

relayer-cli/build.rs
```diff
@@ -1,7 +1,86 @@
...
+}
+
+fn version() -> String {
+    let mut vers = env::var("CARGO_PKG_VERSION").unwrap();
+
+    if let Some(git) = GitHandle::new() {
+        println!("cargo:rustc-rerun-if-changed=.git/HEAD");
+
+        vers.push('+');
+        vers.push_str(&git.branch());
```
hu55a1n1, line 23: I am also thinking about removing the branch-name when the output of the `git.branch()` function is `master` or `HEAD` so that the version looks like -> `0.4.0+760270a8` instead of `0.4.0+master-760270a8`. What do you suggest?

romac:
> > One request: is there a way we can ensure that binaries built on CI and uploaded to the GitHub release show just the version without the `+branch-commit[-dirty]` suffix?
> 
> Thanks for reviewing the PR @romac!
> I have been thinking about how to achieve that. Do you know if there's an env variable that the CI sets or some other way that I can use to figure out if the build script is running on the CI?
> 
> Otherwise, is it an option to change the regex in `acceptance.rs` to version meta?
> 
> ```rust
> fn start_no_args() {
>     // ...
>     let mut runner = RUNNER.clone();
>     let mut cmd = runner.capture_stdout().run();
>     cmd.stdout().expect_regex(
>         format!(
>             "^[^ ]*{} {}", // <-------------------------- remove the trailing `$`?
>             env!("CARGO_PKG_NAME"),
>             env!("CARGO_PKG_VERSION")
>         )
>         .as_str(),
>     );
>     // ...
> }
> ```

Yes, I actually did that in https://github.com/informalsystems/ibc-rs/pull/1101/commits/d3c5360806dd55c5fa2241466f6e6e1ec3be069b, which has fixed the test.

But now the CI is still broken because Rust 1.53 was just released, and comes with new clippy lints. I think we can disregard these failures for now. I will fix them in a separate PR.

relayer-cli/build.rs
romac: I think I would keep `master` for consistency.

romac:
> Do you know if there's an env variable that the CI sets or some other way that I can use to figure out if the build script is running on the CI?

Looks like there's a bunch of them, not sure which one we should use. Perhaps we can check for `GITHUB_JOB == "create-release"`?

romac:
> But now the CI is still broken because Rust 1.53 was just released, and comes with new clippy lints. I think we can disregard these failures for now. I will fix them in a separate PR.

Should be good now, merged #1102 and merged master here.

hu55a1n1:
I just noticed that the new version doesn't appear for `help` subcommands like `hermes tx help`. The reason is that there's a blanket  `impl<C> Command for Help<C> where C: Command`, and due to the way `gumdrop` deals with `help`, this would mean that we would have to write quite a lot of impls for helps manually instead of deriving them.
I think this was the reason the build script overwrites the `CARGO_PKG_NAME` instead of trying to change `Command` impls. I think that is the best solution to modify the version as well, i.e. overwrite `CARGO_PKG_VERSION` with a valid semver. 
If there is no objection, I would like to implement the git hash support in a similar fashion. 🙂

hu55a1n1:
Linking #590 here, because it describes the problem in more detail. Although here, the chances of breaking cargo workflow are quite low since we're not overwriting the version for CI builds.

romac:
@hu55a1n1 What's the status on this PR?

adizere:
Tagged with "backburner" until Shoaib returns from holidays to look at this again.

hu55a1n1:
> @hu55a1n1 What's the status on this PR?
The acceptance test for the CLI was failing because it didn't escape regular expression meta characters in the `CARGO_PKG_VERSION` since this wasn't expected before. I think it is fixed now. 👍

adizere (approved):
Nice work, this is going to be very useful!

PR #1120, opened 2021-06-22
Check header and update state

cezarad:
Closes: #XXX

## Description

- implements the ICS07 verification of the client update ``checkValidityAndUpdateState`` in https://github.com/cosmos/ibc/tree/master/spec/client/ics-007-tendermint-client 
- modifications in the tendermint testgen to make test pass with this extra check. 

______

For contributor use:


adizere (changes requested):
This seems almost ready, nice work Cezara!

Two questions:

- some tests are failing when I run locally `cargo test`, in particular

> test ics02_client::handler::update_client::tests::test_update_client_ok ... FAILED
... 
test ics02_client::handler::update_client::tests::test_update_client_ok_multiple ... FAILED
test ics26_routing::handler::tests::routing_module_and_keepers ... FAILED
test ics18_relayer::utils::tests::client_update_ping_pong ... FAILED


It seems that the cause is the newly added checks in the update client handler. Can you confirm that the tests are failing and see how we can fix them please?

- how do you think we should proceed with the code in `modules/src/ics07_tendermint/header.rs` which seems duplicating logic from tendermint-rs? Should we keep it there, or would it be better to work with the tendermint-rs team to adapt their interfaces, so that we avoid duplicating code?

romac:
> * Should we keep it there, or would it be better to work with the tendermint-rs team to adapt their interfaces, so that we avoid duplicating code?

I think we should sort this out in tendermint-rs directly and adapt the light client code so that we do not need to duplicate this code.

romac:
Note: there is also duplicated code between `modules/src/ics07_tendermint/header.rs` and `modules/src/ics07_tendermint/predicates.rs`

adizere:
> > * Should we keep it there, or would it be better to work with the tendermint-rs team to adapt their interfaces, so that we avoid duplicating code?
> 
> I think we should sort this out in tendermint-rs directly and adapt the light client code so that we do not need to duplicate this code.

cezarad:
> Note: there is also duplicated code between `modules/src/ics07_tendermint/header.rs` and `modules/src/ics07_tendermint/predicates.rs`

We can remove the predicates.rs I put all In headers

cezarad:
mode to PR verification update #1252

Issue #1117, opened 2021-06-22
Configuration reload


romac:
## Crate

`ibc-relayer`, `ibc-relayer-cli`

## Summary

Add ability for Hermes to reload the configuration from disk without restarting it.

## Problem Definition

At the moment, for Hermes to pick any change to its configuration, operators need to stop and restart Hermes.

We want to provide a mechanism to trigger a reloading of Hermes configuration without having to restart it, eg. via a POSIX signal (typically `SIGHUP`).

## Proposal

- [x] Add ability to parse the configuration from disk at runtime
- [x] Perform a diff from the currently loaded configuration and the new one
- [x] Spawn or shutdown the appropriate workers based on the configuration changes
- [x] Hook up the `SIGHUP` signal to trigger a config reload

## Acceptance Criteria

The relayer reacts to a `SIGHUP` signal by reloading the configuration and spawning or shutting down the affected workers.
____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #1123, opened 2021-06-23
Configuration reload

romac:
Closes: #1117 

## Description

- [x] Add ability to parse the configuration from disk at runtime
- [x] Perform a diff from the currently loaded configuration and the new one
- [x] Spawn or shutdown the appropriate workers based on the configuration changes
- [x] Hook up the `SIGHUP` signal to trigger a config reload
- [x] Hook up the `SIGUSR1` signal to dump the supervisor state to the console
- [x] Fix bug where the client worker for the counterparty client is not spawned when a chain is added

### Manual test notes:
https://gist.github.com/ancazamfir/1ece9dadadc4eba70f12add103fc6c93
______

For contributor use:


relayer/src/supervisor/spawn.rs
```diff
@@ -0,0 +1,544 @@
...
+            // On config reload, we need to spawn the Packet worker for the counterparty as well,
+            // otherwise we may never do it, eg. if the counterparty chain was not updated in the
+            // config.
+
+            let remote_channel = counterparty_channel.and_then(|c| {
+                c.remote
+                    .channel_id
+                    .as_ref()
+                    .map(|cid| (c.remote.port_id.clone(), cid.clone()))
+            });
```
ancazamfir, line 492: I think there is a problem here as the remote of the counterparty channel is the local one. We should just take the remote from channel. During testing the worker for direction from other chain to the added chain is not spawned.

romac:
CI is failing because of [this minor security advisory](https://rustsec.org/advisories/RUSTSEC-2021-0073.html), will fix it by upgrading to prost 0.8 in an another PR.

adizere (approved):

Issue #600, opened 2021-02-02
Config validation


romac:
## Crate

`ibc-relayer`, `ibc-relayer-cli`

## Summary

- Implement configuration validation in `ibc-relayer`
- Implement `config validate` command in `ibc-relayer-cli`
- Stretch goal (or followup): semantic validation: `simulate` or `dry-run` txs to ensure the URL endpoints are reachable

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1128, opened 2021-06-24
Syntactic validation of configuration file

adizere:
Closes: #600 

## Description

- [x] config validation upon startup for any command
- [x] added new CLI `hermes config validate` which yield a small diagnostic in case a syntactic error is found in the config file

Validation handled:
  - [x] `log_level` field
  - [x] error signalled if no chain is specified for command `start`
  - [x] duplicate chain id specified
  - [X] prevent unknown fields in the configuration

______

For contributor use:

### DIFF
--- a/relayer-cli/src/commands.rs
+++ b/relayer-cli/src/commands.rs
@@ -16,10 +16,10 @@ use crate::config::Config;
 use crate::DEFAULT_CONFIG_PATH;
 
 use self::{
-    create::CreateCmds, keys::KeysCmd, listen::ListenCmd, query::QueryCmd, start::StartCmd,
-    tx::TxCmd, update::UpdateCmds, upgrade::UpgradeCmds, version::VersionCmd,
+    config::ConfigCmd, create::CreateCmds, keys::KeysCmd, listen::ListenCmd,
+    misbehaviour::MisbehaviourCmd, query::QueryCmd, start::StartCmd, tx::TxCmd, update::UpdateCmds,
+    upgrade::UpgradeCmds, version::VersionCmd,
 };
-use crate::commands::misbehaviour::MisbehaviourCmd;
 
 mod config;
 mod create;
@@ -38,17 +38,16 @@ pub fn default_config_file() -> Option<PathBuf> {
     dirs_next::home_dir().map(|home| home.join(DEFAULT_CONFIG_PATH))
 }
 
-// TODO: Re-add the `config` subcommand
-// /// The `config` subcommand
-// #[options(help = "manipulate the relayer configuration")]
-// Config(ConfigCmd),
-
 /// Cli Subcommands
 #[derive(Command, Debug, Options, Runnable)]
 pub enum CliCmd {
     /// The `help` subcommand
     #[options(help = "Get usage information")]
     Help(Help<Self>),
+
+    /// The `config` subcommand
+    #[options(help = "Validate Hermes configuration file")]
+    Config(ConfigCmd),
 
     /// The `keys` subcommand
     #[options(help = "Manage keys in the relayer for each chain")]
@@ -68,7 +67,7 @@ pub enum CliCmd {
 
     /// The `start` subcommand
     #[options(help = "Start the relayer in multi-chain mode. \
-                      Relays packets and channel handshake messages between all chains in the config.")]
+                      Relays packets and open handshake messages between all chains in the config.")]
     Start(StartCmd),
 
     /// The `query` subcommand
--- a/relayer-cli/src/components.rs
+++ b/relayer-cli/src/components.rs
@@ -96,7 +96,7 @@ fn enable_ansi() -> bool {
 
 /// Builds a tracing filter based on the input `log_level`.
 /// Enables tracing exclusively for the relayer crates.
-/// Returns error if the given `log_level` is invalid.
+/// Returns error if the filter failed to build.
 fn build_tracing_filter(log_level: String) -> Result<EnvFilter, FrameworkError> {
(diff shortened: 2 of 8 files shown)


relayer-cli/src/components.rs
```diff
@@ -92,12 +94,33 @@ fn enable_ansi() -> bool {
...
         .map(|&c| format!("{}={}", c, log_level))
         .reduce(|a, b| format!("{},{}", a, b))
-        .unwrap()
+        .unwrap();
+
+    // Build the filter directive
+    match EnvFilter::try_new(directive_raw.clone()) {
+        Ok(out) => Ok(out),
+        Err(e) => {
+            let our_err = config::Error::InvalidLogLevel(log_level.clone(), e.to_string());
```
romac, line 114: Maybe we should make our enum for the `log_level` property and let serde raise an error if the level is invalid.

ancazamfir:
Should we check also that all options included in the config file are supported by the current version? And issue at least a warning? Right now they are ignored.

relayer-cli/src/config.rs
```diff
@@ -4,20 +4,46 @@
...
-    assert!(args.next().is_some(), "expected one argument but got zero");
-    let args = args.collect::<Vec<_>>();
-    let app = EntryPoint::<CliCmd>::parse_args_default(args.as_slice())?;
-    let config_path = app.config.ok_or("no config file specified")?;
-    Ok(config_path)
+/// Specifies all the possible syntactic errors
+/// that a Hermes configuration file could contain.
+#[derive(Error, Debug)]
+pub enum Error {
+    /// No chain was configured
```
ancazamfir, line 18:
```suggestion
    /// No chain is configured
```

relayer-cli/src/config.rs
```diff
@@ -4,20 +4,46 @@
...
+    /// The log level is invalid
+    #[error("config file specifies an invalid log level ('{0}'), caused by: {1}")]
+    InvalidLogLevel(String, String),
+
+    /// Duplicate chains configured
+    #[error("config file has duplicate entry for the chain with id {0}")]
+    DuplicateChains(ChainId),
+}
+
+/// Validation method for syntactic validation of the input
```
ancazamfir, line 31:
```suggestion
/// Method for syntactic validation of the input
```

relayer-cli/src/config.rs
```diff
@@ -4,20 +4,46 @@
...
+/// configuration file.
+pub fn validate_config(config: &Config) -> Result<(), Error> {
+    // Check for empty or solitary chains config.
```
ancazamfir, line 34: what is "solitary chains"?

relayer-cli/src/components.rs
adizere: Good idea 38cc315

relayer-cli/src/config.rs
adizere: That was a stale comment that I forgot to delete! Should be fixed.

relayer-cli/src/commands.rs
```diff
@@ -38,18 +38,17 @@ pub fn default_config_file() -> Option<PathBuf> {
...
 /// Cli Subcommands
 #[derive(Command, Debug, Options, Runnable)]
 pub enum CliCmd {
     /// The `help` subcommand
     #[options(help = "Get usage information")]
     Help(Help<Self>),
 
+    /// The `config` subcommand
+    #[options(help = "Validate Hermes configuration file")]
+    Config(ConfigCmd),
```
romac, line 50: Can we make a `validate` subcommand for this? We may want to add other subcommands related to config, eg. `generate` so it might good to namespace it already.

relayer/src/config.rs
```diff
@@ -83,26 +85,55 @@ impl Default for Strategy {
     }
 }
 
+/// Log levels are wrappers over [`tracing_core::Level`].
+///
+/// [`tracing_core::Level`]: https://docs.rs/tracing-core/0.1.17/tracing_core/struct.Level.html
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "lowercase")]
+pub enum LogLevel {
```
romac, line 93: 🎉

relayer/src/config.rs
```diff
@@ -83,26 +85,55 @@ impl Default for Strategy {
...
+            LogLevel::Debug => write!(f, "debug"),
+            LogLevel::Info => write!(f, "info"),
+            LogLevel::Warn => write!(f, "warn"),
+            LogLevel::Error => write!(f, "error"),
+        }
+    }
+}
+
 #[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(default, deny_unknown_fields)]
```
romac, line 120: 💯 

Did you check that this is transitive and doesn't only apply to the fields on `GlobalConfig` itself?

relayer-cli/src/commands.rs
adizere: I think there is already a namespace for Config commands:


https://github.com/informalsystems/ibc-rs/blob/9c733d7af1ee7c2dee07086c24f2a94f4d169825/relayer-cli/src/commands/config.rs#L18

Is this what you meant?

Perhaps the command `help` is misleading because it only talks about validating configuration file; but I wrote it like that on purpose, since there is no other functionality yet. Once we add more subcommands (e.g., generate), we can rephrase the wording for the `help`. WDYT?

romac: Right, my bad! Yeah that's fine :)

romac (approved):

Issue #1062, opened 2021-06-08
New channel query to get channel, connection and client info in a single CLI


ancazamfir:
## Crate
relayer, relayer-cli

## Summary
DEX had asked us to provide a CLI that given a `chain-id`, `port-id`, and `channel-id` we return the `ChannelEnd`, `ConnectionEnd` and `AnyClientState` on both ends. We should probably define a new struct to include the information such that it can be read easily.

## Problem Definition
Currently the user needs to do ~6 queries to get this information. 

## Proposal

We can introduce a new query CLI, tentatively called `query trace` taking 3 parameters: a `chain-id`, `port-id`, and `channel-id`. 

The most challenging part is how to assemble the output in a simple, meaningful way, given that there are multiple queries that this will involve, as follows:

1. `hermes query channel end chain-id port-id channel-id`
    - extract from the result the connection hop identifier, for example `connection-id`
    - also extract the counterparty channel/port, `counterparty-port-id` and `counterparty-channel-id`

2. `hermes query connection end chain-id connection-id`
    - extract the client identifier of the underlying client, for example `client-id`
    - extract also the counterparty client identifier, `counterparty-client-id`
    - extract also the counterparty conn identifier, `counterparty-connection-id`

3. `hermes query client state chain-id client-id`
    - extract the chain identifier from the client state, `counterparty-chain-id`
    - now we have the full trace information on the side of `chain-id`
    - proceed to fetch the trace information from the counterparty side

4. `hermes query client state counterparty-chain-id counterparty-client-id`
5. `hermes query connection end counterparty-chain-id counterparty-connection-id`
6. `hermes query channel end counterparty-chain-id counterparty-port-id counterparty-channel-id`

It probably makes sense to have two modes for the output:

- **default**: Hermes outputs the whole structure of every field (not only the identifiers) for instance:

```rust
struct Trace {
  channel_end: ChannelEnd,
  connection_end: ConnectionEnd,
  client_state: AnyClientState,
  counterparty_channel_end: ChannelEnd,
  counterparty_connection_end: ConnectionEnd,
  counterparty_client_state: AnyClientState, 
}
```

- **summarised**: Hermes _only_ returns the identifier information, in a compact format, for instance:

```rust
struct TraceSummary {
  port_id: PortId,
  channel_id: ChannelId,
  connection_id: ConnectionId,
  client_id: ClientId,
  counterparty_port_id: PortId,
  counterparty_channel_id: ChannelId,
[...]

PR #1130, opened 2021-06-28
Implement query channel ends command

soareschen:
Closes: #1062

## Description

In contrast to the the description & acceptance criteria in the linked issue, we ended up deciding on the following

- CLI is `query channel ends CHAIN-ID PORT-ID CHAN-ID [-v/--verbose]
- the output has the same structure as in the issue description, but slightly different names:
    - for `-v` option: `ChannelEnds { ... }`
    - without `-v`, default: `ChannelEndsSummary { .. }`

______

For contributor use:

### DIFF
--- a/e2e/e2e/channel.py
+++ b/e2e/e2e/channel.py
@@ -272,7 +272,7 @@ class QueryChannelEnd(Cmd[ChannelEnd]):
     def process(self, result: Any) -> ChannelEnd:
         return from_dict(ChannelEnd, result)
 
-@ cmd("query channel ends -s")
+@ cmd("query channel ends")
 @ dataclass
 class QueryChannelEnds(Cmd[ChannelEnds]):
     chain_id: ChainId
--- a/e2e/e2e/channel.py
+++ b/e2e/e2e/channel.py
@@ -244,6 +244,20 @@ class ChannelEnd:
     state: str
     version: str
 
+@ dataclass
+class ChannelEnds:
+    chain_id: str
+    client_id: str
+    connection_id: str
+    channel_id: str
+    port_id: str
+
+    counterparty_chain_id: str
+    counterparty_client_id: str
+    counterparty_connection_id: str
+    counterparty_channel_id: str
+    counterparty_port_id: str
+
 
 @ cmd("query channel end")
 @ dataclass
@@ -258,6 +272,18 @@ class QueryChannelEnd(Cmd[ChannelEnd]):
     def process(self, result: Any) -> ChannelEnd:
         return from_dict(ChannelEnd, result)
 
+@ cmd("query channel ends -s")
+@ dataclass
+class QueryChannelEnds(Cmd[ChannelEnds]):
+    chain_id: ChainId
+    port_id: PortId
+    channel_id: ChannelId
+
+    def args(self) -> List[str]:
+        return [self.chain_id, self.port_id, self.channel_id]
+
+    def process(self, result: Any) -> ChannelEnds:
+        return from_dict(ChannelEnds, result)
 
 # =============================================================================
 # CHANNEL handshake
@@ -463,6 +489,9 @@ def handshake(
             f'Channel end with id {b_chan_id} on chain {side_b} is not in Open state, got: {b_chan_end.state}')
         exit(1)
 
+    a_chan_ends = query_channel_ends(c, side_a, port_id, a_chan_id)
+    l.debug(f'query channel ends result: {a_chan_ends}')
+
     return a_chan_id, b_chan_id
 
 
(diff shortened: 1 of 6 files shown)


e2e/e2e/channel.py
```diff
@@ -463,6 +489,32 @@ def handshake(
             f'Channel end with id {b_chan_id} on chain {side_b} is not in Open state, got: {b_chan_end.state}')
         exit(1)
 
+    a_chan_ends = query_channel_ends(c, side_a, port_id, a_chan_id)
+    l.debug(f'query channel ends result: {a_chan_ends}')
+
+    assert a_chan_ends.chain_id == side_a
```
adizere, line 495: Nicely done, very thorough! 💯

relayer-cli/src/commands/query/channel_ends.rs
```diff
@@ -0,0 +1,182 @@
...
+    let chain_id = cmd.chain_id.clone();
+    let port_id = cmd.port_id.clone();
+    let channel_id = cmd.channel_id.clone();
+
+    let chain_config = config
+        .find_chain(&chain_id)
+        .ok_or_else(|| format!("chain '{}' not found in configuration file", chain_id))?;
+
+    let rt = Arc::new(TokioRuntime::new().unwrap());
+    let chain = CosmosSdkChain::bootstrap(chain_config.clone(), rt.clone())?;
```
adizere, line 75: I think we can use a simpler approach here, along the lines of:

```suggestion
    let chain = spawn_chain_runtime(&config, &chain_id)?;
```

relayer-cli/src/commands/query/channel_ends.rs
```diff
@@ -0,0 +1,182 @@
...
+        None => chain.query_latest_height()?,
+    };
+
+    let channel_end = chain.query_channel(&port_id, &channel_id, chain_height)?;
+
+    let connection_id = channel_end
+        .connection_hops
+        .first()
+        .ok_or_else(|| format!("missing connection_hops"))?
+        .clone();
```
adizere, line 88: FYI there are some clippy complaints.

relayer-cli/src/commands/query/channel_ends.rs
```diff
@@ -0,0 +1,182 @@
...
+    let counterparty_chain_id = client_state.chain_id.clone();
+
+    let chain_config_b = config.find_chain(&counterparty_chain_id).ok_or_else(|| {
+        format!(
+            "chain '{}' not found in configuration file",
+            counterparty_chain_id
+        )
+    })?;
+
+    let counterparty_chain = CosmosSdkChain::bootstrap(chain_config_b.clone(), rt).unwrap();
```
adizere, line 126: It would be ideal to avoid `unwrap`. Perhaps try using `spawn_chain_runtime(..)?`.

relayer-cli/src/commands/query/channel_ends.rs
```diff
@@ -0,0 +1,182 @@
...
+        Output::success(res).exit();
+    }
+
+    Ok(())
+}
+
+impl Runnable for QueryChannelEndsCmd {
+    fn run(&self) {
+        match do_run(self) {
+            Ok(()) => {}
```
adizere, line 178: I like your approach to call in the `Ok(out)` branch the `Output::success(out)` (started to use the same in some [PRs](https://github.com/informalsystems/ibc-rs/pull/1122/commits/f97a5ba342f90b7701dd4c050fa4f8d3c6717a15)) and was wondering why did you choose to omit this approach in this case?

relayer-cli/src/commands/query/channel_ends.rs
soareschen: I was copying the code from the other command implementations. Just looked at `spawn_chain_runtime` and seems like it returns a `Box<dyn ChainHandle>` which erases the type and cause some information to be unavailable. It also spawns other threads such as the event monitor, which are probably not needed here.

relayer-cli/src/commands/query/channel_ends.rs
soareschen: Fixed.

relayer-cli/src/commands/query/channel_ends.rs
soareschen: Fixed.

relayer-cli/src/commands/query/channel_ends.rs
soareschen: The command returns two different structs depending on the summary flag. Wrapping them inside another enum will cause the auto-derived `Serialize` implementation to output different json. So I thought just outputting the result directly is simpler in this case.

adizere: Got it, I didn't notice that!

relayer-cli/src/commands/query/channel_ends.rs
adizere: I agree that the event monitor is not necessary. We can still use `CosmosSdkChain::bootstrap` if you think it's better, but it important to avoid `unwrap`s.

adizere (approved):
Looks good to go, cool stuff Soares, thank you!

Issue #1127, opened 2021-06-24
Full inline documentation for the default config.toml


adizere:
## Crate

`ibc-relayer-cli`

## Summary & Proposal

The default `config.toml` configuration file for Hermes has currently no documentation, except for
the strategy field.

https://github.com/informalsystems/ibc-rs/blob/f9278382143f44b8c1e1a6a99893b54b5999a30e/config.toml#L1-L6

This is a problem because the expectation is that users will consult
https://hermes.informal.systems/config.html, but they rarely do.
Most users will take the shortcut of trying to infer the meaning of different fields and use defaults throughout.

Having documentation alongside each field will provide more clarity to users.

## Acceptance Criteria

- [ ] the `config.toml` file has every field documented.
- [ ] ~~consider trimming the section https://hermes.informal.systems/config.html#parameters-2~~

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

PR #1136, opened 2021-06-29
Add inline documentation to example config.toml

soareschen:
Closes: #1127

## Description

The descriptions are mostly copied from https://hermes.informal.systems/config.html. I have also added some missing fields with their default value.

______

For contributor use:

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -1,5 +1,4 @@
 # The global section has parameters that apply globally to the relayer operation.
-# https://hermes.informal.systems/config.html
 [global]
 
 # Specify the strategy to be used by the relayer. Default: 'packets'
@@ -53,11 +52,12 @@ rpc_timeout = '10s'
 account_prefix = 'cosmos'
 
 # Specify the name of the private key to use for signing transactions. Required
-# See the Adding Keys chapter for for more information about managing signing keys:
+# See the Adding Keys chapter for more information about managing signing keys:
 #   https://hermes.informal.systems/commands/keys/index.html#adding-keys
 key_name = 'testkey'
 
 # Specify the store prefix used by the on-chain IBC modules. Required
+# Recommended value for Cosmos SDK: 'ibc'
 store_prefix = 'ibc'
 
 # Specify the maximum amount of gas to be used as the gas limit for a transaction.
@@ -66,10 +66,7 @@ max_gas = 3000000
 
 # Specify the price per gas used of the fee to submit a transaction and
 # the denomination of the fee. Required
-gas_price = {
-  price = 0.001,
-  denom = 'stake'
-}
+gas_price = { price = 0.001, denom = 'stake' }
 
 # Specify by ratio to increase the gas estimate used to compute the fee,
 # to account for potential estimation error. Default: 0.1, ie. 10%.
@@ -97,10 +94,7 @@ trusting_period = '14days'
 # Specify the trust threshold for the light client, ie. the maximum fraction of validators
 # which have changed between two blocks.
 # Default: { numerator = '1', denominator = '3' }, ie. 1/3.
-trust_threshold = {
-  numerator = '1',
-  denominator = '3'
-}
+trust_threshold = { numerator = '1', denominator = '3' }
 
 [[chains]]
 id = 'ibc-1'
--- a/config.toml
+++ b/config.toml
@@ -1,29 +1,101 @@
+# The global section has parameters that apply globally to the relayer operation.
+# https://hermes.informal.systems/config.html
 [global]
-# Valid strategies:
-#   'packets': restricts relaying to packets only, or
-#   'all': relay from all types of events
+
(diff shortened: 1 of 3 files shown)


romac:
Great stuff! Can we also document whether or not an option is required, and if it has a default value? While the current values match the default values, it's not yet 100% clear which of the options have a default value and which have not.

soareschen:
Ok then. Done.

romac:
Perfect, thank you :)

romac (approved):

config.toml
```diff
@@ -1,29 +1,106 @@
...
+
+# Specify the prefix used by the chain. Required
 account_prefix = 'cosmos'
+
+# Specify the name of the private key to use for signing transactions. Required
+# See the Adding Keys chapter for for more information about managing signing keys:
+#   https://hermes.informal.systems/commands/keys/index.html#adding-keys
 key_name = 'testkey'
+
+# Specify the store prefix used by the on-chain IBC modules. Required
```
adizere, line 60: add recommended option for SDK: `ibc`.

config.toml
```diff
@@ -1,29 +1,106 @@
+# The global section has parameters that apply globally to the relayer operation.
+# https://hermes.informal.systems/config.html
```
adizere, line 2: To make it easier to maintain, I suggest we no longer keep
the description of fields in
https://hermes.informal.systems/config.html.
Instead, we just present this file directly in https://hermes.informal.systems/config.html (with no additional discussion of the fields).

I added the label "needs guide update" to remember to update the guide correspondingly.

config.toml
```diff
@@ -1,29 +1,106 @@
...
+
+# Specify the maximum amount of time (duration) that the RPC requests should
+# take before timing out. Default: 10s (10 seconds)
 rpc_timeout = '10s'
+
+# Specify the prefix used by the chain. Required
 account_prefix = 'cosmos'
+
+# Specify the name of the private key to use for signing transactions. Required
+# See the Adding Keys chapter for for more information about managing signing keys:
```
adizere, line 56:
```suggestion
# See the Adding Keys chapter for more information about managing signing keys:
```

config.toml
```diff
@@ -1,29 +1,106 @@
...
+# Default: 300000
 max_gas = 3000000
-gas_price = { price = 0.001, denom = 'stake' }
+
+# Specify the price per gas used of the fee to submit a transaction and
+# the denomination of the fee. Required
+gas_price = {
+  price = 0.001,
+  denom = 'stake'
+}
```
adizere, line 72: I think adding newline here will not parse correctly.

Specifically, this is the error:

```
> hermes -c config.toml query clients ibc-0
error: hermes fatal error: parse error: expected a table key, found a newline at line 69 column 14
```

```suggestion
gas_price = { price = 0.001, denom = 'stake' }
```

config.toml
```diff
@@ -1,29 +1,106 @@
...
 trusting_period = '14days'
-trust_threshold = { numerator = '1', denominator = '3' }
+
+# Specify the trust threshold for the light client, ie. the maximum fraction of validators
+# which have changed between two blocks.
+# Default: { numerator = '1', denominator = '3' }, ie. 1/3.
+trust_threshold = {
+  numerator = '1',
+  denominator = '3'
+}
```
adizere, line 103: Same as the `gas_price` comment.

```suggestion
trust_threshold = { numerator = '1', denominator = '3' }
```

adizere (changes requested):

config.toml
soareschen: Done.

config.toml
soareschen: I have now removed the whole Sections section.

config.toml
soareschen: Fixed.

config.toml
soareschen: Interesting. Didn't know toml has such a quirk. Fixed.

config.toml
soareschen: Fixed.

guide/src/config.md
```diff
@@ -16,163 +16,16 @@ name of the command to run, eg. `hermes -c my_config.toml query connection chann
 hermes [-c CONFIG_FILE] COMMAND
 ```
 
-## Sections
```
adizere, line 19: Can we please leave this file untouched for the moment?

There are two reasons for delaying changes to this files:
- there may still be v0.5.0 users which do not have the new config.toml with inline documentation
- but more importantly: it seems the DeX team had a misconfiguration of `trust_threshold` in config.toml and it may be better to try and simplify this file (e.g., by removing sensitive options that have clear defaults) while documenting _all_ possible options in some place (here).


I will open another issue to to follow-up on the second bullet above.

adizere (changes requested):

adizere (approved):

Issue #1143, opened 2021-07-01
Hermes `refresh_schedule` bug


adizere:
## Crate

`ibc-relayer`

## Summary of Bug

The `refresh_schedule` method in the relayer library has the job to trim out events which are scheduled for sending but which are outdated (because they were handled in the meantime). It does so by removing some of [the TransitMessages that are registered in `self.dst_operational_data`](https://github.com/informalsystems/ibc-rs/blob/c78b793d99571df4781cec4c2cfcb18ed68098d1/relayer/src/link.rs#L1381).

The way it is currently implemented, this method can exit in two ways:
-  an early return in case no new timeouts message have been detected (that's line 1360 below). 
- If timeouts have been detected, then the method effectively removes items from `self.dst_operational_data` (line 1381).

The problem is that the modification for line 1381 should always be applied, regardless of the early exit or not. 

https://github.com/informalsystems/ibc-rs/blob/c78b793d99571df4781cec4c2cfcb18ed68098d1/relayer/src/link.rs#L1360-L1381

## Version

c509f889e57728bf2115b87cf5cd8ebaef33c2ff

This was observed in production while doing a log walkthrough with  @mircea-c  and @greg-szabo.

The original logs capturing how this problem manifested are as follows:

```
Jun 29 13:50:24 hermes hermes[2534]: Jun 29 13:50:24.298  INFO [cosmoshub-4:transfer/channel-141 -> osmosis-1] scheduling op. data with 2 msg(s) for Destination chain (height 4-6757828)
Jun 29 13:50:24 hermes hermes[2534]: Jun 29 13:50:24.318 DEBUG [cosmoshub-4:transfer/channel-141 -> osmosis-1] WriteAcknowledgement - h:4-6757827, seq:10169, path:channel-0/transfer->channel-141/transfer, toh:4-6757911, tos:Timestamp(NoTimestamp)) already handled
Jun 29 13:50:24 hermes hermes[2534]: Jun 29 13:50:24.365  INFO [cosmoshub-4:transfer/channel-141 -> osmosis-1] relay op. data to Destination, proofs height 4-6757827, (delayed by: 67.138219ms) [try 1/5]
```

These logs are imperfect, but roughly speaking on the second line Hermes detects that the packet with sn `10169` was already handled it should avoid sending it (by removing that packet from its operational data). On the last line in the log, Hermes proceed to send the same packet (so the packet was never removed from op. data).

## Acceptance Criteria

[...]

PR #1144, opened 2021-07-01
Fix for schedule refreshing bug

adizere:
Closes: #1143 

Adds two important changes:

- 796e5f1 fix for the linked issue (#1143)
- [uni_chan_path.rs](https://github.com/informalsystems/ibc-rs/pull/1144/files#diff-fe5221ddd8cf9efdaff272e1f5d057a4dfbb3b3a8e1882e1321c83948600f3ccR60-R83) -- adjusted the backoff (`thread::sleep`) in the uni-path worker so that the worker does _no backoff_ in case there were some incoming commands from the supervisor

## Description

______

For contributor use:


relayer/src/worker/uni_chan_path.rs
```diff
@@ -76,6 +76,11 @@ impl UniChanPathWorker {
                     .into());
                 }
             }
+
+            // If there are no incoming commands, it's safe to backoff.
+            if self.cmd_rx.is_empty() {
+                thread::sleep(Duration::from_millis(200));
+            }
```
romac, line 83: Maybe we can refactor this by using [`select!` and `after`](https://docs.rs/crossbeam-channel/0.5.1/crossbeam_channel/fn.after.html#examples)? Do you think it would make the code more readable?

ie. we would write something like this instead of `let maybe_cmd = self.cmd_rx.try_recv().ok();`:

```rust
const TIMEOUT = Duration::from_millis(200);

let maybe_cmd = select! {
  recv(self.cmd_rx) -> cmd => Some(cmd),
  recv(after(TIMEOUT)) -> _ => None,
};
```

ancazamfir (approved):
Looks good to me. I also like @romac 's suggestion.

Issue #1153, opened 2021-07-06
Hermes schedule execution routine can drop op. data


adizere:
This bug was observed during operation for the Osmosis app. @greg-szabo pointed out the underlying issue and we uncovered it looking directly at logs.

## Crate

`ibc-relayer`

## Summary of Bug

The `execute_schedule` method can potentially drop an operational data, thus Hermes will never relay the messages
associated with that data. The method is as follows

https://github.com/informalsystems/ibc-rs/blob/c78b793d99571df4781cec4c2cfcb18ed68098d1/relayer/src/link.rs#L1289-L1302

The scenario that can lead to a dropped op. data is as follows:

1. line 1290: Hermes removes some op. data that have been scheduled earlier, and retains X operational data to be sent to the Destination chain in the vector `dst_ods`.
    - this means that `len(dst_ods) = X`
    - it is important that X is strictly larger than 1, let's take as example 2

2. line 1294: Hermes takes the first element (among 2 of them) from the vector `dst_ods`
3. line 1295: Hermes attempts to run `relay_from_operational_data(..)?` on that first element
    - suppose that the method `relay_from_operational_data` fails
    - in this case, the whole method `execute_schedule` returns because of the `?`
4. the second element from vector `dst_ods` is never handled

Similarly, any elements from the vector `src_ods` is never handled either. So Hermes never sends those messages.

## Version

de62a4b9329dc00df11720627846d3ef9e3ada87

## Acceptance Criteria

- [ ] clarify if `relay_from_operational_data` can fail
- [ ] reproduce the problem in an automated (integration) test if possible
    - the main difficult in reproducing is in making the method `relay_from_operational_data` fail: unclear if that is easily made to fail
- [ ] `execute_schedule` should not remove multiple elements from the scheduled op. data, instead should remove them one by one

____

## For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

seanchen1991:
Adi mentioned that @ancazamfir is convinced that the `relay_from_operational` data method is (mostly) infallible, such that in most cases no operational data will be dropped. Looking at the method in a vacuum, it seems like it will only drop operational data when the max number of retries have been attempted, or some undocumented non-recoverable error occurs: 

[...]

PR #1154, opened 2021-07-06
Fix for execute schedule bug

adizere:
Closes: #1153

## Description

______

For contributor use:

### DIFF
--- a/relayer/src/link.rs
+++ b/relayer/src/link.rs
@@ -1284,18 +1284,21 @@ impl RelayPath {
         }
     }
 
-    /// Checks if there are any operational data items ready, and if so performs the relaying
-    /// of corresponding packets to the target chain.
+    /// Checks if there is any operational data items ready,
+    /// and if so performs the relaying of corresponding packets
+    /// to the target chain.
     pub fn execute_schedule(&mut self) -> Result<RelaySummary, LinkError> {
-        let (src_ods, dst_ods) = self.try_fetch_scheduled_operational_data();
-
         let mut summary = RelaySummary::empty();
 
-        for od in dst_ods {
+        // Handle first messages heading to the destination chain.
+        let dst_od = self.try_fetch_scheduled_operational_data_dst();
+        if let Some(od) = dst_od {
             summary.extend(self.relay_from_operational_data(od)?);
         }
 
-        for od in src_ods {
+        // Now handle messages heading to the source chain.
+        let src_od = self.try_fetch_scheduled_operational_data_src();
+        if let Some(od) = src_od {
             summary.extend(self.relay_from_operational_data(od)?);
         }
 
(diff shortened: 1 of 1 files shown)


relayer/src/link.rs
```diff
@@ -1426,34 +1429,42 @@ impl RelayPath {
         Ok(())
     }
 
-    /// Pulls out the operational elements with elapsed delay period and that can
+    /// Destination-chain fetching of operational data.
+    ///
+    /// Pulls out an operational elements with elapsed delay period and that can
     /// now be processed. Does not block: if no OD fulfilled the delay period (or none is
-    /// scheduled), returns immediately with `vec![]`.
-    fn try_fetch_scheduled_operational_data(
```
adizere, line 1432: To explain the changes here at a high level, since the diff is a big messy:

I split the method `try_fetch_scheduled_operational_data` into two separate methods:

1. `try_fetch_scheduled_operational_data_dst` which fetches 1 op. data specifically for the destination chain
2. `try_fetch_scheduled_operational_data_src` which fetches 1 op. data specifically for the source chain.

relayer/src/link.rs
```diff
@@ -1426,34 +1429,42 @@ impl RelayPath {
...
+        }
+    }
 
-        let dst_ods: Vec<OperationalData> = self
-            .dst_operational_data
+    /// Source-chain fetching of operational data.
+    ///
+    /// Equivalent to [`try_fetch_scheduled_operational_data_dst`]
+    /// but for the source chain operational data.
+    fn try_fetch_scheduled_operational_data_src(&mut self) -> Option<OperationalData> {
```
adizere, line 1456: We no longer return a vector of op. data, but at most one.

The rationale is as follows: Returning a vector was problematic because one of them might fail to process, and then Hermes would never get a chance to process the subsequent ones.

romac:
The fix seems sound to me. I wonder what we lose (if anything) perf-wise by executing the schedule op. data once per worker-loop iteration rather than as batches. Probably not too bad since we refresh and execute the scheduled op. data once per `step` (ie. whenever of the packet worker.

By the way, looking at the (existing) code I wonder if we are doing the right thing here:

https://github.com/informalsystems/ibc-rs/blob/3da7dc12c060271176c8a8870aa399a8eea64820/relayer/src/worker/uni_chan_path.rs#L62-L64

It looks to me like we are dropping commands when they fail, as for every retry step, we try and block before we receive a _new_ command on `self.cmd_rx`, and the command that originally failed is therefore discarded.

romac:
> The fix seems sound to me. I wonder what we lose (if anything) perf-wise by executing the schedule op. data once per worker-loop iteration rather than as batches. Probably not too bad since we refresh and execute the scheduled op. data once per `step` (ie. whenever of the packet worker.
> 
> By the way, looking at the (existing) code I wonder if we are doing the right thing here:
> 
> https://github.com/informalsystems/ibc-rs/blob/3da7dc12c060271176c8a8870aa399a8eea64820/relayer/src/worker/uni_chan_path.rs#L62-L64
> 
> It looks to me like we are dropping commands when they fail, as for every retry step, we try and block before we receive a _new_ command on `self.cmd_rx`, and the command that originally failed is therefore discarded.

Oh seems like this was already known and there is a fix in #1144. Nevermind then! :)

relayer/src/link.rs
```diff
@@ -1426,34 +1429,42 @@ impl RelayPath {
...
-    fn try_fetch_scheduled_operational_data(
-        &mut self,
-    ) -> (Vec<OperationalData>, Vec<OperationalData>) {
-        // The first elements of the op. data vector contain the oldest entry.
-        // Remove and return the elements with elapsed delay.
-
-        let src_ods: Vec<OperationalData> = self
-            .src_operational_data
+    /// scheduled), returns immediately with `None`.
+    fn try_fetch_scheduled_operational_data_dst(&mut self) -> Option<OperationalData> {
```
hu55a1n1, line 1437: The design of these methods is reminiscent of iterators because they are essentially returning the next element until there are none, so it might be more pragmatic to implement them as custom iterators.

soareschen:
I have tried to simplify the logic and still do bulk processing of all operation data if no error occurred:

```rust

    // Try to execute an iterated list of `OperationData`. Returns the following:
    //   - A summary of processed data,
    //   - A vector of unprocessed data, either because:
    //       - The scheduled time has not elapsed, or
    //       - An error has occured before or during the processing of the given operation
    //   - An optional LinkError, which is returned if any error occurred.
    fn do_execute_schedule<I: Iterator<Item = OperationalData>>
        (&mut self, mut operations: I) ->
        (RelaySummary, Vec<OperationalData>, Option<LinkError>)
    {
        let connection_delay = self.channel.connection_delay;
        let mut summaries = RelaySummary::empty();
        let mut unprocessed = vec![];

        while let Some(operation) = operations.next() {
            if operation.scheduled_time.elapsed() > connection_delay {
                match self.relay_from_operational_data(operation.clone()) {
                    Ok(summary) => {
                        summaries.extend(summary);
                    }
                    Err(e) => {
                        unprocessed.push(operation);
                        unprocessed.extend(operations);

                        return (summaries, unprocessed, Some(e));
                    }
                }
            } else {
                unprocessed.push(operation);
            }
        }

        (summaries, unprocessed, None)
    }

    /// Checks if there is any operational data items ready,
    /// and if so performs the relaying of corresponding packets
    /// to the target chain.
    pub fn execute_schedule(&mut self) -> Result<RelaySummary, LinkError> {
        let mut dst_operational_data = core::mem::take(&mut self.dst_operational_data);
        let (mut dst_summaries, dst_unprocessed, err) =
            self.do_execute_schedule(dst_operational_data.drain(0..));
        self.dst_operational_data = dst_unprocessed;

        if let Some(e) = err {
            return Err(e)
        }

        let mut src_operational_data = core::mem::take(&mut self.src_operational_data);
        let (src_summaries, src_unprocessed, err) =
            self.do_execute_schedule(src_operational_data.drain(0..));
        self.src_operational_data = src_unprocessed;

        if let Some(e) = err {
            return Err(e)
        }

        dst_summaries.extend(src_summaries);

        Ok(dst_summaries)
    }
```

A diff is available here: https://github.com/informalsystems/ibc-rs/compare/adi/1153_execute_schedule_bug...soares/1153_execute_schedule_bug?expand=1

Let me know if this makes sense.

adizere:
Closing for the moment.

PR #1163, opened 2021-07-09
Client state filter

adizere:
Closes: #1165

## Description

For basic testing:

- set `trust_threshold = { numerator = '3', denominator = '3' }` for `ibc-0`
- `hermes create channel ibc-0 ibc-1 --port-a transfer --port-b transfer`
    - so `channel-0` on both ends will be denied
- set `trust_threshold = { numerator = '1', denominator = '3' }` (default) for `ibc-0`
- `hermes create channel ibc-0 ibc-1 --port-a transfer --port-b transfer`
    - so `channel-1` on both ends will be allowed

______

For contributor use:


relayer/src/supervisor/client_state_filter.rs
```diff
@@ -0,0 +1,195 @@
...
+
+/// A cache storing filtering status (allow or deny) for
+/// arbitrary identifiers.
+#[derive(Default, Debug)]
+pub(crate) struct FilterCache {
+    /// A cache associating a generic identifier, such as
+    /// client id, channel id, or connection id, with a
+    /// boolean, representing either allow (if `true`) or
+    /// deny (if `false`) that identifier.
+    permission_cache: HashMap<Identifier, bool>,
```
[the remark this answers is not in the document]
romac: And instead of using a `String`, let's perhaps define an enum for the keys?

```rust
enum CacheKey {
  Client(ChainId, ClientId),
  Channel(ChainId, ChannelId),
  Connection(ChainId, ConnectionId),
}
```

relayer/src/supervisor/client_state_filter.rs
```diff
@@ -0,0 +1,195 @@
...
+use ibc::ics04_channel::error::Kind;
+use ibc::ics24_host::identifier::{ChainId, ChannelId, ClientId, ConnectionId, PortId};
+use ibc::Height;
+
+type Identifier = String;
+
+/// A cache storing filtering status (allow or deny) for
+/// arbitrary identifiers.
+#[derive(Default, Debug)]
+pub(crate) struct FilterCache {
```
romac, line 18: This struct is not strictly a cache since it contains the filtering logic as well, maybe rename it to `FilterPolicy` or something like that?

We can also rename the file/module to `filter_policy` or something.

romac (changes requested):
Great stuff, left some suggestions but the logic is sound!

config.toml
```diff
@@ -100,7 +100,7 @@ trusting_period = '14days'
 # Specify the trust threshold for the light client, ie. the maximum fraction of validators
 # which have changed between two blocks.
 # Default: { numerator = '1', denominator = '3' }, ie. 1/3.
-trust_threshold = { numerator = '1', denominator = '3' }
+trust_threshold = { numerator = '3', denominator = '3' }
```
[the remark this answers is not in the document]
adizere: This was necessary for testing. Indeed we'll reverted it before commit, thanks for pointing it out! Will keep your comment un-resolved until we're ready to merge..

relayer/src/supervisor/client_state_filter.rs
```diff
@@ -0,0 +1,274 @@
...
+}
+
+impl Permission {
+    fn and(self, other: &Permission) -> Permission {
+        if matches!(self, Permission::Allow) && matches!(other, Permission::Allow) {
+            self
+        } else {
+            Permission::Deny
+        }
+    }
```
romac, line 28:
```suggestion
    fn and(self, other: &Self) -> Self {
        if matches!(self, Self::Allow) && matches!(other, Self::Allow) {
            Self::Allow
        } else {
            Self::Deny
        }
    }
```

relayer/src/supervisor/client_state_filter.rs
```diff
@@ -0,0 +1,274 @@
...
+
+        let permission = self.control_connection_end_and_client(
+            registry,
+            &chain_id,
+            &client_state,
+            &connection_end,
+            &conn_id,
+        )?;
+
+        let key = CacheKey::from((chain_id.clone(), port_id.clone(), channel_id.clone()));
```
romac, line 221: Do we really need these `From` instances? It obscures a bit which type of `CacheKey` will be created. How about doing this instead?

```suggestion
        let key = CacheKey::Channel(chain_id.clone(), port_id.clone(), channel_id.clone());
```

adizere: Good point. I think I was just enthusiastic about it and didn't realize this simpler approach, nice!

3abdf1de

relayer/src/supervisor/spawn.rs
```diff
@@ -39,24 +40,32 @@ pub struct SpawnContext<'a> {
     config: &'a RwArc<Config>,
     registry: &'a mut Registry,
     workers: &'a mut WorkerMap,
+    client_state_filter: &'a mut FilterPolicy,
```
adizere, line 43: FYI @romac I had to propagate the mutable `client_state_filter` from the supervisor into the SpawnContext. Did not see a simpler approach than this.

PR #1169, opened 2021-07-11
Nixify proto-compiler

JonathanLorimer:
## Description

This PR adds nix support for the `ibc-proto-compiler`. Originally I was going to try and just nixify everything, but I realized the PR would be rather large, so I figured I would submit PR's one sub-directory at a time.

The goal of nixifying ibc-rs is that I want to run my own IBC relayer, and the easiest way to spin up infra (IMO) is with nix. Nixifying this repo will make things a bit easier downstream (for me), will facilitate adding this software to nixpkgs, and might provide some development benefits to the development team if they choose to use nix for a development environment!

I strongly encourage the reviewer to try the new workflow! I think the fact that compiling is now a one-liner is pretty compelling `nix-shell --run "compile" 
______

For contributor use:

I didn't create an issue for this. Should I?

No unit tests right now, and I don't think nix-tests make sense. But a nice benefit of nix is it allows you to spin up a virtualized machine and run E2E tests, so that is something I could potentially provide after nixifying the whole thing!

Do I need to create an issue?

Updated the `proto-compiler/README.md`

### DIFF
--- a/proto-compiler/README.md
+++ b/proto-compiler/README.md
@@ -15,7 +15,7 @@ $ nix-shell # you can run "nix develop" if you have experimental support for fla
 $ compile
 ```
 
-### Clone the Cosmos SDK
+### Using Cargo
 
 From within the `proto-compiler` directory, compile the binary using the `--locked` flag:
 
--- a/proto-compiler/README.md
+++ b/proto-compiler/README.md
@@ -4,6 +4,17 @@ The `ibc-proto-compiler` is a simple command-line tool to automate the compilati
 
 ## Usage
 
+### Nix Shell
+
+There is a script wrapper provided by nix that takes care of the `protoc` dependency imposed by `prost-build` and provides a shell wrapper that handles cloning the cosmos go dependencies.
+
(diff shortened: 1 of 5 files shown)


proto-compiler/README.md
```diff
@@ -4,7 +4,18 @@ The `ibc-proto-compiler` is a simple command-line tool to automate the compilati
...
+There is a script wrapper provided by nix that takes care of the `protoc` dependency imposed by `prost-build` and provides a shell wrapper that handles cloning the cosmos go dependencies.
+
+Instruction on how to install nix can be found [here](https://nixos.org/download.html)
+
+```shell
+$ nix-shell # you can run "nix develop" if you have experimental support for flakes
+$ compile
+```
+
+### Using Cargo
```
JonathanLorimer, line 18: I'll admit, switching `Clone the Cosmos SDK` to `Using Cargo` is taking a bit of a liberty. But I personally found it a bit confusing to have `Clone the Cosmos SDK` as a header, when that is really just 1 step of the process, and it is handled by a rust program anyways, `Using Cargo` felt more semantic. Happy to change this back if you disagree!

PR #1170, opened 2021-07-11
Nixify ibc-proto-compiler

JonathanLorimer:
## Description

Doesn't close, but gets us closer to closing #1171

This PR adds nix support for the `ibc-proto-compiler`. Originally I was going to try and just nixify everything, but I realized the PR would be rather large, so I figured I would submit PR's one sub-directory at a time.

The goal of nixifying ibc-rs is that I want to run my own IBC relayer, and the easiest way to spin up infra (IMO) is with nix. Nixifying this repo will make things a bit easier downstream (for me), will facilitate adding this software to nixpkgs, and might provide some development benefits to the development team if they choose to use nix for a development environment!

I strongly encourage the reviewer to try the new workflow! I think the fact that compiling is now a one-liner is pretty compelling `nix-shell --run "compile" `

(also removes the need to clutter your global environment with `protoc`, a dependency which wasn't mentioned in the `Usage` instructions :wink: )
______

For contributor use:

~~I didn't create an issue for this. Should I?~~ just read CONTRIBUTING.md

No unit tests right now, and I don't think nix-tests make sense (yet). But a nice benefit of nix is it allows you to spin up a virtualized machine and run E2E tests, so that is something I could potentially provide after nixifying the whole thing!

### DIFF
--- a/proto-compiler/shell.nix
+++ b/proto-compiler/shell.nix
@@ -1,0 +1,12 @@
+# This file exists for legacy nix-shell
+# https://nixos.wiki/wiki/Flakes#Using_flakes_project_from_a_legacy_Nix
+# You generally do *not* have to modify this ever.
+(import (
+  let
+    lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+  in fetchTarball {
+    url = "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
+    sha256 = lock.nodes.flake-compat.locked.narHash; }
+) {
+  src =  ./.;
+}).shellNix
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+@JonathanLorimer added a nix expression to `ibc-proto-compiler` ([#1170])
+
 Many thanks to Fraccaroli Gianmarco (@Fraccaman) for helping us improve the
 reliability of Hermes ([#697]).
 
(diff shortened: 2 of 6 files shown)


proto-compiler/shell.nix
```diff
@@ -0,0 +1,12 @@
+# This file exists for legacy nix-shell
+# https://nixos.wiki/wiki/Flakes#Using_flakes_project_from_a_legacy_Nix
```
soareschen, line 2: Since this is a new project, would it be better if we just use Nix Flakes from the get go and forget about legacy Nix? I understand that Nix Flakes is still experimental, but we currently do not have the baggage of needing to support legacy Nix either.

soareschen (changes requested):
Thanks for the contribution! I think for the case of the protobuf compiler, a better Nix integration would be to run the protobuf compiler all inside Nix and then output the generated Rust code. e.g. something like:

```bash
$ nix build
$ ls result
cosmos.auth.v1beta1.rs
cosmos.base.v1beta1.rs
cosmos.upgrade.v1beta1.rs
ibc.core.connection.v1.rs
...
```

We can then simply copy the generated Rust code and then create a commit for the changes.

To manage the version of cosmos-sdk and ibc-go, we could probably lock the version inside `flake.lock`. Alternatively, we could make it such that the commit ID are passed to the Nix build script, e.g.

```bash
$ nix-build --arg sdk-commit 7648bfca45b9d0897103ec739210607dce77c4fb
```

I'm not sure which approach would be better, so we might have to think about it more.

That being said, we should probably first discuss more at  #1171 on how do we want to organize the Nix code before doing more work on this.

JonathanLorimer:
Closed due to change in direction. See
- https://github.com/informalsystems/ibc-rs/issues/1171
- https://github.com/informalsystems/cosmos.nix

Issue #1192, opened 2021-07-13
Some conversions for Mock failed


yito88:
## Crate
IBC

## Summary of Bug
I found two bugs in using `mock`.
- `ClientType::Mock` couldn't be converted with `from_str()` when using `ClientType::as_string()`
  - Looks inconsistent between [`as_string()`](https://github.com/informalsystems/ibc-rs/blob/3ae2f105a64b67bd3a75c80d9a9fee40e059f3da/modules/src/ics02_client/client_type.rs#L23) and [`from_str()`](https://github.com/informalsystems/ibc-rs/blob/3ae2f105a64b67bd3a75c80d9a9fee40e059f3da/modules/src/ics02_client/client_type.rs#L42)
- When converting `MockHeader` to `RowMockHeader`, stack overflow happened
  - It seems that the [`From` implementation](https://github.com/informalsystems/ibc-rs/blob/3ae2f105a64b67bd3a75c80d9a9fee40e059f3da/modules/src/mock/header.rs#L43) causes this bug

## Version
0.5.0

## Steps to Reproduce
- Try to get the client type from the string
```rust
let client_type = ClientType::Mock;
let type_string = client_type.as_string();
let client_type_from_str = ClientType::from_str(type_string).expect("Failed")
```

- Try to encode `MockHeader` wrapped by `AnyHeader`
```rust
let height = Height::new(1, 10);
let header = MockHeader::new(height);
// `wrap_any()` isn't implemented for now
let any_header = AnyHeader::from(header);
let bytes = header.encode_vec().expect("Stack overflow happened");
```

## Acceptance Criteria
- `ClientType::Mock` can be converted from a proper string
- `MockHeader` can be encoded

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1195, opened 2021-07-13
Fix conversions related to `MockHeader` and `ClientType::Mock`

romac:
Closes: #1192

### DIFF
--- a/modules/src/ics02_client/client_type.rs
+++ b/modules/src/ics02_client/client_type.rs
@@ -14,13 +14,16 @@ pub enum ClientType {
 }
 
 impl ClientType {
+    const TENDERMINT_STR: &'static str = "07-tendermint";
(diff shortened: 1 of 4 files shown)


modules/src/ics02_client/client_type.rs
```diff
@@ -39,7 +39,7 @@ impl std::str::FromStr for ClientType {
             "07-tendermint" => Ok(Self::Tendermint),
 
             #[cfg(any(test, feature = "mocks"))]
-            "mock" => Ok(Self::Mock),
+            "9999-mock" => Ok(Self::Mock),
```
adizere, line 42: maybe we can make these str `const` ?

adizere (approved):

modules/src/ics02_client/client_type.rs
adizere: also: changelog

Issue #1200, opened 2021-07-14
Hermes does not clear packets on start


ancazamfir:
## Crate
relayer

## Summary of Bug
Hermes does not clear packets on start

## Version
master, `v0.6.0` and after. Probably `v0.6.0-rc2` also.

## Steps to Reproduce
- start two gaia chains
- create a channel
- send packets
- start hermes
- check unrelayed packets

(see also 5bab89b with the e2e script changes that show the problem)

## Acceptance Criteria
`hermes start` should clear packets immediately

____

## For Admin Use

- [x] Not duplicate issue
- [ ] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #1196, opened 2021-07-13
Hermes issues found by Dex scalability tests on v0.6.0


ancazamfir:
## Crate
relayer

## Summary of Bug
Following up on the Dex scalability test issue reported, @nodebreaker0-0  has provided access to his setup and showed me the setup, config and the commands he was running.
I was able to reproduce this locally and after debugging found that there are a few issues in v0.6.0:
- hermes doesn't seem to clear packets on start
- when more than ~150 transactions are included in a block, hermes workers only receive Tx Rpc event notifications for ~15 of them and stop receiving anything after (this is also present in v0.5.0)

## Version
hermes v0.6.0, v0.5.0 
Tested with both gaia v4.2.1 and v5.0.0, reproducible with both.

## Steps to Reproduce
- use 10sec timeouts in `one-chain` script (see `anca/dex_rpc_event`)
- start two gaia chains using the `dev-env` script:
    - `./scripts/dev-env  ~/.hermes/config.toml ibc-0 ibc-1`
- create one channel:
   - `hermes create channel ibc-0 ibc-1 --port-a transfer --port-b transfer`
- with these [config files](https://gist.github.com/ancazamfir/2a626613fc177e2c5af54519199cad4c):
  - `hermes -c hermes_config.toml start`
  - `hermes -c ft_transfer_config.toml tx raw ft-transfer ibc-1 ibc-0 transfer channel-0 9999 -o 1000 -n 150 -k user2`
  -  check unreceived packets and acks (only 12-15 events are received and handled, the rest are lost someplace)
  - try to send another transfer and notice in the logs that the event is not generated:
      -  `hermes -c config_1.toml tx raw ft-transfer ibc-1 ibc-0 transfer channel-0 9999 -o 1000 -n 1 -k user2`
  - restart hermes and see that the packets are not cleared on next block (wait > 10sec)
      -  made some quick changes in `anca/dex_rpc_event` for clearing the packets and that part seems ok now
  - the missing Tx events can be seen with 120 transfers also but haven't looked for exact threshold . It works for 100, 110 for example.
   
For the missing rpc events I traced back to `stream_batches()`:
https://github.com/informalsystems/ibc-rs/blob/3f0d4bc38a3d737cbd289c2087d09228bdb7b07f/relayer/src/event/monitor.rs#L393

[...]

PR #1201, opened 2021-07-14
Clear packets on hermes start

ancazamfir:
Closes: #1200
Partially closes: #1196

## Description
Hermes does not clear packets on start if there are no IBC events and no `clear_packets_interval` trigger.
The e2e script masks this error because it sends some packets, starts hermes and then sends more packets (issue masked here because this triggers packet clearing) before checking that all packets are cleared. 

- [x] Fixed script to catch the bug
- [x] Fixed the bug

______

For contributor use:

### DIFF
--- a/relayer/src/worker/packet.rs
+++ b/relayer/src/worker/packet.rs
@@ -98,14 +98,7 @@ impl PacketWorker {
     fn step(&self, cmd: Option<WorkerCmd>, link: &mut Link, index: u64) -> RetryResult<Step, u64> {
         if let Some(cmd) = cmd {
             let result = match cmd {
-                WorkerCmd::IbcEvents { batch } => {
-                    // Try to clear old packets.
-                    link.a_to_b
-                        .schedule_packet_clearing(batch.height, false)
-                        .and_then(|_|
-                        // Update scheduled batches
-                        link.a_to_b.update_schedule(batch))
-                }
+                WorkerCmd::IbcEvents { batch } => link.a_to_b.update_schedule(batch),
 
                 // Handle the arrival of an event signaling that the
                 // source chain has advanced to a new block.
--- a/relayer/src/worker/packet.rs
+++ b/relayer/src/worker/packet.rs
@@ -109,12 +109,13 @@ impl PacketWorker {
                     height,
                     new_block: _,
                 } => {
-                    // Schedule the clearing of pending packets
+                    // Schedule the clearing of pending packets at start and
                     // at predefined block intervals.
-                    if self.clear_packets_interval != 0
-                        && height.revision_height % self.clear_packets_interval == 0
+                    if link.a_to_b.clear_packets()
+                        || self.clear_packets_interval != 0
+                            && height.revision_height % self.clear_packets_interval == 0
                     {
-                        link.a_to_b.clear_packets(height)
+                        link.a_to_b.do_clear_packets(height)
                     } else {
                         Ok(())
                     }
--- a/relayer/src/worker/packet.rs
+++ b/relayer/src/worker/packet.rs
@@ -99,8 +99,12 @@ impl PacketWorker {
         if let Some(cmd) = cmd {
             let result = match cmd {
                 WorkerCmd::IbcEvents { batch } => {
-                    // Update scheduled batches.
-                    link.a_to_b.update_schedule(batch)
+                    // Try to clear old packets.
+                    link.a_to_b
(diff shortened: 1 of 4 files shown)


relayer/src/worker/packet.rs
```diff
@@ -109,12 +109,13 @@ impl PacketWorker {
                     height,
                     new_block: _,
                 } => {
-                    // Schedule the clearing of pending packets
+                    // Schedule the clearing of pending packets at start and
                     // at predefined block intervals.
-                    if self.clear_packets_interval != 0
-                        && height.revision_height % self.clear_packets_interval == 0
+                    if link.a_to_b.clear_packets()
```
adizere, line 114: Two remarks for this change:

1. It can trigger `link.a_to_b.do_clear_packets` repeatedly until the `clear_packets` flag is set to false. This flag is set to false only when the worker receives a command with a batch of events (that is the first match arm above, on [line 103](https://github.com/informalsystems/ibc-rs/pull/1201/files#diff-d7d106630bd632b82613cdf6ec54b5623af0041717ff82a92b7f8579da2629cfR103)).

2. If I understand correctly, with this change there may be two separate invocations to `link.a_to_b.do_clear_packets`:

    - one invocation upon NewBlock (this call site), which would clear packets for that `NewBlock`'s height
    - another invocation with the call site within `update_schedule(batch)`, which would clear packets for the height of that `batch`, see the snippet below:

https://github.com/informalsystems/ibc-rs/blob/872a5030f9bcc46cfbbb71803554731b3a12ca16/relayer/src/link/relay_path.rs#L299-L306

If the height of `NewBlock` event and the `batch` height are the same, then we'll clear packets twice for the same height. Worst case, this will schedule twice the execution of the same operational data.

BTW I tried to reproduce point (2) but was not able. I was only able to reproduce problem (1) above.

adizere: I attempted an alternative fix
 628425b

Feel free to revert if this approach is not appropriate.

ancazamfir: > If the height of NewBlock event and the batch height are the same, then we'll clear packets twice for the same height. Worst case, this will schedule twice the execution of the same operational data.

Ah yes, good point! This is the case in v0.5.0 and older also. We used to get the IBC event before the New Block event so we had to clear on IBC event also. With the changes in the scheduler and monitor we now always get the NewBlock before the IBC event. I think we can clear packets on NewBlock only. I will try some changes along these lines so we simplify the code quite a bit.

>  I was only able to reproduce problem (1) above.
How?

adizere: > How?

Steps to reproduce (1)

1. dev-env with two chains, and create a channel between the two;
2. `git co 872a503` and `start` hermes. The logs show

>Jul 15 09:24:16.570  INFO [ibc-1:transfer/channel-0 -> ibc-0] clearing pending packets from events before height 1-312
Jul 15 09:24:16.581  INFO [ibc-1:transfer/channel-0 -> ibc-0] finished scheduling the clearing of pending packets
Jul 15 09:24:16.886  INFO [ibc-0:transfer/channel-0 -> ibc-1] clearing pending packets from events before height 0-324
Jul 15 09:24:16.902  INFO [ibc-0:transfer/channel-0 -> ibc-1] finished scheduling the clearing of pending packets
Jul 15 09:24:17.723  INFO [ibc-1:transfer/channel-0 -> ibc-0] clearing pending packets from events before height 1-313
Jul 15 09:24:17.733  INFO [ibc-1:transfer/channel-0 -> ibc-0] finished scheduling the clearing of pending packets
Jul 15 09:24:18.091  INFO [ibc-0:transfer/channel-0 -> ibc-1] clearing pending packets from events before height 0-325
Jul 15 09:24:18.101  INFO [ibc-0:transfer/channel-0 -> ibc-1] finished scheduling the clearing of pending packets
Jul 15 09:24:18.872  INFO [ibc-1:transfer/channel-0 -> ibc-0] clearing pending packets from events before height 1-314
Jul 15 09:24:18.880  INFO [ibc-1:transfer/channel-0 -> ibc-0] finished scheduling the clearing of pending packets
Jul 15 09:24:19.190  INFO [ibc-0:transfer/channel-0 -> ibc-1] clearing pending packets from events before height 0-326

The problem was: the flag `link.a_to_b.clear_packets` is never set to `false`, so upon every `NewBlock` event, the method `do_clear_packets` is invoked.

adizere (approved):
LGTM!

Issue #1224, opened 2021-07-21
Release Hermes v0.6.1


romac:
⚡

## Release PR

- [ ] #1226 

## Tasks

- [x] Create new release section in CHANGELOG and move "unreleased" items into this section.
- [x] Bump all crate versions to the new version.
- [ ] Reassign unfinished issues of previous milestone to the next milestone.
- [x] Update Cargo.lock file (if re-publishing `ibc-relayer-cli`)

PR #1077, opened 2021-06-11
Add bulk upgrade CLI 

hu55a1n1:
Closes: #763 

## Description

Add a bulk upgrade CLI command that takes a source chain's id as input and upgrades clients for all chains in config target the specified source chain client.
______

For contributor use:

- [ ] Updated the __Unreleased__ section of [CHANGELOG.md](https://github.com/informalsystems/ibc-rs/blob/master/CHANGELOG.md) with the issue.
- [ ] If applicable: Unit tests written, added test to CI.
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Updated relevant documentation (`docs/`) and code comments.
- [ ] Re-reviewed `Files changed` in the Github PR explorer.

PR #1226, opened 2021-07-21
Release Hermes v0.6.1

romac:
Closes: #1224

## Guide updates

- [ ] #1077 (@romac)

> Postponed due to #1229 

- [X] Fix for `unwrap` in the `query tx events` CLI (@adizere)

####  Problem with `unwrap` in the `query tx events` CLI 

This unwrap can provoke a panic in case an invalid hash is provided to the CLI:

https://github.com/informalsystems/ibc-rs/blob/47899196c271f523a1db395202d77072beaa0593/relayer-cli/src/commands/query/tx/events.rs#L48

Example:

> $ hermes query tx events ibc-0 bb
The application panicked (crashed).
Message:  called `Result::unwrap()` on an `Err` value: Context { kind: Parse, source: Some("hash length"), backtrace: Some(   0: backtrace::backtrace::libunwind::trace
...
Location: relayer-cli/src/commands/query/tx/events.rs:48

After fix (ec9c7050fc68c6ff0e11f45245278e276cc4502f):

> $ hermes query tx events ibc-0 bb
Error: CLI argument error: could not parse 'bb' into a valid hash: parse error: hash length

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,10 @@ being closed under its feet by Tendermint.
 
 Upgrading from version `0.6.0` to `0.6.1` requires no explicit steps.
 
-> **WARNING:** Due to a regression ([#1229]), the `upgrade client`
-> and `tx raw upgrade-chain` commands have been temporarily disabled
-> in this version. They will be re-enabled in the next version.
+> **WARNING:** Due to a regression ([#1229]), the `upgrade client`,
+> `tx raw upgrade-clients`, and `tx raw upgrade-chain` commands have
+> been temporarily disabled in this version.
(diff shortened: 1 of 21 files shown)


adizere (commented):
I think we only need to update the guide to include a brief mention of the `upgrade clients` CLI. Note that we also have the command `tx raw upgrade-clients` command that is identical to `upgrade clients`, but I think we should only document the latter one (and leave the tx raw undocumented).


Going forward, it may make sense to leaving `tx raw` undocumented, or even trim the guide to remove them. The reason is that they are an additional burden to maintain and they are not designed to be used by users, but by developers (who can read the code & comments).

adizere (approved):

PR #1236, opened 2021-07-27
Add LegacyError type to replace BoxError in no-std mode

DaviRain-Su:
> I'd propose that we define a type like LegacyError, which is either a newtype wrapper to Box<dyn std::error::Error> in std mode, or a newtype wrapper to String in no-std mode. By always using a newtype wrapper, we can avoid accidental implicit coercion from Box<dyn std::error::Error> to LegacyError in std mode, which will prevent no-std mode to work.

Relate issue: https://github.com/informalsystems/ibc-rs/issues/1158
Relate Pr: https://github.com/informalsystems/ibc-rs/pull/1167

### DIFF
--- a/modules/Cargo.toml
+++ b/modules/Cargo.toml
@@ -16,10 +16,11 @@ description = """
 [features]
 # This feature grants access to development-time mocking libraries, such as `MockContext` or `MockHeader`.
 # Depends on the `testgen` suite for generating Tendermint light blocks.
-default = ["std", "eyre_tracer"]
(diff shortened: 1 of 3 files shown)


adizere:
What is the status of this PR in your opinion @soareschen ?

soareschen:
I think this can be closed now since we have managed to remove almost all sources of `Box<dyn std::error::Error>` in #988. There are only a few remaining sources coming from tendermint-rs side, and we can fix that directly from there.

Thanks @DaviRain-Su anyway for the changes!

Issue #1238, opened 2021-07-27
Separate generated code from `prost` and `tonic` in `proto-compiler`


soareschen:
## Crate

`ibc-proto`, `ibc-proto-compiler`

## Summary

This is a follow up on #1164. We want to separate the generated code from `prost` and `tonic`, and put the `tonic` code behind an `std` feature flag. This is so that the generated `prost` code can still be used in no_std mode, as `prost` supports no_std but `tonic` doesn't.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
One awkward solution to have separate code is to generate two sets of files:

1. one set including both prost and tonic (i.e., the same set of file we currently generate in ibc-proto);
2. one set with just prost, without tonic's client- and server-side logic, i.e., by setting `false` to buid_client and build_server

https://github.com/informalsystems/ibc-rs/blob/f9a24edb74808f3be23b087d7a4ce6fc86fe251c/proto-compiler/src/cmd/compile.rs#L106-L107

This solution is sub-optimal but not that difficult to implement.

romac:
`tonic-build` 0.5 can add custom attributes on the Rust modules containing the gRPC code (see https://docs.rs/tonic-https://docs.rs/tonic-build/0.5.2/tonic_build/struct.Builder.html#method.server_mod_attributebuild/0.5.2/tonic_build/struct.Builder.html#method.client_mod_attribute).

So I guess what we could do is:
1. wait for Prost 0.9 and Tonic 0.6
2. add an `std` feature to the `ibc-proto` crate
3. use the aforementioned feature to disable the gRPC code when the `std` feature is disabled

PR #1239, opened 2021-07-27
decouple proto and grpc 

DaviRain-Su:
relate issue : https://github.com/informalsystems/ibc-rs/issues/1238

______

For contributor use:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1374,6 +1374,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "ibc-proto-rpc"
+version = "0.1.0"
+
+[[package]]
 name = "ibc-relayer"
 version = "0.6.1"
 dependencies = [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1370,12 +1370,16 @@ dependencies = [
  "prost",
  "prost-types",
  "tendermint-proto",
- "tonic",
 ]
 
 [[package]]
 name = "ibc-proto-rpc"
 version = "0.1.0"
+dependencies = [
+ "ibc-proto",
+ "tendermint-proto",
+ "tonic",
+]
 
 [[package]]
 name = "ibc-relayer"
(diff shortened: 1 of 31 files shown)


romac:
Thanks for this PR, splitting the two crates makes sense to me!

Is this the result of manual work? If so, we need to figure out a way to do this programmatically in the [`proto-compiler`](https://github.com/informalsystems/ibc-rs/tree/master/proto-compiler) tool that is part of this repo, and which we use to generate these Rust Protobuf definitions.

romac:
@soareschen What's your take on this? Would you rather bundle the gRPC code in `ibc-proto` or does having two crates make sense to you?

DaviRain-Su:
> Is this the result of manual work?

It's not particularly clear what the proto-compiler does? It seems that the rpc in proto is also written out manually.

romac:
> It seems that the rpc in proto is also written out manually.

It's not. 

> It's not particularly clear what the proto-compiler does? 

`proto-compiler` is a command-line tool we use to generate the Rust code from the Protobuf definitions. The tool can pull the Cosmos SDK and the IBC Go repositories at a specific commit, and will then uses Prost and Tonic to generate the Rust code from the .proto files bundled in the Cosmos SDK/IBC Go repositories. You can check out the README for how to use it.

DaviRain-Su:
> proto-compiler is a command-line tool we use to generate the Rust code from the Protobuf definitions. The tool can pull the Cosmos SDK and the IBC Go repositories at a specific commit, and will then uses Prost and Tonic to generate the Rust code from the .proto files bundled in the Cosmos SDK/IBC Go repositories. You can check out the README for how to use it.

Which is very good, then here you can learn to split proto with rpc with tendermint-rs.

soareschen:
I think it is beneficial to separate the code into two module namespaces or crates. Though this will depend on how well we can generate the code separately. I am looking at the documentation of `prost_build` and `tonic_build` to see if we can invoke the two code generations separately.

romac:
Closing this PR until we manage to generate two separate crates with our `proto-compiler` tool.

Issue #1021, opened 2021-05-31
indicate config file error when applicable


colin-axner:
## Crate

relayer-cli

## Summary of Bug

```
hermes help start
    Finished release [optimized] target(s) in 0.10s
     Running `target/release/hermes help start`
error: hermes fatal error: parse error: unknown variant `naive`, expected `packets` or `all` for key `global.strategy` at line 31 column 1
```

## Version

master

## Steps to Reproduce

build master, run:
```
hermes -c config_example.toml start cosmoshub-4 akashnet-2 -p transfer -c channel-95
    Finished release [optimized] target(s) in 0.11s
     Running `target/release/hermes -c config_example.toml start cosmoshub-4 akashnet-2 -p transfer -c channel-95`
error: unexpected free argument `cosmoshub-4`

hermes 0.3.2
Informal Systems <hello@informal.systems>
For help, run `hermes [<FLAGS>] help <COMMAND>`, for example: `hermes help create client`

FLAGS:
    -c, --config CONFIG       path to configuration file
    -j, --json                enable JSON output
```

follow the provided hint ``For help, run `hermes [<FLAGS>] help <COMMAND>``

## Acceptance Criteria

The help command displays the help page for `hermes start`

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

colin-axner:
Ah! The issue was in parsing my configuration file! Is there any way to indicate this in the error message?

adizere:
> Ah! The issue was in parsing my configuration file! Is there any way to indicate this in the error message?

Indeed. There was a breaking change which we didn't signal very well (#924), specifically the `global.strategy` field. 

https://github.com/informalsystems/ibc-rs/blob/cdecd3674af564bb069681c2335185853e878a70/config.toml#L1-L5

Unfortunately, the message which Hermes prints upon doing

```
hermes help start
    Finished release [optimized] target(s) in 0.10s
     Running `target/release/hermes help start`
error: hermes fatal error: parse error: unknown variant `naive`, expected `packets` or `all` for key `global.strategy` at line 31 column 1
```
is not at all a help message. The config.toml could not be parsed, which prevents executing the `help` command. Hermes should be more explicit about this..

colin-axner:
[...]

PR #1240, opened 2021-07-27
More explicit config error message

hu55a1n1:
Closes: #1021 

## Description

Modifies the configuration file parse error message, so an error that previously looked like so ->
```bash
$ hermes start
error: hermes fatal error: parse error: unknown variant `naive`, expected `packets` or `all` for key `global.strategy` at line 62 column 1
```

LE @adizere -- Now looks like ->
```bash
$ hermes start
The Hermes configuration file at path '/Users/adi/.hermes/config.toml' is invalid, reason: parse error: unknown variant `naive`, expected `packets` or `all` for key `global.strategy` at line 49 column 1
Please see the example configuration for detailed information about the supported configuration options: https://github.com/informalsystems/ibc-rs/blob/master/config.toml
```

______

For contributor use:

### DIFF
--- a/relayer-cli/src/application.rs
+++ b/relayer-cli/src/application.rs
@@ -148,7 +148,8 @@ impl Application for CliApp {
 
         let config = config_path
             .map(|path| self.load_config(&path))
-            .transpose()?
+            .transpose()
+            .map_err(|err| FrameworkErrorKind::ConfigError.context(format!("{}", err)))?
             .unwrap_or_default();
 
         // Update the `json_output` flag used by `conclude::Output`
--- a/relayer-cli/src/application.rs
+++ b/relayer-cli/src/application.rs
@@ -149,7 +149,16 @@ impl Application for CliApp {
         let config = config_path
             .map(|path| self.load_config(&path))
             .transpose()
-            .map_err(|err| FrameworkErrorKind::ConfigError.context(format!("{}", err)))?
+            .map_err(|err| {
+                eprintln!("The Hermes configuration file is invalid, reason: {}", err);
+                eprintln!(
+                    "Please see the example configuration for detailed information about the \
+                    supported configuration options: \
+                    https://github.com/informalsystems/ibc-rs/blob/master/config.toml"
+                );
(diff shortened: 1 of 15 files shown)


hu55a1n1:
@adizere and @romac - do you think this fix is sufficient or should we also print the help output in such cases?

romac:
Can we instead catch the config error and print a more informative message? Maybe something like:

> the Hermes configuration file is invalid, reason: etc.
> Please see the example configuration for detailed information about the supported configuration options: [link to config.toml]

hu55a1n1:
> Can we instead catch the config error and print a more informative message? 

Good point! Done! 👍

romac:
Thanks! One last thing, can you please update the PR description with the new error message? LGTM otherwise :)

romac (approved):

relayer-cli/src/application.rs
```diff
@@ -148,7 +148,17 @@ impl Application for CliApp {
 
         let config = config_path
             .map(|path| self.load_config(&path))
-            .transpose()?
+            .transpose()
+            .map_err(|err| {
+                eprintln!("The Hermes configuration file is invalid, reason: {}", err);
```
adizere, line 153: Just wanted to double-check: using `tracing::error` here is not an option here, right (because it's not initialized yet)? If it is an option, we should consider using that.

adizere (approved):
Just had a question, otherwise looks good!

Please consider updating the changelog also (we now use [`unclog`](https://github.com/informalsystems/unclog)).

relayer-cli/src/application.rs
hu55a1n1: Exactly, I tried to use it before but then realized that it needs to get the log level from the config file for initialization. The actual log init is done later in the same function, so I don't think it would be possible to use it here.

adizere (approved):
Looks great, thanks Shoaib! I did one minor aesthetic change here d17d3e5c570a8a3d126e6c46475304ed9ffa4359. Approving and will let you do the merge if you agree with the changes I made.

Issue #1132, opened 2021-06-28
Add filtering to `query channels` to show channels between a pairs of chains


ancazamfir:
## Crate
relayer-cli

## Summary
Current `hermes query channels <chain-id>` shows all channels on the specified chains.
It would be very useful to show all channels between two chains.

## Problem Definition
Needed for operations (Dex, osmosis, etc).

## Proposal
Extend `query channels` to take an optional destination chain. In addition add the `--verbose` option to show client and connection ids on the counterparty chain. In this case the output should use, for each channel, the same format as in #1062 

```query channels <chain-id> [--destination-chain <chain-id>] [--verbose]```
e.g.:

```
$ query channels ibc-0 -d ibc-1

Success: [
    PortChannelId {
        channel_id: ChannelId(
            "channel-0",
        ),
        port_id: PortId(
            "transfer",
        ),
    },
    PortChannelId {
        channel_id: ChannelId(
            "channel-1",
        ),
        port_id: PortId(
            "transfer",
        ),
    },
]
```

```
$ query channels ibc-0 -d ibc-1 --verbose
...
```

## Acceptance Criteria

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

**Comment by adizere** [comment]
Found out that the filtering implemented in #1243 does not work. 

Note: The option `--destination-chain` was renamed into `--counterparty-chain` as part of ADR 010 and its implementation.

Examples:

- we have a single channel opened between `ibc-0` and `ibc-1` in these examples;
- filtering by the same chain as the chain we're querying
```
$ hermes query channels --chain ibc-0 --counterparty-chain ibc-0
    Finished dev [unoptimized + debuginfo] target(s) in 0.25s
     Running `target/debug/hermes query channels --chain ibc-0 --counterparty-chain ibc-0`
2022-06-24T08:03:53.395055Z  INFO ThreadId(01) using default configuration from '/Users/adi/.hermes/config.toml'
2022-06-24T08:03:53.396324Z DEBUG ThreadId(01) Options: QueryChannelsCmd { chain_id: ChainId { id: "ibc-0", version: 0 }, dst_chain_id: Some(ChainId { id: "ibc-0", version: 0 }), verbose: false }
[...]

PR #1243, opened 2021-07-28
Add optional destination chain and verbose opts for CLI query channels

hu55a1n1:
Closes: #1132 

## Description

Add optional `destination chain` and `verbose` opts for CLI query channels. If a `--destination-chain/-d` is specified we only output those channels whose counterparty is the destination chain. If the `--verbose/-v` flag is set, the output has the same structure as the `ChannelEnds` described in #1130. 

______

For contributor use:


hu55a1n1:
There are some opportunities for code refactor/reuse/deduplication here. For example -> 
* The `impl Runnable` for `QueryChannelEndsCmd` and `QueryChannelsCmd` are both using a lot of the same code i.e. getting the counterparty chain from a channel.
* Also the `channel_connection_client()` is using much of the same code although it takes a `impl ChainHandle` as input instead of `impl Chain`. 
I think we can separate out common trait methods in the `Chain` and `ChainHandle` traits into a super trait, e.g. 
```rust
pub(crate) trait Queryable {
    fn query_commitment_prefix(&self) -> Result<CommitmentPrefix, Error>;

    fn query_compatible_versions(&self) -> Result<Vec<Version>, Error>;

    fn query_latest_height(&self) -> Result<ICSHeight, Error>;
    
    // etc.
}

pub trait Chain: Sized + Queryable + Provable {}
```
This should allow us to reuse more code in the future.

adizere:
@soareschen could you please help with reviewing and testing this PR?

relayer-cli/src/commands/query/channels.rs
```diff
@@ -1,60 +1,184 @@
...
+    let rt = Arc::new(TokioRuntime::new()?);
+    let chain = CosmosSdkChain::bootstrap(chain_config.clone(), rt.clone())?;
+    let chain_height = chain.query_latest_height()?;
+
+    let req = QueryChannelsRequest {
+        pagination: ibc_proto::cosmos::base::query::pagination::all(),
+    };
+    let identified_channels = chain.query_channels(req)?;
+
+    for identified_channel in identified_channels {
```
soareschen, line 56: I think the body of this for loop should be split into two more functions: one when `verbose` is true and one when `verbose` is false.

relayer-cli/src/commands/query/channels.rs
```diff
@@ -1,60 +1,184 @@
...
+            })?
+            .clone();
+
+        let connection_end = chain.query_connection(&connection_id, chain_height)?;
+        let client_id = connection_end.client_id().clone();
+        let client_state = chain.query_client_state(&client_id, chain_height)?;
+        let counterparty_chain_id = client_state.chain_id.clone();
+
+        match &cmd.destination_chain {
+            Some(dst_chain_id) if dst_chain_id != &counterparty_chain_id => {
```
soareschen, line 87: If there is a mismatch, should this be an error? Also, if `cmd.destination_chain` is `None` or if `verbose` is false, there is no need to call `query_connection` and `query_client_state`.

relayer-cli/src/commands/query/channels.rs
```diff
@@ -1,60 +1,184 @@
...
+            })?;
+
+            let chain_config_b = config.find_chain(&counterparty_chain_id).ok_or_else(|| {
+                format!(
+                    "counterparty chain '{}' not found in configuration file",
+                    counterparty_chain_id
+                )
+            })?;
+
+            let counterparty_chain = CosmosSdkChain::bootstrap(chain_config_b.clone(), rt.clone())?;
```
soareschen, line 129: Since this is inside the loop, would it cause issues if the same chain is bootstrapped multiple times? Can we use something like `Registry` to help manage the chain runtimes?

relayer-cli/src/commands/query/channels.rs
```diff
@@ -1,60 +1,184 @@
...
+            });
+        }
+    }
+
+    let output = if !cmd.verbose {
+        QueryChannelsOutput::Summary(output)
+    } else {
+        QueryChannelsOutput::Verbose(output_verbose)
+    };
+    Ok(output)
```
soareschen, line 158: We can do `Output::success` here to skip having to define `QueryChannelsOutput`.

soareschen (changes requested):

relayer-cli/src/commands/query/channels.rs
romac: I actually like having a well-defined return type for the actual command that is only converted to an `Output` at the top-level. Makes the code more explicit and easier to read and reason about imho.

relayer-cli/src/commands/query/channels.rs
romac: https://github.com/informalsystems/ibc-rs/pull/1243/commits/78a617c6313a4b733b72b7b32fd8677ac79f6d40

relayer-cli/src/commands/query/channels.rs
romac: https://github.com/informalsystems/ibc-rs/pull/1243/commits/4e084a86c8a1ace8ac702ab6e094dd66feeb8bfb

relayer-cli/src/commands/query/channels.rs
soareschen: Sounds good. I'm fine either way.

relayer-cli/src/commands/query/channels.rs
romac: @soareschen @adizere In your opinion, should we emit an error and a) abort or b) continue?

romac: I've pushed a commit which just emits an `error!` and continues, so that the user can still get all the other non-failed channel ends.

soareschen (approved):
Looks good and works for me!

PR #812, opened 2021-04-09
Refactor MBT for ICS02 & ICS03 using new modelator Rust API

andrey-kuprianov:
Closes: #713 

## Description

Refactoring of MBT for ICS02 & ICS03 using the new `modelator` Rust API 
______

For contributor use:

- [ ] Updated the __Unreleased__ section of [CHANGELOG.md](https://github.com/informalsystems/ibc-rs/blob/master/CHANGELOG.md) with the issue.
- [ ] If applicable: Unit tests written, added test to CI.
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Updated relevant documentation (`docs/`) and code comments.
- [ ] Re-reviewed `Files changed` in the Github PR explorer.

PR #1249, opened 2021-07-29
Update `modelator` to 0.2.0

jtremback:
## Description

This updates to the new Modelator 0.2.0. Some caveats:

- I am manually setting the model checker to TLC (the default in Modelator 0.2.0 is Apalache), because the TLA+ spec seems not to run in Apalache.
- I had to get rid of an assert I don't understand on https://github.com/informalsystems/ibc-rs/commit/60e3185e9249e1fb499758f37cc72a728e3c187b#diff-ec1a9a4dbc3d8d3aa74543450bb465b3c01d62036a5449fbba077e185f4ce323L66 because it was failing.

Let me know if this looks OK otherwise, and I can add the unclog file and other housekeeping stuff to this PR.

______

For contributor use:

- [ ] (not applicable) If applicable: Unit tests written, added test to CI.
- [ ] (not applicable) Linked to Github issue with discussion and accepted design OR link to spec that describes this work.

### DIFF
--- a/.changelog/unreleased/improvements/1249-update-modelator.md
+++ b/.changelog/unreleased/improvements/1249-update-modelator.md
@@ -1,3 +1,3 @@
-- Update Modelator to 0.2.0
+- Update Modelator to 0.2.0 ([#1249])
 
 [#1249]: https://github.com/informalsystems/ibc-rs/pull/1249
--- a/.changelog/unreleased/improvements/1249-update-modelator.md
+++ b/.changelog/unreleased/improvements/1249-update-modelator.md
@@ -1,0 +1,3 @@
+- Update Modelator to 0.2.0
+
+[#1249]: https://github.com/informalsystems/ibc-rs/pull/1249
--- a/modules/tests/runner/mod.rs
+++ b/modules/tests/runner/mod.rs
@@ -52,7 +52,6 @@ impl IbcTestRunner {
     }
 
     /// Create a `MockContext` for a given `chain_id`.
-    /// Panic if a context for `chain_id` already exists.
     pub fn init_chain_context(&mut self, chain_id: String, initial_height: u64) {
         let chain_id = Self::chain_id(chain_id);
         // never GC blocks
--- a/modules/tests/runner/mod.rs
+++ b/modules/tests/runner/mod.rs
@@ -436,8 +436,8 @@ impl IbcTestRunner {
     }
 }
 
-impl modelator::runner::TestRunner<Step> for IbcTestRunner {
-    fn initial_step(&mut self, step: Step) -> bool {
+impl modelator::step_runner::StepRunner<Step> for IbcTestRunner {
+    fn initial_step(&mut self, step: Step) -> Result<(), String> {
         assert_eq!(step.action, Action::None, "unexpected action type");
         assert_eq!(
             step.action_outcome,
@@ -448,10 +448,10 @@ impl modelator::runner::TestRunner<Step> for IbcTestRunner {
         for (chain_id, chain) in step.chains {
             self.init_chain_context(chain_id, chain.height);
         }
-        true
+        Ok(())
     }
 
-    fn next_step(&mut self, step: Step) -> bool {
+    fn next_step(&mut self, step: Step) -> Result<(), String> {
         let result = self.apply(step.action);
         let outcome_matches = match step.action_outcome {
             ActionOutcome::None => panic!("unexpected action outcome"),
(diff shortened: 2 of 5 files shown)


modules/tests/runner/mod.rs
```diff
@@ -63,7 +63,7 @@ impl IbcTestRunner {
             max_history_size,
             Height::new(Self::revision(), initial_height),
         );
-        assert!(self.contexts.insert(chain_id, ctx).is_none());
+        self.contexts.insert(chain_id, ctx);
```
vitorenesduarte, line 66: I believe this assert fails in the new version because the `IbcTestRunner` is shared across tests: `run_tla_steps` now takes a `&mut`, reusing it on all tests, while before `run` took an owned value that was cloned for each test. (This probably means that the assert only fails when the second test is run.)

modules/tests/runner/mod.rs
```diff
@@ -448,11 +448,11 @@ impl modelator::runner::TestRunner<Step> for IbcTestRunner {
...
             self.init_chain_context(chain_id, chain.height);
         }
-        true
+        Ok(())
     }
 
-    fn next_step(&mut self, step: Step) -> bool {
-        let result = self.apply(step.action);
+    fn next_step(&mut self, step: Step) -> Result<(), String> {
+        let result = self.apply(step.clone().action);
```
vitorenesduarte, line 455: This `.clone()` shouldn't be needed.

modules/tests/runner/mod.rs
jtremback: Is it important? Does it need to go back in and be fixed?

modules/tests/runner/mod.rs
jtremback: Thanks! You're right.

modules/tests/runner/mod.rs
vitorenesduarte: In my view this is fine given the current behaviour of modelator. But I think that the behaviour in the previous version is more intuitive as it allows this kind of asserts. (In this case we're saying that each chain should be initialized only once, which is true within a single test, but false across tests).

jtremback: Ok, leaving the assert out

jtremback:
Security audit is failing but that appears to be a Tendermint dep?

adizere:
I'm not sure I understand the relation between this PR and the previously-opened one #812. Could you clarify if the current PR complete supersedes (so we can close and ignore) #812?

adizere:
> Security audit is failing but that appears to be a Tendermint dep?

Indeed, not exactly Tendermint dep, but fixing the security audit is blocked by prost & tonic (ref https://github.com/informalsystems/ibc-rs/pull/1166#issuecomment-879173781)

jtremback:
> I'm not sure I understand the relation between this PR and the previously-opened one #812. Could you clarify if the current PR complete supersedes (so we can close and ignore) #812?

We should probably wait for @andrey-kuprianov to weigh in, but #812 brings in a whole new API (the EventRunner API). It is also significantly behind master in other respects, which makes completing and merging it hard. 

This PR uses the same API as the Modelator tests in master (the StepRunner API). This is a very straightforward PR, which just involved changing some function names. It unblocks my work on the ICS07 upgrade client mbt tests, since there were some bugs in the old modelator currently used on master.

We may want to finish #812 switch over to the EventRunner API at some point, but IMO that is a separate task and this PR should be merged first.

adizere:
> > I'm not sure I understand the relation between this PR and the previously-opened one #812. Could you clarify if the current PR complete supersedes (so we can close and ignore) #812?
> 
> We should probably wait for @andrey-kuprianov to weigh in, but #812 brings in a whole new API (the EventRunner API). It is also significantly behind master in other respects, which makes completing and merging it hard.
> 
> This PR uses the same API as the Modelator tests in master (the StepRunner API). This is a very straightforward PR, which just involved updating the modelator version and changing some function names etc. It unblocks my work on the ICS07 upgrade client mbt tests, since there were some bugs in the old modelator currently used on master.
> 
> We may want to finish #812 and switch over to the EventRunner API in the future, but IMO that is a separate task and this PR should be merged first.

Got it, thanks for the explanation Jehan, that's very helpful!

Copying @cezarad here if she wants to weight in by reviewing this PR.

.changelog/unreleased/improvements/1249-update-modelator.md
```diff
@@ -0,0 +1,3 @@
+- Update Modelator to 0.2.0
```
romac, line 1:
```suggestion
- Update Modelator to 0.2.0 ([#1249])
```

romac (approved):

PR #1252, opened 2021-08-02
ICS07: Add verification method for client update handler

cezarad:
Closes: #XXX

## Description

This is the former PR check_client_and_update_state: 

- implements the ICS07 verification of the client update checkValidityAndUpdateState in https://github.com/cosmos/ibc/tree/master/spec/client/ics-007-tendermint-client
 
- modifications in the tendermint testgen to make test pass with this extra check.

- added timestamp to Anyclient: On one hand in AnyClientState we have the method```expired``` but in AnyHeader we do not have the timestamp so we cannot check in ICS02 if the update is expired. On the other hand the pseudo code of ICS02 talks only about the height. 

TODO! 
- use directly tendermint-rs functions not to duplicate the code: The duplicates are : the function ```is voting_power_in``` and the function that is calls ```total_power_of``` and ```non_absent_vote``` from ics07_tendermint/header.rs 
The return type of ```is voting_power_in``` is a bit simplified but is a modification easily revertible.  
 
______

For contributor use:


adizere:
Update: This is currently blocked on some changes on tm-rs side. The problem is that we’d like to merge this PR without duplicating the code from tendermint-rs. After talking with @thanethomson, the plan is:

1. We use approach of adding `Clone` and other necessary trait bounds so that we can use `VerificationPredicates` as a trait object
2. Thane can do a bit of refactoring of Cezara’s code to remove the duplicate code and clean up the PR. I will help reviewing this.


Many thanks to @thanethomson and @cezarad for shepherding this!

modules/src/ics07_tendermint/client_def.rs
```diff
@@ -25,17 +36,117 @@ impl ClientDef for TendermintClient {
...
+                cs => AnyConsensusState::Tendermint
+            )
+            .ok_or_else(|| Ics02Error::client_args_type_mismatch(ClientType::Tendermint))?;
+
+            if consensus_state != ConsensusState::from(header.clone()) {
+                //freeze the client and return the installed consensus state
+                return Ok((
+                    client_state.with_set_frozen(header.height()),
+                    consensus_state,
+                ));
```
ancazamfir, line 59: We should freeze the client only if the conflicting header verifies. i.e. we need to run the checks/ validity predicates (from the new tendermint-rs verify function) below. If the header in update doesn't verify we should return an error.

modules/src/ics07_tendermint/client_def.rs
```diff
@@ -25,17 +36,117 @@ impl ClientDef for TendermintClient {
...
+            } else {
+                return Ok((client_state, consensus_state));
+            }
+        };
+
+        let latest_consensus_state =
+            match ctx.consensus_state(&client_id, client_state.latest_height) {
```
ancazamfir, line 66: We have to get the consensus state based on the header trusted height which might not be the latest:
```suggestion
        let trusted_consensus_state =
            match ctx.consensus_state(&client_id, header.trusted_height) {
```

modules/src/ics07_tendermint/client_def.rs
```diff
@@ -25,17 +36,117 @@ impl ClientDef for TendermintClient {
...
+                None => {
+                    return Err(Ics02Error::consensus_state_not_found(
+                        client_id.clone(),
+                        client_state.latest_height,
+                    ));
+                }
+            };
+
+        monotonicity_checks(latest_consensus_state, header.clone(), client_state.clone())
+            .map_err(Ics02Error::tendermint_handler_error)?;
```
ancazamfir, line 81: The `monotonicity..` check should also deal with in-the-middle updates where a consensus state with higher height than the header is present. This requires a new ctx api to get the next consensus state, something like this
```suggestion
        let next_higher_consensus_state: Option<ConsensusState> = ctx.next_consensus_state(&client_id, header.height())?;
        monotonicity_checks(trusted_consensus_state, next_higher_consensus_state, header.clone(), client_state.clone())
            .map_err(Ics02Error::tendermint_handler_error)?;
```

modules/src/ics07_tendermint/client_def.rs
```diff
@@ -25,17 +36,117 @@ impl ClientDef for TendermintClient {
...
+                ts => AnyConsensusState::Tendermint
+            )
+            .ok_or_else(|| Ics02Error::client_args_type_mismatch(ClientType::Tendermint))?,
+            None => {
+                return Err(Ics02Error::consensus_state_not_found(
+                    client_id,
+                    header.trusted_height,
+                ))
+            }
+        };
```
ancazamfir, line 103: this can be removed as we should use this above instead of `latest_consensus_state`.

modules/src/ics07_tendermint/client_def.rs
```diff
@@ -25,17 +36,117 @@ impl ClientDef for TendermintClient {
...
+
+        if header.height() == header.trusted_height.increment() {
+            //adjacent
+
+            // check that the header's trusted validator set is
+            // the next_validator_set of the trusted consensus state
+            if Set::hash(&header.validator_set) != trusted_consensus_state.next_validators_hash {
```
ancazamfir, line 110: This check should be done for both adjacent and non-adjacent. And you need the trusted validator instead:
```suggestion
            if Set::hash(&header.trusted_validator_set) != trusted_consensus_state.next_validators_hash {
```
This check should be done in the tendermint light client API.

modules/src/ics07_tendermint/client_def.rs
```diff
@@ -25,17 +36,117 @@ impl ClientDef for TendermintClient {
...
+
+            // check that the validators that sign the commit of the untrusted header
+            // have 2/3 of the voting power of the current validator set.
+            if let Err(e) = voting_power_in(
+                &header.signed_header,
+                &header.validator_set,
+                TrustThresholdFraction::TWO_THIRDS,
+            ) {
+                return Err(Error::insufficient_voting_power(e.to_string()).into());
+            }
```
ancazamfir, line 126: same here, the check should happen for both adjacent and non-adjacent headers. In fact you have this below in the `else` branch.

modules/src/ics07_tendermint/client_def.rs
```diff
@@ -25,17 +36,117 @@ impl ClientDef for TendermintClient {
...
+            // check that the validators that sign the commit of the untrusted header
+            // have 2/3 of the voting power of the current validator set.
+            if let Err(e) = voting_power_in(
+                &header.signed_header,
+                &header.validator_set,
+                TrustThresholdFraction::TWO_THIRDS,
+            ) {
+                return Err(Error::insufficient_voting_power(e.to_string()).into());
+            };
+        }
```
ancazamfir, line 149: All these checks should happen in the `tendermint-rs` light client API. It should also include the following checks:
- trusted header is not expired, i.e. it is within trusting period (note that we cannot provide a full trusted header to the API, just what we can get from the consensus state: height, time and next val hash) 
- update header timestamp is not beyond `now()` plus clock drift

Probably most of the monotonicity check between trusted and update headers should happen in the tendermint light client API. The only one left in this function should be between the update and next higher headers, just check the timestamps are monotonically increased and nothing else.

ancazamfir (changes requested):
I only reviewed `modules/src/ics07_tendermint/client_def.rs` for now. A lot of those changes should go in the tendermint light client API imho.

romac:
Superseded by #1321.

PR #1254, opened 2021-08-02
Fix for regression #1229 using new type ics02::TrustThreshold

### LINKED ISSUE #1229: CLIs for upgrades regression following update to tm-rs 0.21
## Crate

`ibc-relayer`

## Summary of Bug

Following the update to tm-rs 0.21 (#1223), in particular this change mentioned [here](https://github.com/informalsystems/ibc-rs/pull/1223#discussion_r674113798), the `upgrade-chain` functionality fails. The `zero_custom_fields` method can no longer enforce the trust threshold to be 0/0, because that is not a valid domain type construction of the `TrustThresholdFraction` type.

There is a possible workaround to make the `upgrade-chain` command work successfully, documented in this commit
https://github.com/informalsystems/ibc-rs/commit/27c1a7e29c05cfd7475509cee9951568d319dbea
This workaround achieves the following:
- construct a `prost_types::Any` directly from a raw `ClientState`
- the raw `ClientState` has not type checks, so we can enforce zero trust_level
- enables the `upgrade-chain` tx to succeed

The drawback of this solution is that it's partial: the `upgrade client(s)` CLIs still don't work, because they required reading and decoding the client state from the upgraded chain, and this state includes the zero value, which cannot be decoded. 

>  hermes upgrade client ibc-1 07-tendermint-0
Jul 22 16:50:48.047  INFO [ibc-0 -> ibc-1:07-tendermint-0] upgrade Height: 0-2401
Jul 22 16:50:48.072 DEBUG [ibc-0 -> ibc-1:07-tendermint-0] MsgUpdateAnyClient from trusted height 0-231 to target height 0-2401
Error: failed while trying to upgrade client id 07-tendermint-0 hosted on chain id ibc-1 with error: failed while fetching from chain ibc-0 the upgraded client state: GRPC error: invalid raw client state: error converting message type into domain type: invalid client state trust threshold: trust threshold is too large (must be <= 1)

Two possible long-term solution:

- modify the tm-rs type `TrustThresholdFraction` to permit construction with zero field
- replicate the domain type `TrustThresholdFraction` in the `relayer` crate and handle its validation there, permitting construction with zero fields

__As a short term solution, for 0.6.1, we are explicitly disabling the `upgrade-chain` and `upgrade client` functionality.__

### More details on the underlying problem

A successful governance proposal would have a zero trust level:

> gaiad query gov proposal 1 --home data/ibc-0/
...
      trust_level:
        denominator: "0"
        numerator: "0"

Using the changes from a0d9d4b5823a9f0bb9850c990fe9dfdfeaeb00e0, the chain accepts the upgrade proposal, but the upgraded trust level is not zero:

> gaiad query gov proposal 1 --node http://127.0.0.1:26557
...
      trust_level:
        denominator: "3"
        numerator: "1"

The `trust_level` field, like all the other fields specified in the `zero_custom_fields` method, are custom and differ from client to client (these are chosen by the relayer when that client was created). Since these are client-specific field, the upgrading chain must set them to zero, and must only change those fields that are chain-specific (chain id, unbonding period).

After discussing with @colin-axner, here are further details:
- the x/gov module stores the governance proposal, so when we query for it, we see the 1/3 trust level details
- the [client state gets zeroed out](https://github.com/cosmos/cosmos-sdk/blob/84c33215658131d87daf3c629e909e12ed9370fa/x/upgrade/keeper/keeper.go#L104) when the chain stores it to the ibc store keeper, so effectively the stored trust level (and its proof) are for 0/0 trust level
- it seems that the on-chain handler for MsgUpgradeClient does not zero out the submitted client state, so if a relayer submits a client state with 1/3, the proof won't pass
- the `upgrade client` CLI subsequently fails with an error specifying that proof verification failed, [log here](https://gist.github.com/adizere/70e67aa38227e14fddfb1a426082aeac).

## Version

3c383cf5650063c2dbfb4af52239af6fc8782e0b

## Acceptance Criteria

- [ ] re-enable `upgrade-chain` and `upgrade client` CLIs

____

## For Admin Use

- [ ] Not duplicate issue

adizere:
Closes #1229

## Description

- [X] Added a new domain type `TrustThreshold` in ICS02, which is an IBC-compliant representation of trust threshold.
- [X] re-enabled the chain and client upgrade CLIs

______

For contributor use:

### DIFF
--- a/modules/src/ics02_client/trust_threshold.rs
+++ b/modules/src/ics02_client/trust_threshold.rs
@@ -52,8 +52,10 @@ impl TrustThreshold {
     /// is in the range`[0, 1)`.
     pub fn new(numerator: u64, denominator: u64) -> Result<Self, Error> {
         // The two parameters cannot yield a fraction that is bigger or equal to 1
-        if (numerator > denominator) || (denominator == 0 && numerator != 0) ||
-            (numerator == denominator && numerator != 0) {
+        if (numerator > denominator)
+            || (denominator == 0 && numerator != 0)
+            || (numerator == denominator && numerator != 0)
+        {
             return Err(Error::invalid_trust_threshold(numerator, denominator));
         }
 
--- a/modules/src/ics02_client/trust_threshold.rs
+++ b/modules/src/ics02_client/trust_threshold.rs
@@ -1,0 +1,128 @@
+//! IBC Domain type definition for [`TrustThreshold`]
+//! represented as a fraction with valid values in the
+//! range `[0, 1)`.
+
+use std::{convert::TryFrom, fmt};
+
+use serde::{Deserialize, Serialize};
+use tendermint_proto::Protobuf;
+
+use ibc_proto::ibc::lightclients::tendermint::v1::Fraction;
+use tendermint::trust_threshold::TrustThresholdFraction;
+
+use crate::ics02_client::error::Error;
+
+/// [`TrustThreshold`] defines the level of trust that a client has
+/// towards a set of validators of a chain.
+///
+/// A trust threshold is represented as a fraction, i.e., a numerator and
+/// and a denominator.
+/// A typical trust threshold is 1/3 in practice.
+/// This type accepts even a value of 0, (numerator = 0, denominator = 0),
+/// which is used in the client state of an upgrading client.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct TrustThreshold {
+    numerator: u64,
+    denominator: u64,
+}
+
+impl TrustThreshold {
+    /// Constant for a trust threshold of 1/3.
+    pub const ONE_THIRD: Self = Self {
+        numerator: 1,
+        denominator: 3,
+    };
+
+    /// Constant for a trust threshold of 2/3.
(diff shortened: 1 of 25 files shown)


modules/src/ics02_client/trust_threshold.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// Instantiate a TrustThreshold with the given denominator and
+    /// numerator.
+    ///
+    /// The constructor succeeds if long as the resulting fraction
+    /// is in the range`[0, 1)`.
+    pub fn new(numerator: u64, denominator: u64) -> Result<Self, Error> {
+        // The two parameters cannot yield a fraction that is bigger or equal to 1
+        if (numerator > denominator)
+            || (denominator == 0 && numerator != 0)
+            || (numerator == denominator && numerator != 0)
```
romac, line 57: I don't think the parens are necessary but we can keep them if you think it improve readability.

romac (approved):
Very nice! Left a couple nitpicks but pre-approved

modules/src/ics02_client/trust_threshold.rs
adizere: Hmm, the parens did help me while writing this code.

Issue #1231, opened 2021-07-23
When I send IBC packets at endblock handle, the relayer cannot work normally


lyqingye:
## Crate
relayer
relayer-cli

## Summary of Bug
I want to connect src with dst. src sends IBC packets to dst at ABCI Endblock, that code look like this

```go
// EndBlocker updates the current price
func EndBlocker(ctx sdk.Context, k keeper.Keeper) {
	// ... do something
       k.channelKeeper.SendPacket(ctx, channelCap, packet);
}
```
In the relayer log, you can see the following information
```bash
DEBUG [src:oracle/channel-0 -> dst] recv packets to send out to dst of the ones with commitments on source src: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 (first 10 shown here; total=23)
Jul 23 15:42:01.857  
INFO [src:oracle/channel-0 -> dst] found unprocessed SendPacket events for [] (first 10 shown here; total=0)
```
The relayer can query the comments in SRC that are not acked, as described in the first log
However, the relayer cannot found unprocessed sendpacket events according to the sequence corresponding to the commitments
I read the relayer code, https://github.com/informalsystems/ibc-rs/blob/master/relayer/src/link/relay_path.rs#L836

```rust
events_result = self.src_chain().query_txs(query)?;
```
This line of code is to query the corresponding transaction information according to the information of commitments, and extract IBC event from the transaction information.
because my IBC packet is sent at endblock, this line of code will return empty events
so the relayer cannot forward IBC packets normally

I think this is a bug because the IBC send_packet events may not exist in Tx or in beginblock / endblock (NewBlock)

## Version
0.6.0

## Acceptance Criteria

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
Hi,
Thanks for documenting this issue and for your patience!

> This line of code is to query the corresponding transaction information according to the information of commitments, and extract IBC event from the transaction information.

That's correct. Hermes uses the Tendermint ABCI [`tx_search`](https://docs.tendermint.com/master/rpc/#/Info/tx_search) to query each packet (by its source and destination port/channel id and its seq. nr.) obtain packet information.

> because my IBC packet is sent at endblock, this line of code will return empty events

[...]

PR #1172, opened 2021-07-12
Get events from RPC event's data instead of events

ancazamfir:
Closes: #1191

## TODOs

- [x] Please remember to delete any constants that we might have added for matching `message.action`, e.g., [here](https://github.com/informalsystems/ibc-rs/pull/948/files/4bf511c974a0032c36a78f244f10fd7369589602#diff-e60b388a28a7d3e1fd6a6f60b68e496e48f3ebdb949787758fddf7a1b3180cf8).
    - (Cross-posted from PR https://github.com/informalsystems/ibc-rs/pull/948)

Context for this PR: Simplifies greatly the relayer's RPC events parsing architecture. Additionally, this is a low-hanging fruit to improve the ergonomics of testing with Hermes and Interchain Accounts (ICS27). Main customer benefiting from this: @seantking.

## Description

______

For contributor use:

- [x] Updated the __Unreleased__ section of [CHANGELOG.md](https://github.com/informalsystems/ibc-rs/blob/master/CHANGELOG.md) with the issue.
- [ ] If applicable: Unit tests written, added test to CI.
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Updated relevant documentation (`docs/`) and code comments.
- [x] Re-reviewed `Files changed` in the Github PR explorer.

PR #1277, opened 2021-08-12
simplify get all event from tx logic

lyqingye:
reuse code  [from_tx_response_event](https://github.com/informalsystems/ibc-rs/blob/master/modules/src/events.rs#L168) for [get_all_events](https://github.com/informalsystems/ibc-rs/blob/master/relayer/src/event/rpc.rs#L15)

## Description

extract IBC packets directly from events of tx result 

______

For contributor use:

### DIFF
--- a/relayer/src/event/rpc.rs
+++ b/relayer/src/event/rpc.rs
@@ -1,16 +1,10 @@
-use std::{collections::HashMap, convert::TryFrom};
-
+use ibc::events::from_tx_response_event;
 use tendermint_rpc::event::{Event as RpcEvent, EventData as RpcEventData};
 
 use ibc::ics02_client::events::NewBlock;
 use ibc::ics02_client::height::Height;
 use ibc::ics24_host::identifier::ChainId;
-use ibc::{
-    events::{Error as EventError, IbcEvent, RawObject},
-    ics02_client::events as ClientEvents,
-    ics03_connection::events as ConnectionEvents,
-    ics04_channel::events as ChannelEvents,
-};
+use ibc:: events::IbcEvent;
 
 pub fn get_all_events(
     chain_id: &ChainId,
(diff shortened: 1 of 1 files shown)


RiccardoM:
@adizere Can we have an update on this? IBC integration is going to play a big role inside the Cosmos ecosystem, and they rely on end block events for sending back the result of their async work

romac:
@RiccardoM Can you please provide some rationale as to why this change is needed and how it helps with your use case?

romac:
@lyqingye @RiccardoM @RiccardoM Can you please provide some rationale as to why this change is needed and how it helps with your use case?

RiccardoM:
> @lyqingye @RiccardoM Can you please provide a more detailed rationale as to why this change is needed and how it helps with the use case?

I think [this](https://github.com/informalsystems/ibc-rs/issues/1231#issuecomment-897294980) better explains it

lyqingye:
> @lyqingye @RiccardoM Can you please provide a more detailed rationale as to why this change is needed and how it helps with the use case?

The purpose of this change is to extract IBC send_packets in any transaction. The difference [build_event](https://github.com/informalsystems/ibc-rs/blob/master/relayer/src/event/rpc.rs#L59) is that [build_event](https://github.com/informalsystems/ibc-rs/blob/master/relayer/src/event/rpc.rs#L59) only extracts send_packets from specified action of transactions, such as "transfer" | "register" | "send".

Because IBC send_packets of some chains are generated by custom transactions, the Hermes cannot relay this packets

romac:
@ancazamfir Can this be closed now that https://github.com/informalsystems/ibc-rs/pull/1172 is in?

ancazamfir:
> @ancazamfir Can this be closed now that #1172 is in?

Yes!

romac:
Superseded and fixed in #1172.

PR #1285, opened 2021-08-13
Correct small TypeOk bug in IBC.tla

jtremback:
Closes: #XXX (I'm not doing a separate issue because it's a one-line fix that I noticed and fixed)

## Description

This PR fixes one line in the IBC.tla testing spec. There was a typo in the typeOK invariant which stops this spec from running correctly. It did not affect the tests because they import a different spec, which imports this spec, but whose config file does not check the invariant with the typo.
______

For contributor use:


adizere (approved):
Good catch!

PR #1286, opened 2021-08-15
Attempt to document the associated chains for each packet field

soareschen:
## Description

This is my attempt to document the associated chains for each packet field to understand how the relayer code really works. I am pretty sure some of the documentation are completely wrong. So the hope is through the reviews I can get a clearer picture of which packet field really belongs to which chain.

______

For contributor use:

### DIFF
--- a/modules/src/ics04_channel/msgs.rs
+++ b/modules/src/ics04_channel/msgs.rs
@@ -39,10 +39,47 @@ pub enum ChannelMsg {
     ChannelCloseConfirm(MsgChannelCloseConfirm),
 }
 
+/*
+    `PacketMsg` represents packet messages sent between an ALPHA chain
+    and a BETA chain.
+
+    The roles of ALPHA and BETA are FIXED and never change during
+    the relaying session. DIFFERENT relaying sessions are executed
+    for the case where the ALPHA and BETA roles are FLIPPED
+
+    Depending on the variant, a packet can flow in *either* direction,
+    and the exact flow is documented below.
+
+    All variants are parameterized by the SAME ALPHA and BETA
+    chains, i.e. the chain roles DO NOT FLIP.
+
+    In the same relaying session, the packet's ALPHA chain is
+    ALWAYS the SOURCE chain, and the packet's BETA
+    chain is ALWAYS the DESTINATION chain.
+*/
 #[derive(Clone, Debug, PartialEq)]
 pub enum PacketMsg {
+    /*
+        A `RecvPacket` represents the ALPHA chain RECEIVING
+        an INCOMING packet from the BETA chain.
+     */
     RecvPacket(MsgRecvPacket),
+
+    /*
+        An `AckPacket` represents the ALPHA chain SENDING
+        an acknowlegement as OUTGOING packet to the BETA chain.
+    */
     AckPacket(MsgAcknowledgement),
+
+    /*
+        A `ToPacket` represents the ALPHA chain RECEIVING
+        a timeout INCOMING packet from the BETA chain.
+     */
     ToPacket(MsgTimeout),
+
+    /*
+        A `ToClosePacket` represents the ALPHA chain RECEIVING
+        a timeout-on-close INCOMING packet from the BETA chain.
+     */
     ToClosePacket(MsgTimeoutOnClose),
 }
--- a/modules/src/ics04_channel/msgs/acknowledgement.rs
+++ b/modules/src/ics04_channel/msgs/acknowledgement.rs
@@ -12,14 +12,31 @@ use crate::tx_msg::Msg;
 
 pub const TYPE_URL: &str = "/ibc.core.channel.v1.MsgAcknowledgement";
 
-///
-/// Message definition for packet acknowledgements.
-///
+/*
+    A `MsgRecvPacket` is part of a `PacketMsg` which represents
+    an OUTGOING acknowledgment packet SENT by the ALPHA chain to
+    the BETA chain.
+ */
 #[derive(Clone, Debug, PartialEq)]
 pub struct MsgAcknowledgement {
+    /*
+        An OUTGOING packet with the SAME ALPHA and BETA chains.
+     */
     pub packet: Packet,
+
+    /*
+        The acknowlegement PRODUCED BY the ALPHA chain.
+     */
     pub acknowledgement: Vec<u8>, // TODO(romac): Introduce a newtype for this
+
+    /*
+        The proof of acknowledgement PRODUCED BY the ALPHA chain.
+     */
     pub proofs: Proofs,
+
+    /*
+        The signer FOR the BETA chain.
+     */
     pub signer: Signer,
 }
 
--- a/modules/src/ics04_channel/msgs/recv_packet.rs
+++ b/modules/src/ics04_channel/msgs/recv_packet.rs
@@ -12,13 +12,28 @@ use crate::tx_msg::Msg;
 
 pub const TYPE_URL: &str = "/ibc.core.channel.v1.MsgRecvPacket";
 
-///
-/// Message definition for the "packet receiving" datagram.
-///
+/*
+    Message definition for the "packet receiving" datagram.
+
+    A `MsgRecvPacket` is part of a `PacketMsg` which represents
+    an INCOMING packet RECEIVED by the ALPHA chain
+    and SENT from the BETA chain.
+ */
 #[derive(Clone, Debug, PartialEq)]
 pub struct MsgRecvPacket {
+    /*
+        An INCOMING packet with the ALPHA and BETA chains remain the SAME.
+     */
     pub packet: Packet,
+
+    /*
+        A proof that is produced FROM the BETA chain.
+     */
     pub proofs: Proofs,
+
+    /*
+        A signer FOR the ALPHA chain.
+     */
     pub signer: Signer,
 }
 
--- a/modules/src/ics04_channel/msgs/timeout.rs
+++ b/modules/src/ics04_channel/msgs/timeout.rs
@@ -12,14 +12,33 @@ use crate::tx_msg::Msg;
 
 pub const TYPE_URL: &str = "/ibc.core.channel.v1.MsgTimeout";
 
-///
-/// Message definition for packet timeout domain type.
-///
+/*
+    Message definition for packet timeout domain type.
+
+    A `MsgTimeout` is part of `PacketMsg` representing an
+    INCOMING timeout packet RECEIVED BY the ALPHA chain
+    and SENT FROM the BETA chain.
+ */
 #[derive(Clone, Debug, PartialEq)]
 pub struct MsgTimeout {
+    /*
+        An INCOMING packet with the SAME ALPHA and BETA chains.
+     */
     pub packet: Packet,
+
+    /*
+        The sequence number ON the ALPHA chain.
+     */
     pub next_sequence_recv: Sequence,
+
+    /*
+        The proof FROM the BETA chain proving the timeout.
+     */
     pub proofs: Proofs,
+
+    /*
+        The signer FOR the ALPHA chain.
+     */
     pub signer: Signer,
 }
 
--- a/modules/src/ics04_channel/packet.rs
+++ b/modules/src/ics04_channel/packet.rs
@@ -15,13 +15,28 @@ use super::handler::{
     timeout::TimeoutPacketResult, write_acknowledgement::WriteAckPacketResult,
 };
 
-/// Enumeration of proof carrying ICS4 message, helper for relayer.
+/*
+    Enumeration of proof carrying ICS4 message, helper for relayer.
+
+    The `PacketMsgType` is used for identifying the message type
+    for packets sent between an ALPHA chain an a BETA chain.
+
+    It also identifies the DIRECTION on the packet, on whether
+    it is an INCOMING or OUTGOING packet.
+ */
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum PacketMsgType {
+    // INCOMING packet
     Recv,
+
+    // OUTGOING packet
     Ack,
     TimeoutUnordered,
+
+    // INCOMING packet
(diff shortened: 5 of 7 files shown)


adizere (commented):
Had a quick look, hope the comments I left help!

I would like to encourage this effort, Soares! I think it's very valuable having a specification of the relayer's workflow with respect to packets, which is very complex and poorly documented.

Reading through this, I'm starting to learn that probably we _cannot_ get away from having multiple naming schemes, e.g., beta vs. alpha <> party vs. counterparty <> src vs. destination. The domain is very complex and each separate scheme seems to be the most appropriate in different contexts.

modules/src/ics04_channel/msgs.rs
```diff
@@ -39,10 +39,47 @@ pub enum ChannelMsg {
     ChannelCloseConfirm(MsgChannelCloseConfirm),
 }
 
+/*
+    `PacketMsg` represents packet messages sent between an ALPHA chain
```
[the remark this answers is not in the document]
soareschen: > I'm not sure the right place to document these are in the modules (as we're doing here).

Yeah, I think we can put all explanation in one place, or cross reference to other places where it is used. The documentation should also target cargo doc so that readers can read from the generated Rustdoc.

> The exact details of how packets flow depend quite a bit on the relayer's internal bookkeeping: a RecvPacket could equally be flowing from what we call a source to a destination, or the other way around.

Here I try to persist the chain roles across module boundaries. So for example if a `RelayPath` object treats its source chain as Alpha and its destination chain as Beta, then my understanding is that the same `RelayPath` should not suddenly treat its source chain as Beta and its destination chain as Alpha. If my assumption is wrong, then it is even more important to determine where the roles are flipped.

modules/src/ics04_channel/msgs/acknowledgement.rs
```diff
@@ -12,14 +12,31 @@ use crate::tx_msg::Msg;
...
+     */
     pub packet: Packet,
+
+    /*
+        The acknowlegement PRODUCED BY the ALPHA chain.
+     */
     pub acknowledgement: Vec<u8>, // TODO(romac): Introduce a newtype for this
+
+    /*
+        The proof of acknowledgement PRODUCED BY the ALPHA chain.
```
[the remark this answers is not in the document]
soareschen: Thanks for clarifying. Yeah perhaps it is more appropriate to state it as that the proofs are assembled from the Alpha chain. Nevertheless, the main point is to determine whether we are using the state from chain Alpha or chain Beta to construct the proofs.

modules/src/ics04_channel/msgs/timeout.rs
```diff
@@ -12,14 +12,33 @@ use crate::tx_msg::Msg;
...
 pub const TYPE_URL: &str = "/ibc.core.channel.v1.MsgTimeout";
 
-///
-/// Message definition for packet timeout domain type.
-///
+/*
+    Message definition for packet timeout domain type.
+
+    A `MsgTimeout` is part of `PacketMsg` representing an
+    INCOMING timeout packet RECEIVED BY the ALPHA chain
```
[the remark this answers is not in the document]
soareschen: Thanks for clarifying. I guess the correct explanation is that the timeout packet is constructed from the evidence that chain Beta's height/time state has exceeded the timeout specified. The main point is we are constructing the information from chain Beta and then submit it to chain Alpha.

soareschen:
> Reading through this, I'm starting to learn that probably we cannot get away from having multiple naming schemes, e.g., beta vs. alpha <> party vs. counterparty <> src vs. destination. The domain is very complex and each separate scheme seems to be the most appropriate in different contexts.

I recently realized that the reason we need different terminologies are because there are 3 parties involved, but each terminology focus on the interaction between only 2 parties. As a result, depending on which two parties are involved and the directions, we need to use different terminologies for the situation. For example:

- Relayer -> Chain - The relayer is submitting transaction to one of the 2 chains. A suitable term may be "Relayer -> Host Chain", or less ideally "Relayer -> Destination Chain".
- Relayer <- Chain - The relayer is querying the state or events emitted by one of the two chains. A suitable term may be "Relayer <- Host Chain" , or less ideally "Relayer <- Source Chain".
- Chain -> Counterparty Chain - A chain registers in its state that it is sending a packet to its IBC client. On the counterparty chain it registers its state as its IBC client sending a packet to it. A suitable term may be "Source Chain -> Destination Chain", or "Alpha Chain -> Beta Chain"
- Chain <- Counterparty Chain - A chain registers in its state that it receives a packet from its IBC client. On the counterparty chain it registers its state as it sends a packet to its IBC client. A suitable term may be "Alpha Chain <- Beta Chain". The term "Source Chain <- Destination Chain" completely breaks down in this case.

modules/src/ics04_channel/msgs.rs
```diff
@@ -39,10 +39,47 @@ pub enum ChannelMsg {
...
+
+    In the same relaying session, the packet's ALPHA chain is
+    ALWAYS the SOURCE chain, and the packet's BETA
+    chain is ALWAYS the DESTINATION chain.
+*/
 #[derive(Clone, Debug, PartialEq)]
 pub enum PacketMsg {
+    /*
+        A `RecvPacket` represents the ALPHA chain RECEIVING
+        an INCOMING packet from the BETA chain.
```
ancazamfir, line 64:
```suggestion
        A `RecvPacket` represents the BETA chain RECEIVING
        an INCOMING packet from the ALPHA chain.
```

modules/src/ics04_channel/msgs.rs
```diff
@@ -39,10 +39,47 @@ pub enum ChannelMsg {
...
+     */
     RecvPacket(MsgRecvPacket),
+
+    /*
+        An `AckPacket` represents the ALPHA chain SENDING
+        an acknowlegement as OUTGOING packet to the BETA chain.
```
ancazamfir, line 70:
```suggestion
        An `AckPacket` represents the BETA chain SENDING
        an acknowlegement as OUTGOING packet to the ALPHA chain.
```

modules/src/ics04_channel/msgs.rs
```diff
@@ -39,10 +39,47 @@ pub enum ChannelMsg {
...
+    */
     AckPacket(MsgAcknowledgement),
+
+    /*
+        A `ToPacket` represents the ALPHA chain RECEIVING
+        a timeout INCOMING packet from the BETA chain.
```
ancazamfir, line 76: Here and in general...a chain doesn't send anything. It is the relayer that sends the messages. In the timeout case is more obvious because the chain B in this case doesn't receive any kind of packet or notification. It is the relayer that determines this and sends it to A.
```suggestion
        A `ToPacket` relays a notification to the ALPHA chain about a timeout that occurred due to the BETA chain advancing over the packet timeout height.
```

modules/src/ics04_channel/msgs.rs
ancazamfir: I would avoid using the terms "OUTGOING" and "INCOMING". 
As I mentioned before, all ics04 packet messages include a `Packet` that is identical for all the messages for a given channel and packet sequence. So here, if we consider A the source chain (where `packet_src_channel` and `packet_src_port` are) and B is the destination chain (where `packet_dst_channel` and `packet_dst_port`) are.

```suggestion
        An `AckPacket` relays information to the source chain (A) about the destination chain (B) writing
        an acknowlegement for the specified packet included in the message.
```

modules/src/ics04_channel/msgs.rs
```diff
@@ -39,10 +39,47 @@ pub enum ChannelMsg {
...
+     */
     ToPacket(MsgTimeout),
+
+    /*
+        A `ToClosePacket` represents the ALPHA chain RECEIVING
+        a timeout-on-close INCOMING packet from the BETA chain.
```
ancazamfir, line 82:
```suggestion
A `ToClosePacket` relays a notification to the ALPHA chain about a timeout that occurred due to the channel on BETA chain being closed
```

modules/src/ics04_channel/msgs/acknowledgement.rs
```diff
@@ -12,14 +12,31 @@ use crate::tx_msg::Msg;
 
 pub const TYPE_URL: &str = "/ibc.core.channel.v1.MsgAcknowledgement";
 
-///
-/// Message definition for packet acknowledgements.
-///
+/*
+    A `MsgRecvPacket` is part of a `PacketMsg` which represents
+    an OUTGOING acknowledgment packet SENT by the ALPHA chain to
+    the BETA chain.
```
ancazamfir, line 18:
```suggestion
    A `MsgRecvPacket` is a type of `PacketMsg` that relays information about a sent packet stored on the ALPHA chain with source chain being ALPHA and destination chain being BETA.
```

modules/src/ics04_channel/msgs/acknowledgement.rs
```diff
@@ -12,14 +12,31 @@ use crate::tx_msg::Msg;
...
+ */
 #[derive(Clone, Debug, PartialEq)]
 pub struct MsgAcknowledgement {
+    /*
+        An OUTGOING packet with the SAME ALPHA and BETA chains.
```
ancazamfir, line 23: This is a bit confusing. "same" as compared to what?

modules/src/ics04_channel/msgs/acknowledgement.rs
```diff
@@ -12,14 +12,31 @@ use crate::tx_msg::Msg;
...
+     */
     pub packet: Packet,
+
+    /*
+        The acknowlegement PRODUCED BY the ALPHA chain.
```
ancazamfir, line 28:
```suggestion
        The acknowlegement PRODUCED BY the BETA chain.
```

modules/src/ics04_channel/msgs/recv_packet.rs
```diff
@@ -12,13 +12,28 @@ use crate::tx_msg::Msg;
...
 
-///
-/// Message definition for the "packet receiving" datagram.
-///
+/*
+    Message definition for the "packet receiving" datagram.
+
+    A `MsgRecvPacket` is part of a `PacketMsg` which represents
+    an INCOMING packet RECEIVED by the ALPHA chain
+    and SENT from the BETA chain.
```
ancazamfir, line 20:
```suggestion
    A `MsgRecvPacket` is a type of a `PacketMsg` that relays to BETA the information that a packet with source chain ALPHA for destination chain BETA has been stored on ALPHA chain.
```

modules/src/ics04_channel/msgs/recv_packet.rs
```diff
@@ -12,13 +12,28 @@ use crate::tx_msg::Msg;
...
+ */
 #[derive(Clone, Debug, PartialEq)]
 pub struct MsgRecvPacket {
+    /*
+        An INCOMING packet with the ALPHA and BETA chains remain the SAME.
+     */
     pub packet: Packet,
+
+    /*
+        A proof that is produced FROM the BETA chain.
```
ancazamfir, line 30:
```suggestion
        A proof that is produced FROM the ALPHA chain.
```

modules/src/ics04_channel/packet.rs
```diff
@@ -15,13 +15,28 @@ use super::handler::{
...
+
+    The `PacketMsgType` is used for identifying the message type
+    for packets sent between an ALPHA chain an a BETA chain.
+
+    It also identifies the DIRECTION on the packet, on whether
+    it is an INCOMING or OUTGOING packet.
+ */
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum PacketMsgType {
+    // INCOMING packet
```
ancazamfir, line 29:
```suggestion
    // OUGOING packet
```

modules/src/ics04_channel/packet.rs
```diff
@@ -15,13 +15,28 @@ use super::handler::{
...
     Recv,
+
+    // OUTGOING packet
```
ancazamfir, line 32:
```suggestion
    // INCOMING packet
```

modules/src/ics04_channel/packet.rs
```diff
@@ -99,16 +114,84 @@ impl std::fmt::Display for Sequence {
...
+    a HOST chain and a COUNTERPARTY chain.
+    It is in fact used in TWO DIRECTIONS: INCOMING and OUTGOING.
+
+    When using a packet, one must identify the HOST chain,
+    the COUNTERPARTY chain, *and* also the DIRECTION.
+    Inside a relaying session, the HOST and COUNTERPARTY
+    chains of the packets remain FIXED and are NEVER FLIPPED.
+
+    Depending on the packet mode, some fields in `Packet`
+    belongs to DIFFERENT chain. This is further noted below.
```
ancazamfir, line 128: This is really confusing to me. Maybe we should discuss this in a call.

ancazamfir (changes requested):
There are a lot of comments spread across multiple files which doesn't help with keeping things consistent. There are quite a few ones i don't understand. I am not thrilled with using "incoming", "outgoing", "ALPHA" and "BETA".  We currently have the notion of packet's source and destination, and the relay path's source and destination. I thought that as long as we know which we are talking about the meaning of src and dest should be clear.

Maybe we should we do a sitdown review for this?

modules/src/ics04_channel/packet.rs
andynog: also adding some comments. We should try to be more in-line with the terms used in the ecosystem (other relayers, specs, docs). For example

* ts-relayer: uses `source | destination` - https://github.com/confio/ts-relayer/blob/main/spec/ibc-setup.md

* golang relayer: uses `source | destination` - https://github.com/cosmos/relayer/blob/112205bab8224295cb046badcda512d37ea1c7b8/cmd/paths.go#L47

and docs also use `chain A | chain B` so I think those are the most popular.

Trying to use terms like `Host | Guest` or `Alpha | Beta` in my opinion might cause some confusion.

Issue #1155, opened 2021-07-06
Secp256k1 signature verification (BandChain)


greg-szabo:
## Crate

relayer
relayer-cli

## Summary of Bug

hermes seems to be incompatible with chains that use `tendermint/PubKeySecp256k1` keys.

## Version

commit ID: `3da7dc12c060271176c8a8870aa399a8eea64820`
(master as of opening the issue)

## Steps to Reproduce

1. Add the secp256k1 feature to the tendermint dependency in `relayer-cli/Cargo.toml`
1. Build (`cargo build --release`)
1. Set up two chains (for example gaiad and band testnets)
1. Add keys to hermes (`gm hermes keys`)

Run create channel (`"/opt/bin/hermes_3da7dc12c060271176c8a8870aa399a8eea64820_secp" create channel ibc-0 ibc-8 --port-a transfer --port-b transfer`)

Result (see the warning messages with "received corrupt message" and the final error message):
```
Jul 06 15:47:26.211  INFO using default configuration from '/root/.hermes/config.toml'
Jul 06 15:47:26.228  WARN Hermes health check failed for service tendermint::ServiceClient on the gRPC interface of chain ibc-0:https://localhost:27002/; caused by: transport error: error trying to connect: received corrupt message
Jul 06 15:47:26.229  WARN some Hermes features may not work in this mode!
Jul 06 15:47:26.230 TRACE subscribing to query: tm.event = 'Tx' chain.id=ibc-0
Jul 06 15:47:26.231 TRACE subscribing to query: tm.event = 'NewBlock' chain.id=ibc-0
Jul 06 15:47:26.232 TRACE subscribed to all queries chain.id=ibc-0
Jul 06 15:47:26.233 DEBUG starting event monitor chain.id=ibc-0
Jul 06 15:47:26.237  WARN Hermes health check failed for service tendermint::ServiceClient on the gRPC interface of chain ibc-8:https://localhost:27082/; caused by: transport error: error trying to connect: received corrupt message
Jul 06 15:47:26.237  WARN some Hermes features may not work in this mode!
Jul 06 15:47:26.239 TRACE subscribing to query: tm.event = 'Tx' chain.id=ibc-8
Jul 06 15:47:26.239 TRACE subscribing to query: tm.event = 'NewBlock' chain.id=ibc-8
Jul 06 15:47:26.240 TRACE subscribed to all queries chain.id=ibc-8
[...]

PR #1339, opened 2021-09-09
Update to tendermint-rs `master`

romac:
Closes: #1155

This is a prerequisite for #1321.

- [x] We might be able to re-introduce the `PartialEq` and `Eq` bounds on the various error types where they've been removed once https://github.com/informalsystems/tendermint-rs/pull/970 is merged.
- [x] Enable support for Secp256k1 consensus signatures (#1155)


adizere (approved):
LGTM!

Issue #1290, opened 2021-08-16
Workers should retry indefinitely


greg-szabo:
## Crate
hermes

## Summary of Bug
sentry node restart took too long and it killed hermes worker:
```
Aug 16 13:34:34 hermes hermes[25520]: Aug 16 13:34:34.641 ERROR [cosmoshub-4:transfer/channel-141 -> osmosis-1] worker: handling command encountered error: failed during query to chain id osmosis-1 with underlying error: Light client error for RPC address http://10.10.51.151:26657/: node at http://10.10.51.151:26657/ running chain osmosis-1 not caught up path=packet::channel-141/transfer:cosmoshub-4->osmosis-1
Aug 16 13:34:34 hermes hermes[25520]: Aug 16 13:34:34.642 ERROR [core-1:transfer/channel-6 -> osmosis-1] worker: handling command encountered error: failed during query to chain id osmosis-1 with underlying error: Light client error for RPC address http://10.10.51.151:26657/: node at http://10.10.51.151:26657/ running chain osmosis-1 not caught up path=packet::channel-6/transfer:core-1->osmosis-1
Aug 16 13:34:34 hermes hermes[25520]: Aug 16 13:34:34.876 ERROR [cosmoshub-4:transfer/channel-141 -> osmosis-1] worker: handling command encountered error: failed during query to chain id osmosis-1 with underlying error: Light client error for RPC address http://10.10.51.151:26657/: node at http://10.10.51.151:26657/ running chain osmosis-1 not caught up path=packet::channel-141/transfer:cosmoshub-4->osmosis-1
Aug 16 13:34:34 hermes hermes[25520]: Aug 16 13:34:34.876 ERROR [packet::channel-141/transfer:cosmoshub-4->osmosis-1#103] worker aborted with error: Packet worker failed after 7 retries
```

## Version
0.6.1

## Acceptance Criteria
Workers should retry infinitely

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
Looking at logs from gravity dex/emeris, the same problem shows up there. We'll on this ASAP, thanks for reporting Greg!

romac:
As discussed during today's call, there are multiple (sometimes orthogonal) solutions for this:

a) Retry indefinitely
[...]

PR #1340, opened 2021-09-10
Modified packet worker to use stubborn strategy

adizere:
Closes: #1290 

## Description

The following describes worker behavior after this PR:

- packet worker:
    - old behavior: used a default retry strategy that retried for 6 iterations and then gave up;
      - the individual retry backoff varied between 200ms and 500ms
      - **the cumulative backoff time was 2 seconds**
    - new behavior: using a stubborn strategy, which retries for infinite iterations and never gives up
      - the individual backoff time (between subsequent retries) is 1sec in the beginning, and increases steadily by 10ms at every step

~~The rationale for this choice is as follows. I expect past a certain duration (I choose in this PR 6 hours), Hermes might as well give up, as there is little chance that the connection can become healthy again without operator intervention, but not sure if my intuition is appropriate.~~

---

Unchanged behavior:

- client worker
    - this worker is in charge of regularly updating clients (unless there is already channel traffic, which will trigger client updates), at a frequency of 2/3 of trusting period
    - **this worker is already stubborn**: never quits unless the supervisor instructs it to do so, or if the client is expired or frozen

- channel and connection workers
    - these are in charge of finishing up initialized (but incomplete) connection handshakes or initialized channel handshakes
    - **these workers don't need to retry stubbornly**, because their activity is time-sensitive

______

For contributor use:

### DIFF
--- a/relayer/src/worker/retry_strategy.rs
+++ b/relayer/src/worker/retry_strategy.rs
@@ -18,17 +18,17 @@ pub fn worker_default_strategy() -> impl Iterator<Item = Duration> {
 
 /// A stubborn worker retry strategy.
 ///
-/// Initial retry delay is hardcoded to 500ms, and
-/// the delay grows by 500ms at every step. The
-/// strategy delay is capped at 20 sec. The
-/// total number of retries is capped to 200.
+/// Initial retry delay is hardcoded to 1s, and
+/// the delay grows by 1s at every step. The
+/// strategy delay is capped at 60 sec. The
+/// total number of retries is capped to 400.
 /// Therefore the overall amount of time spent
-/// retrying will be approx. 1 hour.
+/// retrying will be approx. 6 hours.
 ///
 /// See the `stubbord_strategy` test below.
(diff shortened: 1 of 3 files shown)


relayer/src/worker/retry_strategy.rs
```diff
@@ -1,12 +1,93 @@
...
+
+
+#[cfg(test)]
+mod tests {
+    use std::time::Duration;
+
+    use crate::worker::retry_strategy::{worker_stubborn_strategy, worker_default_strategy};
+
+    #[test]
+    fn default_strategy() {
```
adizere, line 42: This test is almost identical to this test
https://github.com/informalsystems/ibc-rs/blob/85446586daf6dba652271eef31dae29f35da86c2/relayer/src/util/retry.rs#L153

but I only realized after writing it.

I still found to keep it, for the purpose of understanding exactly the details of the strategy we use in practice.

hu55a1n1 (approved):
👌💯

Issue #1264, opened 2021-08-06
Sequence number caching in Hermes is unreliable


gamarin2:
## Crate

Hermes

## Summary

Query the state for every chain in config every X sec for the relayer account's seq number, and update cached seq number with it. Make X a configurable param. 

## Problem Definition

At some point (around height 1008 on Cosmos hub), for some reason, the cached seq number on our relayer instance got incremented, but the actual on-chain seq number wasn't. From there, our relayer hasn't been able to relay, producing the following error:

```
Aug 06 15:51:41.681  INFO [irishub-1:transfer/channel-12 -> cosmoshub-4] result events:
    ChainErrorEv(deliver_tx on chain cosmoshub-4 for Tx hash EC50AD139769F2585F351CDE69636B0C82F522A5B29ADDD1C279737088F4E098 reports error: code=Err(32), log=Log("account sequence mismatch, expected 1008, got 1358: incorrect account sequence"))
```

While we don't know what caused this mismatch to happen, we couldn't detect it and thought our relayer was running properly. It would be good to figure out why the mismatch happened, but it would also be good to have a failsafe that updates the cached seq number with the actual on-chain seq number periodically so that the relayer can recover on its own. 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

andynog:
I've done a preliminary investigation. Here's some thoughts on what might be happening

The sequence is incremented at the of send_tx here https://github.com/informalsystems/ibc-rs/blob/142c7dc29c7636d1c9ece51ce8fd6f7c50ca3477/relayer/src/chain/cosmos.rs#L384

then the response from the `broadcast_tx_sync` is returned from `send_tx` but there are no checks or validation on the response at this stage, so I'm assuming if the response has an "unsuccessful" or "unexpected" response the internal sequence increments anyways so this might potentially cause some situations that the account sequence will mismatch because the internal sequence is ahead of the sequence on chain.

I believe the sequence is only retrieved once and cached here https://github.com/informalsystems/ibc-rs/blob/142c7dc29c7636d1c9ece51ce8fd6f7c50ca3477/relayer/src/chain/cosmos.rs#L521

An account is returned in the logic below and the sequence is retrieved from this account 
[...]

PR #1349, opened 2021-09-14
Retry send_tx on account sequence mismatch

soareschen:
Closes: #1264

## Description

This PR makes Hermes retry on `send_tx` for up to 5 times, in the event that the submission failed with a `incorrect account sequence` error. This error can occur either at the gas estimation step (`send_tx_simulate`), or at the transaction submission step (`broadcast_tx_sync`). In both cases, we parse the error message to match on its details and infer if it is account sequence mismatch.

- For the case when the error appears at the gas estimation step, we match on the "account sequence mismatch" status message. We do not retry in the runtime, but rather at the worker level (`step` function).

- For the case when the account sequence mismatch occurs at transaction submission step, we detect the error by checking for the error code 32 defined in [Cosmos SDK](https://github.com/cosmos/cosmos-sdk/blob/cb5bfb7678985142b79150c03ca99694fe3ff46c/types/errors/errors.go#L115-L117), and then perform the retry accordingly directly in the runtime.

______

For contributor use:


relayer/src/chain/cosmos.rs
```diff
@@ -373,13 +379,65 @@ impl CosmosSdkChain {
...
+            Code::Ok => {
+                self.incr_account_sequence();
+                Ok(response)
+            }
+            // Cosmos SDK defines an account sequence mismatch with the error code 32
+            Code::Err(32) => {
+                if retries < MAX_ACCOUNT_SEQUENCE_RETRY {
+                    warn!("send_tx failed with incorrect account sequence. retrying with account sequence refetched from the chain.");
+
+                    // Refetch account information from the chain.
```
adizere, line 412: I appreciate the thorough comments! Also, the solution of refreshing only on error is simpler than what I had in mind (parsing error message) and probably more future-proof. Cool work Soares!

I have two concerns:

- were you able to reproduce the error, and could you describe the scenario? Is it as simple as using two instances of hermes, a `start` and an `ft-transfer` concurrently, or did you stumble upon more advanced scenarios that we should be aware while testing?
- seems like a tail-recursive method call, but the max depth is `5` so not too big. My shallow understanding of this is that it shouldn't pose problems from a memory-use perspective, so I just wanted to run it by you.

soareschen:
> were you able to reproduce the error, and could you describe the scenario? Is it as simple as using two instances of hermes, a start and an ft-transfer concurrently, or did you stumble upon more advanced scenarios that we should be aware while testing?

Yes, you can test it out with a two raw tx transfers using the same wallet as the relayer. In the log you should see the relayer refetch the account sequence when the first attempt failed.

```
Sep 15 11:50:46.655 DEBUG send_messages_and_wait_check_tx with 3 messages
Sep 15 11:50:46.655 DEBUG [ibc-0] send_tx: sending 3 messages using nonce 12
Sep 15 11:50:46.670 DEBUG [ibc-0] send_tx: broadcast_tx_sync: Response { code: Err(32), data: Data([]), log: Log("account sequence mismatch, expected 13, got 12: incorrect account sequence"), hash: transaction::Hash(8AD74142F7A0EF604FBFE91291F212DE4353CC88B21BE2130210556CA4A58FDA) }
Sep 15 11:50:46.670  WARN send_tx failed with incorrect account sequence. retrying with account sequence refetched from the chain.
Sep 15 11:50:46.674 DEBUG [ibc-0] send_tx: retrieved account sequence=13 number=0
Sep 15 11:50:46.674 DEBUG [ibc-0] send_tx: sending 3 messages using nonce 13
Sep 15 11:50:46.690 DEBUG [ibc-0] send_tx: broadcast_tx_sync: Response { code: Ok, data: Data([]), log: Log("[]"), hash: transaction::Hash(AD0F9E06FDCED226E19FD78A60F4AD2BD3AD59C9FB95128D7D6801B275578EF7) }
```

relayer/src/chain/cosmos.rs
soareschen: > seems like a tail-recursive method call, but the max depth is 5 so not too big.

It shouldn't be an issue, and usually the retry should succeed after one try anyway. The only time when the max retry would exceed is if the relayer is configured with a shared wallet that is being used concurrently all the time, which in that case is considered a misconfiguration.

relayer/src/chain/cosmos.rs
```diff
@@ -78,12 +78,17 @@ use crate::keyring::{KeyEntry, KeyRing, Store};
...
 use crate::{chain::QueryResponse, event::monitor::TxMonitorCmd};
 
 use super::ChainEndpoint;
 
 mod compatibility;
 
+// Maximum number of retries for send_tx in the case of
+// an account sequence mismatch.
+const MAX_ACCOUNT_SEQUENCE_RETRY: u32 = 5;
+
```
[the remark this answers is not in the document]
soareschen: Done

relayer/src/chain/cosmos.rs
```diff
@@ -373,13 +379,65 @@ impl CosmosSdkChain {
...
         debug!("[{}] send_tx: broadcast_tx_sync: {:?}", self.id(), response);
 
-        self.incr_account_sequence()?;
+        match response.code {
+            Code::Ok => {
+                self.incr_account_sequence();
+                Ok(response)
+            }
+            // Cosmos SDK defines an account sequence mismatch with the error code 32
+            Code::Err(32) => {
```
romac, line 408: Can we extract this "magic number" as a constant?

soareschen: > Can we extract this "magic number" as a constant?

Fixed

relayer/src/chain/cosmos.rs
```diff
@@ -546,9 +613,13 @@ impl CosmosSdkChain {
...
-    fn incr_account_sequence(&mut self) -> Result<(), Error> {
-        self.account()?.sequence += 1;
-        Ok(())
+    fn incr_account_sequence(&mut self) {
+        match &mut self.account {
+            Some(account) => {
+                account.sequence += 1;
+            }
+            None => {}
+        }
```
romac, line 622: I think this could be written slightly more simply as:

```suggestion
        if let Some(account) = &mut self.account {
                account.sequence += 1;
        }
```

relayer/src/chain/cosmos.rs
```diff
@@ -373,13 +383,64 @@ impl CosmosSdkChain {
...
 
         debug!("[{}] send_tx: broadcast_tx_sync: {:?}", self.id(), response);
 
-        self.incr_account_sequence()?;
+        match response.code {
+            Code::Ok => {
+                self.incr_account_sequence();
+                Ok(response)
+            }
+            Code::Err(code) if code == INCORRECT_ACCOUNT_SEQUENCE_ERR => {
```
romac, line 411: You can use constants in pattern matches:

```suggestion
            Code::Err(INCORRECT_ACCOUNT_SEQUENCE_ERR) => {
```

soareschen: I tried before and it didn't work. Rust does not allow pattern matching with constants.

romac: Really? It works for me, at least in some capacity, see: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2bfcd350cf71b52e4954a8c127a1784d

romac: Just tried in the actual code and it works as well :)

soareschen: Ok. Probably it was because I mistyped the constant as `u64` instead of `u32` before, and then got mistaken of the actual error message when I tried to do inline constant pattern matching. Fixed it now.

adizere:
FYI -- 
there is a slightly more complex scenario that occurs in production, which is very painful for the ops team to deal with. I detailed a way to [reproduce the problem easily](https://github.com/informalsystems/ibc-rs/issues/1264#issuecomment-923095155). I think the fix in this PR should be informed by the problem documented there, and at the very least we should adjust Hermes logs to describe the issue more accurately. Concretely, it would be ideal if we follow one or more of these ideas:

- Hermes reports (e.g., in the log) that the full node has a certain number of tx stuck in the mempool, which may confuse the relayer
    - unclear if Hermes should retry upon encountering the "account sequence mismatch", but probably is a good idea to keep the retry mechanism
    - to find if the mempool has pending tx-es we have access to:
        - https://docs.tendermint.com/master/rpc/#/Info/unconfirmed_txs
        - https://docs.tendermint.com/master/rpc/#/Info/num_unconfirmed_txs
- Hermes reports that the full node has misconfigured min-gas-prices
- Hermes team follows-up on https://github.com/tendermint/tendermint/issues/6958 to improve dealing with this problem with a long-term principled fix

adizere:
I tried to summarize the status of the issue in here https://github.com/informalsystems/ibc-rs/issues/1397.

The solution in the present PR is incomplete, and we'll deprioritize this until we investigate further on the tm-go side. In the meantim, let's close this PR. We'll reuse some of the code later if it proves useful.

karzak:
@adizere - thanks for your work on this! 

We're running this branch on testnet and it has been avoiding the account sequence deadlock that we were seeing on v0.9.0 so far. 

On `v0.9.0` whenever we got into this state:

```
2021-12-17T18:55:38.202492Z ERROR ThreadId(20) [loan-wars] estimate_gas: failed to simulate tx. propagating error to caller: gRPC call failed 2021-12-17T18:55:38.202492Z ERROR ThreadId(20) [loan-wars] estimate_gas: failed to simulate tx. propagating error to caller: gRPC call failed with status: status: Unknown, message: "account sequence mismatch, expected 10190, got 10191: incorrect account sequence", details: [], metadata: MetadataMap { headers: {"date": "Fri, 17 Dec 2021 18:55:38 GMT", "content-type": "application/grpc", "server": "nginx/1.18.0 (Ubuntu)", "x-cosmos-block-height": "31204"} }
2021-12-17T18:55:38.202539Z  WARN ThreadId(20) send_tx failed at estimate_gas step mismatching account sequence: dropping the tx & refreshing account sequence number
2021-12-17T18:55:38.215376Z ERROR ThreadId(26) [loan-wars-ibc:transfer/channel-0 -> loan-wars] worker: schedule execution encountered error: failed with underlying error: gRPC call failed with status: status: Unknown, message: "account sequence mismatch, expected 10190, got 10191: incorrect account sequence", details: [], metadata: MetadataMap { headers: {"date": "Fri, 17 Dec 2021 18:55:38 GMT", "content-type": "application/grpc", "server": "nginx/1.18.0 (Ubuntu)", "x-cosmos-block-height": "31204"} } status: status: Unknown, message: "account sequence mismatch, expected 10190, got 10191: incorrect account sequence", details: [], metadata: MetadataMap { headers: {"date": "Fri, 17 Dec 2021 18:55:38 GMT", "content-type": "application/grpc", "server": "nginx/1.18.0 (Ubuntu)", "x-cosmos-block-height": "31204"} }
```

It was un-resolvable without restarting the `hermes` process. Running this branch, I've seen what looks like graceful recovery from this state. Will keep monitoring and update if that changes.

adizere (approved):
Got positive feedback from 2 relayer operators, still waiting on 1 of them. Specifically, the feedback is that this PR seems to alleviate the problem of needing to restart Hermes when acct. seq. mismatch occurs. I will merge this within the day, unless a relayer operator highlights any issue.

PR #1355, opened 2021-09-15
Update `gm` to work with chains using two different binaries for daemon and CLI (eg. gaiad/gaiacli)

romac:
Closes: #XXX

## Description

______

For contributor use:


adizere:
I'm curious which chain build version splits the binary in two?

romac:
Perhaps this is badly worded or maybe I understand this wrong. From what I can tell, there's pretty much always been two binaries (going back to at least v0.14), but some commands have recently (I think in v0.43) been moved from `[gaia]d` to `[gaia]cli` and vice-versa which means that one now needs to call into the two executables to spawn and setup a chain.

adizere:
Can we close this? It doesn't seem like a problem.

`gm` is getting some maintenance work in #1413 and there's no mention in that PR of split gaia binaries.

PR #1364, opened 2021-09-19
Enable global use of telemetrics

romac:
## Description

Enable the use of telemetrics globally across the whole `ibc-relayer` codebase without having to thread a `Arc<TelemetryState>` everywhere.

To this end, this PR introduces a global (`static`) `Arc<TelemetryState>` which can be accessed with `ibc_telemetry::global()` from anywhere in the `ibc-relayer` crate.

Updating a metric now looks like this:

```rust
telemetry!(ibc_telemetry::global().update_some_metric(foo, bar, 42));
```

---

Note: Internal change: no need for issue + changelog

### DIFF
--- a/relayer-cli/src/commands/start.rs
+++ b/relayer-cli/src/commands/start.rs
@@ -11,7 +11,6 @@ use ibc_relayer::config::reload::ConfigReload;
 use ibc_relayer::config::Config;
 use ibc_relayer::rest;
 use ibc_relayer::supervisor::{cmd::SupervisorCmd, Supervisor};
-use ibc_relayer::telemetry::Telemetry;
 
 use crate::conclude::json;
 use crate::conclude::Output;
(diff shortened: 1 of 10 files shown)


romac:
We could now potentially improve the macro a bit if people think it's useful, eg. to be able to write instead:

```rust
telemetry!(update_some_metric, foo, bar, 42);
```

What do you think @adizere @soareschen?

adizere:
> We could now potentially improve the macro a bit if people think it's useful, eg. to be able to write instead:
> 
> ```rust
> telemetry!(update_some_metric, foo, bar, 42);
> ```
> 
> What do you think @adizere @soareschen?

This sounds like a nice idea, but not a priority, so I guess if it's a low hanging fruit let's do it!

soareschen:
> We could now potentially improve the macro a bit if people think it's useful, eg. to be able to write instead:
> 
> ```rust
> telemetry!(update_some_metric, foo, bar, 42);
> ```

I don't see a need of defining `telemetry!` as a macro. We can just define a `telemetry` function whose implementation is empty when the telemetry feature is disable. Rust should be smart enough to optimize that function call away when it is empty. So I think it could instead be written as:

```rust
telemetry("update_some_metric", "foo", "bar", 42);
```

thanethomson:
With this function-based option:

```rust
telemetry("update_some_metric", "foo", "bar", 42);
```

How does one cater for variable numbers of arguments? Since different metrics would potentially need different numbers of arguments.

adizere (approved):
Awesome work!

romac:
After discussing with @adizere and @soareschen, we will keep the macro-based solution (for now at least).

soareschen:
> How does one cater for variable numbers of arguments? Since different metrics would potentially need different numbers/types of arguments.

Right, I think an alternative is to use trait method like:

```rust
telemetry().update_some_metric("foo", "bar", 42);
```

But it's minor so let's leave that for now.

PR #1378, opened 2021-09-22
Support for toggling tx confirmation mechanism

adizere:
Closes: #1380

## Description

______

For contributor use:

### DIFF
--- a/relayer-cli/src/commands/tx/packet.rs
+++ b/relayer-cli/src/commands/tx/packet.rs
@@ -37,7 +37,7 @@ impl Runnable for TxRawPacketRecvCmd {
             src_port_id: self.src_port_id.clone(),
             src_channel_id: self.src_channel_id.clone(),
         };
-        let mut link = match Link::new_from_opts(chains.src, chains.dst, opts) {
+        let mut link = match Link::new_from_opts(chains.src, chains.dst, opts, false) {
             Ok(link) => link,
             Err(e) => return Output::error(format!("{}", e)).exit(),
         };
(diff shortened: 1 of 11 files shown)


relayer-cli/src/commands/tx/packet.rs
```diff
@@ -37,7 +37,7 @@ impl Runnable for TxRawPacketRecvCmd {
             src_port_id: self.src_port_id.clone(),
             src_channel_id: self.src_channel_id.clone(),
         };
-        let mut link = match Link::new_from_opts(chains.src, chains.dst, opts) {
+        let mut link = match Link::new_from_opts(chains.src, chains.dst, opts, false) {
```
ancazamfir, line 40: why not use the config setting?

ancazamfir (commented):
after a quick look, I think that if we hit the periodic clear packet with pending transactions we will submit those again. Maybe @mircea-c could do some eval, maybe in the end it is a small price to pay.

relayer-cli/src/commands/tx/packet.rs
adizere: This is the CLI for doing `tx raw packet-recv`. This means under the hood we're using the `SyncSender` which synchronously waits to confirm transactions. We are not using the `AsyncSender` here, so not sure having tx confirmations would be appropriate.

ancazamfir: right! missed that one. I hope we move all CLIs and workers to async at some point.

relayer/src/worker/client.rs
```diff
@@ -51,30 +51,37 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> ClientWorker<ChainA, ChainB> {
...
-                        ibc_client_update,
-                        &self.client.dst_chain_id,
-                        &self.client.dst_client_id,
-                        1
-                    );
-                }
-                Err(e) => {
-                    if let ForeignClientErrorDetail::ExpiredOrFrozen(_) = e.detail() {
-                        warn!("failed to refresh client '{}': {}", client, e);
+            if last_refresh.elapsed() > Duration::from_secs(60) {
```
soareschen, line 60: Maybe add a comment that explain that we do not need to refresh too frequently.

soareschen (approved):

PR #1387, opened 2021-09-24
Fix compilation of `proto-compiler`

romac:


adizere (approved):
Top, thanks for the fixes!

Issue #1359, opened 2021-09-16
Tracking issue for no-std compliance for dependencies in ibc-rs and tendermint-rs


soareschen:
This issue is part of #1158 to track the no-std support in dependencies that are used by ibc-rs and tendermint-rs. To fully support ibc-rs to be used in no-std environment like Substrate, we must also fix the dependencies by either fixing it upstream, or remove the dependencies from ibc-rs and tendermint-rs and replace them with alternatives that support no_std.

The checks are done by running [no-std-check](https://github.com/informalsystems/ibc-rs/tree/master/ci/no-std-check) with only the specific unsupported dependency activated. Currently the checks are done from the branch at #1358.

- [x] `ics23` - Upstream fix yet to be merged in https://github.com/informalsystems/ics23/tree/soares/no-std.
- [ ] `safe-regex` - Upstream fix yet to be merged in https://github.com/informalsystems/safe-regex.
- [x] `crossbeam-channel` - Requires `std` and is currently used by `tendermint-light-client`.
  - Removed
- [x] `tonic`
  - Fails in all checks. We will have to remove the dependency to `tonic` from ibc-proto and tendermint-proto
  - Removed in #1439
- [x] `socket2`
  - Fails in all checks. It uses `std` and does not support Wasm targets. 
- [x] `getrandom`
  - Fails in `check-wasm` and `check-substrate`.
- [x] `ics23`
  - Fix is merged in https://github.com/confio/ics23/pull/41 but the changes are still unreleased.
- [x] `thiserror`
  - Fails in all checks. We should have already removed all use of `thiserror` with `flex-error`, but still needs to double check.
  - Removed in #1390.
- [x] `regex`
  - Fails in `check-wasm` and `check-substrate`.
  - Use `safe-regex` as replacement in #1395
- [x] `sled`
  - Fails in all checks. We may have to remove dependency to `sled`.
- [x] `tokio`
  - Fails in all checks.
- [x] `toml`
  - Fails in all checks.
  - Dependency removed in informalsystems/tendermint-rs#986
- [x] `url` 
  - Fails in all checks.
  - Pending PR: servo/rust-url/pull/717
  - Dependency removed in informalsystems/tendermint-rs#986
____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

[...]

PR #1395, opened 2021-09-28
Use safe-regex instead of regex for regular expression

soareschen:
Part of #1359

## Description

Since the `regex` crate cannot work without `std`, we can use [`safe-regex`](https://docs.rs/safe-regex/0.2.4/safe_regex/) as an no_std alternative for regular expressions.

______

For contributor use:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1291,7 +1291,7 @@ dependencies = [
  "modelator",
  "prost",
  "prost-types",
- "regex",
+ "safe-regex",
  "serde",
  "serde_derive",
  "serde_json",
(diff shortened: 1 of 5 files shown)


adizere:
Thanks Soares! Looks good, let's just check on the failing tests otherwise this should be an easy one.

```
test src/ics24_host/identifier.rs - ics24_host::identifier::ChainId::is_epoch_format (line 79) ... FAILED
475
test src/ics24_host/identifier.rs - ics24_host::identifier::ChainId::chain_version (line 56) ... FAILED
```

soareschen:
The tests are fixed now.

romac (approved):

Issue #1354, opened 2021-09-15
Hermes always uses `max_gas` for transactions


Galadrin:
## Crate

Hermes ibc_relayer v0.7.1

## Summary of Bug

Hermes transaction always use max_gas amount.

https://www.mintscan.io/osmosis/txs/E742F9EFE2E97A5FCD71FDB3CCDF2FF02BEE47DE5B2D0A51110805736F67C0E4

## Version

v0.7.1

## Acceptance Criteria

Hermes use automatic gas adjustment and `gas_adjustment` from config

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
Can you explain in a bit more detail the problem?

Was Hermes configured with `max_gas` 60,000,000 (or thereabout)? It would be good if you can share the gas-related configuration details from config.toml for chain Osmosis.

Note that Hermes will by default try to simulate each tx, thus obtaining an approximate amount of gas it should use, and thenit will subsequently submit the transaction with that approximated amount (the amount is adjusted up using `gas_adjustment`). If the tx simulation fails, then Hermes will simply use the configured `max_gas`.

https://github.com/informalsystems/ibc-rs/blob/5d38857ac41affd93573c100fedf333b979dfcf3/relayer/src/chain/cosmos.rs#L335-L343

Galadrin:
Hi,
my config is the same as issue 1342

https://github.com/informalsystems/ibc-rs/issues/1342#issuecomment-918590572

for Osmosis, Crypto.org and Cosmos, the max_gas value is used for every transaction.
the transaction shared is an example.

hu55a1n1:
Hi @Galadrin! It appears that you have configured `gas_adjustment=5` for the `osmosis-1` chain in your [config.toml](https://pastebin.com/Pbca30tz). And that probably caused `hermes` to use the `max_gas` based on this calculation ->
https://github.com/informalsystems/ibc-rs/blob/5d38857ac41affd93573c100fedf333b979dfcf3/relayer/src/chain/cosmos.rs#L405-L410

Can you try a `gas_adjustment < 1` i.e. `0.5`?

Galadrin:
I have the same symptom on Crypto.org and Cosmos chain.
and the issue is not present using v0.7.0

Galadrin:
[...]

PR #948, opened 2021-05-15
Upgrade to SDK v0.43+

ancazamfir:
Closes: #947
Closes: #918

## Description

The proto files in this branch are currently aligned with:

- ibc-go v1.0.0-alpha2: [d70f49c8f612d60f1b7e2f1d1f160f28988962e1](https://github.com/cosmos/ibc-go/commit/d70f49c8f612d60f1b7e2f1d1f160f28988962e1)
- cosmos-sdk v0.43.0-beta1: [8cfa2c26738276d895caf9eb98b3f70616218e17](https://github.com/cosmos/cosmos-sdk/commit/8cfa2c26738276d895caf9eb98b3f70616218e17)

______

For contributor use:

- [x] Updated the __Unreleased__ section of [CHANGELOG.md](https://github.com/informalsystems/ibc-rs/blob/master/CHANGELOG.md) with the issue.
- [ ] If applicable: Unit tests written, added test to CI.
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Updated relevant documentation (`docs/`) and code comments.
- [x] Re-reviewed `Files changed` in the Github PR explorer.

PR #1405, opened 2021-09-30
Fix tx simulation for chains based on Cosmos SDK 0.42.x

romac:
Closes: #1354

## Description

In Cosmos SDK 0.43+, the Protocol Buffers message for `SimulateRequest` was updated, now included a `tx_bytes` field and deprecating the `tx` field:

```diff
pub struct SimulateRequest {
     /// tx is the transaction to simulate.
+    /// Deprecated. Send raw tx bytes instead.
+    #[deprecated]
     #[prost(message, optional, tag = "1")]
     pub tx: ::core::option::Option<Tx>,
+    /// tx_bytes is the raw transaction.
+    #[prost(bytes = "vec", tag = "2")]
+    pub tx_bytes: ::prost::alloc::vec::Vec<u8>,
 }
```

https://github.com/informalsystems/ibc-rs/commit/c4b6e3035c3b717f9f8ca09ec97c15beaf3d7cd9#diff-9ea64978cd0695e7d1321fc30fabcaa46a5857c514b638a22f6bad5f9ca25870R240-R248

As part of #948, we have changed the logic for building the `SimulateRequest` message accordingly, but did not realize that chains based on Cosmos SDK <0.43 would now reject the message with `invalid empty tx: invalid request`.

This would cause the gRPC call to fail and Hermes to then use the `max_gas` specified in the config to compute the fee for sending the actual transaction.

This PR fixes this by including both the `tx` and the `tx_bytes` fields so that chains requiring either field can process the request:

```rust
        let req = SimulateRequest {
             tx: Some(tx), // needed for simulation to go through with Cosmos SDK <  0.43
             tx_bytes,     // needed for simulation to go through with Cosmos SDk >= 0.43
         };
```

https://github.com/informalsystems/ibc-rs/pull/1405/files#diff-c12c5f52a7e35d9e3d8220ad3701dbe0f33dc1c21c7bf44518c44a5864f46907R406-R409

We may have to revisit this if the deprecated `tx` field is eventually removed, but this is outside the scope of this PR.

______

For contributor use:

### DIFF
--- a/relayer/src/chain/cosmos.rs
+++ b/relayer/src/chain/cosmos.rs
@@ -6,7 +6,7 @@ use core::{
     str::FromStr,
     time::Duration,
 };
-use std::{thread, time::Instant};
+use std::{fmt, thread, time::Instant};
 
 use bech32::{ToBase32, Variant};
 use bitcoin::hashes::hex::ToHex;
@@ -24,7 +24,7 @@ use tendermint_rpc::query::{EventType, Query};
 use tendermint_rpc::{endpoint::broadcast::tx_sync::Response, Client, HttpClient, Order};
 use tokio::runtime::Runtime as TokioRuntime;
 use tonic::codegen::http::Uri;
-use tracing::{debug, trace, warn};
+use tracing::{debug, error, trace, warn};
 
 use ibc::downcast;
 use ibc::events::{from_tx_response_event, IbcEvent};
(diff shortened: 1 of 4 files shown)


Galadrin:
tested in production environment

https://www.mintscan.io/osmosis/txs/65E1A765970312E10A35F3DAE69A26B9F7BE8BB60455BA5A479D5510CE5F1344
https://www.mintscan.io/cosmos/txs/DB46FF1B3A85A2A6739C95AF376CD2F3516F1696A3526C232B0E766FC8F9DF6F

romac:
Seems like the gas was estimated correctly there, taking into account the gas adjustment parameter.

So, would you say this fixed the issue you were seeing? If so, we can release this fix as part of Hermes 0.7.3 in the next couple days.

Sdpierret:
I will keep it running during the next 24h

Galadrin:
I will keepit running during the next 24h

soareschen (approved):

mircea-c:
Just started running this in production (kek). Looks to be working as advertised. All our fees are now lower.

ancazamfir:
@romac could you describe what the issue was exactly? Not clear to me after a first look at the diffs. 
Was the regression introduced  by #948?

relayer/src/chain/cosmos.rs
```diff
@@ -379,12 +398,15 @@ impl CosmosSdkChain {
...
+        // The `tx` field of `SimulateRequest` was deprecated in Cosmos SDK 0.43 in favor of `tx_bytes`.
         let mut tx_bytes = vec![];
         prost::Message::encode(&tx, &mut tx_bytes).unwrap();
+
         #[allow(deprecated)]
-        let req = SimulateRequest { tx: None, tx_bytes };
+        let req = SimulateRequest {
+            tx: Some(tx), // needed for simulation to go through with Cosmos SDK <  0.43
+            tx_bytes,     // needed for simulation to go through with Cosmos SDk >= 0.43
+        };
```
ancazamfir, line 409: ah, I guess this was the problem. Were you able to test the fix with both `< 0.43` and `>= 0.43`?

romac:
> @romac could you describe what the issue was exactly? Not clear to me after a first look at the diffs.
> Was the regression introduced by #948?

Just updated the PR description.

relayer/src/chain/cosmos.rs
romac: Yes, I tested with both Gaia 5.0.5 and Gaia 6.0.0-rc1

Issue #1402, opened 2021-09-30
Incorrect account sequence logic upon gas errors


adizere:
## Summary of Bug

When Hermes encounters an error submitting a transaction, it mistakenly increments the (cached) account sequence number.

See the trace below, provided from Gravity DeX ops:

```
Sep 30 13:53:44 hznr1 hermes[1800189]: Sep 30 13:53:44.780 DEBUG ThreadId(69) [impacthub-3] send_tx: retrieved account sequence=189 number=214
Sep 30 13:53:44 hznr1 hermes[1800189]: Sep 30 13:53:44.780 DEBUG ThreadId(69) [impacthub-3] send_tx: sending 11 messages using nonce 189
Sep 30 13:53:44 hznr1 hermes[1800189]: Sep 30 13:53:44.795 DEBUG ThreadId(69) [impacthub-3] send_tx: broadcast_tx_sync: Response { code: Err(11), data: Data([]), log: Log("out of gas in location: txSize; gasWanted: 500000, gasUsed: 597058: out of gas"), hash: transaction::Hash(9A3FB110201A811D8B4E469B19E03935040EFE97181E6EA1CA5C115B2C9DB085) }
Sep 30 13:53:46 hznr1 hermes[1800189]: Sep 30 13:53:46.379 DEBUG ThreadId(69) send_messages_and_wait_check_tx with 2 messages
Sep 30 13:53:46 hznr1 hermes[1800189]: Sep 30 13:53:46.379 DEBUG ThreadId(69) [impacthub-3] send_tx: sending 2 messages using nonce 190
Sep 30 13:53:46 hznr1 hermes[1800189]: Sep 30 13:53:46.388 DEBUG ThreadId(69) [impacthub-3] send_tx: broadcast_tx_sync: Response { code: Err(32), data: Data([]), log: Log("account sequence mismatch, expected 189, got 190: incorrect account sequence"), hash: transaction::Hash(ED1C62FCEF2DFDBDE5B86B2AE820441D6028AD164D9CE9760799C4C715B1BCD7) }
Sep 30 13:53:48 hznr1 hermes[1800189]: Sep 30 13:53:48.573 DEBUG ThreadId(69) send_messages_and_wait_check_tx with 2 messages
Sep 30 13:53:48 hznr1 hermes[1800189]: Sep 30 13:53:48.573 DEBUG ThreadId(69) [impacthub-3] send_tx: sending 2 messages using nonce 191
[...]

Issue #1404, opened 2021-09-30
Packet clearing function does not actually clear packets


mircea-c:
## Crate

Hermes

## Summary of Bug

Hermes attempts to clear pending packets, but fails to do so repeatedly:
```
Sep 30 12:03:01.645  INFO ThreadId(3549) [osmosis-1:transfer/channel-15 -> iov-mainnet-ibc] clearing pending packets height=Some(Height { revision: 1, height: 1395619 })
 Sep 30 12:03:01.658 DEBUG ThreadId(3549) [osmosis-1:transfer/channel-15 -> iov-mainnet-ibc] packets that still have commitments on osmosis-1: 2936, 2937, 2938, 2939, 2940, 2941, 2942, 2943 (first 10 shown here; total=8)
 Sep 30 12:03:01.659 DEBUG ThreadId(3549) [osmosis-1:transfer/channel-15 -> iov-mainnet-ibc] recv packets to send out to iov-mainnet-ibc of the ones with commitments on source osmosis-1: 2936, 2937, 2938, 2939, 2940, 2941, 2942, 2943 (first 10 shown here; total=8)
 Sep 30 12:03:16.289  INFO ThreadId(3549) [osmosis-1:transfer/channel-15 -> iov-mainnet-ibc] found unprocessed SendPacket events for [] (first 10 shown here; total=0)
 Sep 30 12:03:16.942  INFO ThreadId(3549) [osmosis-1:transfer/channel-15 -> iov-mainnet-ibc] finished scheduling pending packets clearing height=Some(Height { revision: 1, height: 1395619 })
 Sep 30 12:05:31.991  INFO ThreadId(3549) [osmosis-1:transfer/channel-15 -> iov-mainnet-ibc] clearing pending packets height=Some(Height { revision: 1, height: 1395639 })
 Sep 30 12:05:32.164 DEBUG ThreadId(3549) [osmosis-1:transfer/channel-15 -> iov-mainnet-ibc] packets that still have commitments on osmosis-1: 2936, 2937, 2938, 2939, 2940, 2941, 2942, 2943 (first 10 shown here; total=8)
 Sep 30 12:05:32.164 DEBUG ThreadId(3549) [osmosis-1:transfer/channel-15 -> iov-mainnet-ibc] recv packets to send out to iov-mainnet-ibc of the ones with commitments on source osmosis-1: 2936, 2937, 2938, 2939, 2940, 2941, 2942, 2943 (first 10 shown here; total=8)
[...]

PR #1406, opened 2021-10-01
Fixed account seq. nr logic in the error case

adizere:
Closes: #1402

## Description

- [x] fixes the logic for account seq. nr, to only increment it if the call to `broadcast_tx_sync` is successful
- [x] fixes following the discussion from issue #1404 
    - lowers the log-level for a couple of log lines in the packet clearing methods, 
    - avoid printing "found unprocessed" log in the packet clearing methods if no event was found

______

For contributor use:

### DIFF
--- a/relayer/src/chain/cosmos.rs
+++ b/relayer/src/chain/cosmos.rs
@@ -284,8 +284,16 @@ impl CosmosSdkChain {
             }
             tendermint::abci::Code::Err(_) => {
                 // Avoid increasing the account s.n. if CheckTx failed
-                // Simply log the error
+                // Log the error
                 error!("[{}] send_tx: broadcast_tx_sync: {:?}", self.id(), response);
+                // The primary reason (we know of) causing broadcast_tx_sync to fail
+                // is due to "out of gas" errors. These are unrecoverable at the moment
+                // on the Hermes side. We'll inform the user to check for misconfig.
+                error!(
+                    "[{}] the price configuration for this chain may be too low!\\
+                    please check the `gas_price.price` Hermes config.toml",
+                    self.id()
+                );
             }
         }
 
--- a/relayer/src/chain/cosmos.rs
+++ b/relayer/src/chain/cosmos.rs
@@ -71,7 +71,6 @@ use ibc_proto::ibc::core::connection::v1::{
     QueryClientConnectionsRequest, QueryConnectionsRequest,
 };
 
-use crate::config::{AddressType, ChainConfig, GasPrice};
 use crate::error::Error;
 use crate::event::monitor::{EventMonitor, EventReceiver};
 use crate::keyring::{KeyEntry, KeyRing, Store};
@@ -79,6 +78,10 @@ use crate::light_client::tendermint::LightClient as TmLightClient;
 use crate::light_client::LightClient;
 use crate::light_client::Verified;
 use crate::{chain::QueryResponse, event::monitor::TxMonitorCmd};
+use crate::{
+    config::{AddressType, ChainConfig, GasPrice},
+    sdk_error::sdk_error_from_tx_sync_error_code,
+};
 
 use super::{ChainEndpoint, HealthCheck};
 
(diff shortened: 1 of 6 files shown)


relayer/src/link/relay_path.rs
```diff
@@ -864,25 +864,32 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+                "[{}] found zero unprocessed SendPacket events on source chain, nothing to do",
+                self
+            );
+        } else {
+            let mut packet_sequences = vec![];
+            for event in events_result.iter() {
+                match event {
+                    IbcEvent::SendPacket(send_event) => {
+                        packet_sequences.push(send_event.packet.sequence);
+                        if packet_sequences.len() > 10 {
```
hu55a1n1, line 878:
```suggestion
                        if packet_sequences.len() >= 10 {
```

relayer/src/link/relay_path.rs
```diff
@@ -946,22 +953,29 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+                "[{}] found zero unprocessed WriteAcknowledgement events on source chain, nothing to do",
+                self
+            );
+        } else {
+            let mut packet_sequences = vec![];
+            for event in events_result.iter() {
+                match event {
+                    IbcEvent::WriteAcknowledgement(write_ack_event) => {
+                        packet_sequences.push(write_ack_event.packet.sequence);
+                        if packet_sequences.len() > 10 {
```
hu55a1n1, line 967:
```suggestion
                        if packet_sequences.len() >= 10 {
```

relayer/src/chain/cosmos.rs
```diff
@@ -276,9 +276,25 @@ impl CosmosSdkChain {
...
+                debug!("[{}] send_tx: broadcast_tx_sync: {:?}", self.id(), response);
+            }
+            tendermint::abci::Code::Err(_) => {
+                // Avoid increasing the account s.n. if CheckTx failed
+                // Log the error
+                error!("[{}] send_tx: broadcast_tx_sync: {:?}", self.id(), response);
+                // The primary reason (we know of) causing broadcast_tx_sync to fail
+                // is due to "out of gas" errors. These are unrecoverable at the moment
+                // on the Hermes side. We'll inform the user to check for misconfig.
+                error!(
```
soareschen, line 292: Should we check for the error code 11 to display the out of gas error message? https://github.com/cosmos/cosmos-sdk/blob/v0.42.10/types/errors/errors.go#L50

soareschen (approved):

relayer/src/chain/cosmos.rs
```diff
@@ -276,9 +276,25 @@ impl CosmosSdkChain {
 
         let response = self.block_on(broadcast_tx_sync(self, tx_bytes))?;
 
-        debug!("[{}] send_tx: broadcast_tx_sync: {:?}", self.id(), response);
-
-        self.incr_account_sequence()?;
+        match response.code {
+            tendermint::abci::Code::Ok => {
+                // A success means the account s.n. was increased
```
hu55a1n1, line 281: Just a question - does a failure during DeliverTx also result in incrementing the account sequence? And is it possible that CheckTx succeeds while DeliverTx fails? 
I'm asking because I notice from my tests that a failure in DeliverTx (with a non-zero code) is still interpreted as `tendermint::abci::Code::Ok`.

relayer/src/chain/cosmos.rs
```diff
@@ -276,9 +276,25 @@ impl CosmosSdkChain {
 
         let response = self.block_on(broadcast_tx_sync(self, tx_bytes))?;
 
-        debug!("[{}] send_tx: broadcast_tx_sync: {:?}", self.id(), response);
-
-        self.incr_account_sequence()?;
+        match response.code {
```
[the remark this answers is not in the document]
hu55a1n1: Never mind - just realized this is a tendermint-rs thing. 😅

relayer/src/link/relay_path.rs
adizere: Nice catch!

relayer/src/chain/cosmos.rs
adizere: >does a failure during DeliverTx also result in incrementing the account sequence?

A failure in CheckTx means the transactions was not accepted into the mempool. The s.n., remains unchanged. A success in CheckTx (and I think regardless of the outcome of DeliverTx) means the s.n. increased.

> And is it possible that CheckTx succeeds while DeliverTx fails?

Yes, for instance DeliverTx can result in an error with insufficient fees, or "already executed" error.

relayer/src/chain/cosmos.rs
adizere: This is a very good point I overlooked, thanks Soares, fixing now!

romac (approved):

Issue #620, opened 2021-02-08
Improve crates.io and docs.rs documentation


adizere:
## Crate

ibc, relayer, and relayer-cli

## Summary

The [crates.io](https://crates.io/crates/ibc-relayer) and [docs.rs](https://docs.rs/ibc-relayer-cli/0.1.0/ibc_relayer_cli/) presentation of our main crates is minimalistic. Especially for the [ibc library](https://docs.rs/ibc/0.1.0/ibc/), we should improve documentation and presentation.

[ibc-proto](https://crates.io/crates/ibc-proto) is better in this regard.

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

romac:
Few other things to look for:

- Some modules don't have any module-level documentation (see first few modules in https://docs.rs/ibc/0.1.0/ibc/index.html)
- Some modules don't export anything (https://docs.rs/ibc/0.1.0/ibc/ics03_connection/handler/conn_open_ack/index.html)

PR #1407, opened 2021-10-01
Improve crates.io documentation

seanchen1991:
Addresses #620. 

### DIFF
--- a/modules/src/lib.rs
+++ b/modules/src/lib.rs
@@ -15,19 +15,31 @@
 )]
 #![forbid(unsafe_code)]
 
-//! Implementation of the following ICS modules:
+//! This library implements the _I_nter_B_lockchain _C_ommunication (IBC) protocol in Rust. IBC is
+//! a distributed protocol that enables communication between distinct sovereign blockchains.
+//! Loose analogies may be drawn between the IBC protocol and the TCP/UDP protocols that enable
+//! communication over the internet via packet streaming. Indeed, IBC also encodes the notion of
+//! ordered and unordered packet streams.
 //!
-//! - ICS 02: Client
-//! - ICS 03: Connection
-//! - ICS 04: Channel
-//! - ICS 05: Port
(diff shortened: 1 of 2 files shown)


modules/src/lib.rs
```diff
@@ -15,19 +15,31 @@
...
-//! - ICS 05: Port
-//! - ICS 07: Tendermint Client
-//! - ICS 18: Basic relayer functions
-//! - ICS 23: Vector Commitment Scheme
-//! - ICS 24: Host Requirements
-//! - ICS 26: Routing
-//! - Applications:
-//!    - ICS 20: Fungible Token Transfer
+//! At its highest level, IBC can be categorized into on-chain modules and off-chain processes.
+//! On-chain modules, so called because they are the building blocks upon which IBC chains are
```
adizere, line 25:
```suggestion
//! The present crate `ibc` is a Rust implementation of the on-chain modules.
//! On-chain module, so called because it is the main building blocks upon which IBC chains are
```

adizere (commented):
Great start!

I just left a comment to clarify how the crate fits within the documentation's content.

Seems like it needs a quick `fmt` fix.

Should we also cover the concerns from this comment? https://github.com/informalsystems/ibc-rs/issues/620#issuecomment-776003378

seanchen1991:
Do you think the scope, i.e. covering Clients, Connections, and Channels in brief, is appropriate for the root-level documentation of this crate? Should it be expanded to cover more components of the crate? 

Yea, we can definitely cover the concerns listed in #620. Another synchronous session to cover modules with missing documentation would be super helpful when it comes to this 🙂

adizere:
> Do you think the scope, i.e. covering Clients, Connections, and Channels in brief, is appropriate for the root-level documentation of this crate? Should it be expanded to cover more components of the crate?

We're missing some functionalities, so the mentioned scope is appropriate for the current state of the `ibc` crate.

Issue #1412, opened 2021-10-04
gm-connector for e2e testing in Rust


greg-szabo:
## Crate
relayer
gm-connector

## Summary
gm-connector is a Rust library that allows Rust tests to start/stop/manage SDK networks through `gm` (Gaiad Manager).

## Problem Definition
e2e testing in Rust is complex and it has dependencies completely unrelated to Rust (or to the code we oversee): network binaries, configuration, ports, etc.

During developer testing it is possible to contain these dependencies with `gm` Gaiad Manager. `gm` allows the user to describe multiple network nodes and their dependencies on each other in a simple TOML file.

`gm-connector` should allow the developer to do the same in Rust tests either ran locally or in CI. `gm-connector` deals with telling `gm` to start/stop any network nodes as well as it returns endpoints to the developer for RPC, gRPC and other connection types.
Subsequently, `gm-connector` should also allow the developer to read/change the network configuration of each node.

What problems may be addressed by introducing this feature?

e2e testing now uses a lot of different kind of dependencies (Python, network binaries, etc). The relevant dependencies are lowered.

What benefits does IBC-rs stand to gain by including this feature?

Easier e2e testing using Rust native tools.

Are there any disadvantages of including this feature?

Yet another Rust library to maintain. (Fortunately, it is only used in testing.)

## Proposal
Write an initial version of `gm-connector` and see if it resolves dependencies as expected.

## Acceptance Criteria
At least one e2e test can be run successfully without invoking Python.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
After discussing with @soareschen and @greg-szabo, we have concluded that:

- There is no need for `gm-connector`, so we can close this issue;
- `gm` definitely valuable now, and will definitely be needed in the future for developer network setup & manual/complex tests
- `gm` may grab new users in the future (e.g., gaiad or ibc-go team) and then we may consider moving it in its own repo to cover those use-case.

PR #1413, opened 2021-10-04
gm-connector initial release

greg-szabo:
Closes: #1412

## Description
gm-connector initial implementation with updates to `gm` to support it.

______

For contributor use:


adizere:
Many thanks Greg for broaching this!

@soareschen is the main customer for this feature. I suspect we'll make good use of the gm-connector in bootstrapping and managing chain binaries, in particular for PR #1273.

For context, this PR may be blocked for a few days. Reason is that Soares is sprinting to finish no_std, so it may be a few days until we get his input here. I think that shouldn't be a problem.

greg-szabo:
All improvements are welcome, it was essentially written in a few nights and it's more a line of thought than a crate of code. :) (Option<> instead of Result<>? What was I smoking?) Think of it more as a conversation starter than a PR. (Well, the `gm` part contains a few fixes that are worth merging, but the Rust code is more of a topic of discussion.)

Regarding separate crate: sure, I don't mind either way. Writing a PoC was easier in a separate crate. How much of our e2e testing toolset will be reusable outside of Hermes testing? `gm` was written with Hermes in mind, but the target audience is "anyone who wants to run testnets" and people on the Go side of things picked it up too.

Calling `gaiad` directly: it's up to you. I had the same idea for years (setting up a testnet is like four `gaiad` commands total) but somehow I kept running into people not going through the motions and complaining about how hard it is setting up testnets. `gm` solves issues around configuring these networks: validator key, validator gentx, wallet with available funds, managing endpoints on the host, etc. If you don't need those for testing, we can simply close this ticket without merging. (I'll move the `gm`-related fixes over to a new PR.) As I mentioned, it's more of a topic to discuss.

I'm not sure what you mean by async dispatch of `gm`. The code doesn't exit until `gm` is finished and `gm` goes through the nodes one-by-one and starts them. You don't need to wait after calling the `start()` method, your network is online. At least, that's what I had in mind. I might have missed the target, maybe?

scripts/gm/bin/lib-gm
```diff
@@ -5,9 +5,9 @@ if [ "${DEBUG:-}" = "2" ]; then
 fi
 
 version() {
-  VERSION="v0.0.9"
-  if [ "${OUTPUT:-}" == "json" ]; then
-    printf '{"status": "success", "message": "%s"}' "$VERSION"
+  VERSION="v0.1.0"
```
adizere, line 8: 0.1 is a relatively big moment! Great work with gm Greg!

adizere (approved):
Cool work here!

I read through the code & comments and I have two remarks:

1.  It's a bit of a bummer that we are not reusing code between the crates `gm-connector` and `relayer-test` (#1521). We will soon have two crates that accomplish very similar work, i.e., chain scaffolding (or driver or manager).
    - One interpretation of the situation: we probably got here because of the very explicit purpose of `relayer-test` to focus on reproducing relayer logic/core bugs, as opposed of a focus on enabling interoperability with different gaia versions and different chain binaries (regen, akash) which `gm` offers. Given that purpose, I think we probably did the pragmatic decision to add in `relayer-test` support similar to that of `gm-connector`.
    - That being said, it's not a problem at the moment, but in the future I doubt it's a good idea, and it's preferable I believe to have more collaboration and more code reuse. Let me know if I'm misinterpreting anything here.


2. Repository layout:
    - What do you think about moving `gm-connector` out of the top-level and into a `scripts` or `tests` sub-directory? I think the same applies for #1521 BTW (@soareschen).
    - Context: I think at the top level we'll strive to keep either ibc-related artifacts (`module`, `proto`, `relayer`, `relayer-cli`) or simple categories of supporting artifacts (`tests`, `CI`, `docs`) to make navigation simpler. We'll probably refactor everything soon, so let's start experimenting with this scheme already in this PR. WDYT?

scripts/gm/bin/lib-gm
greg-szabo: Thanks. I wanted to make it count. :)

greg-szabo:
I agree with Adi's concerns, hence I wanted to have more discussion around this.

Regarding folder structure: in tendermint-rs we created a `tools` folder which is essentially its own set of Rust crates completely separated from the main repository. It contains the proto-compiler, some testing frameworks and other tools that required structured upkeep (i.e. `Cargo.toml`) but we didn't want to mix it with the "main" topic of the repository. Check it out: https://github.com/informalsystems/tendermint-rs/tree/master/tools
I would do something similar here and move `gm-connector`, `proto-compiler` and possibly `e2e` into it. I don't consider the scripts folder structured enough for that. (Especially, since it mixes configuration files with unversioned scripts as well as `gm` which is versioned and documented.)

I would like to merge the efforts of `relayer-test` and `gm-connector`, if possible but I could not get any requirements in Berlin. This would also have a big effect on the future of `gm`, so it is crucial to get this in order.

adizere:
Closing in favor of #1606.

This comment https://github.com/informalsystems/ibc-rs/issues/1412#issuecomment-976334390 provides more context on why some of the changes in this PR are not necessary at the moment.

PR #1430, opened 2021-10-06
Conversion from `IbcEvent` to `AbciEvent`

yito88:
Closes: informalsystems/ibc-rs-tmp#81

## Description
- Add the conversion from `IbcEvent` to `AbciEvent`
- Make consistent with `IbcEventType` and `IbcEvent`

______

For contributor use:


romac:
@yito88 Thanks for the modifications!

romac:
I think I managed to fix the issue when decoding Channel and Connection events, but it appears that the decoding of Packet-related event is also somewhat broken at the moment. I will keep investigating.

As for connection and channels, I am planning to push some commits to revert back to the previous behavior where we accept invalid connection and channel identifiers for all events, and we can then follow-up on that in a separate PR to only allow those at specific steps of the handshake, ie. only allow empty counterparty connection identifiers when we are at the ConnInit step.

hu55a1n1:
Here's the fix for the tests, along with some cleanup ->
```patch
diff --git a/modules/src/ics04_channel/events.rs b/modules/src/ics04_channel/events.rs
index 8a34da02..809a8c2f 100644
--- a/modules/src/ics04_channel/events.rs
+++ b/modules/src/ics04_channel/events.rs
@@ -7,7 +7,6 @@ use crate::ics24_host::identifier::{ChannelId, ConnectionId, PortId};
 use crate::prelude::*;
 
 use serde_derive::{Deserialize, Serialize};
-use subtle_encoding::hex;
 use tendermint::abci::tag::Tag;
 use tendermint::abci::Event as AbciEvent;
 
@@ -312,9 +311,8 @@ impl From<Packet> for Vec<Tag> {
                 .unwrap(),
         };
         attributes.push(timeout_timestamp);
-        let encoded = hex::encode(p.data);
         let val =
-            String::from_utf8(encoded).expect("hex-encoded string should always be valid UTF-8");
+            String::from_utf8(p.data).expect("hex-encoded string should always be valid UTF-8");
         let packet_data = Tag {
             key: PKT_DATA_ATTRIBUTE_KEY.parse().unwrap(),
             value: val.parse().unwrap(),
@@ -735,9 +733,8 @@ impl From<WriteAcknowledgement> for IbcEvent {
 impl From<WriteAcknowledgement> for AbciEvent {
     fn from(v: WriteAcknowledgement) -> Self {
         let mut attributes = Vec::<Tag>::from(v.packet);
-        let encoded = hex::encode(v.ack);
         let val =
-            String::from_utf8(encoded).expect("hex-encoded string should always be valid UTF-8");
+            String::from_utf8(v.ack).expect("hex-encoded string should always be valid UTF-8");
         // No actual conversion from string to `Tag::Key` or `Tag::Value`
         let ack = Tag {
             key: PKT_ACK_ATTRIBUTE_KEY.parse().unwrap(),
diff --git a/modules/src/ics04_channel/packet.rs b/modules/src/ics04_channel/packet.rs
index 919ef2a5..0ff7370c 100644
--- a/modules/src/ics04_channel/packet.rs
+++ b/modules/src/ics04_channel/packet.rs
@@ -100,7 +100,7 @@ impl core::fmt::Display for Sequence {
     }
 }
 
-#[derive(Clone, Debug, Hash, PartialEq, Deserialize, Serialize)]
+#[derive(Clone, Debug, Default, Hash, PartialEq, Deserialize, Serialize)]
 pub struct Packet {
     pub sequence: Sequence,
     pub source_port: PortId,
@@ -138,21 +138,6 @@ impl core::fmt::Display for Packet {
     }
 }
 
-impl Default for Packet {
-    fn default() -> Self {
-        Packet {
-            sequence: Sequence(0),
-            source_port: Default::default(),
-            source_channel: Default::default(),
-            destination_port: Default::default(),
-            destination_channel: Default::default(),
-            data: Vec::new(),
-            timeout_height: Default::default(),
-            timeout_timestamp: Default::default(),
-        }
-    }
-}
-
 impl TryFrom<RawPacket> for Packet {
     type Error = Error;
 
@@ -261,7 +246,7 @@ mod tests {
             Test {
                 name: "Good parameters".to_string(),
                 raw: default_raw_msg.clone(),
-                want_pass: true
+                want_pass: true,
             },
             Test {
                 name: "Src port validation: correct".to_string(),

```

Issue #1408, opened 2021-10-04
Upgrade protos, upgrade gaiad in CI [Vega upgrade]


adizere:
## Summary

Gaiad RC for 6.0.0 is out https://github.com/cosmos/gaia/releases/tag/v6.0.0-rc1

This is the first RC for Vega update, which will go live on the hub soon (should have been in Q3 2021).
https://github.com/cosmos/gaia/blob/main/docs/roadmap/cosmos-hub-roadmap-2.0.md#vega-upgrade-expected-q3-2021

## Proposal

We should prepare for 6.0.0:
- [ ] update our CI suite to use the new chain binary
- [ ] upgrade our generated .proto files to include [ibc-go 1.2](https://github.com/cosmos/ibc-go/blob/v1.2.0/CHANGELOG.md), with the associated compatibility fixes; for these fixes, it could be instructive to read on the [versioning discussion from the ibc-go repo](https://github.com/cosmos/ibc-go/discussions/409)

____

#### For Admin Use

- [x] Not duplicate issue
- [X] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

andynog:
Running the e2e test locally against gaia v6.0.0-rc2, the relayer test fails with:
````
{"timestamp":"2021-11-09T20:11:36.075264Z","level":"TRACE","fields":{"message":"[ibc-0 -> ibc-1:07-tendermint-1] finished misbehaviour verification after 5.074784876s"},"threadId":"ThreadId(22)"}
{"timestamp":"2021-11-09T20:11:36.130906Z","level":"TRACE","fields":{"message":"light client verification","trusted":"1-85","target":"1-96"},"threadId":"ThreadId(19)"}
{"timestamp":"2021-11-09T20:11:36.376720Z","level":"TRACE","fields":{"message":"light client verification","trusted":"1-78","target":"1-85"},"threadId":"ThreadId(19)"}
{"timestamp":"2021-11-09T20:11:36.624447Z","level":"TRACE","fields":{"message":"light client verification","trusted":"1-64","target":"1-78"},"threadId":"ThreadId(19)"}
{"timestamp":"2021-11-09T20:11:36.876632Z","level":"TRACE","fields":{"message":"light client verification","trusted":"1-61","target":"1-64"},"threadId":"ThreadId(19)"}
[...]

PR #1438, opened 2021-10-07
Upgrade proto to ibc-go v1.2 and sdk v0.44

adizere:
Closes: #1408 

### Proto files

Proto files are even with [ibc-go v1.2.2](https://github.com/cosmos/ibc-go/releases/tag/v1.2.2) and [SDK v0.44.3](https://github.com/cosmos/cosmos-sdk/releases/tag/v0.44.3).

### Compatibility

Hermes is fully compatible with ibc-go v1.2 and SDK v0.44 **with one exception**:
- there is unsupported behavior when two chains having ibc-go version pre-v1.2 and the other post-v1.2 try to handshake on a new channel and the channel's port identifier has length bigger than 64 characters.
    - we will tackle this corner case in a separate PR

### Bootstrap method semantics

The method [`CosmosSdk::bootstrap`](https://github.com/informalsystems/ibc-rs/blob/652e8f0b6455ca4dc26f6a735ddb1e041d8ed5eb/relayer/src/chain/cosmos.rs#L762) can potentially return with error if it fails to query the SDK version from the full node (via gRPC `GetNodeInfoRequest`).
  - This means that **the semantics of `bootstrap` changed**: this method returns successfully only if the the full node is up-and-running and Hermes obtains a successful response to `GetNodeInfoRequest`.
  - Before this PR, the semantics of `bootstrap` were more lax and this method returned Ok even if the full node was entirely inaccessible.
  - For this reason, this PR eliminates some of the `bootstrap(..).unwrap()` calls. These unwraps would result in a panic when the full node is inaccessible.

______
### TODOs

- [x] document in a separate issue the incompatibilty corner-case we're not handling yet re: port id length > 64
______

For contributor use:


relayer/src/chain/cosmos.rs
```diff
@@ -1469,7 +1475,18 @@ impl ChainEndpoint for CosmosSdkChain {
...
+            // if < v1.2 overwrite with empty `packet_commitment_sequences` field
+            let mut r = request;
+            r.packet_commitment_sequences = vec![];
+            r
+        } else {
+            // if >= v1.2 leave the request unchanged
+            request
+        };
+
+        let request = tonic::Request::new(version_compatible_request);
```
ancazamfir, line 1489: After looking more into this, doing more tests, reading on proto, I think we should undo this change. See [unknowns](https://developers.google.com/protocol-buffers/docs/proto3#unknowns).
It turns out unknown fields are just dropped or retained as unknown fields (depending on the proto and parser versions) but the parser should still be able to work with all fields present in the old proto version. I tested without these checks with old gaia versions and all looks good.
(fwiw for ibc-rs/modules and basecoin, prost [drops](https://github.com/tokio-rs/prost/issues/2) unknown fields iiuc)
Would still keep the `versions_specs` boilerplate to deal with incompatibilities not handled by proto, e.g. `chann-init` with >64 len ports`PortId`-s where we need to check the version of the chain potentially receiving the `chann-try`.

adizere (commented):
LGTM!

romac (approved):
🎉

PR #1491, opened 2021-10-25
Structured logs for relayer logic

adizere:
Closes: #1537

## Description

______

For contributor use:


relayer-cli/src/commands.rs
```diff
@@ -140,6 +141,14 @@ impl Configurable<Config> for CliCmd {
             ccfg.memo_prefix.apply_suffix(&suffix);
         }
 
+        // For all commands except for `start` Hermes retries
+        // for a prolonged period of time.
+        if !matches!(self, CliCmd::Start(_)) {
```
adizere, line 146: We may consider being more restrictive here (only some commands, e.g., `create-*`).

MZ: Anything that is interactive could have a separate config parameter, so we avoid the hardcoded value here.

relayer/src/connection.rs
```diff
@@ -754,9 +755,11 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Connection<ChainA, ChainB> {
     pub fn build_conn_init_and_send(&self) -> Result<IbcEvent, ConnectionError> {
         let dst_msgs = self.build_conn_init()?;
 
+        let tm = TrackedMsgs::new(dst_msgs, "create connection");
```
adizere, line 758: MZ: COuld be an enum, not a hardcoded string.

AS: These changes are not very interesting, just a workaround to using tracked messages. This code pertains to interactive CLIs (e.g., `hermes create channel --port-a transfer..`).

relayer/src/connection.rs
```diff
@@ -1062,9 +1074,11 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Connection<ChainA, ChainB> {
     pub fn build_conn_confirm_and_send(&self) -> Result<IbcEvent, ConnectionError> {
         let dst_msgs = self.build_conn_confirm()?;
 
+        let tm = TrackedMsgs::new(dst_msgs, "create connection");
```
mzabaluev, line 1077: Also lots of places using "create connection". Should they all be refactored to a helper method that would do it in one place?

relayer/src/link/operational_data.rs
```diff
@@ -109,21 +166,24 @@ impl OperationalData {
             None => self.batch.iter().map(|gm| gm.msg.clone()).collect(),
         };
 
+        let tm = TrackedMsgs::new(msgs, &self.tracking_id);
+
         info!(
-            "[{}] assembled batch of {} message(s)",
+            "[{}] assembled batch of {} msgs",
```
mzabaluev, line 172: Why abbreviate in this INFO level message?

relayer/src/link/relay_path.rs
```diff
@@ -390,17 +393,24 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
-        let mut dst_od = OperationalData::new(src_height, OperationalDataTarget::Destination);
+        let mut dst_od = OperationalData::new(
+            src_height,
+            OperationalDataTarget::Destination,
+            events.tracking_id(),
+        );
 
         for event in input {
-            debug!("[{}] {} => {}", self, self.src_chain().id(), event);
+            trace!("[{}] {} => {}", self, self.src_chain().id(), event);
```
mzabaluev, line 409: This could use more KV pairs for legibility. The span already gives the tracking ID, so the `self` parameter replicates it (and adds more non-KV clutter)?

relayer/src/link/relay_path.rs
```diff
@@ -370,15 +376,12 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+        events: TrackedEvents,
     ) -> Result<(Option<OperationalData>, Option<OperationalData>), LinkError> {
-        if !input.is_empty() {
-            info!(
-                "[{}] generate messages from batch with {} events",
-                self,
-                input.len()
-            );
-        }
+        let span = span!(Level::DEBUG, "generate", id = %events);
```
mzabaluev, line 381: Less cryptic IMHO:

```suggestion
        let span = span!(Level::DEBUG, "generate", id = %events.tracking_id());
```

relayer/src/link/relay_path.rs
```diff
@@ -501,15 +511,15 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
         initial_od: OperationalData,
     ) -> Result<S::Reply, LinkError> {
         // We will operate on potentially different operational data if the initial one fails.
+        let span = span!(Level::INFO, "relay", id = %initial_od);
```
mzabaluev, line 514:
```suggestion
        let span = span!(Level::INFO, "relay", id = %initial_od.tracking_id());
```

relayer/src/link/relay_path.rs
```diff
@@ -501,15 +511,15 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+        let _enter = span.enter();
+
         let mut odata = initial_od;
 
         for i in 0..MAX_RETRIES {
-            info!(
-                "[{}] relay op. data of {} msgs(s) to {} (height {}), delayed by: {:?} [try {}/{}]",
+            debug!(
+                "[{}] delayed by: {:?} [try {}/{}]",
```
mzabaluev, line 521: Could use more KV pairs.

relayer/src/link/relay_path.rs
```diff
@@ -565,7 +575,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
         &self,
         initial_odata: OperationalData,
     ) -> Option<OperationalData> {
-        info!(
+        warn!(
```
mzabaluev, line 578: Could this benefit from the tracking ID of `initial_odata` under the same `id` key as elsewhere?

relayer/src/link/relay_path.rs
```diff
@@ -1343,34 +1364,32 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
                         } else if let Some(new_msg) =
                             self.build_timeout_from_send_packet_event(e, &dst_status)?
                         {
-                            debug!(
-                                "[{}] refreshing schedule: found a timed-out msg in the op data {}",
-                                self, odata
-                            );
-                            timed_out.entry(odata_pos).or_insert_with(Vec::new).push(
-                                TransitMessage {
+                            debug!("[{}] found a timed-out msg in the op data {}", self, odata);
```
mzabaluev, line 1371: The id of `odata` could be put into KV here, too, rather than formatting in-message.

relayer/src/link/relay_path.rs
```diff
@@ -1419,6 +1433,9 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
     /// If the relaying path has non-zero packet delays, this method also updates the client on the
     /// target chain with the appropriate headers.
     fn schedule_operational_data(&self, mut od: OperationalData) -> Result<(), LinkError> {
+        let span = span!(Level::INFO, "schedule", id = %od);
```
mzabaluev, line 1436:
```suggestion
        let span = span!(Level::INFO, "schedule", id = %od.tracking_id());
```

relayer/src/link/relay_path.rs
```diff
@@ -1427,21 +1444,23 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+                }
+                OperationalDataTarget::Destination => {
+                    self.update_client_dst(target_height, &od.tracking_id)?
+                }
             };
+        } else {
+            debug!(
+                "[{}] connection delay is zero: client update message will be prepended later",
+                self
+            );
```
mzabaluev, line 1463: All this formatting of `self` maybe needs a more systematic alternative with a helper making use of spans and/or KVs, but I don't have a well formed proposal at the moment.

relayer/src/supervisor/spawn.rs
```diff
@@ -407,7 +407,7 @@ impl<'a, Chain: ChainHandle + 'static> SpawnContext<'a, Chain> {
         );
 
         if conn_state_src.is_open() && conn_state_dst.is_open() {
-            debug!(
+            trace!(
                 "connection {} on chain {} is already open, not spawning Connection worker",
                 connection.connection_id,
                 chain.id()
```
mzabaluev, line 413: Both args here could usefully be KVs:

```suggestion
            trace!(
                connection_id=%connection.connection_id,
                chain_id=%chain.id(),
                "connection is already open, not spawning Connection worker"
```

config.toml
```diff
@@ -121,6 +121,8 @@ websocket_addr = 'ws://127.0.0.1:26657/websocket'
 
 # Specify the maximum amount of time (duration) that the RPC requests should
 # take before timing out. Default: 10s (10 seconds)
+# Note: Hermes uses this parameter _only_ in `start` mode; for all other CLIs,
+# Hermes uses a large preconfigured timeout (on the order of minutes).
```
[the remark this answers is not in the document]
adizere: Hmm not sure I see the benefits of configuring the CLI-specific (non-`start`) timeout. Let's keep that as a nice-to-have for later, if the need arises.

relayer/src/chain/cosmos.rs
```diff
@@ -1009,13 +1010,14 @@ impl ChainEndpoint for CosmosSdkChain {
...
-        proto_msgs: Vec<Any>,
+        tracked_msgs: TrackedMsgs,
     ) -> Result<Vec<Response>, Error> {
         crate::time!("send_messages_and_wait_check_tx");
-        debug!(
-            "send_messages_and_wait_check_tx with {} messages",
-            proto_msgs.len()
-        );
+
+        let span = span!(Level::DEBUG, "send", id = %tracked_msgs);
```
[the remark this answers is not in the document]
adizere: I chose the names `send` and `send_w` for brevity, agree not the most straightforward choice.

Alternatives, feel free to choose whichever seems more appropriate:
- `send_messages_commit` (instead of `send_w`) and `send_messages_check`
- `send_commit` (instead of `send_w`) and `send_check`

relayer/src/chain/tx.rs
```diff
@@ -0,0 +1,41 @@
...
+    fn from(tm: TrackedMsgs) -> Vec<Any> {
+        tm.msgs
+    }
+}
+
+impl fmt::Display for TrackedMsgs {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}; len={}", self.tracking_id, self.msgs.len())
+    }
+}
```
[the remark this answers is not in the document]
adizere: I might have abused `Display`, indeed.

> perhaps msgs.tracking_id() and msgs.len() could be used in format args instead of the whole value for better clarity?

Agree. It would be more appropriate to concretely specify which fields we want to display, namely `len` and `tracking_id`.

I added a comment (1d15085e) to at least document the purpose of this type. feel free to replace the use of `Display` with the proposed approach.

relayer/src/link/operational_data.rs
```diff
@@ -27,8 +29,57 @@ impl fmt::Display for OperationalDataTarget {
...
+            tracking_id: nanoid!(10),
+        }
+    }
+}
+
+impl fmt::Display for TrackedEvents {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.tracking_id)
+    }
+}
```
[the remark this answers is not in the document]
adizere: 60e66e64

relayer/src/link/operational_data.rs
adizere: No particular reason (maybe sloppiness). Feel free to fix.

relayer/src/link/relay_path.rs
adizere: 60e66e6438d87b0b5738d16610f4fb318816c82a

relayer/src/link/relay_path.rs
adizere: This is how this log line looks in practice:

> 2021-12-22T12:18:40.052114Z TRACE ThreadId(443) generate{id=BS6CBKeio4}: [ibc-0:transfer/channel-0 -> ibc-1] ibc-0 => SendPacketEv(SendPacket - h:0-25087, seq:9, path:channel-0/transfer->channel-0/transfer, toh:0-0, tos:Timestamp(2021-12-22T12:21:29.002094Z)))


I think `self` does not replicate the tracking ID, but displays the relaying path: `ibc-0:transfer/channel-0 -> ibc-1` in this case.

What would be a concrete suggestion to replace this log using KV syntax?

relayer/src/link/relay_path.rs
adizere: Note that this log provides a bit more context beyond the tracking identifier, namely:

> 2021-12-22T12:18:40.359804Z DEBUG ThreadId(444) relay{id=BS6CBKeio4 ->Destination @0-25087; len=1}: [ibc-0:transfer/channel-0 -> ibc-1] delayed by: 290.181236ms [try 1/5].

- it also shows whether the relay operation is targeting the Source or Destination (`->Destination` in this case)
- and the height where clients proofs are built `0-25087`
- and the # of messages the operation comprises.

I think we do want all of this information in there.
Should we structure it more clearly with KVs?

relayer/src/link/relay_path.rs
adizere: Same as my comment for the `relay` span. There is additional context beyond the tracking ID which we'd like to represent.

> 2021-12-22T12:18:40.069556Z DEBUG ThreadId(443) schedule{id=BS6CBKeio4 ->Destination @0-25087; len=1}: [ibc-0:transfer/channel-0 -> ibc-1] connection delay is zero: client update message will be prepended later

Shall we have KVs?

relayer/src/connection.rs
adizere: I think it may actually be wrong to call this "create connection", and instead we should be more accurate, and also provide the concrete name of the handshake step, in this case it would be: "create connection confirm". Agree with the refactoring ideea.

relayer/src/channel.rs
```diff
@@ -1190,9 +1200,11 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
     pub fn build_chan_close_confirm_and_send(&self) -> Result<IbcEvent, ChannelError> {
         let dst_msgs = self.build_chan_close_confirm()?;
 
+        let tm = TrackedMsgs::new(dst_msgs, "create channel");
```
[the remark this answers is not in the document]
adizere: Agree this is confusing. We should specify the step name, e.g. "create channel confirm".

relayer/src/link/relay_path.rs
adizere: I think the `Display` impl of odata provides more context, and it's usually useful to have those additional details (the target, the height, and the batch size).

mzabaluev: OK, it's just some logging backends allow you to do filtering by e.g. the `tracking_id` key, but baking it in would have the filter miss that.

relayer/src/link/operational_data.rs
```diff
@@ -50,38 +101,44 @@ pub struct OperationalData {
...
         }
     }
 
-    pub fn events(&self) -> Vec<IbcEvent> {
-        self.batch.iter().map(|gm| gm.event.clone()).collect()
+    pub fn push(&mut self, msg: TransitMessage) {
+        self.batch.push(msg)
+    }
+
+    pub fn events(&self) -> TrackedEvents {
```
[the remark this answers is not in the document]
mzabaluev: And changing the name revealed that the only place where it was used only wants the number of events, the rest of the copied data goes to waste. The whole place could be less clone-happy with the entire OperationalData value, too.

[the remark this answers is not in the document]
mzabaluev: OK, there is also another place that needs the batch, but it can move from the original struct and all its members are public :shrug:

relayer/src/link/relay_path.rs
mzabaluev: In 85c60fa8f1b37e0f823aaa3cf03c79bafd8ef994, I have opted to use the `odata` key to display the operational data information without commingling with `id` used just for tracking IDs. Here and in the other places.

relayer/src/link/operational_data.rs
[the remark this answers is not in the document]
mzabaluev: Changed to `into_events` as there is a lot of data that needs to be moved rather than copied.

relayer/src/link/relay_path.rs
mzabaluev: Used `odata` in 85c60fa8f1b37e0f823aaa3cf03c79bafd8ef994.

relayer/src/link/relay_path.rs
mzabaluev: Used `odata` in 85c60fa8f1b37e0f823aaa3cf03c79bafd8ef994.

relayer/src/chain/cosmos.rs
```diff
@@ -1111,7 +1112,7 @@ impl ChainEndpoint for CosmosSdkChain {
     ) -> Result<Vec<Response>, Error> {
         crate::time!("send_messages_and_wait_check_tx");
 
-        let span = span!(Level::DEBUG, "send", id = %tracked_msgs);
+        let span = span!(Level::DEBUG, "send_tx_check", id = %tracked_msgs.tracking_id());
```
adizere, line 1115: Thanks, this is clearer!

relayer/src/channel.rs
```diff
@@ -1119,7 +1119,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
     pub fn build_chan_close_init_and_send(&self) -> Result<IbcEvent, ChannelError> {
         let dst_msgs = self.build_chan_close_init()?;
 
-        let tm = TrackedMsgs::new(dst_msgs, "create channel");
+        let tm = TrackedMsgs::new(dst_msgs, "create channel with ChannelCloseInit");
```
adizere, line 1122: I think there's a typo here: the channel is not being created. But we can skip the "create channel" part altogether.
```suggestion
        let tm = TrackedMsgs::new(dst_msgs, "ChannelCloseInit");
```

relayer/src/channel.rs
```diff
@@ -1200,7 +1200,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
     pub fn build_chan_close_confirm_and_send(&self) -> Result<IbcEvent, ChannelError> {
         let dst_msgs = self.build_chan_close_confirm()?;
 
-        let tm = TrackedMsgs::new(dst_msgs, "create channel");
+        let tm = TrackedMsgs::new(dst_msgs, "create channel with ChannelCloseConfirm");
```
adizere, line 1203:
```suggestion
        let tm = TrackedMsgs::new(dst_msgs, "ChannelCloseConfirm");
```

relayer/src/channel.rs
```diff
@@ -704,7 +704,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
     pub fn build_chan_open_init_and_send(&self) -> Result<IbcEvent, ChannelError> {
         let dst_msgs = self.build_chan_open_init()?;
 
-        let tm = TrackedMsgs::new(dst_msgs, "create channel");
+        let tm = TrackedMsgs::new(dst_msgs, "create channel with ChannelOpenInit");
```
adizere, line 707: I would be in favor of keeping this message short, since it's used as prefix on many log messages. Suggestion below:

```suggestion
        let tm = TrackedMsgs::new(dst_msgs, "ChannelOpenInit");
```

relayer/src/channel.rs
```diff
@@ -870,7 +870,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
     pub fn build_chan_open_try_and_send(&self) -> Result<IbcEvent, ChannelError> {
         let dst_msgs = self.build_chan_open_try()?;
 
-        let tm = TrackedMsgs::new(dst_msgs, "create channel");
+        let tm = TrackedMsgs::new(dst_msgs, "create channel with ChannelOpenTry");
```
adizere, line 873:
```suggestion
        let tm = TrackedMsgs::new(dst_msgs, "ChannelOpenTry");
```

relayer/src/channel.rs
```diff
@@ -957,7 +957,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
         ) -> Result<IbcEvent, ChannelError> {
             let dst_msgs = channel.build_chan_open_ack()?;
 
-            let tm = TrackedMsgs::new(dst_msgs, "create channel");
+            let tm = TrackedMsgs::new(dst_msgs, "create channel with ChannelOpenAck");
```
adizere, line 960:
```suggestion
            let tm = TrackedMsgs::new(dst_msgs, "ChannelOpenAck");
```

relayer/src/channel.rs
```diff
@@ -1055,7 +1055,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
         ) -> Result<IbcEvent, ChannelError> {
             let dst_msgs = channel.build_chan_open_confirm()?;
 
-            let tm = TrackedMsgs::new(dst_msgs, "create channel");
+            let tm = TrackedMsgs::new(dst_msgs, "create channel with ChannelOpenConfirm");
```
adizere, line 1058:
```suggestion
            let tm = TrackedMsgs::new(dst_msgs, "ChannelOpenConfirm");
```

relayer/src/connection.rs
```diff
@@ -768,7 +768,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Connection<ChainA, ChainB> {
     pub fn build_conn_init_and_send(&self) -> Result<IbcEvent, ConnectionError> {
         let dst_msgs = self.build_conn_init()?;
 
-        let tm = TrackedMsgs::new(dst_msgs, "create connection");
+        let tm = TrackedMsgs::new(dst_msgs, "create connection with ConnectionOpenInit");
```
adizere, line 771:
```suggestion
        let tm = TrackedMsgs::new(dst_msgs, "ConnectionOpenInit");
```

relayer/src/connection.rs
```diff
@@ -899,7 +899,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Connection<ChainA, ChainB> {
     pub fn build_conn_try_and_send(&self) -> Result<IbcEvent, ConnectionError> {
         let dst_msgs = self.build_conn_try()?;
 
-        let tm = TrackedMsgs::new(dst_msgs, "create connection");
+        let tm = TrackedMsgs::new(dst_msgs, "create connection with ConnectionOpenTry");
```
adizere, line 902:
```suggestion
        let tm = TrackedMsgs::new(dst_msgs, "ConnectionOpenTry");
```

relayer/src/connection.rs
```diff
@@ -828,7 +828,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Connection<ChainA, ChainB> {
             .map_err(|e| ConnectionError::chain_query(self.dst_chain().id(), e))?;
         let client_msgs = self.build_update_client_on_src(src_client_target_height)?;
 
-        let tm = TrackedMsgs::new(client_msgs, "create connection");
+        let tm = TrackedMsgs::new(client_msgs, "update client on source for ConnectionOpenTry");
```
adizere, line 831: I can't think of a shorter way to formulate this without losing context. I guess we can keep it as it is.

relayer/src/connection.rs
```diff
@@ -995,7 +995,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Connection<ChainA, ChainB> {
     pub fn build_conn_ack_and_send(&self) -> Result<IbcEvent, ConnectionError> {
         let dst_msgs = self.build_conn_ack()?;
 
-        let tm = TrackedMsgs::new(dst_msgs, "create connection");
+        let tm = TrackedMsgs::new(dst_msgs, "create connection with ConnectionOpenAck");
```
adizere, line 998:
```suggestion
        let tm = TrackedMsgs::new(dst_msgs, "ConnectionOpenAck");
```

relayer/src/connection.rs
```diff
@@ -1074,7 +1074,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Connection<ChainA, ChainB> {
     pub fn build_conn_confirm_and_send(&self) -> Result<IbcEvent, ConnectionError> {
         let dst_msgs = self.build_conn_confirm()?;
 
-        let tm = TrackedMsgs::new(dst_msgs, "create connection");
+        let tm = TrackedMsgs::new(dst_msgs, "create connection with ConnectionOpenConfirm");
```
adizere, line 1077:
```suggestion
        let tm = TrackedMsgs::new(dst_msgs, "ConnectionOpenConfirm");
```

relayer/src/link/relay_path.rs
adizere: > Could this benefit from the tracking ID of initial_odata under ..

Hmm I'm not sure what you mean by "this". Do you mean the `self` object, which is of type `RelayPath`? I think for this type of object we don't have tracking `id`s.

On line 586 below, the tracking `id` is propagated from `initial_data` into new `OperationalData` objects, which happens in method `generate_operational_data`.

adizere (commented):
Thanks for the adjustments Mikhail! I think we should just adjust a bit the tracking `id`s, because they can make the output lines very long -- I left suggestions for that.

Can we also signal the changes in the changelog? https://github.com/informalsystems/ibc-rs/blob/master/CONTRIBUTING.md#examples

relayer/src/channel.rs
mzabaluev: Yes, I just mechanically changed things without noticing that this line looks out of place. Will fix together with the other messages.

adizere (commented):
Once we do the release (merge #1712), let's merge this. Thanks Mikhail!

adizere:
Post-merge artifact that might need our attention: when we encounter an account sequence number mismatch error, we see the following error message missing any tracking id:

```
2022-01-10T13:45:14.341974Z WARN ThreadId(1723) task PacketWorker(ibc-1:transfer/channel-0 -> ibc-0) encountered ignorable error: link errror: failed with underlying error: gRPC call failed with status: status: InvalidArgument, message: "account sequence mismatch, expected 86, got 85: incorrect account sequence: invalid request", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }
```

mzabaluev:
@adizere The problem with that tracing entry is that it is [produced](https://github.com/informalsystems/ibc-rs/blob/914373c1560ce9c66cac6513c795306755b153e6/relayer/src/util/task.rs#L118) higher up in the call stack than the specific task that creates the tracking id.

relayer/src/util/task.rs
```diff
@@ -90,14 +90,19 @@ pub fn spawn_background_task<E: Display>(
     interval_pause: Option<Duration>,
     mut step_runner: impl FnMut() -> Result<Next, TaskError<E>> + Send + Sync + 'static,
 ) -> TaskHandle {
-    info!("spawning new background task {}", task_name);
+    let span = span!(tracing::Level::ERROR, "task", name = %task_name);
+    let _entered = span.enter();
+
+    info!("spawning");
```
adizere, line 96: Cool! 

I noticed that we can now get into a 3-level span, e.g., `task{..}:task{..}:relay{..}`.


Sample log output:
```
2022-01-11T08:36:32.047359Z  INFO ThreadId(01) Hermes has started
2022-01-11T08:36:47.169785Z  INFO ThreadId(2847) task{name=supervisor_batch}:task{name=PacketCmdWorker(ibc-0:transfer/channel-0 -> ibc-1)}: spawning
2022-01-11T08:36:47.170052Z  INFO ThreadId(2847) task{name=supervisor_batch}:task{name=PacketWorker(ibc-0:transfer/channel-0 -> ibc-1)}: spawning
2022-01-11T08:36:47.433581Z  INFO ThreadId(2851) task{name=supervisor_batch}:task{name=PacketCmdWorker(ibc-0:transfer/channel-0 -> ibc-1)}:relay{odata=BF24AQfUzl ->Destination @0-18454; len=1}: [ibc-0:transfer/channel-0 -> ibc-1] assembled batch of 2 message(s)
2022-01-11T08:36:47.436739Z  INFO ThreadId(18) [ibc-1] refresh: retrieved account sequence=109 number=1
2022-01-11T08:36:47.448264Z  INFO ThreadId(2851) task{name=supervisor_batch}:task{name=PacketCmdWorker(ibc-0:transfer/channel-0 -> ibc-1)}:relay{odata=BF24AQfUzl ->Destination @0-18454; len=1}: [Async~>ibc-1] response(s): 1; Ok:78DD8796BA5C98F34EE531F6156BE608A3456A3CE2F7EE2C52B05030BAA86A0A

2022-01-11T08:36:47.448364Z  INFO ThreadId(2851) task{name=supervisor_batch}:task{name=PacketCmdWorker(ibc-0:transfer/channel-0 -> ibc-1)}:relay{odata=BF24AQfUzl ->Destination @0-18454; len=1}: [ibc-0:transfer/channel-0 -> ibc-1] success
2022-01-11T08:36:55.751340Z  INFO ThreadId(2847) task{name=supervisor_batch}:task{name=PacketCmdWorker(ibc-1:transfer/channel-0 -> ibc-0)}: spawning
2022-01-11T08:36:55.751602Z  INFO ThreadId(2847) task{name=supervisor_batch}:task{name=PacketWorker(ibc-1:transfer/channel-0 -> ibc-0)}: spawning
2022-01-11T08:36:56.010186Z  INFO ThreadId(2871) task{name=supervisor_batch}:task{name=PacketCmdWorker(ibc-1:transfer/channel-0 -> ibc-0)}:relay{odata=LNDPziRenQ ->Destination @1-18452; len=1}: [ibc-1:transfer/channel-0 -> ibc-0] assembled batch of 2 message(s)
2022-01-11T08:36:56.013194Z  INFO ThreadId(17) [ibc-0] refresh: retrieved account sequence=135 number=1
2022-01-11T08:36:56.024201Z  INFO ThreadId(2871) task{name=supervisor_batch}:task{name=PacketCmdWorker(ibc-1:transfer/channel-0 -> ibc-0)}:relay{odata=LNDPziRenQ ->Destination @1-18452; len=1}: [Async~>ibc-0] response(s): 1; Ok:FFD0C8779F0C188A0828CF5FEE9E29A336E6BE60EEA76A8075FAA80717F3D6EF

2022-01-11T08:36:56.024290Z  INFO ThreadId(2871) task{name=supervisor_batch}:task{name=PacketCmdWorker(ibc-1:transfer/channel-0 -> ibc-0)}:relay{odata=LNDPziRenQ ->Destination @1-18452; len=1}: [ibc-1:transfer/channel-0 -> ibc-0] success
```

Feedback:

I think this improves debugging, as the relations among task names becomes explicit in each log line.

As a disadvantage: this can make log lines very long. So let's improve that.
- In particular the span `task{name=supervisor_batch}` not sure how it improves observability, so maybe we can skip entering that span.
- The span `task{name=PacketCmdWorker(ibc-0:transfer/channel-0 -> ibc-1)}` seems useful, and we should probably keep it. We have redundant information now, however, since the path details `[ibc-0:transfer/channel-0 -> ibc-1]` appears twice on most lines, see eg. lines with "assembled batch of 2 message(s)". This information appears first from the `span` we just introduced, and second from `RelayPath:fmt::Display` impl ([example](https://github.com/informalsystems/ibc-rs/blob/85c60fa8f1b37e0f823aaa3cf03c79bafd8ef994/relayer/src/link/operational_data.rs#L179)). I think we should eliminate the second and keep the span only.

mzabaluev: Thanks, I haven't noticed there is already a "task" span. I can revert the top-level messages in `spawn_background_task` to use the `bg_task` keyword with the background task name.

mzabaluev: Oh, this is more interesting; there are two layers of background tasks, one spawning another. In principle, this indirect recursion can go arbitrarily deep. So maybe a span is not a good solution.

mzabaluev: > * This information appears first from the `span` we just introduced, and second from `RelayPath:fmt::Display` impl ([example](https://github.com/informalsystems/ibc-rs/blob/85c60fa8f1b37e0f823aaa3cf03c79bafd8ef994/relayer/src/link/operational_data.rs#L179)).

Yes, there are many `[{}]` constructs in the log macro invocations, using a `Display` impl of this or that object to provide context information. I think they should all be converted to use spans instead.

mzabaluev: I have reworked the span use in `spawn_background_task`, it should look much better now.

relayer/src/worker/client.rs
```diff
@@ -52,40 +58,47 @@ pub fn detect_misbehavior_task<ChainA: ChainHandle, ChainB: ChainHandle>(
...
 
     let handle = spawn_background_task(
-        format!("DetectMisbehaviorWorker({})", client),
+        span!(
+            tracing::Level::ERROR,
+            "DetectMisbehaviorWorker",
+            client = %client.id,
+            src_chain = %client.src_chain.id(),
+            dst_chain = %client.dst_chain.id(),
+        ),
```
mzabaluev, line 88: This is a bit more repetitive than the previous approach, but this allows filtering logs by any of the individual ID fields.

If this level of detail is not desirable, we could implement `tracing::Value` on the objects that were logged with `Display` (removing the issue of implementing a general-purpose formatting trait just for logging purposes), and resurrect the previous compact string representation.

relayer/src/worker/client.rs
```diff
@@ -52,40 +58,47 @@ pub fn detect_misbehavior_task<ChainA: ChainHandle, ChainB: ChainHandle>(
...
 
     {
-        debug!("[{}] doing first misbehavior check", client);
+        let _span = span!(
+            tracing::Level::DEBUG,
+            "DetectMisbehaviorFirstCheck",
+            client = %client.id,
+            src_chain = %client.src_chain.id(),
+            dst_chain = %client.dst_chain.id(),
+        )
```
mzabaluev, line 74: This could be folded into the "DetectMisbehaviorWorker" span below, but I thought it would be useful to distinguish these contexts since the other one is performed in a background worker task.

relayer/src/link/relay_path.rs
mzabaluev: With the span rework, most of this information should be redundant.

relayer/src/util/task.rs
mzabaluev: > * We have redundant information now, however, since the path details `[ibc-0:transfer/channel-0 -> ibc-1]` appears twice on most lines, see eg. lines with "assembled batch of 2 message(s)". This information appears first from the `span` we just introduced, and second from `RelayPath:fmt::Display` impl ([example](https://github.com/informalsystems/ibc-rs/blob/85c60fa8f1b37e0f823aaa3cf03c79bafd8ef994/relayer/src/link/operational_data.rs#L179)). I think we should eliminate the second and keep the span only.

I have removed the `Display` impl for `RelayPath` and converted its uses into the same information laid out with spans.

adizere:
Mikhail and I had one more sync. session on this PR and we're both in agreement that this is ready for merging. It's a great first step in making the relayer's logs less grating on the eyes (and the codebase more idiomatic). The last nit is the changelog.

Cool stuff.

PR #1514, opened 2021-10-27
Output errors on a single line if backtraces are disabled

romac:
Closes: #1529

## Description

This uses our fork of [`simple-eyre`](https://crates.io/crates/simple-eyre), called [`oneline-eyre`](https://github.com/informalsystems/oneline-eyre).

---

### Output with `RUST_BACKTRACE = "full"`:

```
2021-10-27T22:00:13.861230Z ERROR ThreadId(27) [ibc-0] failed to collect events:
   0: WebSocket driver failed
   1:
   1:    0: web socket error: failed to read from WebSocket connection
   1:    1: WebSocket protocol error: Connection reset without closing handshake

   1: Location:
   1:    /Users/coromac/.cargo/registry/src/github.com-1ecc6299db9ec823/flex-error-0.4.4/src/tracer_impl/eyre.rs:10

   1:   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
   1:                                 ⋮ 10 frames hidden ⋮

...
```

### Output without `RUST_BACKTRACE` unset or set to `"0"` or `""`:

```
2021-10-27T22:01:37.228355Z ERROR ThreadId(27) [ibc-0] failed to collect events: WebSocket driver failed: web socket error: failed to read from WebSocket connection: WebSocket protocol error: Connection reset without closing handshake
```

______

For contributor use:

### DIFF
--- a/relayer-cli/src/bin/hermes/main.rs
+++ b/relayer-cli/src/bin/hermes/main.rs
@@ -6,17 +6,31 @@
 use ibc_relayer_cli::application::APPLICATION;
 use ibc_relayer_cli::components::enable_ansi;
 
-/// Boot Cli
 fn main() -> eyre::Result<()> {
-    if compact_logs() {
-        oneline_eyre::install()?;
-    } else if enable_ansi() {
-        color_eyre::install()?;
-    }
+    install_error_reporter()?;
 
     abscissa_core::boot(&APPLICATION);
 }
 
-fn compact_logs() -> bool {
-    matches!(std::env::var("RUST_BACKTRACE").as_deref(), Ok("" | "0"))
+fn install_error_reporter() -> eyre::Result<()> {
+    if !backtrace_enabled() {
+        // If backtraces are disabled, display errors in single line.
+        oneline_eyre::install()
+    } else if enable_ansi() {
+        // Else, if backtraces are enabled and we are in a terminal
+        // supporting color, display full error logs in color.
+        color_eyre::install()
+    } else {
+        // Otherwise, backtraces are enabled and we are piping to logs, so use the
+        // default error report handler, which displays multiline errors
+        // without color.
+        Ok(())
+    }
+}
+
+fn backtrace_enabled() -> bool {
+    match std::env::var("RUST_BACKTRACE").as_deref() {
+        Ok("" | "0") | Err(_) => false,
+        Ok(_) => true,
+    }
 }


mircea-c (approved):

soareschen:
This feels a bit overly restrictive that it would be impossible to capture full back trace in log files if it is ever needed. Should we only use the one-line reporter if `RUST_BACKTRACE` is `""` or not set?

relayer-cli/src/bin/hermes/main.rs
```diff
@@ -8,9 +8,15 @@ use ibc_relayer_cli::components::enable_ansi;
...
+        oneline_eyre::install()?;
+    } else if enable_ansi() {
         color_eyre::install()?;
     }
 
     abscissa_core::boot(&APPLICATION);
 }
+
+fn compact_logs() -> bool {
+    matches!(std::env::var("RUST_BACKTRACE").as_deref(), Ok("" | "0"))
```
soareschen, line 21: I think in this case if `RUST_BACKTRACE` is not set, the error case `Err(_)` will return false when it should be true.

This should be clearer with comment explaining what should happen in different branches:

```rust
fn main() -> eyre::Result<()> {
    if enable_ansi() {
        // If we are in a terminal supporting color, display
        // full error logs in color
        color_eyre::install()?;
    } else if !backtrace_enabled() {
        // Else if we are piping logs to file and backtrace is *not* enabled,
        // display errors in single line.
        oneline_eyre::install()?;
    } else {
        // Else backtrace is enabled and we are piping to logs, so use the
        // default error report handler, which displays multiline errors
        // without color.
    }

    abscissa_core::boot(&APPLICATION);
}

fn backtrace_enabled() -> bool {
    match std::env::var("RUST_BACKTRACE").as_deref() {
        Ok("" | "0") | Err(_) => false,
        Ok(_) => true,
    }
}
```

soareschen (changes requested):

relayer-cli/src/bin/hermes/main.rs
romac: Good catch, thanks! Shouldn't we invert the first two branches? ie. if backtraces are not enabled, we use one-line logs, otherwise if there is a tty we use colors, else we use the default reporter

soareschen: I think it depends on what we want the default behavior to be. Most of the time the relayer will run without `RUST_BACKTRACE` set, even for terminal cases. Do we want those to display one-line errors as well?

mircea-c: Most of the time the relayer will be piping to logs. So one line logs should be the default case.

soareschen (approved):

PR #1521, opened 2021-10-29
IBC Relayer Integration Test Suite

### LINKED ISSUE #797: Port the Python end-to-end integration test suite to Rust
## Crate

n/a

## Summary

Port the Python end-to-end integration test suite to Rust.

## Problem Definition

As the set of test in the end-to-end integration test suite is growing, we end up duplicating pretty much all the data structures found in the ibc-rs Rust codebase in Python. We could instead go the other way and rewrite the test suite in Rust in order to leverage all the existing data structures, code, etc.

## Proposal

TBD 

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate milestone (priority) applied

soareschen:
Closes #797 

## Description

A preview for the documentation is available here: https://lucid-hamilton-df13e0.netlify.app/ibc_integration_test/

______

For contributor use:


soareschen:
This PR is now ready for review and merge. A full documentation for writing relayer tests is available as Rustdoc, which you can view using `cargo doc`. A preview for the documentation is available here: https://lucid-hamilton-df13e0.netlify.app/ibc_relayer_test/

adizere:
Solid work Soares!

The same comment as I had for #1413 : could we move this new crate into a top-level directory called e.g., `tools`? Will follow-up a review!

relayer-test/src/chain/builder.rs
```diff
@@ -26,7 +26,7 @@ pub struct ChainBuilder {
     */
     pub command_path: String,
 
-    /*
+    /**
```
romac, line 29: Shouldn't these comments start with `///` ?

soareschen: `/**` also starts a doc comment block: https://doc.rust-lang.org/reference/comments.html

romac: Oh I didn't know about that, thanks! Any reason for picking block comments over line comments here?

soareschen: It is more convenient to always use multi-line comments so that when additional documentation needs to be added in the future, there is no need to re-edit the first line to use multi-line comment or use `///` on every line.

relayer-test/src/tests/memo.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    let memo_field = &tx_info["txs"][0]["tx"]["body"]["memo"];
+
+    info!("memo field value: {}", memo_field);
+
+    let memo_str = memo_field
+        .as_str()
+        .ok_or_else(|| eyre!("expect memo string field to be present in JSON"))?;
+
+    assert_eq!(memo_str, expected_memo);
```
adizere, line 86: Cool!

adizere (changes requested):

Cargo.toml
```diff
@@ -7,6 +7,7 @@ members = [
     "relayer",
     "relayer-cli",
     "relayer-rest",
+    "relayer-test",
```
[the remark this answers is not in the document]
soareschen: I think an issue for that is that `cargo build` will no longer automatically build `relayer-test`, and so code changes that break the test will go undetected by the developer until it goes into the CI. This has been a major annoyance for me working on the tendermint-rs project, as there are so many custom commands I need to run to find out if my code breaks anything, and I never managed to remember them.

relayer-test/Cargo.toml
```diff
@@ -0,0 +1,46 @@
+[package]
+name       = "ibc-relayer-test"
```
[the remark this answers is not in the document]
soareschen: I now call it `ibc-integration-test`. Right now I setup the `tracing_subscriber` in the test so that it only show logs from crates with the `ibc_` prefix, so the test crate itself also have to start with `ibc-`.

Cargo.toml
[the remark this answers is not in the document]
soareschen: I moved the crate into `tools/` but still keep it as a member in the root `Cargo.toml`. That way we avoid the pitfall in tendermint-rs where the projects in `tools/` are not built automatically.

soareschen:
I have added a command line wrapper `test_setup_with_binary_channel` that sets up the test chains as a CLI instead of test.

When the command is executed, you should see log messages such as
following near the end:

```bash
$ cargo run --bin test_setup_with_binary_channel
...
INFO ibc_integration_test::framework::binary::channel: written channel environment to /path/to/ibc-rs/data/test-3742758098/binary-channels.env
WARN ibc_integration_test::util::suspend: suspending the test indefinitely. you can still interact with any spawned chains and relayers
```

The `binary-channels.env` file generated contains the environment variables
that are essential for accessing the test chains. You can source it and
run the relayer commands in a separate terminal such as:

```bash
$ source /path/to/ibc-rs/data/test-1790156739/binary-channels.env
$ cargo run --bin hermes -- -c $RELAYER_CONFIG tx raw ft-transfer \
    $CHAIN_ID_B $CHAIN_ID_A $PORT_A $CHANNEL_ID_A 9999 -o 1000 \
    -k $NODE_A_WALLETS_USER1_KEY_ID -d $NODE_A_DENOM
```

More info at https://lucid-hamilton-df13e0.netlify.app/test_setup_with_binary_channel/index.html.

tools/integration-test/src/framework/nary/mod.rs
```diff
@@ -0,0 +1,13 @@
+/*!
+   Experimental work to define test cases with N-ary chains.
```
soareschen, line 2: I added the N-ary test chains implementation in this PR as well, but hide is behind an `"experimental"` flag so that we can still merge this even though it is not too ready for review yet due to documentation not yet being written.

Alternatively I can move the N-ary modules into a separate draft PR so that they can be reviewed separately.

soareschen:
I have just updated the bootstrap chain code so that the client/connection/channel IDs generated are random between 1-8. When you run the tests you should see the logs such as:

```
INFO ibc_integration_test::bootstrap::binary::channel: created new channel chain/client/connection/channel from ibc-alpha-a58a63ff/07-tendermint-3/connection-2/channel-4 to ibc-beta-c83842bf/07-tendermint-7/connection-3/channel-3
INFO ibc_integration_test::framework::binary::channel: written channel environment to /path/to/ibc-rs/tools/integration-test/data/test-2710114153/binary-channels.env
```

You can then run `source /path/to/ibc-rs/tools/integration-test/data/test-2710114153/binary-channels.env` to expose the environment variables on your terminal to further test on the relayer manually. The current exported environment variables look like:

```
CHAIN_ID_A=ibc-alpha-a58a63ff
CHAIN_ID_B=ibc-beta-c83842bf
CHANNEL_ID_A=channel-4
CHANNEL_ID_B=channel-3
CLIENT_ID_A=07-tendermint-3
CLIENT_ID_B=07-tendermint-7
CONNECTION_ID_A=connection-2
CONNECTION_ID_B=connection-3
NODE_A_CMD=gaiad
NODE_A_DENOM=coind964d90a
NODE_A_GRPC_ADDR=http://localhost:28406
NODE_A_HOME=/mnt/gamma/development/informal/ibc-rs-2/tools/integration-test/data/test-2710114153/ibc-alpha-a58a63ff
NODE_A_RPC_ADDR=http://localhost:49311
NODE_A_WALLETS_RELAYER_ADDRESS=cosmos1ntzvftcteeqdxmfa9ayzyeefem4sna3s84yctm
NODE_A_WALLETS_RELAYER_KEY_ID=relayer-45936844
NODE_A_WALLETS_USER1_ADDRESS=cosmos1mk42l0707a4jqnv3d7ne6hx6yu55nv4t3cd8ua
NODE_A_WALLETS_USER1_KEY_ID=user1-c3e5547
NODE_A_WALLETS_USER2_ADDRESS=cosmos1nsql3jg9hn5aqv7v9a7hhax5dr6y5l0geesh3g
NODE_A_WALLETS_USER2_KEY_ID=user2-699178b4
NODE_A_WALLETS_VALIDATOR_ADDRESS=cosmos14z9fyeyp5yfz7jh4t0w58stxll0uyveeezfpat
NODE_A_WALLETS_VALIDATOR_KEY_ID=validator-2e639555
NODE_B_CMD=gaiad
NODE_B_DENOM=coin86c992ce
NODE_B_GRPC_ADDR=http://localhost:26583
NODE_B_HOME=/mnt/gamma/development/informal/ibc-rs-2/tools/integration-test/data/test-2710114153/ibc-beta-c83842bf
NODE_B_RPC_ADDR=http://localhost:31734
NODE_B_WALLETS_RELAYER_ADDRESS=cosmos1xtm2x3x9xv2vg8n5cy3lr29nqzjjc77jxegfhe
NODE_B_WALLETS_RELAYER_KEY_ID=relayer-2335cf72
NODE_B_WALLETS_USER1_ADDRESS=cosmos1p2e6cvqv2f2u0hucn5rh2alap06mcwd7m2m0pl
NODE_B_WALLETS_USER1_KEY_ID=user1-227c4619
NODE_B_WALLETS_USER2_ADDRESS=cosmos15twlf3t6777qhwr0jpqfn7ukzc2rkvxv79ewuz
NODE_B_WALLETS_USER2_KEY_ID=user2-8e72a99e
NODE_B_WALLETS_VALIDATOR_ADDRESS=cosmos1zh66gag8nlq5mz8cs3h54mld5tv0r2r5y9ct2a
NODE_B_WALLETS_VALIDATOR_KEY_ID=validator-48c6bb76
PORT_A=transfer
PORT_B=transfer
RELAYER_CONFIG=/mnt/gamma/development/informal/ibc-rs-2/tools/integration-test/data/test-2710114153/relayer-config.toml
```

adizere:
> I have just updated the bootstrap chain code so that the client/connection/channel IDs generated are random between 1-8. When you run the tests you should see the logs such as:

Cool! I would propose we stop adding features here and focus on cleaning up & merging. New features to the testing framework would ideally be motivated directly by specific test cases.

romac (approved):
Incredible work, @soareschen! Such a beautiful infrastructure, and many thanks for the extensive comments and documentation :)

I have all but two very minor nitpicks:

1) `Binary` is used in a bunch of places to mean `two`, and given that `binary` is already an overloaded term in CS, what do you think of just using `two` instead? Though this may be moot after #1572 lands?

2) In the rest of the ibc-rs project, we use `///` as a doc comment prefix whereas most of the new code here uses `/**`, which introduces an inconsistency (and gives me Java flashbacks but that's not really the point). Can we harmonize this by switching over to `///`?

Again, amazing work!

PS: Happy to see `mononym` being put to good use already

tools/integration-test/src/relayer/chain.rs
```diff
@@ -0,0 +1,362 @@
...
+use ibc_proto::ibc::core::connection::v1::QueryClientConnectionsRequest;
+use ibc_proto::ibc::core::connection::v1::QueryConnectionsRequest;
+use ibc_relayer::chain::handle::{ChainHandle, ChainRequest, Subscription};
+use ibc_relayer::chain::{HealthCheck, StatusResponse};
+use ibc_relayer::config::ChainConfig;
+use ibc_relayer::error::Error;
+use ibc_relayer::{connection::ConnectionMsgType, keyring::KeyEntry};
+
+use crate::types::tagged::*;
+
```
romac, line 42: This module does not export any items, so perhaps we should hide it from the docs. Otherwise it appears empty in the documentation:
Screen Shot 2021-11-18 at 11 40 19

soareschen: I just added more documentation to it.

soareschen:
>     1. `Binary` is used in a bunch of places to mean `two`, and given that `binary` is already an overloaded term in CS, what do you think of just using `two` instead? Though this may be moot after #1572 lands?

Yeah I agree that using the word "binary" is not very fitting here, though I can't think of a better term especially considering that we need a good correspondence to the 'N-ary" setup as well. 

The most common place where test writers are going to use it is through the traits like `BinaryChannelTest` or `NaryChannelTest`. Do you think referring them as `TwoChannelTest` and `MultiChannelTest` would be better?
 
>     2. In the rest of the ibc-rs project, we use `///` as a doc comment prefix whereas most of the new code here uses `/**`, which introduces an inconsistency (and gives me Java flashbacks but that's not really the point). Can we harmonize this by switching over to `///`?

I would be against using `///` for doc comments, as they are very inconvenient for writing new comments, and even more annoying for editing existing comments. If I need to wrap around comment lines due to line limits, the `///` blocks would get in the way of me combining the two lines. In the long term this would just discourage developers to write or edit comments, which is bad for the project.

> Again, amazing work!
> 
> PS: Happy to see `mononym` being put to good use already

Thanks! Actually tagged identifiers are different from named values in `mononym`. In the case of ibc-rs, the tags we use are non-unique and are only there to prevent accidental mix ups. But the tagged constructs does not prevent programmers from assigning the wrong tags to the data types, and so there are still chance for mistakes to be made.

PR #1523, opened 2021-10-30
Increase max length for port identifier to 128

litvintech:
Following ibc-go spec the default length for port identifier is 2-128 

[ibc-go/PortIdentifierValidator](https://github.com/cosmos/ibc-go/blob/main/modules/core/24-host/validate.go#L86)

Tested using hermes(0.7.3) during cw20-ics20 transfer where port_id was wasm.bostrom14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9sww4mxt

Closes: #XXX

## Description

______

For contributor use:

### DIFF
--- a/modules/src/core/ics24_host/validate.rs
+++ b/modules/src/core/ics24_host/validate.rs
@@ -61,10 +61,10 @@ pub fn validate_connection_identifier(id: &str) -> Result<(), Error> {
 
 /// Default validator function for Port identifiers.
 ///
-/// A valid Identifier must be between 2-64 characters and only contain lowercase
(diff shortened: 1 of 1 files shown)


adizere:

This is an important change we'd like to support in Hermes. We'd like maintain backwards compatibility, however, with chains running on ibc-go version <1.2 (which have port id limited to 64 characters). We'd also like to emit a warning/error when two chains with incompatible port limits try to handshake on a new channel. We plan to do some of these changes in PR #1438, so switching the base branch for your work to the developer branch for that PR, hope that makes sense.

adizere (approved):
LGTM!

PR #1531, opened 2021-11-02
Fix broken links in changelog

romac:


hu55a1n1 (approved):
Thanks, @romac! 👍 
Should we update these links in the epilogue.md as well?

romac:
> Thanks, @romac! 👍
> Should we update these links in the epilogue.md as well?

Good point, done!

Issue #1518, opened 2021-10-28
Allow more granular configuration of Hermes modes of operation


ancazamfir:
## Crate
relayer

## Summary
Allow more granular configuration of hermes client operation.
[cc @mankenavenkatesh]

## Problem Definition
[Principal problem is described in #1489. This issue is a first step over one track towards fixing #1489. There will be others, e.g. migration to async, that will deal with performance.]
- client misbehaviour worker impacts performance of the packet relaying
- there are applications that require live packets to be relayed in real-time while old packets can be left to timeout, clear at a later stage, etc. An example is injective chains that send requests periodically to get token pricing information 

## Proposal
Allow the user to specify at a lower granularity the mode of operation for a given instance. For this issue we want to be able to specify if client workers should run misbehaviour and/or periodic client refresh.
After an initial discussion with @romac and @adizere, the proposal here is that we take this opportunity and remove `strategy` and add new configuration section `mode` (or something else). We also add a clients and packets sections to the configuration.
Ideally the config will be changed to something like:

```
[global]
log_level = 'info'

[mode]

[mode.clients]
enabled = true
refresh = true
misbehaviour = true

[mode.connections]
enabled = false

[mode.channels]
enabled = false

[mode.packets]
enabled = true
clear_interval = 100
clear_on_start = true 
filter = false
tx_confirmation = true

[rest]
...

[telemetry]
...
```

- for example hermes instance that runs refresh and misbehavior only
```
[mode]

[mode.clients]
enabled = true
refresh = true
misbehaviour = true

[mode.connections]
enabled = false

[mode.channels]
enabled = false

[mode.packets]
enabled = false

```

## Acceptance Criteria
Ability to disable client misbehavior.
Ability to not clear packets on start.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #1539, opened 2021-11-03
Allow more granular configuration of Hermes modes of operation

hu55a1n1:
Closes: #1518 

## Description
* Default config for `mode` would look like -> 
```toml
[mode]

[mode.clients]
enabled = true
refresh = true
misbehaviour = true

[mode.connections]
enabled = false

[mode.channels]
enabled = false

[mode.packets]
enabled = true
clear_interval = 100
clear_on_start = true
filter = false
tx_confirmation = true
```

* `strategy` and `clear_packets_interval` options would be removed. 
* ...

#### TODO (aside from acceptance criteria)
- [x] Guide changes.
- [x] Fix E2E tests.
- [x] Fix `gm` related stuff. 
______

For contributor use:


relayer/src/supervisor.rs
```diff
@@ -171,6 +171,28 @@ impl<Chain: ChainHandle + 'static> Supervisor<Chain> {
         }
     }
 
+    fn collect_event<F>(
```
romac, line 174: 😍

relayer/src/worker.rs
```diff
@@ -107,8 +114,9 @@ impl<ChainA: ChainHandle + 'static, ChainB: ChainHandle + 'static> Worker<ChainA
                     path.clone(),
                     chains,
                     cmd_rx,
-                    config.global.clear_packets_interval,
-                    config.global.tx_confirmation,
+                    config.mode.packets.clear_interval,
+                    config.mode.packets.clear_on_start,
+                    config.mode.packets.tx_confirmation,
```
romac, line 119: Same here.

relayer/src/worker.rs
```diff
@@ -91,9 +91,16 @@ impl<ChainA: ChainHandle + 'static, ChainB: ChainHandle + 'static> Worker<ChainA
...
-                Self::Client(id, ClientWorker::new(client.clone(), chains, cmd_rx))
-            }
+            Object::Client(client) => Self::Client(
+                id,
+                ClientWorker::new(
+                    client.clone(),
+                    chains,
+                    cmd_rx,
+                    config.mode.clients.misbehaviour,
+                    config.mode.clients.refresh,
```
romac, line 101: How about we pass the `Packets` struct directly in order to avoid potential issues with [boolean blindness](https://github.com/ostera/reason-design-patterns/blob/master/patterns/boolean-blindness.md)?

relayer/src/worker/packet.rs
```diff
@@ -26,6 +26,7 @@ pub struct PacketWorker<ChainA: ChainHandle, ChainB: ChainHandle> {
     chains: ChainHandlePair<ChainA, ChainB>,
     cmd_rx: Receiver<WorkerCmd>,
     clear_packets_interval: u64,
+    clear_packets_on_start: Option<bool>,
```
romac, line 29: Maybe we should define an `enum` to give more semantic meaning to the three possible states here?

eg.
```rust
enum OnStart {
  ClearPackets,
  DoNothing,
}

impl Default for OnStart {
  fn default() -> Self {
    Self::DoNothing
  }
}
```

and map

```rust
None => OnStart::default()
Some(true) => OnStart::ClearPackets
Some(false) => OnStart::DoNothing
```

romac (commented):
Amazing stuff @hu55a1n1 🥇

Here are a few comments for discussion after a first pass over the code.

relayer/src/worker/packet.rs
hu55a1n1: Nice idea! I'll add this. 👍

relayer/src/worker.rs
hu55a1n1: https://github.com/informalsystems/ibc-rs/pull/1539/commits/a12be44bd4852b74c0d5fb936d6ff9021e96f96a

relayer/src/worker.rs
hu55a1n1: https://github.com/informalsystems/ibc-rs/pull/1539/commits/a12be44bd4852b74c0d5fb936d6ff9021e96f96a

config.toml
```diff
@@ -1,13 +1,45 @@
...
-#   - 'all': Relay packets and perform channel and connection handshakes.
-#   - 'packets': Relay packets only.
-strategy = 'packets'
+# Specify the verbosity for the relayer logging output. Default: 'info'
+# Valid options are 'error', 'warn', 'info', 'debug', 'trace'.
+log_level = 'info'
+
+
+# Specify the mode to be used by the relayer. [Required]
+[mode]
```
hu55a1n1, line 10: Should we add a test that matches `Config::default()` against this file?

relayer-cli/src/config.rs
```diff
@@ -68,6 +75,19 @@ pub fn validate_config(config: &Config) -> Result<(), Error> {
...
+    Ok(())
+}
+
+fn validate_mode(mode: &ModeConfig) -> Result<(), Error> {
+    if mode.clients.enabled && !mode.clients.refresh && !mode.clients.misbehaviour {
+        return Err(Error::invalid_mode(
+            "either `refresh` or `misbehaviour` must be set to true if `clients.enabled` is set to true".to_string(),
+        ));
+    }
+
```
hu55a1n1, line 90: Would the following be considered valid config? 
```toml
[mode]

[mode.clients]
enabled = false

[mode.connections]
enabled = false

[mode.channels]
enabled = false

[mode.packets]
enabled = false
```

relayer/src/worker/packet.rs
```diff
@@ -129,18 +123,18 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> PacketWorker<ChainA, ChainB> {
                     height,
                     new_block: _,
                 } => {
-                    let (first_run, clear_on_start) = {
-                        match self.clear_packets_on_start.take() {
-                            None => (false, false),
-                            Some(clear_on_start) => (true, clear_on_start),
-                        }
+                    let clear_on_start = if self.packets_cfg.clear_on_start {
+                        self.packets_cfg.clear_on_start = false;
```
romac, line 127: I'd rather we maintain a `first_run` variable or field somewhere to make this behavior a bit more explicit if that makes sense?

hu55a1n1: https://github.com/informalsystems/ibc-rs/pull/1539/commits/891251130a421c9146182f30fb399cee5367a53e

relayer/src/worker/packet.rs
```diff
@@ -25,37 +26,45 @@ pub struct PacketWorker<ChainA: ChainHandle, ChainB: ChainHandle> {
...
             chains,
             cmd_rx,
-            clear_packets_interval,
-            with_tx_confirmation,
+            packets_cfg,
+            first_run: true,
+        }
+    }
+
+    fn clear_packets_on_start(&mut self) -> bool {
```
romac, line 49:
```suggestion
    fn clear_packets(&mut self) -> bool {
```

adizere:
@greg-szabo we would appreciate your quick look here, at the `gm` related changes.

greg-szabo:
Looks good to me. The only thing missing is the gm Changelog update but I can deal with that separately in my own PR.

config.toml
romac: That would be nice to have but I don't think it's necessary, also it might cause problems if devs modify this file when testing etc.

relayer-cli/src/config.rs
romac: > @ancazamfir: it’s valid and actually interesting to just watch what’s going on. I would use this :slightly_smiling_face: But a warning is definitely required.

ancazamfir:
@hu55a1n1 I tried with a config that has:
```
[mode.packets]
# Whether or not to enable the packet workers. [Required]
enabled = true
# Parametrize the periodic packet clearing feature.
# Interval (in number of blocks) at which pending packets
# should be eagerly cleared. A value of '0' will disable
# periodic packet clearing. [Default: 100]
clear_interval = 10
# Whether or not to clear packets on start. [Default: false]
clear_on_start = true
```
all other modes disabled.
Create a channel, send some packets then start hermes. The packets are not cleared immediately. Also not cleared on the `clear_interval`.  If I send a packet after hermes start then everything is back to normal.
Didn't look yet at the code but I think it should be straight forward to fix.

Also  e2e  fails with this  config.

hu55a1n1:
> The packets are not cleared immediately. Also not cleared on the clear_interval.

Thanks for catching that @ancazamfir! I think this was a bug in the way packet workers were spawned. I just pushed a possible fix -> 4ec957b

relayer-cli/src/config.rs
romac: https://github.com/informalsystems/ibc-rs/pull/1539/commits/2e3bca3cc9617f102cfff02a56b1aa4bc0264f19

relayer/src/worker/client.rs
```diff
@@ -49,7 +53,8 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> ClientWorker<ChainA, ChainB> {
         );
 
         // initial check for evidence of misbehaviour for all updates
-        let skip_misbehaviour = self.detect_misbehaviour(&client, None);
+        let skip_misbehaviour =
```
ancazamfir, line 56: Just above is this log:
```
        info!(
            "[{}] running client worker & initial misbehaviour detection",
            client
        );
```
can we drop "initial misbehavior detection" part if misbehavior is not configured? log "refresh" and "misbehavior" depending on the config.

ancazamfir (commented):
Tested a bunch of scenarios, it looks very good!! 
I will look again through the changes next but I think we should be good to go.

relayer/src/worker/client.rs
romac: https://github.com/informalsystems/ibc-rs/pull/1539/commits/db3b6a5a11eff6f85f7f5f3ee64d4e87bc092a87

romac (approved):
Amazing work @hu55a1n1 🌷

ancazamfir (approved):
Amazing work @hu55a1n1!

Issue #987, opened 2021-05-25
Deny uses of `unwrap` and `panic!` in non-test code (second pass)


adizere:
## Crates

In order of priority:

- [ ] `ibc`
- [ ] `ibc-relayer`
- [ ] `ibc-telemetry`
- [ ] `ibc-rest`
- [ ] `ibc-relayer-cli`

## Summary

Follow-up to https://github.com/informalsystems/ibc-rs/issues/871, which took care of some of unwraps in crates `ibc` and `relayer`.

## Acceptance Criteria

Enable `#![cfg_attr(not(test), deny(clippy::unwrap_used))]`

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #1540, opened 2021-11-03
Clean up some unwraps in `relayer`

seanchen1991:
Addresses: #987 

## Description

Remove usages of `unwrap` from non-test code in the `ibc` crate. The focus is on cleaning up unwraps in `relayer/src/chain/cosmos.rs` and `relayer/src/chain/runtime.rs`, as these are the cause of crashes in chain runtimes. A crashed chain runtime has ripple effects in all of the of workers that are holding a handle to the runtime. 
______

For contributor use:

### DIFF
--- a/relayer/src/chain/cosmos.rs
+++ b/relayer/src/chain/cosmos.rs
@@ -308,7 +308,7 @@ impl CosmosSdkChain {
         };
 
         let mut tx_bytes = Vec::new();
-        prost::Message::encode(&tx_raw, &mut tx_bytes).unwrap();
+        prost::Message::encode(&tx_raw, &mut tx_bytes).map_err(|e| Error::protobuf_encode(String::from("Transaction"), e))?;
 
         let response = self.block_on(broadcast_tx_sync(self, tx_bytes))?;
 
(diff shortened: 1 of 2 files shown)


relayer/src/chain/cosmos.rs
```diff
@@ -308,7 +308,7 @@ impl CosmosSdkChain {
         };
 
         let mut tx_bytes = Vec::new();
-        prost::Message::encode(&tx_raw, &mut tx_bytes).unwrap();
+        prost::Message::encode(&tx_raw, &mut tx_bytes).map_err(|e| Error::protobuf_encode(String::from("Transaction"), e))?;
```
adizere, line 311: I guess it's very tricky if not impossible to reproduce this error? I'd be curious to see how it propagates up the call stack and at which level do we retry (it should be ideally as low-level as possible).

adizere (commented):
Thanks Sean! What does the PR miss to mark it ready for review?

I left one question, but it's not a blocking concern.

relayer/src/chain/cosmos.rs
romac: It's not clear to me how that can even happen, would have to dig through the `prost` code to figure it out. Regardless, we wan't to avoid crashing Hermes if that happens and therefore have to bubble up the error here. We don't act on this error in any specific way so this will just trigger retries in the workers, and very likely fail repeatedly so it will be seen in the logs. As we won't ever get to submitting a tx, it won't drain the wallet either which makes the current solution acceptable for now. What do you think?

romac (approved):
Thanks @seanchen1991 🌻

Issue #1479, opened 2021-10-19
Hermes resubmits already received packets


mircea-c:
## Crate

ibc-rs

## Summary of Bug

When gas estimation fails with `packet sequence (142099) already has been received` Hermes will submit the transaction using `max_gas` anyway.

## Version

git hash: `f328157ae87a1ca90b960e5b8a9004a2d720c629`

## Steps to Reproduce

Run two instances of `hermes start` submit one transaction to be relayed. The following log entries are seen:
```
Oct 19 10:09:17.860 DEBUG ThreadId(42) [cosmoshub-4] simulated gas: Err(
    0: #033[91mGRPC call return error status status: InvalidArgument, message: "failed to execute message; message index: 1: receive packet verification failed: packet sequence (142099) already has been received: invalid packet: invalid request", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }#033[0m
 
 Location:
    #033[35m/usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/flex-error-0.4.3/src/tracer_impl/eyre.rs#033[0m:#033[35m10#033[0m
 
 Backtrace omitted.
 Run with RUST_BACKTRACE=1 environment variable to display it.
 Run with RUST_BACKTRACE=full to include source snippets.)
 Oct 19 10:09:17.860 ERROR ThreadId(42) [cosmoshub-4] failed to estimate gas, falling back on max gas, error: 
    0: #033[91mGRPC call return error status status: InvalidArgument, message: "failed to execute message; message index: 1: receive packet verification failed: packet sequence (142099) already has been received: invalid packet: invalid request", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }#033[0m
 
 Location:
    #033[35m/usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/flex-error-0.4.3/src/tracer_impl/eyre.rs#033[0m:#033[35m10#033[0m
 
 Backtrace omitted.
 Run with RUST_BACKTRACE=1 environment variable to display it.
 Run with RUST_BACKTRACE=full to include source snippets.
 Oct 19 10:09:17.872 DEBUG ThreadId(42) [cosmoshub-4] send_tx: broadcast_tx_sync: Response { code: Ok, data: Data([]), log: Log("[]"), hash: transaction::Hash(98BA2278A493254F0DF38734E1B0CADD5253AF4FBC78B10F7C38CB3AF392EF9C) }
[...]

PR #1542, opened 2021-11-04
Avoid submitting txs when simulation fails, except for one special case

romac:
Closes: #1479
Related: #712

## Description

This PR changes the way the relayer reacts to tx simulation failures in the following way:
- if the tx simulation failed because the client state height is too low, the relayer will fallback on the default gas setting and attempt to submit the tx anyway.

   > @ancazamfir: https://github.com/informalsystems/ibc-rs/pull/1542#issuecomment-961766790
   > Regarding the first case, it's been a lot on my mind as we see this very often in the packet worker. It happens when we have to split N msgs in a number of Tx-es with n (aka max_msg_num) msgs each, with n < N and when the first message is UpdateClient.
    > We may have the worker determine n from the config and prepend the UpdateClient before every batch of n-1 packets. But by the time it gets serialized and into the runtime, it might have to split again due to max_tx_size. The other possibility that we discussed before is to not serialize messages at all and let the runtime ensure that each batch in the Tx starts with UpdateClient.
    > I think that if we could fix this then we can drop the Tx on all simulation errors.

- if the tx simulation failed for any other reason, we abort the submission of the tx and bubble the error up, and let the worker deal with the error. At the moment, workers do not handle these errors in any specific manner, and will just attempt to send the tx again for a fixed number of times. In that case, the overall behavior of the relayer does not change, except that we do not waste fees unnecessarily anymore because we won't attempt to actually submit the tx using the default gas.

## How to test

1. Create a channel

   ```
   ❯ hermes create channel ibc-0 ibc-1 --port-a transfer --port-b transfer
   ```

2. In the Hermes config, set `max_msg_num = 5` for chain `ibc-0`

3. Start Hermes

   ```
   ❯ hermes start
   ```

4. Send N > 5 packets to `ibc-0`

   ```
   ❯ hermes tx raw ft-transfer ibc-0 ibc-1 transfer channel-0 9999 -o 1000 -n 10 -k user2
   ```

5. Check the `hermes start` logs, you should find something like:

   ```
   2021-11-05T14:12:09.633184Z  WARN ThreadId(30) [ibc-1] estimate_gas: failed to simulate tx, falling back on default gas because the error is potentially recoverable: gRPC call failed with status: status: InvalidArgument, message: "failed to execute message; message index: 0: acknowledge packet verification failed: packet acknowledgement verification failed: failed packet acknowledgement verification for client (07-tendermint-0): client state height < proof height ({0 243} < {0 554}): invalid height: invalid request", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }
   2021-11-05T14:12:09.633290Z DEBUG ThreadId(30) [ibc-1] send_tx: using 900000000 gas, fee Fee { amount: "900000stake", gas_limit: 900000000 }
   2021-11-05T14:12:09.639044Z DEBUG ThreadId(30) [ibc-1] send_tx: broadcast_tx_sync: Response { code: Ok, data: Data([]), log: Log("[]"), hash: transaction::Hash(BA94AE4CA198F56E27B4A44DA5E508A2E2207E306F475E5285D873296D892170) }
   ```

______

For contributor use:

### DIFF
--- a/relayer/src/chain/cosmos.rs
+++ b/relayer/src/chain/cosmos.rs
@@ -286,13 +286,13 @@ impl CosmosSdkChain {
             signatures: vec![signed_doc],
         };
 
-        let estimated_gas = self.estimate_gas(simulate_tx)?;
-        let adjusted_fee = self.fee_with_gas(estimated_gas);
+        let simulated_gas = self.simulate_gas(simulate_tx)?;
+        let adjusted_fee = self.fee_with_gas(simulated_gas);
 
         debug!(
             "[{}] send_tx: using {} gas, fee {}",
             self.id(),
-            estimated_gas,
+            simulated_gas,
             PrettyFee(&adjusted_fee)
         );
 
@@ -340,10 +340,10 @@ impl CosmosSdkChain {
     /// [`MsgUpdateClient`, `MsgRecvPacket`, ..., `MsgRecvPacket`]
     /// if the batch is split in two TX-es, the second one will fail the simulation in `deliverTx` check
     /// In this case we use the `default_gas` param.
-    fn estimate_gas(&self, tx: Tx) -> Result<u64, Error> {
-        let estimated_gas = self.send_tx_simulate(tx).map(|sr| sr.gas_info);
+    fn simulate_gas(&self, tx: Tx) -> Result<u64, Error> {
+        let simulated_gas = self.send_tx_simulate(tx).map(|sr| sr.gas_info);
 
-        if let Ok(ref gas) = estimated_gas {
+        if let Ok(ref gas) = simulated_gas {
             debug!(
                 "[{}] send_tx: tx simulation successful, simulated gas: {:?}",
                 self.id(),
@@ -351,7 +351,7 @@ impl CosmosSdkChain {
             );
         }
 
-        let estimated_gas = estimated_gas.map_or_else(
+        let simulated_gas = simulated_gas.map_or_else(
             |e| {
                 error!(
                     "[{}] send_tx: failed to estimate gas, falling back on default gas, error: {}",
@@ -364,17 +364,17 @@ impl CosmosSdkChain {
             |gas_info| gas_info.map_or(self.default_gas(), |g| g.gas_used),
         );
 
-        if estimated_gas > self.max_gas() {
-            debug!(estimated = ?estimated_gas, max = ?self.max_gas(), "[{}] send_tx: estimated gas is higher than max gas", self.id());
+        if simulated_gas > self.max_gas() {
+            debug!(simulated = ?simulated_gas, max = ?self.max_gas(), "[{}] send_tx: simulated gas is higher than max gas", self.id());
 
             return Err(Error::tx_simulate_gas_estimate_exceeded(
                 self.id().clone(),
-                estimated_gas,
+                simulated_gas,
                 self.max_gas(),
             ));
         }
 
-        Ok(estimated_gas)
+        Ok(simulated_gas)
     }
 
     /// The default amount of gas the relayer is willing to pay for a transaction,
--- a/relayer/src/chain/cosmos.rs
+++ b/relayer/src/chain/cosmos.rs
@@ -200,7 +200,7 @@ impl CosmosSdkChain {
         // Check that the configured max gas is lower or equal to the consensus params max gas.
         let consensus_max_gas = result.consensus_params.block.max_gas;
 
-        // If the consensus max gas is < 0, this check does not make sense.
+        // If the consensus max gas is < 0, we don't need to perform the check.
         if consensus_max_gas >= 0 {
             let consensus_max_gas: u64 = consensus_max_gas
                 .try_into()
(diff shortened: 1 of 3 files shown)


adizere:
Feedback so far: love the PR title! 💯 

Quick idea we can iterate on:

> if the tx simulation failed because the chain sent an account sequence mismatch error, the relayer will fallback on the default gas setting and attempt to submit the tx anyway. This is because we have seen cases in production where this happened, for reasons which are yet to be clarified.

Since we noticed this special case in production, I wonder if it makes sense to quickly parse some production logs: (1) grep for `sequence mismatch error` at the simulation step, and (2) derive a rough success vs. failure ratio for the submitted transactions. If many tx-es succeed, the solution here is sound! Otherwise we can think about revising the approach. Would that make sense?

ancazamfir:
A few thoughts:
- we have to continue for errors that include `..verification failed: ...client state height < proof height `:
```
2021-11-04T13:28:00.277866Z ERROR ThreadId(26) [ibc-0] send_tx: failed to estimate gas, falling back on default gas, error: GRPC call return error status status: InvalidArgument, message: "failed to execute message; message index: 0: acknowledge packet verification failed: packet acknowledgement verification failed: failed packet acknowledgement verification for client (07-tendermint-0): client state height < proof height ({1 323} < {1 400}): invalid height: invalid request", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }
```
- for account sequence mismatch we need to re-cache but with the understanding that there is still no guarantee that we get it right as other Tx-es may be in the mempool. The only cases I know where this happens are:
    - the wallet is being used outside of this hermes instance
    - a previous Tx has passed CheckTx but dropped (e.g. by the mempool) before DeliverTx has a chance to run. One example is  when we send max gas > block max gas
Continuing with sending the Tx will not solve this I think.

- for all other errors is probably good to just drop the Tx with error like you propose

Regarding the first case, it's been a lot on my mind as we see this very often in the packet worker. It happens when we have to split `N` msgs in a number of Tx-es with `n` (aka `max_msg_num`) msgs each, with `n < N` and when the first message is `UpdateClient`. 
We may have the worker determine `n` from the config and prepend the `UpdateClient` before every batch of `n-1` packets. But by the time it gets serialized and into the runtime, it might have to split again due to `max_tx_size`. The other possibility that we discussed before is to not serialize messages at all and let the runtime ensure that each batch in the Tx starts with `UpdateClient`. 
I think that if we could fix this then we can drop the Tx on all simulation errors. 

So in summary I thinking we should consider:
- fix first issue
- recache account sequence on `SDK 32` error 
- drop Tx for ALL simulation errors

relayer/src/chain/cosmos.rs
```diff
@@ -336,45 +336,58 @@ impl CosmosSdkChain {
...
-    /// they do not individually verify. For example for the following batch
+    /// they do not individually verify. For example for the following batch:
     /// [`MsgUpdateClient`, `MsgRecvPacket`, ..., `MsgRecvPacket`]
-    /// if the batch is split in two TX-es, the second one will fail the simulation in `deliverTx` check
+    ///
+    /// If the batch is split in two TX-es, the second one will fail the simulation in `deliverTx` check.
     /// In this case we use the `default_gas` param.
-    fn estimate_gas(&self, tx: Tx) -> Result<u64, Error> {
-        let estimated_gas = self.send_tx_simulate(tx).map(|sr| sr.gas_info);
+    fn simulate_gas(&self, tx: Tx) -> Result<u64, Error> {
```
ancazamfir, line 344: Not sure why the change? The Tx is being "simulated" and it returns the `gas_used`. This is a gas estimation at this point as when the actual Tx is executed the gas might be slightly different.

relayer/src/chain/cosmos.rs
```diff
@@ -336,45 +336,58 @@ impl CosmosSdkChain {
...
+                    e.detail()
+                );
 
-            return Err(Error::tx_simulate_gas_estimate_exceeded(
-                self.id().clone(),
-                estimated_gas,
-                self.max_gas(),
-            ));
+                Err(e)
+            }
```
ancazamfir, line 389: We should still check that the estimated gas is smaller than the configured max gas, no?

romac: https://github.com/informalsystems/ibc-rs/pull/1542/commits/fedd6c6af9c51f84c7d55075073a3024cecf77ae & https://github.com/informalsystems/ibc-rs/pull/1542/commits/fedd6c6af9c51f84c7d55075073a3024cecf77ae

relayer/src/chain/cosmos.rs
```diff
@@ -197,6 +197,24 @@ impl CosmosSdkChain {
...
+                .try_into()
+                .expect("cannot over or underflow because it is positive");
+
+            if self.max_gas() > consensus_max_gas {
+                return Err(Error::config_validation_max_gas_too_high(
+                    self.id().clone(),
+                    self.max_gas(),
+                    result.consensus_params.block.max_gas,
+                ));
+            }
```
adizere, line 215: Excellent!

adizere (approved):
This looks like solid work Romain, I'm still testing and will merge soon!

Issue #1555, opened 2021-11-09
relayer fee estimation amount may overflow


tomtau:
## Crate

relayer

## Summary of Bug

The fee amount is estimated using u64: https://github.com/informalsystems/ibc-rs/blob/9d1f86b9e7bdb06a3fb2c5b3ed05fc8004650784/relayer/src/chain/cosmos.rs#L2286

This may not be work in certain networks and it will result in transactions not being relayed due to an insufficient fee being paid.

## Version

v0.8.0

## Steps to Reproduce

1. network A: start up a Cosmos SDK-based network and configure its fee token to have many decimal places (e.g. 15) and configure a non-zero gas price
2. configure Hermes to create and relay on a channel to a network B (it can be anything)
3. start sending many IBC transfers of any coin from the network B to the network A

## Acceptance Criteria

Large batches of transactions being relayed successfully on networks with fee tokens that have higher decimal places, such as networks working with Ethermint where a gas token has by default 15 decimal places.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1568, opened 2021-11-15
Prevent overflow in fee amount computation

romac:
Closes: #1555 

## TODO

- [x] Write unit test showcasing the potential overflow

## Description

Prevent a potential overflow when computing the fee for a tx to be sent to a chain which uses a denomination which high decimal places.

This is done by using infinite precision integers and rationals in the `mul_ceil` function. The resulting `BigInt` is then formatted as a string in the `Fee` proto.

Because the computation is now done will full precision, and because some decimal numbers cannot be represented precisely as a 64-bit IEEE-754 floating point number, the result of the computation may sometime change, eg. be rounded up to the next integer. This should not be a problem as the gas and fee computations are not fully precise anyway, given that they rely on the simulated gas.

______

For contributor use:

### DIFF
--- a/relayer/src/chain/cosmos.rs
+++ b/relayer/src/chain/cosmos.rs
@@ -452,10 +452,10 @@ impl CosmosSdkChain {
     /// The actual gas cost, when a transaction is executed, may be slightly higher than the
     /// one returned by the simulation.
     fn apply_adjustment_to_gas(&self, gas_amount: u64) -> u64 {
-        debug_assert!(self.gas_adjustment() <= 1.0);
+        assert!(self.gas_adjustment() <= 1.0);
 
         let (_, digits) = mul_ceil(gas_amount, self.gas_adjustment()).to_u64_digits();
-        debug_assert!(digits.len() == 1);
+        assert!(digits.len() == 1);
 
         let adjustment = digits[0];
         let gas = gas_amount.checked_add(adjustment).unwrap_or(u64::MAX);
@@ -2388,7 +2388,7 @@ fn calculate_fee(adjusted_gas_amount: u64, gas_price: &GasPrice) -> Coin {
 
 /// Multiply `a` with `f` and round the result up to the nearest integer.
 fn mul_ceil(a: u64, f: f64) -> BigInt {
-    debug_assert!(f.is_finite());
+    assert!(f.is_finite());
 
     let a = BigInt::from(a);
     let f = BigRational::from_float(f).expect("f is finite");
--- a/relayer/src/chain/cosmos.rs
+++ b/relayer/src/chain/cosmos.rs
@@ -2007,8 +2007,7 @@ fn packet_from_tx_search_response(
         .tx_result
         .events
         .into_iter()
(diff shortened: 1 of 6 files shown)


adizere (approved):
Ran several tests and looks good!

relayer/src/chain/cosmos.rs
```diff
@@ -449,10 +452,15 @@ impl CosmosSdkChain {
     /// The actual gas cost, when a transaction is executed, may be slightly higher than the
     /// one returned by the simulation.
     fn apply_adjustment_to_gas(&self, gas_amount: u64) -> u64 {
-        min(
-            gas_amount + mul_ceil(gas_amount, self.gas_adjustment()),
-            self.max_gas(),
-        )
+        debug_assert!(self.gas_adjustment() <= 1.0);
```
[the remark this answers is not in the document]
romac: I am thinking we should perhaps make this a standard `assert!` since if this invariant does not hold then it means we have a logic issue in config validation, which we probably want to hear about in production. What do you think?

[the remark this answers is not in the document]
mzabaluev: My approach to `debug_assert` is to check violations that will not break the functionality, but can be a tripwire for finding bugs in development.

relayer/src/chain/cosmos.rs
```diff
@@ -2380,13 +2387,12 @@ fn calculate_fee(adjusted_gas_amount: u64, gas_price: &GasPrice) -> Coin {
 }
 
 /// Multiply `a` with `f` and round to result up to the nearest integer.
-fn mul_ceil(a: u64, f: f64) -> u64 {
-    use fraction::Fraction as F;
+fn mul_ceil(a: u64, f: f64) -> BigInt {
+    debug_assert!(f.is_finite());
```
romac, line 2391:
```suggestion
    assert!(f.is_finite());
```

relayer/src/chain/cosmos.rs
```diff
@@ -449,10 +452,15 @@ impl CosmosSdkChain {
...
     /// one returned by the simulation.
     fn apply_adjustment_to_gas(&self, gas_amount: u64) -> u64 {
-        min(
-            gas_amount + mul_ceil(gas_amount, self.gas_adjustment()),
-            self.max_gas(),
-        )
+        debug_assert!(self.gas_adjustment() <= 1.0);
+
+        let (_, digits) = mul_ceil(gas_amount, self.gas_adjustment()).to_u64_digits();
+        debug_assert!(digits.len() == 1);
```
romac, line 458:
```suggestion
        assert!(digits.len() == 1);
```

romac, line 458:
```suggestion
        assert!(digits.len() == 1);
```

relayer/src/chain/cosmos.rs
romac, line 455: Agreed, so as this will break functionality if not upheld, it should be a runtime assertion in release mode as well.

```suggestion
        assert!(self.gas_adjustment() <= 1.0);
```

PR #1572, opened 2021-11-16
N-ary Test Chains Support

### LINKED ISSUE #1577: Support N-ary chains in integration test
## Crate

`ibc-integration-test`

## Summary

- Follow up on #1521 to allow tests to setup more than 2 chains.
- Write a test case for 2-hops IBC transfer between 3 chains using the N-ary test framework.

____

#### For Admin Use

- [ ] Not duplicate issue

soareschen:
Closes #1577 and #1776

## Description

This is a follow up on #1521 to allow tests to setup N-ary test chains using [const generics](https://doc.rust-lang.org/stable/reference/items/generics.html#const-generics)

Documentation for how to use the N-ary constructs indexed by const generics: https://informalsystems.github.io/ibc-rs-doc/ibc_integration_test/types/nary/index.html

______

For contributor use:


mzabaluev:
Conflicts need to be resolved.

tools/integration-test/src/chain/driver.rs
```diff
@@ -456,7 +456,7 @@ impl ChainDriver {
     ) -> Result<(), Error> {
         assert_eventually_succeed(
             "wallet reach expected amount",
-            20,
+            40,
```
soareschen, line 459: It seems like there is some performance regression when the relayer is setup with multiple channels. I have to increase the timeout limit for `assert_eventual_wallet_amount` to > 20s for the IBC transfer test to not fail.

soareschen:
I have implemented a ternary transfer test `test_ternary_ibc_transfer`, which implements the trait `NaryChannelTest<3>` and test 2-hops and 3-hops IBC transfer.

Another notable change is that the trait `SupervisorOverride` is now removed, with test cases required to spawn the supervisor manually using either `RelayerDriver::spawn_supervisor` or `RelayerDriver::with_supervisor`. This is because in N-ary chains setup, the relayer picks up new channel events exponentially slower as more chains/connections/channels are added.

The original setup had `BinaryChainTest` and `NaryChainTest` spawn the supervisor at the base layer first, before higher layers like `NaryChannelTest` spawn the connections and channels. Because the connections and channels are spawned after the relayer is started, it somehow slows down exponentially in particular in the Nary setup. On the other hand, if the relayer is spawned manually by individual tests after all channels have been setup, packet relaying can be start pretty much immediately.

With the relayer started manually by tests, it is also simpler as compared to having to explicitly disable spawning of the supervisor in `TestOverrides` in tests such as `test_supervisor`.

soareschen:
This branch is currently crashing in Rust Analyzer caused by https://github.com/rust-analyzer/rust-analyzer/issues/11659.

soareschen:
This branch is now working with the pre-release version of Rust Analyzer. You'll have to explicitly enable that in the VSCode extension for it to work:

Screenshot from 2022-03-10 17-45-48
 
This PR is blocked from merging until new official release of Rust Analyzer.

@adizere it might also be a good idea for you to test this on IntelliJ. If it doesn't work there, we'd have to switch back to always having const generic parameters at the last position.

romac (approved):

.github/workflows/cargo-doc.yaml
```diff
@@ -19,7 +19,7 @@ jobs:
       - run: cargo doc
 
       - name: Deploy
-        if: ${{ github.ref == 'ref/head/master' }}
+        if: ${{ github.ref == 'refs/heads/soares/nary-test-chains' }}
```
romac, line 22: Let's revert this before merging?

soareschen: Right, actually I keep having problem finding the correct expression to trigger this step. It is not triggering on master as well.

soareschen: I have reverted the branch to master. It probably still won't work on master, but we can figure out a fix in a separate PR.

Issue #1561, opened 2021-11-10
Allow users to specify `ProofSpecs` for non-SDK chains


hu55a1n1:
## Crate
ibc

## Summary
Currently, we use the Cosmos SDK `ProofSpecs` definition for all Tendermint chains. We should allow users to specify the `ProofSpecs` for non-SDK chains.

## Proposal
We maybe add a config option `[[chains]].proof_spec` that takes a file/text with a serialized (JSON?) `ProofSpec` definition. 

## Acceptance Criteria
Users are able to specify a custom `ProofSpecs` for non-SDK chains
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #853, opened 2021-04-26
Duplicate proto files for ICS23


adizere:
## Crate

`ibc` and `ibc-proto`

## Problem Definition

Currently, the canonical place for the .proto file of ICS23 is [confio/ics23](https://github.com/confio/ics23/blob/master/proofs.proto). The Cosmos-SDK and IBC-go repos both copy the `proofs.proto` file from confio/ics23 as a "third party proto". The `ibc-proto` crate similarly copies the same `proofs.proto` file from Cosmos-SDK. In general, the state of .proto files is quite fractured, but ICS23 is even more so.

This fracturing creates the problem that there are two version of the ICS23 specs:

- in `ibc_proto` we have [this](https://github.com/informalsystems/ibc-rs/blob/master/proto/src/prost/ics23.rs)
- in the "canonical" confio/ics23 there is [this](https://github.com/confio/ics23/blob/master/rust/src/ics23.rs)

The two files are not identical, since the confio/ics23 version has been recently changed, and the updates did not propagate downstream into CosmosSDK nor into ibc-proto (yet). Even if the files were on par, there would still be two legit definitions of the same data structures, for every structure in `ics23.rs`.

## Proposal

Ideally, crate `ibc` should depend on a single .proto definition of the ICS23 files.

## Acceptance Criteria

- The fix is to avoid generating `ics23.rs` in `ibc-proto` crate and rely entirely on confio/ics23. This means that [ibc_proto::ics23](https://github.com/informalsystems/ibc-rs/blob/master/proto/src/prost/ics23.rs) should not exist anymore.

The long-term fix is to have all .proto files related to IBC stored in a single unified place, e.g., repo cosmos/ibc. This is a larger eco-system problem, however, which we don't need to track with this issue.

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #1574, opened 2021-11-17
Custom proof specs in config

hu55a1n1:
Closes: #1561 

## Description
Proof specs are currently read as JSON serialized TOML strings from the config. `Serialize/Deserialize` are derived on the `ibc_proto::ics23` types. 

______

For contributor use:


modules/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -52,6 +51,7 @@ impl ClientState {
         max_clock_drift: Duration,
         latest_height: Height,
         frozen_height: Height,
+        proof_specs: Vec<ibc_proto::ics23::ProofSpec>,
```
hu55a1n1, line 54: Should we instead write our own impls of `Serialize/Deserialize` for `modules/src/core/ics23_commitment/specs.rs::ProofSpecs` (i.e. our domain type) and use that instead of the proto types?

romac: Yeah I think we should have domain types for all types which we expose publicly somehow, either via the API or to users in the config.

hu55a1n1: https://github.com/informalsystems/ibc-rs/pull/1574/commits/3c42a9e9a6e3a41663a58e3de391e21003697f4d

modules/src/core/ics23_commitment/specs.rs
```diff
@@ -1,6 +1,7 @@
 use crate::prelude::*;
 use ibc_proto::ics23::ProofSpec as ProtoProofSpec;
 use ics23::ProofSpec;
```
hu55a1n1, line 3: Just FYI, with regard to #853 - this file (i.e. `modules/src/core/ics23_commitment/specs.rs`) is the only place where we use the `ics23` crate. We only use it to get the definitions for the proof-specs for cosmos and to provide `convert::From` impls for  our domain type (i.e. `ProofSpecs`). For all other cases, we use the `ibc_proto::ics23` defs because we need to derive serde and other traits for these types.

modules/src/core/ics23_commitment/specs.rs
```diff
@@ -44,3 +49,26 @@ impl From<ProofSpecs> for Vec<ProtoProofSpec> {
         raw_specs
     }
 }
+
+impl From<Vec<ProofSpec>> for ProofSpecs {
+    fn from(proto_specs: Vec<ProofSpec>) -> Self {
+        let mut specs = Vec::new();
+        for ds in proto_specs {
+            // Both `ProofSpec` types implement trait `prost::Message`. Convert by encoding, then
```
hu55a1n1, line 57: Regarding the safety of these unwraps - it is quite possible that the `ics23` crate could remove fields that are marked redundant in future versions (e.g. https://github.com/confio/ics23/blob/master/proofs.proto#L77). 
Should we use `TryFrom` instead?

modules/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -99,6 +100,7 @@ impl ClientState {
             max_clock_drift,
             frozen_height,
             latest_height,
+            proof_specs,
```
hu55a1n1, line 103: Should we allow empty proof-specs?

romac: I am not sure I understand what this would entail? Can you elaborate on what happens if we use empty proof-specs?

modules/src/core/ics23_commitment/specs.rs
romac: I think it would be cleaner to implement the conversion by hand rather than rely on the proto encoding, this way we would catch breaking changes at compile time rather than at runtime with `TryFrom`. What do you think?

proto/Cargo.toml
```diff
@@ -27,6 +27,7 @@ prost-types = "0.9"
 bytes = "1.1"
 tonic = "0.6"
 getrandom = { version = "0.2", features = ["js"] }
+serde = "1.0.130"
```
romac, line 30:
```suggestion
serde = "1"
```

relayer/src/config.rs
```diff
@@ -405,6 +408,42 @@ pub struct ChainConfig {
...
+
+    pub fn serialize<S: Serializer>(
+        proof_specs: &ProofSpecs,
+        serializer: S,
+    ) -> Result<S::Ok, S::Error> {
+        let json_str = serde_json::to_string(proof_specs).map_err(ser::Error::custom)?;
+        serializer.serialize_str(&json_str)
+    }
+
+    struct ProofSpecsVisitor;
```
romac, line 424: Very nice :)

modules/src/core/ics23_commitment/specs.rs
mzabaluev: Apart from the gratuitous dynamism, the encode-decode roundtrip can be also a performance wart.

proto/Cargo.toml
mzabaluev: > will avoid dependabot bothering us with updates.

It will anyway because the repo has `Cargo.lock` checked in.
Regardless, there is no need to depend on a semver patch version unless you want to make sure that e.g. a bugfix in the dependency is addressed since that specific version.

relayer/src/config.rs
```diff
@@ -405,6 +408,42 @@ pub struct ChainConfig {
...
+    use core::fmt;
+    use ibc::core::ics23_commitment::specs::ProofSpecs;
+    use serde::{de, ser, Deserializer, Serializer};
+
+    pub fn serialize<S: Serializer>(
+        proof_specs: &ProofSpecs,
+        serializer: S,
+    ) -> Result<S::Ok, S::Error> {
+        let json_str = serde_json::to_string(proof_specs).map_err(ser::Error::custom)?;
+        serializer.serialize_str(&json_str)
```
mzabaluev, line 421: So this is forced to be a JSON string embedded into any serialized format, human-readable or not?

When the JSON serializer is used, this is will end up as a string with escaped JSON inside. Forgive the newbie, but I'd rather like to know a good reason why the serialization can't be structurally composed. Is it because we can't lock down the serialization schema for `ProofSpecs` in the embedding format?

modules/src/clients/ics07_tendermint/client_state.rs
hu55a1n1: Thinking about this again - an empty `ProofSpecs` implies that the chain doesn't have a `ProvableStore` and therefore cannot provide proofs for the data it stores. This is a requirement in ICS024, so I don't think we should allow empty proof-specs.

proto/Cargo.toml
hu55a1n1: Good point! I guess one of us copied the version from crates.io (which always shows the full semver), and we stuck to it for consistency. I just set it to `1.0` for all Cargo.toml's in the repo. https://github.com/informalsystems/ibc-rs/pull/1574/commits/a72ab391e34d00aab88fe307322f79edec0f69d2

relayer/src/config.rs
hu55a1n1: > So this is forced to be a JSON string embedded into any serialized format, human-readable or not?

Right, but this (i.e. `serde(with=proof_specs_serde)`) is only used in the config which is serialized as TOML. 

> When the JSON serializer is used for an outer structure, this is will end up as a string with escaped JSON inside. 

Nice point, I modified the serializing code to serialize as a pretty-printed string, and the TOML serializer defaults to pretty strings (represented with `'''`) to serialize strings with newlines in them. Also tested it locally.

modules/src/clients/ics07_tendermint/client_state.rs
romac: Makes sense, agreed!

modules/src/core/ics23_commitment/specs.rs
hu55a1n1: Done in https://github.com/informalsystems/ibc-rs/pull/1574/commits/12a46a47e233ee6b57d8d69a8017e70d901ec759. 👍 We still need to handle the case where a new field was added to one of the proto-types and the other wasn't updated. 
Should we add an assert that makes sure the size of the structs matches (not sure that will work if the proto structs have different type layouts i.e. `repr`s)?

modules/src/clients/ics07_tendermint/client_state.rs
hu55a1n1: https://github.com/informalsystems/ibc-rs/pull/1574/commits/a229708fdde2510f87913c4f7570bd4634439bb6

relayer/src/config.rs
mzabaluev: The serde impls are agnostic to the serializer and they are part the public API, so there's no stopping anyone from serializing in any other format. I agree it wouldn't make sense for the config to be serialized in something non-readable by humans, though.

To be clear, this does not just delegate to the `ProofSpecs`' serde impls to get it emitted as native JSON in JSON (in case someone wants to have a JSON config), because the primary use case is TOML and it probably would look horrible, unless you have a way to tell the toml-rs serializer to switch to inline JSON at the level of that value, or is there some other reason?

mzabaluev: Sheesh, I forgot that TOML is actively trying not to become YAML, and as such does not support inline JSON.

hu55a1n1: Sorry, but I'm not sure I understand the concern here. The rationale for choosing to (de)serialize using `serde(with=proof_specs_serde)` was that `ProofSpecs` encoded as TOML would look ugly (and take a lot of space). Although `ProofSpecs` is in the public API, the `proof_specs_serde` module is private and meant to be used only for (de)serializing `ProofSpecs` within the `Config`.
Note that `ProofSpecs` also implements (derives) `Serialize`/`Deserialize` itself for other use cases.

modules/src/core/ics23_commitment/specs.rs
romac: I am not sure what you mean by "the case where a new field was added"? Wouldn't we catch this at compile-time whenever we update either the protos or the ics23 crate?

relayer/src/config.rs
hu55a1n1: Aahaa, I get it now! Not sure what would be the best way to solve this (I think a proper solution would require specialization). Found this related issue -> [Serializer-specific struct serialization
](https://github.com/serde-rs/serde/issues/968). Do you have any ideas on how we can solve this problem? 
One good thing is that all components of the config are public themselves, so users can compose their own config types. Also, AFAIK the relayer crate is only meant to be used by other packages in the workspace.

mzabaluev: I think it's not really solvable within serde because it's designed to be largely agnostic about the serialization format (`is_human_readable` being the only concession).

It may be not such a big problem if the crate is not meant to be widely used.

Otherwise, the public API could hide the serde impls by implementing them on private structs, and lock down TOML as the only possible serialization format, but this requires a lot of newtype wrappers, accessor methods and so on. I don't think it's worth it.

romac (approved):

config.toml
```diff
@@ -223,6 +223,60 @@ memo_prefix = ''
 #   ['transfer', 'channel-0'],
 # ]
 
+# Specify custom ICS23 proof-specs for the chain (serialized as JSON).
+# Default: [`ProofSpecs::cosmos()`](modules/src/core/ics23_commitment/specs.rs)
+proof-specs = '''
```
adizere, line 228: Is there a more operator-friendly manner to specify this? I tried to give an alternative suggestion in #1630 .

Issue #1063, opened 2021-06-08
Spawn event monitor only when needed


romac:
## Crate

`ibc-relayer`

## Summary

Spawn the event monitor only for commands which need it.

## Problem Definition

Commands which do not need the event monitor shouldn't need to incur the penalty of spawning it and subscribing to queries.

## Proposal

Spawn the event monitor only for commands which need it.

## Acceptance Criteria

When commands which do not need the event monitor do not spawn it.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #1575, opened 2021-11-17
Improve performance of standalone commands by starting the event monitor on-demand

romac:
Closes: #1063

## Description

Lazily spawn the event monitor and subscribe to events over WebSocket only when `ChainHandle::subscribe()` is called, which improves the startup time of commands which do not need to subscribe to WS events.

______

For contributor use:


relayer/src/chain/runtime.rs
```diff
@@ -1,5 +1,5 @@
 use alloc::sync::Arc;
-use std::thread;
+use std::{option::Option, thread};
```
romac, line 2:
```suggestion
use std::thread;
```

mzabaluev:
Needs rebasing.

relayer/src/event/monitor/error.rs
```diff
@@ -0,0 +1,62 @@
...
+            |_| { "RPC error" },
+    }
+}
+
+impl Error {
+    pub fn canceled_or_generic(e: RpcError) -> Self {
+        use tendermint_rpc::error::ErrorDetail;
+
+        match e.detail() {
+            ErrorDetail::Server(detail) if detail.reason.contains("subscription was cancelled") => {
```
mzabaluev, line 56: Can we demand a more type-checked way to identify this reason from `tendermint-rpc`?
Matching by string is liable to silently stopping to work, unless the tendermint-rpc developers commit to stable reason strings as part of their API.

romac: This error is actually emitted as a string by the Cosmos SDK, so there's not much we can do here aside from moving that very same logic into `tendermint-rpc`.

mzabaluev (approved):
Looks good, other than issues outside of this project's control (I've got to raise an issue against cosmos-sdk).

Issue #134, opened 2020-07-08
Relayer commands don't support help menu for subcommands


ebuchman:
## Summary of Bug
When running `relayer query --help` I expect to see the list of subcommands but I just get:

```
$ relayer query --help
error: unrecognized command `--help`

relayer-cli 0.0.1
Anca Zamfir <anca@interchain.io>, Romain Ruetschi <romain@informal.systems>

FLAGS:
    -c, --config CONFIG       path to configuration file
    -h, --help                print help message
    -v, --verbose             be verbose

```

## Version

v0.0.1

## Steps to Reproduce

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

andynog:
I've noticed this today when testing the query and mentioned on the group call. I'll take a look at that and ensure this works.

andynog:
@ebuchman, I've done some investigation on this one. Haven't been able to solve it yet because was just trying to understand how abscissa works, and how it uses gumdrop for command line. I believe the way our code is currently implemented it should work (show usage for subcommands). Unfortunately the docs for abscissa and gumdrop are minimal on this topic. I'm more familiar with clap and structopt. 

But found some issues that might point that help usage for subcommands might be a limitation

- an old issue that might be related to this (it's closed but the current behaviour seems to point to this)
https://github.com/iqlusioninc/abscissa/pull/187

There are also a couple of issues open related to this

- upgrade gumdrop
https://github.com/iqlusioninc/abscissa/issues/250

- Or replace gumdrop with clap (I really like this idea, I found gumdrop to be very limited compared to clap)
https://github.com/iqlusioninc/abscissa/issues/298

ebuchman:
Ah, looks like abscissa is going in the direction of clap. Maybe we should wait for / help with that. I guess we put this on hold until then?

adizere:
Running this today (from targer/debug) works for me. The error message is gone.

```
adi@styx:ibc-rs $ target/debug/relayer --help
relayer-cli 0.0.1
Anca Zamfir <anca@interchain.io>, Romain Ruetschi <romain@informal.systems>

USAGE:
    relayer-cli <SUBCOMMAND>

SUBCOMMANDS:
    help       get usage information
    start      start the relayer
    listen     listen to IBC events
    config     manipulate the relayer configuration
    version    display version information
    query      query state from chain
[...]

Issue #609, opened 2021-02-04
Hermes Guide: easier maintenance work


adizere:
## Summary

The guide we keep at hermes.informal.systems is pretty amazing.

For easier maintenance, some ideas for future work include:

- [ ] modify the ci action to unpin from the dev branch

https://github.com/informalsystems/ibc-rs/blob/5dc062707b2f5f34722e203f26b74658646b76c0/.github/workflows/guide.yml#L3-L6

- [ ] include a versioning scheme, so we can access a specific version of the guide, for example

`hermes.informal.systems/v1/`
`hermes.informal.systems/v2/`

Note that the current versioning scheme is very crude and it's [just a remark in the guide readme](https://github.com/informalsystems/ibc-rs/blob/master/guide/README.md) file.

- [ ] design a staging area, so that we're keeping multiple instances of the guide online: one that reflects whatever is in the `master` (e.g., via uri `hermes.informal.systems/master/`), and the other versions for all published (tagged) versions. It would be amazing if we can automate the tagging process to deploy new versions out of master.

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

romac:
> modify the ci action to unpin from the dev branch

https://github.com/informalsystems/ibc-rs/pull/610

andynog:
Did some preliminary investigation on how to support versioning with mdbook. There might be a few options. 

Is the idea to have some sort of a drop down box on the documentation site so users can switch between versions (something similar to cargo docs site) ?

romac:
That would be nice feature to have but it's not mandatory imho.

andynog:
I think supporting that will be too complex at the moment with mdBook, probably would need to create a pre-processor or add-in to support that. The easiest way is to have maybe separate jobs in CI (like one to publish the guide when master is changed and push to gh-pages branch) and other that pushes the staging to another place so we can test the updates to documentation on a different place.

adizere:
Beside versioning, I found this principled approach to doing documentation: https://documentation.divio.com.
Would be good if we try to apply some of the ideas in there, in particular these [dimensions](https://documentation.divio.com/introduction/#the-secret)

PR #1576, opened 2021-11-18
Update to abscissa 0.6.0-beta.1 and clap 3.0.0-beta.5

mzabaluev:
Closes: #134

## Description

Overhaul command line argument processing to use the `clap_derive` macro instead of `gumdrop`.

## Caveats

Usage of flags like `-h` and `--version` in subcommands is ambiguous with the flags automatically supported by clap for conventional help and version printouts. This is highlighted by FIXME comments and clap settings disabling the automatically generated flags for these select subcommands. `--help` is also not available for the subcommands that use `-h` for something else, because the clap setting can only suppress both long and short help flags.

gumdrop is still used by tendermint-testgen, so it's not removed from the dependency tree (there's one version less of it now, though)

______

For contributor use:


mzabaluev:
Problems to be resolved:

- [ ] It panics on any valid command that is not internally supported by clap, with some hoo-ha about setting the color theme twice.
- [ ] The `listen` flag `--event` does not work the intended way due to a current limitation in clap: https://github.com/clap-rs/clap/issues/1772

adizere:
I had a cursory look over the PR. Great job so far, Mikhail!

I have a question regarding dependency management: is it appropriate for `ibc-rs` to release a version where we pin on a `-beta` dependency, concretely for `abscissa` and `clap`? I'm thinking in terms of risks of API instability of these beta versions. We'd like to avoid depending on APIs that shift too much, but not sure if this could be a problem here. WDYT?

mzabaluev:
> I have a question regarding dependency management: is it appropriate for `ibc-rs` to release a version where we pin on a `-beta` dependency, concretely for `abscissa` and `clap`? I'm thinking in terms of risks of API instability of these beta versions.

As these are dependencies of the `hermes` executable and some tests, but not of the libraries, and the exact versions are locked in `Cargo.lock`, it's probably OK.

relayer-cli/src/bin/hermes/main.rs
```diff
@@ -5,28 +5,6 @@
...
-    if !backtrace_enabled() {
-        // If backtraces are disabled, display errors in single line.
-        oneline_eyre::install()
-    } else {
-        // Otherwise, backtraces are enabled, so use the
-        // default error report handler. The impl for
-        // Application::framework_components takes care about
-        // initialization of color_eyre.
-        Ok(())
-    }
```
mzabaluev: @roman Sorry, this clobbers your work in #1514 because I could not prevent a panic in the new abscissa trying to set up the global eyre handler (which it probably should not do: https://github.com/iqlusioninc/abscissa/issues/589)

BTW, that issue is wrongly linked in the unclog entry as #1515.

mzabaluev: Reverted this now, but suppressed the color in abscissa terminal components, whatever that does.

epage:
I came across this from the backlink in clap 1772.  I'm curious, any feedback on clap in making this transition?

mzabaluev:
> I came across this from the backlink in clap 1772. I'm curious, any feedback on clap in making this transition?

Overall, It's been smooth and, as previously experienced with `clap` 0.2/`structopt`, we got a lot of extra polish and convenience for free.

The derive macro is not documented in detail in the `clap` crate, so I had to extrapolate from the `structopt` docs. 

One badly needed fix is https://github.com/clap-rs/clap/issues/1772. I had to look into the code of clap-derive to understand how to manually write the `Args` and `FromArgMatches` impls in order to get `multiple_occurrences(true)` where it was needed. These traits are scantily documented and perhaps not intended to be implemented by hand. 

It would be nice to have settings to disable only the small or the long flag for version and help: we've got subcommands that use `-h` for something else, but `--help` could still be auto-generated on the same command. Likewise there's one subcommand with a long `--version` flag, but that should arguably be renamed to unambiguously refer to the subject entity; also, clap-provided `--version` is not supported on subcommands by default.

Thank you @epage for your hard work!

epage:
> The derive macro is not documented in detail in the `clap` crate, so I had to extrapolate from the `structopt` docs.

Yes, this will be addressed before 3.0.

> One badly needed fix is [clap-rs/clap#1772](https://github.com/clap-rs/clap/issues/1772). I had to look into the code of clap-derive to understand how to manually write the `Args` and `FromArgMatches` impls in order to get `multiple_occurrences(true)` where it was needed.

Mind providing feedback on what you want to see from the derives?  I have https://github.com/clap-rs/clap/pull/2993 as one proposal.

> These traits are scantily documented and perhaps not intended to be implemented by hand.

I've been trying to move in this direction, trying to meet the goals:
- crates like [clap-verbosity-flag](https://github.com/rust-cli/clap-verbosity-flag) are usable by non-derive users
- For build-time conscience, allow developers of crates like the above to hand-implement the traits so builder users don't have to pull in the proc macro machinery.

All the breaking changes to meet these goals are complete.  The rest is in documentation which is a lower priority than just getting the derives out in a stable release.

Any feedback on what you want in documentation is welcome!

> It would be nice to have settings to disable only the small or the long flag for version and help: we've got subcommands that use `-h` for something else, but `--help` could still be auto-generated on the same command.

Whats the use case for a non-help `-h`?  Seems like `-h` is too tightly associated with help to reuse.

As a hack, you could re-assign the short with  `mut_arg("help", |a| a.short('t'))`.  I've been contemplating having more functions take `impl Into<Option<T>>` so you can clear things make to default, like `App::help_heading`

> also, clap-provided `--version` is not supported on subcommands by default.

Its been a while since I've dug into this part, is this just an issue of [setting version](https://github.com/clap-rs/clap/issues/3034) or setting [PropagateVersion](https://docs.rs/clap/3.0.0-beta.5/clap/enum.AppSettings.html#variant.PropagateVersion)

mzabaluev:
@epage

> Mind providing feedback on what you want to see from the derives? I have clap-rs/clap#2993 as one proposal.

I support the changes outlined there. The main thing is, a `Vec` should be treated with `multiple_occurrences(true)`.

There may be also a need to override the special type treatment and let the user provide a custom accumulating parser/validator for the strings produced by `ArgMatches::values_of` and friends. This could be also useful for user-supplied collection types that the macro does not recognize as special. But this could be done later.

> Any feedback on what you want in documentation is welcome!

I think basic examples of hand-written implementation for `Args` and `FromArgMatches` would help a lot.

> Whats the use case for a non-help `-h`? Seems like `-h` is too tightly associated with help to reuse.

Well, folks here have been using it as the short for `--height` without giving much thought to it :grin: But I agree, this could be construed as being at odds with the common convention, and the non-help short option should be renamed to `-H` to avoid ambiguity.

> is this just an issue of [setting version](https://github.com/clap-rs/clap/issues/3034) or setting [PropagateVersion](https://docs.rs/clap/3.0.0-beta.5/clap/enum.AppSettings.html#variant.PropagateVersion)

Sorry, I was trying to say that the application-specified `--version` in a subcommand was not a problem in this case as the automatic top-level `--version` is not propagated. Still, the chosen name is confusing and IMHO `--version` should be reserved to just the version printout case.

romac:
> But I agree, this could be construed as being at odds with the common convention, and the non-help short option should be renamed to -H to avoid ambiguity.

👍

romac:
> Sorry, I was trying to say that the application-specified --version in a subcommand was not a problem in this case as the automatic top-level --version is not propagated. Still, the chosen name is confusing and IMHO --version should be reserved to just the version printout case.

@mzabaluev Can you point me to the place where we rely on `--version` being available in a subcommand?

mzabaluev:
> Can you point me to the place where we rely on `--version` being available in a subcommand?

Sure, I have annotated it with a FIXME: https://github.com/informalsystems/ibc-rs/blob/c33820497a319779c4de4c52a0f0038298be1722/relayer-cli/src/commands/create/channel.rs#L60

guide/src/commands/path-setup/channels.md
```diff
@@ -24,7 +24,7 @@ FLAGS:
     --port-a PORT-A           identifier of the side `a` port for the new channel
     --port-b PORT-B           identifier of the side `b` port for the new channel
     -o, --order ORDER         the channel ordering, valid options 'unordered' (default) and 'ordered'
-    -v, --version VERSION     the version for the new channel
+    -v, --channel-version VERSION     the version for the new channel
```
romac, line 27:
```suggestion
    -v, --version VERSION     the version for the new channel
```

guide/src/help.md
```diff
@@ -74,10 +74,13 @@ FLAGS:
     --port-a PORT-A           identifier of the side `a` port for the new channel
     --port-b PORT-B           identifier of the side `b` port for the new channel
     -o, --order ORDER         the channel ordering, valid options 'unordered' (default) and 'ordered'
-    -v, --version VERSION     the version for the new channel
+    -v, --channel-version VERSION     the version for the new channel
```
romac, line 77:
```suggestion
    -v, --version VERSION     the version for the new channel
```

guide/src/help.md
```diff
@@ -74,10 +74,13 @@ FLAGS:
...
 ```
 
-The `help` command is a replacement of the familiar `-h`/ `--help` flag typical for CLI applications.
+Additionally, the `-h`/`--help` flags typical for CLI applications work on
+most commands (with the exception of some commands currently using `-h` for
+other purposes; these should be changed to follow the convention in a
+future version).
```
romac, line 83:
```suggestion
The `help` command is a replacement of the familiar `-h`/ `--help` flag typical for CLI applications.
```

romac (changes requested):
Let's revert the guide updates and do them in the release PR, otherwise the guide will be out of sync with the latest published version of the relayer since it's published automatically from `master`, cf. https://github.com/informalsystems/ibc-rs/issues/609

romac (approved):
@mzabaluev Thank you so much for stepping up to this tedious task 🌻

mzabaluev:
@romac No problem, some light hazing for a newcomer :grin:

Issue #1487, opened 2021-10-25
Refactor the ChainHandle trait to make it more chain agnostic


hu55a1n1:
## Crate
ibc-relayer

## Summary
Refactor ChainHandle trait to make it more chain agnostic

## Problem Definition
The ChainHandle trait methods can be improved and cleaned to provide a more chain agnostic API ->

- https://github.com/informalsystems/ibc-rs/pull/1409#discussion_r731046696 (@ancazamfir 's comment) -> 
> I think these changes show the bad assumptions in the base code (not your fault) given that this should be chain agnostic. [...]
So instead of having query_blocks and query_txs there should be a single query_packet_data() function that is implemented by the chain specific impl and cosmos-sdk chain should query txs and blocks, order the events, etc.
Along these lines, if we get rid of query_txs we will need two additional methodsquery_client_header()
and query_transaction_by_hash(). Again we expect this to make sense for any chain.

- https://github.com/informalsystems/ibc-rs/pull/1409#discussion_r733689063 (@adizere's comment) -> 
> For the request variant QueryPacketEventDataFromTxs both the chain handle- and the chain runtime-methods are called query_blocks

Also see https://github.com/informalsystems/ibc-rs/pull/1482 for initial attempt at replacing `query_blocks()` and parts of `query_txs()` with `query_packet_data()`. 

## Acceptance Criteria
TDB
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #624, opened 2021-02-09
Remove dependency on `tendermint-rpc` in `ibc` crate


romac:
## Crate

`abc`

## Summary

Remove dependency on `tendermint-rpc` in `ibc` crate by moving the `get_all_events` function into the relayer.

## Problem Definition

The `ibc`  crate should not imply a dependency on `tendermint-rpc` and all its transitive dependencies, especially since the function that introduces the dependency is only use in the `ibc-relayer` crate.

## Proposal

- Move  `get_all_events`  into `ibc-relayer`.
- Make `tendermint-rpc` a test-only dependency.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1582, opened 2021-11-22
Upgrade to IBC-go v2

### LINKED ISSUE #1410: Support dynamic version strings in ChanOpen handshake [ibc-go v2]
## Crate

`ibc-relayer`

## Summary

Add support in Hermes for dynamic channel version negotiation for use in channel open handshakes, with application to ICS-027.

## Problem Definition

Whenever Hermes is building one of the first three messages of a channel's handshake (init, try, ack), it must also relay a version parameter which is an application-defined string. The relevant part of the specification is [here](https://github.com/cosmos/ibc/tree/master/spec/core/ics-004-channel-and-packet-semantics#versioning). The ICA protocol ([ics-027](https://github.com/cosmos/ibc/tree/master/spec/app/ics-027-interchain-accounts)) uses the channel version to pass the interchain account between the host chain and the controller chain. Issue https://github.com/cosmos/ibc-go/issues/378 describes accurately what Hermes is supposed to do, in particular:

> a relayer will need to query the host chain for an updated version string prior to calling the OnChanOpenAck endpoint

Further details are provided in this comment: https://github.com/cosmos/ibc-go/issues/136#issuecomment-897512989

The actual implementation of the new query (which Hermes will need to rely on) is in this PR: https://github.com/cosmos/ibc-go/pull/384

## Proposal

The changes we need to do are localized here:

https://github.com/informalsystems/ibc-rs/blob/8d4d73bf6aa35a6f9111bf6cfbf08f23f2ffe7c6/relayer/src/chain.rs#L224-L233

There is also a _note_ with respect to this problem at a call site in the relayer's channel object implementation.

https://github.com/informalsystems/ibc-rs/blob/8d4d73bf6aa35a6f9111bf6cfbf08f23f2ffe7c6/relayer/src/channel.rs#L655-L656

## Acceptance Criteria

- [x] Modified channel handshake implementation to rely on the new gRPC for supporting dynamic version
- [x] Take note of forward and backwards compatibility
    - Take note that closing this issue will depend (e.g., for testing) on ibc-go v2: https://github.com/cosmos/ibc-go/blob/v2.0.0-beta1/CHANGELOG.md
    - Concrete branch to use for testing: https://github.com/cosmos/gaia/pull/1021

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

adizere:
Closes #1410 

## Description

______

For contributor use:


relayer-cli/src/commands/tx/channel.rs
```diff
@@ -152,53 +153,6 @@ pub struct TxRawChanOpenTryCmd {
 
 impl Runnable for TxRawChanOpenTryCmd {
     fn run(&self) {
-        let config = app_config();
```
adizere, line 155: Deleted this lines because it was a coding error. It's enough to call into `tx_chan_cmd!` macro, we don't need these manual steps since the macro handles everything.

relayer/src/channel.rs
```diff
@@ -224,18 +222,20 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
                 connection_id,
                 port_id,
                 channel_id,
+                // The event does not include the version.
+                // The message handlers `build_chan_open..` determine the version from channel query.
+                // TODO(Adi): Investigate proto correctness wrt version.
```
romac, line 227: Reminder for this TODO

relayer/src/channel.rs
```diff
@@ -282,16 +282,18 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
                 a_connection_id.clone(),
                 channel.src_port_id.clone(),
                 Some(channel.src_channel_id.clone()),
+                // TODO(Adi): Investigate proto correctness.
```
romac, line 285: Reminder for this TODO

relayer/src/channel/version.rs
```diff
@@ -0,0 +1,135 @@
...
+    match ctx {
+        ResolveContext::ChanOpenInit => {
+            // When the channel is uninitialized, the relayer will use the
+            // default channel version, depending on the port id.
+            default_by_port(port_id)
+        }
+        ResolveContext::Other => {
+            // Resolve the version by querying the application version on destination chain
+            let dst_chain_id = channel.dst_chain().id();
+            info!("resolving channel version for port id='{}', context='{:?}' by retrieving dst chain '{}' app version", port_id, ctx, dst_chain_id);
```
romac, line 54:
```suggestion
            debug!("resolving channel version for port id='{}', context='{:?}' by retrieving dst chain '{}' app version", port_id, ctx, dst_chain_id);
```

relayer/src/channel/version.rs
```diff
@@ -0,0 +1,135 @@
...
+            // If unimplemented, we can recover by simply using the default version.
+            // This ensures compatibility with ibc-go v2.0.0+ and pre-v2.0.0.
+            match dst_app_version(channel) {
+                Ok(v) => Ok(v),
+                Err(e) => {
+                    if let ErrorDetail::GrpcStatus(s) = e.detail() {
+                        if s.is_unimplemented_port_query() {
+                            // Ensure compatibility & recover from the error,
+                            // by using the default version.
+                            debug!("resorting to default version because dst chain '{}' does not expose app version gRPC endpoint", dst_chain_id);
```
romac, line 69:
```suggestion
                            debug!(chain_id = %dst_chain_id, "resorting to default version because destination chain does not expose app version gRPC endpoint");
```

relayer/src/channel/version.rs
```diff
@@ -0,0 +1,135 @@
...
+        channel_id: cid.to_string(),
+    });
+
+    let proposed_version = channel
+        .src_version()
+        .cloned()
+        .unwrap_or(
+            default_by_port(channel.src_port_id())
+                .map_err(|e| Error::app_version(e.to_string()))?,
+        )
```
romac, line 118:
```suggestion
        .unwrap_or_else(|| {
            default_by_port(channel.src_port_id())
                .map_err(|e| Error::app_version(e.to_string()))?
        })
```

relayer/src/channel/version.rs
```diff
@@ -0,0 +1,135 @@
...
+        .into();
+    debug!(
```
romac, line 120:
```suggestion

    debug!(
```

relayer/src/chain/handle.rs
```diff
@@ -365,9 +367,10 @@ pub trait ChainHandle: Clone + Send + Sync + Serialize + Debug {
 
     fn get_key(&self) -> Result<KeyEntry, Error>;
 
-    fn add_key(&self, key_name: String, key: KeyEntry) -> Result<(), Error>;
+    fn app_version(&self, request: QueryAppVersionRequest)
+        -> Result<ics04_channel::Version, Error>;
```
romac, line 371: Let's rather expose the underlying fields (as domain types) rather than the proto type itself both because we do not want to depend on proto types in public APIs but also because non-Cosmos chains may use different RPC mechanism than gRPC.

```suggestion
    fn app_version(
        &self,
        port_id: PortId,
        connection_id: ConnectionId,
        ordering: Odering,
        counterparty: Option<Counterparty>,
        proposed_version: Version,
    ) -> Result<ics04_channel::Version, Error>;
```

relayer/src/channel.rs
adizere: Looked into it and it should not be a concern, since we eventually call `version::resolve` to obtain the right version to use directly in the `build_chan_open_try` method.

relayer/src/chain/handle.rs
adizere: Hmm I thought about this a lot and I'm hesitant to do the change. I'll describe below why I'm hesitating, but would encourage that in we take your points into account, Romain, when we work on https://github.com/informalsystems/ibc-rs/discussions/1318, #1487, and #1001.

**Reason to change to domain types:** Briefly, for clean APIs and better modularity wrt supporting different chain types. 
> both because we do not want to depend on proto types in public APIs but also because non-Cosmos chains may use different RPC mechanism than gRPC.

**Reason to keep the proto types:** Briefly, for developer experience.

This `QueryAppVersionRequest` appearing in our public API is an IBC.core type. Any IBC-enabled chain will support this query. (It can be implemented differently than gRPC, but not sure that's a good reason to avoid using the `QueryAppVersionRequest` data type).

But `QueryAppVersionRequest` being an IBC core is not the the point. The point is: We want to have simple interfaces in `ChainHandle` trait because of the handle <> runtime architecture. If the `app_version` method will take five parameters, all those parameters have to be carried back and forth through ~8 interface (or maybe more): chain handle, ProdChainHandle, integration test chain handle, runtime twice, `ChainRequest` enum, cosmos.rs and mock.rs impl. This is too much. So I think the methods signatures should be very compact, though the API can suffer as a consequence as it does in this instance (it doesn't have to, though, we can have both clean APIs and good developer experience, but that's maybe for future PR).

Let me know if I missed anything.

relayer/src/channel/version.rs
adizere: I tried following the suggestion but something's off.

```
error[E0277]: the `?` operator can only be used in a closure that returns `Result` or `Option` (or another type that implements `std::ops::FromResidual`)
   --> relayer/src/channel/version.rs:116:98
    |
115 |           .unwrap_or_else(|| {
    |  _________________________-
116 | |             default_by_port(channel.src_port_id()).map_err(|e| Error::app_version(e.to_string()))?
    | |                                                                                                  ^ cannot use the `?` operator in a closure that returns `ibc::core::ics04_channel::Version`
117 | |         })
    | |_________- this function should return `Result` or `Option` to accept `?`
    |
    = help: the trait `std::ops::FromResidual<std::result::Result<std::convert::Infallible, error::Error>>` is not implemented for `ibc::core::ics04_channel::Version`
note: required by `std::ops::FromResidual::from_residual`
   --> /Users/adi/.rustup/toolchains/1.56-x86_64-apple-darwin/lib/rustlib/src/rust/library/core/src/ops/try_trait.rs:339:5
    |
339 |     fn from_residual(residual: R) -> Self;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

Any further thoughts?

damiannolan:
Thank you for the quick updates to this branch, very much appreciated! ❤️ With the latest changes, the handshake flow for `interchain-accounts` is working pretty seemlessly, and have had success with account registration! 

I am still trying to debug some issues wrt. to executing an `interchain-accounts` tx. I am seeing the following error in hermes logs upon packet recv: `account sequence mismatch, expected 7 got 6: incorrect account sequence`.

I had experienced this before and it was a case that the validator address had been used for the relayer wallet. However, in this case I've set it up such that there is separate wallets being used. Will continue with this tomorrow! 
Thanks once again for this 🙌

relayer/src/channel.rs
```diff
@@ -756,14 +756,12 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
...
 
-        let version = version::resolve(ResolveContext::Other, self)?;
-
         let dst_expected_channel = ChannelEnd::new(
             highest_state,
             self.ordering,
             counterparty,
             vec![self.dst_connection_id().clone()],
-            version,
+            Version::from(""),
```
romac, line 764: Can we extract this into a `Version::empty()` method?

adizere: 908c3e86

relayer/src/channel/version.rs
romac: https://github.com/informalsystems/ibc-rs/pull/1582/commits/f25cff663b07ee35d2c85c5b68f07baa1e450415

relayer/src/chain/handle.rs
romac: > It can be implemented differently than gRPC, but not sure that's a good reason to avoid using the QueryAppVersionRequest data type

imho that's the first reason for not exposing the proto type, as it might be Cosmos-specific and may not be relevant for other types of chains.

Additionally, the idea with having both proto and domain types in tendermint-rs and ibc-rs is that the proto types are only relevant to (de)serialization and are otherwise an implementation detail which should not leak in public APIs.

What I would suggest we do here if you are worried about having too many parameters, is to define our own `QueryAppVersion` struct which would have those params as fields and then pass just this single struct around.

Does that make sense?

adizere: Thanks for following-up Romain. I agree with your suggestion, done! dc3e0c2ed6c836c77ebd04c7be90ea3e36703ee0

damiannolan:
Ack: When tested with the changes in [ibc-go #624](https://github.com/cosmos/ibc-go/pull/624) channel workers are now picking up events emitted from ICS-27 handshake initiation and correctly finishing them! 

Looks good 🙌 LGTM when you are happy with it :)

romac (approved):
🚀

modules/src/core/ics04_channel/channel.rs
```diff
@@ -300,8 +300,8 @@ impl From<Counterparty> for RawCounterparty {
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Deserialize, Serialize)]
 pub enum Order {
     None = 0,
-    Unordered,
-    Ordered,
+    Unordered = 1,
```
adizere, line 303: 😍

Issue #1214, opened 2021-07-20
Tracking issue for verification functions of TendermintClient


yito88:
## Crate
IBC

## Summary of Bug
Almost all verification functions in `ibc::ics07_tendermint::client_def::TendermintClient` are not implemented.

## Version
0.6.0

## Steps to Reproduce
Call verification functions such as `ibc::ics03_connection::handler::verify::verify_proofs` with a TendermintClient

## Acceptance Criteria
All verification functions are implemented

- [x] #1452
- [ ] `verify_client_consensus_state`
- [ ] `verify_connection_state`
- [ ] `verify_channel_state`
- [ ] `verify_client_full_state`
____

## For Admin Use

- [X] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
Just a note on progress on this PR: we are on it #1120. It should be done by the end of this milestone (end of July).

lightyear15:
Hello,
I have been following the hermes project quite closely, taking it as a starting point for a non-cosmos-chain project. [see this](https://github.com/informalsystems/ibc-rs/issues/1481).
I am taking this issue as an ice-breaker to ask a few questions: 
- What is the purpose of the ``ClientDef`` trait and their implementation? 
- Isn't verifying states a duty that the client itself is in charge of, not something the relayer should bother about?
- in light of the recently approved new spec document on [wasm-based clients](https://github.com/cosmos/ibc/tree/master/spec/client/ics-008-wasm-client), how will the hermes realyer implementation for this new client type be able to fulfill the requirements for this a trait? (TL;DR for the wasm client: the client is implemented in wasm and runs in the cosmos chain as a smart contract, the ibc-go go module is an opaque sub-module that works as a proxy, forwarding all the client functions calls to the smart contract) How can the ``ics08-wasm`` module provides an implementation of ``ClientDef``?

Thanks

adizere:
Hey, the `ClientDef` trait can be difficult to understand because this trait is a central interface between off-chain (relayer) and on-chain (modules) IBC components. Our repository comprises both on- and off-chain implementations, hence it's not always clear what's essential for the IBC relayer, and what pertains to the IBC module.

> What is the purpose of the ClientDef trait and their implementation?

[...]

PR #1583, opened 2021-11-22
Implement proof verification functions for Tendermint client

yito88:
Closes: #1214

## Description
Add proof verification functions for Tendermint client.
I referred to [ibc-go implementation](https://github.com/cosmos/ibc-go/blob/main/modules/light-clients/07-tendermint/types/client_state.go).
I think we need some tests and discussion. Welcome your feedback.

______

For contributor use:


modules/src/clients/ics07_tendermint/client_def.rs
```diff
@@ -193,116 +197,198 @@ impl ClientDef for TendermintClient {
...
+        expected_consensus_state: &AnyConsensusState,
     ) -> Result<(), Ics02Error> {
-        todo!()
+        let path = Path::ClientConsensusState {
+            client_id: client_id.clone(),
+            epoch: consensus_height.revision_number,
+            height: consensus_height.revision_height,
+        }
+        .to_string();
+        let value = expected_consensus_state.encode_vec().unwrap();
```
yito88, line 214: I'm not sure whether the coding should be hard-corded.

hu55a1n1:
Thanks so much for this PR @yito88! 🙏 Amazing stuff! 👌

modules/src/clients/ics07_tendermint/client_def.rs
```diff
@@ -193,116 +197,198 @@ impl ClientDef for TendermintClient {
...
+        client_state: &Self::ClientState,
+        prefix: &CommitmentPrefix,
+        proof: &CommitmentProofBytes,
+        root: &CommitmentRoot,
+        client_id: &ClientId,
+        consensus_height: Height,
+        expected_consensus_state: &AnyConsensusState,
     ) -> Result<(), Ics02Error> {
-        todo!()
+        let path = Path::ClientConsensusState {
```
hu55a1n1, line 208: I think we're missing the following checks for all `verify_*()` methods ->
```
// check that the client is at a sufficient height
assert(clientState.latestHeight >= height)
// check that the client is unfrozen or frozen at a higher height
assert(clientState.frozenHeight === null || clientState.frozenHeight > height)
```
The first part of the second assertion (i.e. `clientState.frozenHeight === null`) is already checked, so we only need to check the second part -> https://github.com/informalsystems/ibc-rs/blob/6d54a5ac5597a3b08d2384c3849204c8fbf0de7f/modules/src/core/ics03_connection/handler/verify.rs#L133-L143

Also wondering what the best place for this check would be - should we have a separate function that all `verify_*()` functions call?

modules/src/clients/ics07_tendermint/client_def.rs
hu55a1n1: Good question; I see that ibc-go uses a `codec.BinaryCodec` that gets passed into the verification functions. AFAIK, the spec assumes that details of encoding are communicated out-of-band, so it should be all right to hard-code (for now). But will let @adizere and @romac weigh in here. 🙏

modules/src/clients/ics07_tendermint/client_def.rs
```diff
@@ -193,116 +197,198 @@ impl ClientDef for TendermintClient {
...
+        prefix: &CommitmentPrefix,
+        proof: &CommitmentProofBytes,
+        root: &CommitmentRoot,
+        port_id: &PortId,
+        channel_id: &ChannelId,
+        sequence: Sequence,
+        commitment: String,
     ) -> Result<(), Ics02Error> {
-        todo!()
+        let path = Path::Commitments {
```
hu55a1n1, line 272: All packet related verification functions are additionally missing the following checks ->
```
// assert that enough time has elapsed
assert(currentTimestamp() >= processedTime + delayPeriodTime)
// assert that enough blocks have elapsed
assert(currentHeight() >= processedHeight + delayPeriodBlocks)
```

modules/src/clients/ics07_tendermint/client_def.rs
yito88: Thank you!
I think that they should be called in `ics02_client::client_def::ClientDef` because the height validations are also required for other clients.
I'd like to add frozen_height() function at #1619. Then, I will add the height validations.

hu55a1n1: Thanks! Sounds like a plan! :) I just reviewed #1619 and added some comments.

modules/src/clients/ics07_tendermint/client_def.rs
yito88: @hu55a1n1 Thank you for the catch-up!
I'm trying to add the checks. I'm wondering that too many arguments (for times and heights) are given to the verification functions.
I have an idea to add a function like `validate_time()` to `ics02_client::client_def::ClientDef`:
```rust
pub trait ClientDef: Clone {
...
    fn validate_time(
        &self,
        current_time: Timestamp,
        current_height: Height,
        processed_time: Timestamp,
        processed_height: Height,
        delay_period_time: Duration,
        delay_period_blocks: u64,
    ) -> Result<(), Error>;
}
```
But I have a concern. The check is required by Tendermint client. I'm not sure it should be added to the generalized trait.

hu55a1n1: Good idea! 👍 But since this is Tendermint specific, I think it would be better to add this to [`impl ClientState`](https://github.com/informalsystems/ibc-rs/blob/6d54a5ac5597a3b08d2384c3849204c8fbf0de7f/modules/src/clients/ics07_tendermint/client_state.rs#L166). Also, `verify_delay_passed()` might be a more appropriate name because we're dealing with both time and height delays.

yito88: Thanks! I added `verify_delay_passed()` to Tendermint ClientState.

modules/src/clients/ics07_tendermint/client_def.rs
```diff
@@ -185,116 +189,263 @@ impl ClientDef for TendermintClient {
...
+    let processed_time = ctx
+        .processed_time(client_id)
+        .map_err(|_| Error::processed_time_not_found(client_id.clone()))?;
+    let processed_height = ctx
+        .processed_height(client_id)
+        .map_err(|_| Error::processed_height_not_found(client_id.clone()))?;
+
+    let delay_period_time = connection_end.delay_period();
+    // TODO get delay_period_height from delay_period_time
+    //let delay_period_height = get_block_delay(delay_period_time);
```
yito88, line 434: ibc-go gets the parameter `MaxExpectedTimePerBlock` and calculates `delay_period_height` with `delay_period_time`.
I don't know how to calculate it in ibc-rs.

hu55a1n1:
Sorry for the delay in getting to this @yito88. I just pushed some changes -> 
* I moved the `verify_height()` code out of ics02 because I think it's Tendermint specific. 
* I reverted to our old way of passing identifiers to verification functions instead of giving them ready-made paths because I think that is more error-prone and would require an assertion like `assert!(matches!(path, Path::Connections(..)))` to make sure we passed the right path. 
* Also removed the ics23 file changes as they aren't required anymore. 
I will continue to test/review and get back with more feedback.

modules/src/core/ics03_connection/handler/verify.rs
```diff
@@ -42,12 +45,7 @@ pub fn verify_proofs(
 
     // If a consensus proof is attached to the message, then verify it.
     if let Some(proof) = proofs.consensus_proof() {
-        Ok(verify_consensus_proof(
-            ctx,
-            connection_end,
-            proofs.height(),
-            &proof,
-        )?)
+        Ok(verify_consensus_proof(ctx, height, connection_end, &proof)?)
```
hu55a1n1, line 48: Unrelated to this PR, but I think we're only verifying the consensus proof if we receive (a non-empty) one in the IBC message, so this can be a security problem. I think the root cause of this issue is the way `Proofs` is modelled, due to which we no longer know if this function was called for a message that required consensus proof verification or not. One way to solve this problem would be to model `Proofs` as an enum - I'll open an issue for this. Can be addressed in a separate PR.

modules/src/clients/ics07_tendermint/client_def.rs
hu55a1n1: I added `ChannelReader::max_expected_time_per_block()` to resolve this. IMHO, this should ideally be part of a `ParamReader` trait but let's leave it in `ChannelReader` for now. Furthermore, `ConnectionReader` could be a super trait of `ClientReader`, etc. I'll document these ideas in another issue.

modules/src/clients/ics07_tendermint/client_def.rs
hu55a1n1: Actually, ICS24 explicitly states this -> 
> The provableStore:
> MUST use canonical data structure encodings provided in these specifications as proto3 files

modules/src/clients/ics07_tendermint/client_def.rs
```diff
@@ -185,116 +190,278 @@ impl ClientDef for TendermintClient {
...
+    client_state: &ClientState,
+    prefix: &CommitmentPrefix,
+    proof: &CommitmentProofBytes,
+    root: &CommitmentRoot,
+    path: String,
+    value: Vec<u8>,
+) -> Result<(), Ics02Error> {
+    let merkle_path = apply_prefix(prefix, vec![path]).map_err(Error::ics23_error)?;
+    let merkle_proof: MerkleProof = RawMerkleProof::try_from(proof.clone())
+        .map_err(Ics02Error::invalid_consensus_state_proof)?
```
adizere, line 404: Not sure if the error type `invalid_consensus_state_proof` is right here. The proof is a _not_ for a consensus state, but can be for a connection state, channel state, client state, ack, packet, etc. It's actually a general commitment proof. We might need to defined a new proof [variant in `Ics02Error` to use](https://github.com/informalsystems/ibc-rs/blob/aa2336bdd9ae5cb210c61434dc12b8be58f94be7/modules/src/core/ics02_client/error.rs#L19) here, for example:

```rust
        InvalidCommitmentProof
            [ Ics23Error ]
            | _ | { "invalid commitment proof bytes" },
```

modules/src/clients/ics07_tendermint/client_def.rs
```diff
@@ -185,116 +190,278 @@ impl ClientDef for TendermintClient {
...
+fn verify_non_membership(
+    client_state: &ClientState,
+    prefix: &CommitmentPrefix,
+    proof: &CommitmentProofBytes,
+    root: &CommitmentRoot,
+    path: String,
+) -> Result<(), Ics02Error> {
+    let merkle_path = apply_prefix(prefix, vec![path]).map_err(Error::ics23_error)?;
+    let merkle_proof: MerkleProof = RawMerkleProof::try_from(proof.clone())
+        .map_err(Ics02Error::invalid_consensus_state_proof)?
```
adizere, line 427: Same comment as above re: `invalid_consensus_state_proof`.

modules/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -167,6 +167,41 @@ impl ClientState {
...
+    pub fn verify_delay_passed(
+        current_time: Timestamp,
+        current_height: Height,
+        processed_time: Timestamp,
+        processed_height: Height,
+        delay_period_time: Duration,
+        delay_period_blocks: u64,
+    ) -> Result<(), Error> {
+        let time = (processed_time + delay_period_time).map_err(Error::timestamp_overflow)?;
+        if !(current_time == time || current_time.after(&time)) {
```
adizere, line 181:
```suggestion
        let earliest_time = (processed_time + delay_period_time).map_err(Error::timestamp_overflow)?;
        if !(current_time == earliest_time || current_time.after(&earliest_time)) {
```

Nit :)

modules/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -167,6 +167,41 @@ impl ClientState {
...
+            return Err(Error::not_enough_time_elapsed());
```
adizere, line 182: It could be useful to augment the error type here so that we also capture the specific time parameters that caused the error. Otherwise it will be difficult to debug later on. The return statement would look as follows:

```rust
return Err(Error::not_enough_time_elapsed(current_time, earliest_time));
```

modules/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -167,6 +167,41 @@ impl ClientState {
...
+        }
+
+        if current_height < processed_height.add(delay_period_blocks) {
+            return Err(Error::not_enough_blocks_elapsed());
```
adizere, line 186: Same comment as above, about augmenting the error type. Ideally, we'd also capture the heights in the error message:

```rust
            return Err(Error::not_enough_blocks_elapsed(current_height, processed_height, delay_period_blocks));
```

modules/src/core/ics02_client/context.rs
```diff
@@ -124,4 +129,14 @@ pub trait ClientKeeper {
...
+
+    /// Called upon successful client update.
+    /// Implementations are expected to use this to record the current (host) time as the time at
+    /// which this update (or header) was processed.
+    fn store_processed_time(&mut self, client_id: ClientId, height: Height) -> Result<(), Error>;
+
+    /// Called upon successful client update.
+    /// Implementations are expected to use this to record the current (host) height as the height
+    /// at which this update (or header) was processed.
+    fn store_processed_height(&mut self, client_id: ClientId, height: Height) -> Result<(), Error>;
```
adizere, line 141: The terminology is a bit unclear. Would it be more accurate if we call these methods `store_update_*` instead of `store_processed_*`?

```suggestion
    fn store_update_time(&mut self, client_id: ClientId, height: Height) -> Result<(), Error>;

    /// Called upon successful client update.
    /// Implementations are expected to use this to record the current (host) height as the height
    /// at which this update (or header) was processed.
    fn store_update_height(&mut self, client_id: ClientId, height: Height) -> Result<(), Error>;
```

modules/src/core/ics02_client/handler/create_client.rs
```diff
@@ -20,6 +22,8 @@ pub struct Result {
     pub client_type: ClientType,
     pub client_state: AnyClientState,
     pub consensus_state: AnyConsensusState,
+    pub processed_time: Timestamp,
+    pub processed_height: Height,
```
adizere, line 26: I noticed these two fields are not used anywhere. Can we delete?

modules/src/core/ics02_client/msgs/upgrade_client.rs
```diff
@@ -100,9 +100,9 @@ impl TryFrom<RawMsgUpgradeClient> for MsgUpgradeAnyClient {
             client_state: AnyClientState::try_from(raw_client_state)?,
             consensus_state: AnyConsensusState::try_from(raw_consensus_state)?,
             proof_upgrade_client: RawMerkleProof::try_from(c_bytes)
-                .map_err(Error::invalid_upgrade_client_proof)?,
+                .map_err(Error::invalid_client_proof)?,
```
adizere, line 103: It's unclear why we changed these types here. It seems like the types were more accurate in the original code.

modules/src/core/ics04_channel/context.rs
```diff
@@ -71,10 +72,28 @@ pub trait ChannelReader {
     /// Returns the current timestamp of the local chain.
     fn host_timestamp(&self) -> Timestamp;
 
+    /// Returns the time when the client state of the given identifier is stored
```
adizere, line 75:
```suggestion
    /// Returns the time when the client state for the given [`ClientId`] was updated with a header for the given [`Height`]
```

modules/src/core/ics04_channel/context.rs
```diff
@@ -71,10 +72,28 @@ pub trait ChannelReader {
...
+    fn processed_time(&self, client_id: &ClientId, height: Height) -> Result<Timestamp, Error>;
+
+    /// Returns the height when the client state of the given identifier is stored
```
adizere, line 78:
```suggestion
    /// Returns the height when the client state for the given [`ClientId`] was updated with a header for the given [`Height`]
```

modules/src/core/ics04_channel/context.rs
```diff
@@ -71,10 +72,28 @@ pub trait ChannelReader {
...
+    fn processed_height(&self, client_id: &ClientId, height: Height) -> Result<Height, Error>;
+
```
adizere, line 80: For these two methods, I would find it more accurate if they were called `client_update_time` and `client_update_height`. The reason is that we are in the `ChannelReader` here, and other methods follow a similar convention, for example: `client_consensus_state` or `connection_end`.

modules/src/core/ics04_channel/error.rs
```diff
@@ -324,6 +324,22 @@ define_error! {
                              e.port_channel_id.1)
             },
 
+        ProcessedTimeNotFound
+            { client_id: ClientId }
+            | e | {
+                format_args!(
+                    "Processed time for the client {0} not found",
```
adizere, line 331: Nit: It would be good to capture the specific `Height` in the error type, which would help tracking down bugs or error causes.

modules/src/core/ics04_channel/error.rs
```diff
@@ -324,6 +324,22 @@ define_error! {
...
+                    e.client_id)
+            },
+
+        ProcessedHeightNotFound
+            { client_id: ClientId }
+            | e | {
```
adizere, line 337: Same nit as above: it would be good to capture the specific update_height in the error type.

modules/src/core/ics04_channel/handler/verify.rs
```diff
@@ -36,9 +38,10 @@ pub fn verify_channel_proofs(
     client_def
         .verify_channel_state(
             &client_state,
-            proofs.height(),
+            height,
             connection_end.counterparty().prefix(),
             proofs.object_proof(),
+            consensus_state.root(),
```
adizere, line 44: 💯

modules/src/core/ics23_commitment/error.rs
```diff
@@ -11,5 +11,29 @@ define_error! {
...
             |_| { "failed to decode commitment proof" },
+
+        EmptyCommitmentPrefix
+            |_| { "empty commitment prefix" },
+
+        EmptyMerkleProof
+            |_| { "empty merkle proof" },
+
+        EmptyMerkleRoot
+            |_| { "empty merkle proof" },
```
adizere, line 22:
```suggestion
            |_| { "empty merkle root" },
```

modules/src/core/ics23_commitment/merkle.rs
```diff
@@ -24,9 +25,148 @@ pub fn apply_prefix(
...
+                let mut encoded = Vec::new();
+                prost::Message::encode(&p, &mut encoded).unwrap();
+                prost::Message::decode(&*encoded).unwrap()
+            })
+            .collect();
+        Self { proofs }
+    }
+}
+
+impl MerkleProof {
```
adizere, line 59: I'm curious: are the two methods implemented here amenable to simple unit-testing? Intuitively it seems so to me. If yes, we should probably do that (but separate issue/PR is fine, not necessarily here!).

modules/src/mock/context.rs
```diff
@@ -68,6 +69,12 @@ pub struct MockContext {
     /// The set of all clients, indexed by their id.
     clients: BTreeMap<ClientId, MockClientRecord>,
 
+    /// Tracks the processed time for the clients
```
adizere, line 72:
```suggestion
    /// Tracks the processed time for clients header updates
```

adizere (commented):
Amazing work!

I left a bunch of improvement suggestions, none of them major. I think the PR is almost ready!

modules/src/core/ics02_client/msgs/upgrade_client.rs
hu55a1n1: I think this was to reuse these errors in the verification functions. But as you rightly pointed out that those errors are better expressed as a separate variant, I will revert the names of these errors. 👍

modules/src/clients/ics07_tendermint/client_def.rs
hu55a1n1: 11c641df5a34a74d1228f7dee7643648034a9934

modules/src/clients/ics07_tendermint/client_state.rs
hu55a1n1: 5c1ee1d829aa598f0c649ef4534cf6de08d083ab

modules/src/clients/ics07_tendermint/client_state.rs
hu55a1n1: 5c1ee1d829aa598f0c649ef4534cf6de08d083ab

modules/src/core/ics04_channel/context.rs
hu55a1n1: 8dcb33d3db390346c9b96ec98f84fdfa63adb618

modules/src/clients/ics07_tendermint/client_def.rs
hu55a1n1: Good point! 👍 I have made a note of this in the meta issue #1696. Although in this particular case is, the `unwrap()` is alright because the `encode_vec()` method should actually be infallible (https://github.com/informalsystems/tendermint-rs/issues/1064).

modules/src/core/ics03_connection/handler/verify.rs
hu55a1n1: https://github.com/informalsystems/ibc-rs/issues/1696

modules/src/core/ics02_client/handler/create_client.rs
hu55a1n1: This is a nice catch @adizere! This led me to realize that we're not initializing `ClientState` metadata in our impl. See https://github.com/cosmos/ibc-go/blob/main/modules/light-clients/07-tendermint/types/client_state.go#L187
Updated meta-issue #1696 with this. 👍

modules/src/core/ics04_channel/error.rs
hu55a1n1: https://github.com/informalsystems/ibc-rs/pull/1583/commits/7381e02bd64c2c7821380b8de70e180b642d5642

modules/src/core/ics23_commitment/merkle.rs
hu55a1n1: Right, I just opened an issue to deal with this. https://github.com/informalsystems/ibc-rs/issues/1706

hu55a1n1:
I tried to test this code with basecoin-rs and I see that verification is failing for `ConnOpenTry` proofs -> 
```
$ RUST_BACKTRACE=1 hermes create connection ibc-0  basecoin-0
# ...
2021-12-21T19:16:55.367966Z ERROR ThreadId(01) Failed ConnTry ConnectionSide { chain: ProdChainHandle { chain_id: ChainId { id: "basecoin-0", version: 0 }, runtime_sender: Sender { .. } }, client_id: ClientId("07-tendermint-0"), connection_id: None }: 
   0: tx response event consists of an error: deliver_tx for A5AE0DDB8AE0AF6248BF97FD9BBF2378A6F38E26B265C864E0061A64DE70CE72 reports error: code=Err(2), log=Log("deliver failed with error: IBC module error\n\nCaused by:\n    ICS03 connection error\n\n    Caused by:\n       0: the client state proof verification failed for client id 07-tendermint-0\n       1: tendermint error\n       2: ics23 commitment error\n       3: proof verification failed\n\n    Location:\n        /home/hussaini/.cargo/registry/src/github.com-1ecc6299db9ec823/flex-error-0.4.4/src/tracer_impl/eyre.rs:10:9\n\nLocation:\n    /home/hussaini/.cargo/registry/src/github.com-1ecc6299db9ec823/flex-error-0.4.4/src/tracer_impl/eyre.rs:30:9")
```

adizere (approved):
Looks good, thanks Shoaib for the fixes and the follow-up issue!

Will let you merge this Shoaib 💯

hu55a1n1:
Okay, so I spent some time again trying to debug the `ConnOpenTry` proof-verification failure and I see that there is another problem with `ConnOpenAck` verification too. 
```

```
Interestingly, this is also seen on master, so I think it is unrelated to this PR. It is difficult to say whether this is a problem with basecoin-rs, but I'll continue to debug this separately and open an issue if required.

yito88:
@hu55a1n1 Thank you so much for your help!

PR #1629, opened 2021-12-01
Fix example config.toml, add regression test

mzabaluev:
## Description

The example `config.toml` is broken in master: the `proof_specs` field
was mistakenly named `proof-specs`.

Fix this and add a CLI test against future regressions in validity of the
example `config.toml`.

______

### PR author checklist:
- ~Added changelog entry~
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -223,61 +223,6 @@ memo_prefix = ''
 #   ['transfer', 'channel-0'],
 # ]
 
-# Specify custom ICS23 proof-specs for the chain (serialized as JSON).
-# Default: [`ProofSpecs::cosmos()`](modules/src/core/ics23_commitment/specs.rs)
-proof_specs = '''
-[
-  {
-    "leaf_spec": {
-      "hash": 1,
-      "prehash_key": 0,
-      "prehash_value": 1,
-      "length": 1,
-      "prefix": [
-        0
-      ]
-    },
-    "inner_spec": {
-      "child_order": [
-        0,
-        1
-      ],
-      "child_size": 33,
-      "min_prefix_length": 4,
-      "max_prefix_length": 12,
-      "empty_child": [],
-      "hash": 1
-    },
-    "max_depth": 0,
-    "min_depth": 0
-  },
-  {
-    "leaf_spec": {
-      "hash": 1,
-      "prehash_key": 0,
-      "prehash_value": 1,
-      "length": 1,
-      "prefix": [
-        0
-      ]
-    },
-    "inner_spec": {
-      "child_order": [
-        0,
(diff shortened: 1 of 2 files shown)


relayer-cli/tests/acceptance.rs
```diff
@@ -45,3 +45,15 @@ fn start_no_args() {
     ); // Todo: find out how to disable colored output and then remove the `[^ ]*` part from the regexp.
     cmd.wait().unwrap().expect_success();
 }
+
+#[cfg(not(tarpaulin))]
+#[test]
+fn example_configuration_is_valid() {
```
hu55a1n1, line 51: Nit: There's another file `relayer/tests/config/fixtures/relayer_conf_example.toml` which is used for a similar test in `parse_valid_config()` -> `relayer/src/config.rs:448`. Thinking if we should remove that file and test?

config.toml
```diff
@@ -225,7 +225,7 @@ memo_prefix = ''
 
 # Specify custom ICS23 proof-specs for the chain (serialized as JSON).
 # Default: [`ProofSpecs::cosmos()`](modules/src/core/ics23_commitment/specs.rs)
-proof-specs = '''
+proof_specs = '''
```
adizere, line 228: I think we should hide this option (i.e., delete all these lines from the config.toml). It seems this option is strictly for testing (at least at the moment), and it is a distraction towards relayer operators to keep it here.

Any thoughts @hu55a1n1, am I missing something wrt to the use-cases of `proof_specs`?

relayer-cli/tests/acceptance.rs
adizere: +1

config.toml
hu55a1n1: I agree! I think `proof_specs` is only used for testing currently so we can remove it from here. 👍 We can add it back once `hermes` is able to support non-SDK chains.

relayer-cli/tests/acceptance.rs
mzabaluev: I think both are useful: we are interested in the example config to not bit rot, and relayer lib's test probably needs its conf test vector close by.

mzabaluev: There's also a difference in software under test between the two tests: the test here runs the executable, the test in the lib crate is only a unit test.

config.toml
mzabaluev: OK, will remove.

hu55a1n1 (approved):
🌷

PR #1636, opened 2021-12-02
Increase default for max_gas pre defined

dualsystems:
## Description
Increase default for max_gas constant

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/relayer/src/chain/cosmos.rs
+++ b/relayer/src/chain/cosmos.rs
@@ -98,7 +98,7 @@ mod compatibility;
 pub mod version;
 
 /// Default gas limit when submitting a transaction.
-const DEFAULT_MAX_GAS: u64 = 300_000;
+const DEFAULT_MAX_GAS: u64 = 400_000;
 
 /// Fraction of the estimated gas to add to the estimated gas amount when submitting a transaction.
 const DEFAULT_GAS_PRICE_ADJUSTMENT: f64 = 0.1;
--- a/.changelog/unreleased/improvements/ibc-relayer-cli/1636.md
+++ b/.changelog/unreleased/improvements/ibc-relayer-cli/1636.md
@@ -1,0 +1,2 @@
+- Increase the default for `max_gas` from `300_000` to `400_000` ([#1636](https://github.com/informalsystems/ibc-rs/pull/1636))
+
--- a/config.toml
+++ b/config.toml
@@ -155,8 +155,8 @@ store_prefix = 'ibc'
 default_gas = 100000
 
 # Specify the maximum amount of gas to be used as the gas limit for a transaction.
-# Default: 300 000
-max_gas = 300000
+# Default: 400 000
+max_gas = 400000
 
 # Specify the price per gas used of the fee to submit a transaction and
 # the denomination of the fee. Required
@@ -239,7 +239,7 @@ account_prefix = 'cosmos'
 key_name = 'testkey'
 store_prefix = 'ibc'
 default_gas = 100000
-max_gas = 3000000
+max_gas = 400000
 gas_price = { price = 0.001, denom = 'stake' }
 gas_adjustment = 0.1
 max_msg_num = 30


relayer/src/chain/cosmos.rs
```diff
@@ -98,7 +98,7 @@ mod compatibility;
 pub mod version;
 
 /// Default gas limit when submitting a transaction.
-const DEFAULT_MAX_GAS: u64 = 300_000;
+const DEFAULT_MAX_GAS: u64 = 400_000;
```
adizere, line 101: It's unclear what is the exact use-case for having this constant 400K instead of 300K. As a relayer operator, you can override this default directly from your config.toml, on a per chain basis:

https://github.com/informalsystems/ibc-rs/blob/2e8d8390f731866d30d37627fc66925f4b381fda/config.toml#L159

That being said, could you please update your PR so that the value `400_000` is also reflected in the config.toml? There are three lines where this needs to change:

- https://github.com/informalsystems/ibc-rs/blob/2e8d8390f731866d30d37627fc66925f4b381fda/config.toml#L158
- https://github.com/informalsystems/ibc-rs/blob/2e8d8390f731866d30d37627fc66925f4b381fda/config.toml#L159
- https://github.com/informalsystems/ibc-rs/blob/2e8d8390f731866d30d37627fc66925f4b381fda/config.toml#L236

adizere (changes requested):

relayer/src/chain/cosmos.rs
dualsystems: I think default gas different in chains, but max_gas is defined in genesis.json. I see max_gas can be very large as [example txn](https://www.mintscan.io/osmosis/txs/242755FFB46932B6FBF2D2AB7A5A2525E5294C7040F780D89CFC9EE87D1D74E5)

adizere: >I think default gas different in chains, but max_gas is defined in genesis.json. 

Hermes has two parameters to control gas usage, on a per-chain basis:

1. The `default_gas` option is the value that that Hermes allocated by default to transaction, whenever gas estimation fails. It is very rarely that estimation fails.

2. The `max_gas` option in Hermes config.toml is the hard __upper bound__ on how much gas Hermes will try to allocate to any transaction it submits to a chain. If a transaction is estimated to cost more gas than `max_gas`, then Hermes will not submit it. Currently, this upper bound defaults to 300_000 (the hardcoded value you changed in this PR). Any operator can modify that value to be different, e.g., you can change to 3_000_000 for Osmosis and 1_000_000 for Hub if you wish (directly from your config.toml `max_gas`).

Let me know if this answers your concern.

dualsystems: ok, may be situation with "out gas error" split one large txn to several small

```
2021-12-06T10:22:43.705586Z  INFO ThreadId(36) [osmosis-1:transfer/channel-6 -> irishub-1] assembled batch of 36 message(s)
2021-12-06T10:22:43.818092Z ERROR ThreadId(22) [osmosis-1] send_tx: broadcast_tx_sync: Response { code: Err(11), data: Data([]), log: Log("Too much gas wanted: 35390094, maximum is 25,000,000: out of gas"), hash: transaction::Hash(ACA9F671C52CE5F62F3B925AC729242B8860FD428FCFD6DD60D05330AD0F4E7E) }: diagnostic: the price configuration for this chain may be too low! please check the `gas_price.price` Hermes config.toml
2021-12-06T10:22:43.825131Z ERROR ThreadId(22) [osmosis-1] estimate_gas: failed to simulate tx with non-recoverable error: gRPC call failed with status: status: InvalidArgument, message: "failed to execute message; message index: 0: timeout packet verification failed: clientID (07-tendermint-5), height (1-12808885): consensus state not found: invalid request", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }
```

dualsystems: for correct processing of large queues, I suppose to split transactions into smaller parts in the proportions of max_gas settings

adizere: Please use the parameter `max_tx_num` to reduce the maximum batch size.


https://github.com/informalsystems/ibc-rs/blob/fd65565151106c7e074a3a292b9404360b2b0d06/config.toml#L172

adizere (approved):
Thanks @dualsystems, LGTM!

PR #1642, opened 2021-12-03
Bump Rust toolchain to 1.57, fix lints

mzabaluev:
## Description

New rust release, new lints to fix.
Some of them have uncovered little problems in the code, which are fixed at the root cause.

______

### PR author checklist:
- ~Added changelog entry, using [`unclog`](https://github.com/informalsystems/unclog).~
- ~Added tests: integration (for Hermes) or unit/mock tests (for modules).~
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/relayer-cli/src/components.rs
+++ b/relayer-cli/src/components.rs
@@ -1,14 +1,9 @@
-use std::io;
-
 use abscissa_core::{Component, FrameworkError, FrameworkErrorKind};
 use tracing_subscriber::{filter::EnvFilter, util::SubscriberInitExt, FmtSubscriber};
 
 use ibc_relayer::config::GlobalConfig;
 
 use crate::config::Error;
-
-/// Custom types to simplify the `Tracing` definition below
-type StdWriter = fn() -> io::Stderr;
 
 /// A custom component for parametrizing `tracing` in the relayer.
 /// Primarily used for:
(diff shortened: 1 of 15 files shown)


relayer-cli/src/components.rs
```diff
@@ -45,47 +25,38 @@ impl JsonTracing {
...
-    #[allow(trivial_casts)]
     pub fn new(cfg: GlobalConfig) -> Result<Self, FrameworkError> {
         let filter = build_tracing_filter(cfg.log_level.to_string())?;
 
         // Construct a tracing subscriber with the supplied filter and enable reloading.
         let builder = FmtSubscriber::builder()
             .with_target(false)
             .with_env_filter(filter)
-            .with_writer(std::io::stderr as StdWriter)
+            .with_writer(std::io::stderr)
```
adizere, line 52: Just a note: This code should be tested at a basic level in the E2E python scripts, which rely on the JSON output. It would still be useful to run some CLI commands with and without `--json` to make sure we're not changing the structure of the output here.

We can do that, however, in the release PR (which will close #1648): on that occasion, we also would like to update the guide, which requires re-running the CLI anyway.

adizere (approved):
Thanks Mikhail!

Glad to finally see the clippy fixes!

relayer-cli/src/components.rs
mzabaluev: This is just a stylistic change, the cast did nothing.

romac: It was required on a previous version of rustc, glad to see it's not needed anymore :)

mzabaluev: Ah hm, this means we've implicitly bumped our MSRV.
If we care about such things, it may make sense to add a `clippy.toml` and maintain the MSRV there.

romac: Yeah let's do that!

Issue #1543, opened 2021-11-04
Hermes should check client expiration


ancazamfir:
## Crate
relayer

## Summary of Bug
I did some tests with hermes, triggered by a recent discussion with @mircea-c about client expiration. 
We don't check for client expiration in the following cases:
- create connection - we keep trying and in the meantime perform update client on the unexpired client many times.
- create channel - same as above (i think, still need to test)
- when workers are spawned - we spawn client, channel and connection workers that use expired clients
- when workers relay from events (here we try forever) or on clear packets (here we retry on clear packet block interval)
- CLIs don't have an initial check for expired clients but they fail quickly as there are no retries (note i haven't checked them all)
The only place we do check is in the client refresh (part of client worker).

I think we need a good sweep of the code looking for early detection of expired clients.

## Version
master

## Acceptance Criteria
- [ ] do not spawn workers that would attempt to use expired clients
- [ ] exit workers if clients expire (trigger can be an IBC event)
- [ ] graceful handling in the channel and connection handshake CLIs (exit early before attempting and msg)

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
> I think we need a good sweep of the code looking for early detection of expired clients.

Could we have all the expiration-related checks localized in the `ForeignClient::restore` and `find`?

https://github.com/informalsystems/ibc-rs/blob/master/relayer/src/foreign_client.rs#L275

ancazamfir:
Could be, not entirely sure (don't see `find` being used, `restore` is only in the client worker). When starting workers I would go as early as `spawn_workers_from_chain_to_chain` or `spawn_workers_for_client` the latest, check the client state and if expired, don't spawn ANY worker and continue with the next client. For running workers, check the client state on every event. We can push the check deeper but we need a way to exit the worker on client expiration.

But in general I agree with you, ideally this check should be done in only a few places.

PR #1656, opened 2021-12-06
Improve spawning of supervisor worker tasks

soareschen:
## Description

This is a prerequisite for #1543 to make the spawning of supervisor worker tasks more manageable by parallelizing unrelated tasks.

A key challenge of the current design of the supervisor worker tasks is that we are relying on ad hoc multiplexing of multiple unrelated sub-tasks in a single thread. This result in various unintuitive hacks and bottlenecks:

- If one sub-task raise an unrecoverable error, the other sub-tasks are aborted as well.
- If one sub-task need to abort early, it have to disable itself through flags instead of returning.
- A sub-task may block another sub-task for too long, causing the other sub-task to not get executed.
- The shutdown handler is implemented as one of the sub-tasks, which may be blocked and cause the task to not get shutdown properly.
- The retry-on-error logic containing multiple sub-tasks interfere with the main task loop, causing it to not shutdown properly due to the sub-tasks returning error in the retry loop. Furthermore, the error from some sub-tasks cause the retry even when it should have abort, as there is no way to cancel only one sub-task inside the retry loop.
- There is no way to guarantee a worker task always get terminated at the end. Because it rely on sending an explicit shutdown signal, early returns caused by errors or forgetting to call `.shutdown()` will cause the task to dangle.
- There is no way to wait for all tasks to terminate before continuing. This is needed in the integration test, as otherwise at the end of the test, the worker tasks may raise various errors due to the full node being shutdown.

For the case of #1543, the main challenge is in multiplexing the client refresh and the detect misbehavior sub-tasks. When the client trusting period is reduced to tens of seconds, the misbehavior sub-task may interfere with the refresh sub-task to be able to refresh on time. Furthermore, in the case when a client is in fact frozen, it is difficult to modify the current code to abort that particular sub-task.

This PR introduce the `ibc_relayer::task` module, which encapsulate the logic of spawning background tasks and return a `TaskHandle`. The `TaskHandle` provides methods to shutdown a background task, or wait for the task to terminate. The `TaskHandle` also implements a `Drop` handler that always shutdown a background task when it is dropped, preventing dangling tasks.

The `spawn_background_task` function accepts a stepper closure that is executed repeatedly, until the shutdown signal is received or until the stepper closure return abort signal or fatal error. The stepper closure returns a `TaskError<E>`, which wraps around an error type `E` and contain 3 variants to indicate whether to abort, continue with an ignorable error, or abort with a fatal error. This design significantly simplifies how a step function should be implemented, without having to worry about the low level details of managing a background task.

The supervisor code and its workers are refactored to use the new `task` module. The various sub-tasks are also split into dedicated tasks to allow them to run in parallel without interfering with each others. Due to the use of shared state in the existing code, some mutexes are introduced so that the same shared state can be shared across multiple parallel tasks.

This PR does not introduce any change in logic aside from refactoring the supervisor code to use the `task` module. The work for fixing #1543 would instead be done in a separate PR that is based on this PR.

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


adizere:
What is our strategy for merging this?
- Should we have a sync session to discuss it, or
- Is just you Soares + @romac comfortable to merge this?

I think we have to make some progress here. Would like to get to the main dish (#1543) hehe.

soareschen:
I have created #1664 to show the work in progress to handle the client expiry error in workers.

Unfortunately it seems like the refactoring in this PR would not be sufficient to handle the error, as the connection and channel workers are spawned with one per chain, rather than one per IBC client. Without further re-architecture we are face with the difficult choice of either keep looping to handle expired clients, or abort entirely and not able to handle any other new connections or channels on the chain. More details are discussed in #1664.

romac (approved):
:rocket:

soareschen:
@romac I have added some tests to make sure that the connection and channel workers are working as expected. The packet worker is already verified to be working, since it is needed for the IBC transfer tests to pass.

tools/integration-test/src/tests/supervisor.rs
```diff
@@ -0,0 +1,132 @@
...
+            "connection should eventually open",
+            || {
+                let connection_end_b =
+                    query_connection_end(&chains.handle_b, &connection_id_b.as_ref())?;
+
+                if !connection_end_b
+                    .value()
+                    .state_matches(&ConnectionState::Open)
+                {
+                    return Err(eyre!("expeted connection end A to be in open state"));
```
romac, line 61:
```suggestion
                    return Err(eyre!("expected connection end A to be in open state"));
```

tools/integration-test/src/tests/supervisor.rs
```diff
@@ -0,0 +1,132 @@
...
+                    })?;
+
+                let connection_end_a =
+                    query_connection_end(&chains.handle_a, &connection_id_a.as_ref())?;
+
+                if !connection_end_a
+                    .value()
+                    .state_matches(&ConnectionState::Open)
+                {
+                    return Err(eyre!("expeted connection end B to be in open state"));
```
romac, line 77:
```suggestion
                    return Err(eyre!("expected connection end B to be in open state"));
```

tools/integration-test/src/tests/supervisor.rs
```diff
@@ -0,0 +1,132 @@
...
+                        .tagged_counterparty_channel_id()
+                        .ok_or_else(|| {
+                            eyre!("expected counterparty channel id to present on open channel")
+                        })?;
+
+                let channel_end_a =
+                    query_channel_end(&chains.handle_a, &channel_id_a.as_ref(), &port_a.as_ref())?;
+
+                if !channel_end_a.value().state_matches(&ChannelState::Open) {
+                    return Err(eyre!("expeted channel end B to be in open state"));
```
romac, line 121:
```suggestion
                    return Err(eyre!("expected channel end B to be in open state"));
```

tools/integration-test/src/tests/supervisor.rs
```diff
@@ -0,0 +1,132 @@
...
+        )?;
+
+        assert_eventually_succeed(
+            "channel should eventually open",
+            || {
+                let channel_end_b =
+                    query_channel_end(&chains.handle_b, &channel_id_b.as_ref(), &port_b.as_ref())?;
+
+                if !channel_end_b.value().state_matches(&ChannelState::Open) {
+                    return Err(eyre!("expeted channel end A to be in open state"));
```
romac, line 107:
```suggestion
                    return Err(eyre!("expected channel end A to be in open state"));
```

PR #1664, opened 2021-12-09
Handle expired client errors in workers

soareschen:
Closes: #1543

## Description

Handle errors arise from expired client in connection workers.

I have written some manual integration tests to identify the issues and verify that the solution is working.

This PR is based on the work of #1656 to simplify the way tasks can abort gracefully. ~Unfortunately there are still outstanding issues to be addressed, and the refactoring of #1656 does not seem to be sufficient to handle all case of errors.~ (This is [now resolved](https://github.com/informalsystems/ibc-rs/pull/1664#discussion_r768532184))

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


tools/integration-test/src/tests/manual/client_expiration.rs
```diff
@@ -0,0 +1,261 @@
...
+        sleep(Duration::from_secs(10));
+
+        supervisor.shutdown();
+
+        // Even if we respawn the supervisor, it would still abort immediately
+        // due to the first frozen client error. We are effectively stuck at
+        // either having to keep retrying on frozen clients or not able to
+        // handle any other new connection due to frozen clients.
+
+        let _supervisor = spawn_supervisor(chains.config.clone(), chains.registry, None, false)?;
```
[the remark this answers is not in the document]
adizere: > that the connection and channel workers are configured to run with one worker per chain, rather than one worker per IBC client.

Can you expand on this? I don't understand what you mean exactly. Currently there is one worker per connection (or per channel); there isn't a worker per chain. Do you mean that each connection (and channel..) worker should be parametrized with the underlying IBC client?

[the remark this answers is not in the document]
soareschen: Hmm you are right. It was probably my oversight when observing the logs at the end of the day. I have run the tests again and confirmed that the connection handshake worker in an unexpired client is not interrupted the other expired clients.

adizere:
I noticed the E2E with gaia v6 [failing](https://github.com/informalsystems/ibc-rs/actions/runs/1601751988/attempts/1) so I restarted that.

@soareschen is there a particular (i.e., manual) recipe we can use to test the behavior of this work in practice? Or do we rely entirely on integration tests?

soareschen:
The E2E tests seems to be flaky for a while now.

For manual test, the best way is to look at `tools/integration-test/src/tests/client_expiration.rs` and insert `suspend()` at suitable places to observe the behavior at different steps. The key for this changes is to not see repeating error logs, i.e. the errors should only be displayed once and then the worker should abort.

Currently due to the way the event subscription is wired up, it is not possible to test the workers individually without more significant refactoring. Ideally, the test should allow spawning of the connection or channel worker directly, so that it can probe that the worker task terminated as expected. This would require more flexible creation of the event subscription without having to go through everything through the supervisor.

relayer/src/channel.rs
```diff
@@ -622,33 +639,46 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
...
-            (State::TryOpen, State::Init) => Ok(vec![self.build_chan_open_ack_and_send()?]),
-            (State::TryOpen, State::TryOpen) => Ok(vec![self.build_chan_open_ack_and_send()?]),
-            (State::Open, State::TryOpen) => Ok(vec![self.build_chan_open_confirm_and_send()?]),
-            _ => Ok(vec![]),
-        }
-    }
+    pub fn handshake_step(
+        &mut self,
+        state: State,
+    ) -> Result<(Option<IbcEvent>, bool), ChannelError> {
```
romac, line 645: Can we replace this boolean with an enum which describes whether the handshake is completed or not?

eg. 

```rust
enum HandshakeStatus {
  Completed,
  InProgress
}
```

relayer/src/channel.rs
```diff
@@ -622,33 +639,46 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
...
+            (State::Open, State::Open) => return Ok((None, true)),
+            _ => None,
+        };
 
-    pub fn step_state(&mut self, state: State, index: u64) -> RetryResult<(), u64> {
-        let done = '🥳';
+        Ok((res, false))
+    }
 
+    pub fn step_state(&mut self, state: State, index: u64) -> RetryResult<bool, u64> {
```
romac, line 659: Same here

relayer/src/channel.rs
```diff
@@ -622,33 +639,46 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
...
+            Ok((Some(ev), handshake_completed)) => {
+                info!("channel handshake step completed with events: {:#?}\n", ev);
+                RetryResult::Ok(handshake_completed)
             }
+            Ok((None, handshake_completed)) => RetryResult::Ok(handshake_completed),
         }
     }
 
-    pub fn step_event(&mut self, event: IbcEvent, index: u64) -> RetryResult<(), u64> {
+    pub fn step_event(&mut self, event: IbcEvent, index: u64) -> RetryResult<bool, u64> {
```
romac, line 681: Same here

relayer/src/connection.rs
```diff
@@ -606,33 +620,49 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Connection<ChainA, ChainB> {
...
-            (State::TryOpen, State::Init) => Ok(vec![self.build_conn_ack_and_send()?]),
-            (State::TryOpen, State::TryOpen) => Ok(vec![self.build_conn_ack_and_send()?]),
-            (State::Open, State::TryOpen) => Ok(vec![self.build_conn_confirm_and_send()?]),
-            _ => Ok(vec![]),
-        }
-    }
+    pub fn handshake_step(
+        &mut self,
+        state: State,
+    ) -> Result<(Option<IbcEvent>, bool), ConnectionError> {
```
romac, line 626: Same here

relayer/src/connection.rs
```diff
@@ -606,33 +620,49 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Connection<ChainA, ChainB> {
...
+            (State::Open, State::Open) => return Ok((None, true)),
+            _ => None,
+        };
 
-    pub fn step_state(&mut self, state: State, index: u64) -> RetryResult<(), u64> {
-        let done = '🥳';
+        Ok((event, false))
+    }
 
+    pub fn step_state(&mut self, state: State, index: u64) -> RetryResult<bool, u64> {
```
romac, line 640: Same here

relayer/src/connection.rs
```diff
@@ -606,33 +620,49 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Connection<ChainA, ChainB> {
...
+                    ev
+                );
+                RetryResult::Ok(handshake_completed)
             }
+            Ok((None, handshake_completed)) => RetryResult::Ok(handshake_completed),
         }
     }
 
-    pub fn step_event(&mut self, event: IbcEvent, index: u64) -> RetryResult<(), u64> {
+    pub fn step_event(&mut self, event: IbcEvent, index: u64) -> RetryResult<bool, u64> {
```
romac, line 665: Same here

relayer/src/foreign_client.rs
```diff
@@ -1144,6 +1167,10 @@ impl<DstChain: ChainHandle, SrcChain: ChainHandle> ForeignClient<DstChain, SrcCh
             // Check for misbehaviour according to the specific source chain type.
             // In case of Tendermint client, this will also check the BFT time violation if
             // a header for the event height cannot be retrieved from the witness.
+            //
+            // FIXME: This returns error if the update event contains expired client state.
+            // This can happen even if the latest client state is unexpired, but the
+            // update event is from earlier.
```
romac, line 1173: What should we do about this? Leave for future PR?

relayer/src/worker/channel.rs
```diff
@@ -78,25 +74,22 @@ pub fn spawn_channel_worker<ChainA: ChainHandle + 'static, ChainB: ChainHandle +
...
-                } else {
-                    resume_handshake = false;
-                }
             }
 
-            Ok(())
+            Ok(Next::Continue)
         },
     )
 }
```
romac, line 95: Much more readable, great stuff!

relayer/src/worker/packet.rs
```diff
@@ -60,12 +61,20 @@ pub fn spawn_packet_cmd_worker<ChainA: ChainHandle + 'static, ChainB: ChainHandl
...
             }
 
-            Ok(())
+            Ok(Next::Continue)
         },
     )
 }
 
-pub fn spawn_packet_worker<ChainA: ChainHandle + 'static, ChainB: ChainHandle + 'static>(
+fn handle_link_error_in_task(e: LinkError) -> TaskError<RunError> {
```
romac, line 69: Very nice!

tools/integration-test/src/tests/client_expiration.rs
```diff
@@ -0,0 +1,458 @@
...
+
+            sleep(Duration::from_secs(10));
+
+            let balance_a_2 = chains
+                .node_a
+                .chain_driver()
+                .query_balance(&chains.node_a.wallets().user1().address(), &denom_a)?;
+
+            assert_eq(
+                "balance on wallet A should decrease",
```
romac, line 329: Is that really the behavior we expect? Should we not check whether or not the client is expired before we proceed to the transfer? What happens to the funds withdrawn from wallet A otherwise?

tools/integration-test/src/tests/client_expiration.rs
```diff
@@ -0,0 +1,458 @@
...
+    }
+}
+
+impl HasOverrides for PacketExpirationTest {
+    type Overrides = ExpirationTestOverrides;
+
+    fn get_overrides(&self) -> &ExpirationTestOverrides {
+        &ExpirationTestOverrides
+    }
+}
```
romac, line 458: Love these tests, amazing work :)

relayer/src/channel.rs
soareschen: I make it return `task::Next` directly instead.

tools/integration-test/src/tests/client_expiration.rs
soareschen: The test passes, so that is the current behavior of ibc-go. I think we could file an issue in ibc-go to fail an ibc transfer if the local client is expired or frozen.

On the other hand, a chain probably can't check or guarantee that a counter party client is unfrozen, since everything is asynchronous. So the check would still only handle the simple case.

relayer/src/foreign_client.rs
soareschen: I have tried to reproduce the error with a simpler test and added it here as a manual test. I also made the misbehavior task always continue after initial check, so that it can still act on new client update events.

tools/integration-test/src/tests/client_expiration.rs
```diff
@@ -0,0 +1,534 @@
...
+impl BinaryChainTest for MisbehaviorExpirationTest {
+    fn run<ChainA: ChainHandle, ChainB: ChainHandle>(
+        &self,
+        _config: &TestConfig,
+        chains: ConnectedChains<ChainA, ChainB>,
+    ) -> Result<(), Error> {
+        /*
+           This test reproduce the error log when a misbehavior task is
+           first started. The error arise when `detect_misbehaviour_and_submit_evidence`
+           is called with `None`, and the initial headers are already expired.
```
soareschen, line 451: @adizere @romac I added this manual test to reproduce the error log from the misbehavior detection. The key is that when `detect_misbehaviour_and_submit_evidence` is called with `None`, and some update client events contain expired headers, then the error log is shown. Note however the returned result is still `ValidClient`, which is weird.

tools/integration-test/src/tests/client_expiration.rs
romac: I see! Yeah I think it might be worth filing an issue to discuss this further :)

relayer/src/foreign_client.rs
romac: Looks good to me! I guess we can  now remove the FIXME

tools/integration-test/src/tests/client_expiration.rs
soareschen: Hmm wait actually it might be that big of a deal, since the IBC packets have timeout. So any such failed token transfer will eventually be refunded. We should also write integration tests for such cases in the future.

relayer/src/channel.rs
```diff
@@ -622,33 +640,51 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
...
+            (State::Init, State::Uninitialized) => Some(self.build_chan_open_try_and_send()?),
+            (State::Init, State::Init) => Some(self.build_chan_open_try_and_send()?),
+            (State::TryOpen, State::Init) => Some(self.build_chan_open_ack_and_send()?),
+            (State::TryOpen, State::TryOpen) => Some(self.build_chan_open_ack_and_send()?),
+            (State::Open, State::TryOpen) => Some(self.build_chan_open_confirm_and_send()?),
+            (State::Open, State::Open) => return Ok((None, Next::Abort)),
+
+            // If the counterparty state is already Open but current state is TryOpen,
+            // return anyway as the final step is to be done by the counterparty worker.
+            (State::TryOpen, State::Open) => return Ok((None, Next::Abort)),
```
soareschen, line 657: I added a new exit condition for the connection/channel worker. This is required because if there is a race in the handshake step with an external party, such as `bootstrap_connection`, the worker may otherwise get stuck in a loop and keep processing new block events even when the handshake has completed.

tools/integration-test/src/tests/client_expiration.rs
adizere: > The test passes, so that is the current behavior of ibc-go. I think we could file an issue in ibc-go to fail an ibc transfer if the local client is expired or frozen.

I think ibc-go v1.0.0 has protections against this. What chain binary are we using, does it have ibc-go v1.0?

romac (approved):
Great work!

relayer/src/channel.rs
romac: Good catch!

PR #1665, opened 2021-12-13
Update to tendermint =0.23.2 and remove chrono

### LINKED ISSUE #1639: Replace chrono with time 0.3
> Can we have a parallel branch in `ibc-rs` that fix the breaking changes from this branch?

_Originally posted by @soareschen in https://github.com/informalsystems/tendermint-rs/issues/1030#issuecomment-985450953_

Yes we can! :muscle: Furthermore, we should have it as a full de-chronoization branch to fix #1478.

mzabaluev:
Closes #1639

## Description

Update to changes in tendermint-rs 0.23.2 and remove all direct chrono dependencies.

Also renamed or removed `Timestamp` methods:
- `as_nanoseconds` renamed to `nanoseconds` with deprecation of the old convention-breaking name;
- `as_datetime` removed, `into_datetime` comes in its stead.
- `from_datetime` removed, `From<tendermint::Time>` remains as a substitute.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


modules/src/timestamp.rs
```diff
@@ -88,42 +90,48 @@ impl Timestamp {
...
-            util::assemble_in_nanos(s, time.timestamp_subsec_nanos())
+            s.try_into().unwrap()
         })
     }
 
-    /// Convert a `Timestamp` to an optional [`chrono::DateTime<Utc>`]
-    pub fn as_datetime(&self) -> Option<DateTime<Utc>> {
-        self.time
+    /// Convert a `Timestamp` to an optional [`OffsetDateTime`]
+    pub fn into_datetime(self) -> Option<OffsetDateTime> {
```
mzabaluev, line 133: I wonder if we'd be more interested in getting `tendermint::Time` here?
That would also get rid of a public dependency on `time`.

romac: Outside of ics07, we are not supposed to depend on Tendermint-specific code but I guess we can make an exception for the `Time` struct since it's general purpose?

mzabaluev: In my mad scientist's secret world assimilation plan, we'll migrate to a common domain types crate that will provide idiomatic Rust support for Google's well-known protobuf types, and `tendermint::Time` will be morphed into one of the types from that yet unpublished crate. Until then, it could be whatever seems most practical at the moment.

Cargo.toml
```diff
@@ -18,8 +18,15 @@ exclude = [
...
+# tendermint-proto        = { git = "https://github.com/informalsystems/tendermint-rs", branch = "v0.23.x" }
+# tendermint-light-client = { git = "https://github.com/informalsystems/tendermint-rs", branch = "v0.23.x" }
+# tendermint-testgen      = { git = "https://github.com/informalsystems/tendermint-rs", branch = "v0.23.x" }
+
+# [patch.crates-io]
+# tendermint              = { path = "../tendermint-rs/tendermint" }
+# tendermint-rpc          = { path = "../tendermint-rs/rpc" }
+# tendermint-proto        = { path = "../tendermint-rs/proto" }
+# tendermint-light-client = { path = "../tendermint-rs/light-client" }
+# tendermint-testgen      = { path = "../tendermint-rs/testgen" }
```
thanethomson, line 32: Should we remove these?

modules/src/clients/ics07_tendermint/client_def.rs
```diff
@@ -107,7 +107,7 @@ impl ClientDef for TendermintClient {
             untrusted_state,
             trusted_state,
             &options,
-            Time(chrono::Utc::now()),
+            OffsetDateTime::now_utc().try_into().unwrap(),
```
thanethomson, line 110: Should we not perhaps implement a `now_utc()` method for the `Time` struct to avoid directly importing dependencies from the `time` crate? I'd generally prefer to use interfaces that hide their underlying implementations as far as possible, as it reduces entanglement.

modules/src/timestamp.rs
```diff
@@ -88,42 +90,48 @@ impl Timestamp {
...
     /// let ti = Timestamp::from_nanoseconds(min).unwrap();
-    /// let uti = ti.as_nanoseconds();
+    /// let uti = ti.nanoseconds();
     /// assert_eq!(uti, min);
+    /// let tz = Timestamp::default();
+    /// let utz = tz.nanoseconds();
+    /// assert_eq!(utz, 0);
     /// ```
-    pub fn as_nanoseconds(&self) -> u64 {
+    pub fn nanoseconds(self) -> u64 {
```
thanethomson, line 123: Why does this consume `self`?

thanethomson (commented):
Might also be worthwhile to mention in the changelog that we're upgrading to tendermint-rs v0.23.2.

Cargo.toml
mzabaluev: Yeah, this was left over from my work with a locally merged copy.

modules/src/clients/ics07_tendermint/client_def.rs
mzabaluev: The trouble with `now_utc` is, it requires `std`. So I assume it can't be part of Tendermint's core API without an `std` feature gate, too.

modules/src/timestamp.rs
mzabaluev: `Self` is `Copy`, so it does not matter much.

modules/src/clients/ics07_tendermint/client_def.rs
thanethomson: Yeah I can look into providing that mechanism. It's probably preferable to avoid mixing newtypes and the types they're supposed to wrap.

modules/src/timestamp.rs
mzabaluev: If there's a concern about performance, I'd rather put the `#[inline]` attribute on the method so the compiler can always optimize it in the best way.

modules/src/clients/ics07_tendermint/client_def.rs
mzabaluev: We'd still want to provide the `TryFrom`/`Into` conversions from/to `OffsetDateTime` for convenience, and `time` is not an optional dependency at the moment (though it could be after redoing the internal representation of `Time` per [ADR-010](https://github.com/informalsystems/tendermint-rs/blob/master/docs/architecture/adr-010-time-api-sanity.md)).

soareschen (approved):

modules/src/timestamp.rs
```diff
@@ -51,30 +64,19 @@ impl Timestamp {
...
-                chrono::LocalResult::Single(ts) => Ok(Timestamp { time: Some(ts) }),
-                chrono::LocalResult::Ambiguous(_, _) => {
-                    Err(ParseTimestampError::ambiguous_timestamp_conversion(s, ns))
-                }
-            }
+            // As the `u64` representation can only represent times up to
+            // about year 2554, there is no risk of overflowing `Time`.
+            let seconds = (nanoseconds / NANOS_PER_SEC) as i64;
+            let subsec_nanos = (nanoseconds % NANOS_PER_SEC) as u32;
+            let ts = Time::from_unix_timestamp(seconds, subsec_nanos).unwrap();
```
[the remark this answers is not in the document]
mzabaluev: It would, thanks. It would also make sense to expose this method on `Time`, too.

Issue #1442, opened 2021-10-08
Packets stuck due to overflowing Timeout


RaulBernal:
## Crate

hermes relayer

## Summary of Bug
Pending packets are stucked, can't be relayed
```
[bitcanna-1:transfer/channel-0 -> osmosis-1] packets that still have commitments on bitcanna-1: 11, 12, 13, 14, 15, 16, 17, 18, 19, 20 (first 10 shown here; total=53)
Oct 08 13:56:25.293 DEBUG ThreadId(01) [bitcanna-1:transfer/channel-0 -> osmosis-1] recv packets to send out to osmosis-1 of the ones with commitments on source bitcanna-1: 11, 12, 13, 14, 15, 16, 17, 18, 19, 20 (first 10 shown here; total=52)
The application panicked (crashed).
Message:  called `Result::unwrap()` on an `Err` value: 
   0: error converting from u64 to i64
   1: out of range integral type conversion attempted

Location:
   /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/flex-error-0.4.3/src/tracer_impl/eyre.rs:31

Backtrace omitted.
Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
Location: modules/src/ics04_channel/events.rs:139
```

Another logs: 
```
ct 08 14:03:04 BCNA-SentryNode2 hermes[533309]: Oct 08 14:03:04.625  INFO ThreadId(46) [osmosis-1:transfer/channel-41 -> bitcanna-1] clearing pending packets height=Some(Height { revision: 1, height: 1500899 })
Oct 08 14:03:04 BCNA-SentryNode2 hermes[533309]: Oct 08 14:03:04.638 ERROR ThreadId(46) [osmosis-1:transfer/channel-41 -> bitcanna-1] worker: handling command encountered error:
Oct 08 14:03:04 BCNA-SentryNode2 hermes[533309]:    0: clearing of old packets failed 
```
This error is in bucle in the relayer log:
```
Oct 08 13:09:22 BCNA-SentryNode2 hermes[533309]: Oct 08 13:09:22.894 ERROR ThreadId(46) [osmosis-1:transfer/channel-41 -> bitcanna-1] worker: handling command encountered error:
Oct 08 13:09:22 BCNA-SentryNode2 hermes[533309]:    0: clearing of old packets failed
[...]

PR #1694, opened 2021-12-14
Verification with Prusti

zgrannan:
For the past few months I've been looking at using [Prusti](https://www.pm.inf.ethz.ch/research/prusti.html) to verify the correctness of `ibc-rs`, in particular the `modules` and `relayer` crates. The purpose of this PR is just to share a bit of what I've been able to accomplish so far and explain what an integration would look like. Also please let me know if you have any feedback or questions. __This PR is not intended to be merged__.

The Prusti verification operates on entire crates, and is done via an external tool. Prusti verification does not require introducing any new dependencies on the project. Specifications can be encoded as attributes that are only enabled when the "prusti" flag is set, i.e. `#[cfg_attr(feature="prusti", <spec>)]`. The specification syntax is a superset of Rust, with extensions to support `forall` expressions etc.

However, Prusti does not currently support all Rust language features and datatypes. So, it is also possible to provide different implementations for Prusti verification only (i.e. to work-around unsupported language features), without changing the actual implementation.

The following describes what I've done so far. A large caveat is that verification required changes to the code to remove unsupported features and datatypes; and some parts of the code containing unsupported features were excluded from verification entirely. The code in this PR only includes the added specifications, without the changes, in order to clearly demonstrate the annotation syntax and invariants. As a result, Prusti won't be able to verify the code in this PR; a version that has removed unsupported features is available [here](https://github.com/zgrannan/ibc-rs/tree/prusti). Note that it is still a bit messy.

## Verification of Safety Properties

Prusti was able to verify safety properties on the `modules` and `relayer` crate, with the caveat the properties may not hold for code unsupported by Prusti.

### No function will trigger `panic!`

For example in [context.rs](https://github.com/zgrannan/ibc-rs/blob/921d11763f14e028cde4b01d2b16abda18023873/modules/src/mock/context.rs):

```rust
#[cfg_attr(feature="prusti", requires(self.max_history_size > 0))]
#[cfg_attr(feature="prusti", requires(target_height.revision_height <= u64::MAX))]
#[cfg_attr(feature="prusti", requires(target_height.revision_number == self.latest_height.revision_number))]
#[cfg_attr(feature="prusti", requires(target_height.revision_height >= self.latest_height.revision_height))]
pub fn with_height(self, target_height: Height) -> Self {
    if target_height.revision_number > self.latest_height.revision_number {
        unimplemented!()
    } else if target_height.revision_number < self.latest_height.revision_number {
        panic!("Cannot rewind history of the chain to a smaller revision number!")
    } else if target_height.revision_height < self.latest_height.revision_height {
        panic!("Cannot rewind history of the chain to a smaller revision height!")
    } else if target_height.revision_height > self.latest_height.revision_height {
        // Repeatedly advance the host chain height till we hit the desired height
        let mut ctx = MockContext { ..self };
        while ctx.latest_height.revision_height < target_height.revision_height {
            ctx.advance_host_chain_height()
        }
        ctx
    } else {
        // Both the revision number and height match
        self
    }
}
```

The annotations ensure that the function `with_height()` will not panic, and introduces an obligation on the caller to ensure preconditions are satisfied.

### Calls to `Option.unwrap` and `Result.unwrap` will not panic

For example, for serialization:

```rust
#[cfg_attr(feature="prusti", requires(Hex::upper_case().encode_to_string(data).is_ok()))]
pub fn ser_hex_upper<S, T>(data: T, serializer: S) -> Result<S::Ok, S::Error>
where
    S: Serializer,
    T: AsRef<[u8]>,
{
    let hex = Hex::upper_case().encode_to_string(data).unwrap();
    hex.serialize(serializer)
}
```

### Arithmetic Operations will not overflow

This is useful, for example, operations on `Height`, to ensure that it is monotonically increasing.

```rust
#[cfg_attr(feature="prusti", requires(u64::MAX - self.revision_height >= delta))]
pub fn add(&self, delta: u64) -> Height {
    Height {
        revision_number: self.revision_number,
        revision_height: self.revision_height + delta,
    }
}

#[cfg_attr(feature="prusti", requires(self.revision_height < u64::MAX))]
pub fn increment(&self) -> Height {
    self.add(1)
}
```

## Functional Correctness

It was also possible to prove domain-specific invariants and correctness properties.

### Misbehavior Detection

In [tendermint.rs](https://github.com/zgrannan/ibc-rs/blob/921d11763f14e028cde4b01d2b16abda18023873/relayer/src/light_client/tendermint.rs) we have proven that:

- In the event of misbehaviour, the timestamp for the witness header is after the timestamp of all of the supporting headers
  
The invariant is specified as:
```rust
predicate! {
    fn misbehaviour_invariant(m: &MisbehaviourEvidence) -> bool {
    forall(
        |i : usize|
        (0 <= i && i < m.supporting_headers.len() ==>
          get_supporting_header_time(&m.supporting_headers, i) < get_witness_time(&m.misbehaviour)))
    }
}
```
  
- The witness header is within the trust period

This invariant asserts `header_within_trust_period` from the Tendermint crate.

Proving these invariants require adding and proving additional invariants of other functions.

### Client Keeper

In [ibc.rs](https://github.com/zgrannan/ibc-rs/blob/921d11763f14e028cde4b01d2b16abda18023873/ibc.rs), we prove that `store_client_result` in `ClientKeeper`, will correctly maintain an invariant about the stored client data. This is done out of the codebase, because the separation of `ClientKeeper` and `ClientReader` traits.

The client invariant corresponds to:

```rust
fn client_invariant(client: &MockClientRecord) {
    match client.client_state {
        Some(cs) =>
            match client.consensus_states.keys().max() {
                Some(max_height) => cs.latest_height() == max_height,
                None => false
            },
        None => client.consensus_states.is_empty()
    }
}
```

### DIFF
--- a/ibc.rs
+++ b/ibc.rs
@@ -1,0 +1,324 @@
+#![allow(dead_code)]
+#![allow(unused_must_use)]
+#![allow(unused_variables)]
+extern crate prusti_contracts;
+use prusti_contracts::*;
+
+// The following types are essentially uninterpreted
+
+#[derive(Clone, Copy)]
+pub struct AnyConsensusState(u32);
+
+// Cannot be wrapped in Struct due to Prusti Internal Error (fold/unfold)
+type AnyClientState = u32;
+
+#[pure]
+#[trusted]
+fn get_latest_height(cs: AnyClientState) -> Height {
+   unreachable!()
+}
+
+// Cannot be wrapped in Struct due to Prusti Internal Error (fold/unfold)
+type ClientId = u32;
+
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct ClientRecord {
+    client_state: Option<AnyClientState>
+}
+
+impl ClientRecord {
+    #[pure]
+    fn client_state_equals(
+        self,
+        state: AnyClientState
+    ) -> bool {
+        match self.client_state {
+            Some(cs) => cs == state,
+            None     => false
+        }
+    }
+
+    #[pure]
+    #[trusted]
+    fn get_max_consensus_state_height(self) -> Option<Height> {
+        unreachable!()
+    }
+}
+
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct ClientType(u32);
+
+// Collection of clients
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct Clients(u32);
+
+impl Clients {
+    #[pure]
+    #[trusted]
+    fn contains_key(self, client_id: ClientId) -> bool {
+        unreachable!()
+    }
+}
+
+// TODO: Make this an associated function of Clients
+#[pure]
+#[trusted]
+#[requires(clients.contains_key(client_id))]
+fn get_client(clients: Clients, client_id: ClientId) -> ClientRecord {
+   unreachable!()
+}
+
+// Cannot be wrapped in Struct due to Prusti Internal Error (fold/unfold)
+type Height = u32;
+
+
+
+// IBC Relayer Messages
+
+#[derive(Clone, Copy)]
+pub struct CreateResult {
+    pub client_id: ClientId,
+    pub client_type: ClientType,
+    pub client_state: AnyClientState,
+    pub consensus_state: AnyConsensusState,
+}
+
+#[derive(Clone, Copy)]
+pub struct UpdateResult {
+    pub client_id: ClientId,
+    pub client_state: AnyClientState,
+    pub consensus_state: AnyConsensusState,
+}
+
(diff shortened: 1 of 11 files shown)


adizere:
Hi @zgrannan. Thank you for your work here!

Am I correct in understanding that this PR is a proof of concept wrt. Prusti-based testing integration in ibc-rs? Let us know if there are specific topics you'd like us to weigh in or provide feedback.

In particular the panic and arithmetic overflow guards would be very interesting for us, since we had prior experience where this kind of problems can occur and are not always easy to catch (ref #1885 #1555 #1442 #1122).

PR #1739, opened 2022-01-04
Disable clap-provided -h flags on two subcommands

mzabaluev:
## Description

The conflicting flags were overlooked during the original clap porting in #1576, resulting in panics when `query client consensus` and `query client connections` commands are invoked.

______

### PR author checklist:
- ~Added changelog entry, using [`unclog`](https://github.com/informalsystems/unclog).~
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/relayer-cli/src/commands/query/client.rs
+++ b/relayer-cli/src/commands/query/client.rs
@@ -66,6 +66,7 @@ impl Runnable for QueryClientStateCmd {
 
 /// Query client consensus command
 #[derive(Clone, Command, Debug, Clap)]
+#[clap(setting(DisableHelpFlag))]
 pub struct QueryClientConsensusCmd {
     #[clap(required = true, about = "identifier of the chain to query")]
     chain_id: ChainId,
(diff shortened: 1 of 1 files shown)


mzabaluev:
Re tests: evidently there are no integration tests to exercise these subcommands in CI. How difficult it would be to add the tests, considering that the only way to invoke these commands in a do-nothing mode is being removed by this PR?

romac:
That's nice alternative, but if we go this way how about doing so for all commands? This way `-h` will never mean `--help` for one command and `--height` for another.

mzabaluev:
@romac Unfortunately, the option also disables the long `--help` flags for these select subcommands.
So it's a worse solution to your renaming all height-related short flags to `-H`, which I assume can be an acceptable breakage for the 0.10.0 release.

romac:
Oh I didn't realize it would suppress the --help flag as well, then yeah let's stick with -H flags and incur some breakage.

mzabaluev:
Superseded by #1743

Issue #1770, opened 2022-01-14
Handlers should use host's latest block's timestamp instead of host's timestamp


hu55a1n1:
## Crate
ibc

## Summary of Bug
`ChannelReader::host_timestamp()` gets the host's timestamp to do calculations (eg. verify packet delay elapsed and check packet hasn't expired). The problem with this approach is that the host's timestamp is non-deterministic and may vary from one full node to another. ibc-go seems to solve this by using the block time of the latest block on-chain. 

## Version
master

## Acceptance Criteria
- [ ] Handlers use the latest on-chain blocks timestamp. 

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PS: Many thanks to @ancazamfir for helping find this bug.

PR #1771, opened 2022-01-14
Deterministic host timestamp

hu55a1n1:
Closes: #1770 

## Description
`ChannelReader::host_timestamp()` is now a provided method that fetches the latest host consensus state using `ChannelReader::host_consensus_state()` and uses it's timestamp. 
______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/bug-fixes/ibc/1770-deterministic-host-timestamp.md
+++ b/.changelog/unreleased/bug-fixes/ibc/1770-deterministic-host-timestamp.md
@@ -1,0 +1,2 @@
+- IBC handlers now retrieve the host timestamp from the latest host consensus
+  state ([#1770](https://github.com/informalsystems/ibc-rs/issues/1770))
--- a/modules/src/core/ics04_channel/context.rs
+++ b/modules/src/core/ics04_channel/context.rs
@@ -70,7 +70,15 @@ pub trait ChannelReader {
     fn host_height(&self) -> Height;
 
     /// Returns the current timestamp of the local chain.
-    fn host_timestamp(&self) -> Timestamp;
+    fn host_timestamp(&self) -> Timestamp {
+        let latest_consensus_state = self
+            .host_consensus_state(self.host_height())
+            .expect("host must have latest consensus state");
+        latest_consensus_state.timestamp()
+    }
+
+    /// Returns the ConsensusState of the host (local) chain at a specific height.
+    fn host_consensus_state(&self, height: Height) -> Result<AnyConsensusState, Error>;
 
     /// Returns the time when the client state for the given [`ClientId`] was updated with a header for the given [`Height`]
     fn client_update_time(&self, client_id: &ClientId, height: Height) -> Result<Timestamp, Error>;


hu55a1n1:
A number of tests are currently failing with `ics02_client::error::InvalidConsensusStateTimestamp`. I think this has to do with the way our mock chain ctx generates blocks. Currently, blocks are generated using the current timestamp. The mock chain should ideally have a `block_time` config param and should generate blocks with timestamps that differ by `block_time` duration.

modules/src/core/ics02_client/context.rs
```diff
@@ -58,6 +59,19 @@ pub trait ClientReader {
     /// Returns the current height of the local chain.
     fn host_height(&self) -> Height;
 
+    /// Returns the current timestamp of the local chain.
+    fn host_timestamp(&self) -> Timestamp {
+        let latest_consensus_state = self
+            .host_consensus_state(Height::zero())
+            .expect("host must have pending consensus state");
+        latest_consensus_state.timestamp()
```
soareschen, line 67: Do we need a `pending_block` field just to mock the host timestamp? Would it work if we just get the latest block and add a fixed offset to the timestamp of the latest block?

modules/src/mock/context.rs
```diff
@@ -446,26 +470,30 @@ impl MockContext {
         let new_block = HostBlock::generate_block(
             self.host_chain_id.clone(),
             self.host_chain_type,
-            self.latest_height.increment().revision_height,
+            self.pending_block.height().increment().revision_height,
+            self.pending_block.timestamp().add(self.block_time).unwrap(),
         );
 
+        let old_pending_block = replace(&mut self.pending_block, new_block);
```
soareschen, line 477: It does not seem like we need the `pending_block` to have a cached next block. The only place I see it is used is in `host_timestamp`, which should be able to mock the time in other way. Here we can just generate the next block on the fly using the same time offset as used by `host_timestamp`.

modules/src/mock/context.rs
```diff
@@ -991,6 +1016,16 @@ impl ClientReader for MockContext {
         self.latest_height
     }
 
+    fn host_consensus_state(&self, height: Height) -> Result<AnyConsensusState, Ics02Error> {
+        if height.is_zero() {
+            return Ok(self.pending_block.clone().into());
```
soareschen, line 1021: We should avoid using `Height::zero()` as the magic parameter for getting the latest height. If that is needed, it is better to define a separate method `latest_host_consensus_state`. There is also no need for the `pending_block` here. If it is needed, it looks like the mock context should be able to deterministically generate the next block, so no caching should be needed.

I also don't see `host_consensus_state(Height::zero())` being used anywhere other than getting the host timestamp. So perhaps this is not needed if `host_timestamp` can be improved as suggested.

soareschen (changes requested):

modules/src/core/ics02_client/context.rs
hu55a1n1: Thanks for reviewing @Soares! This is an interesting observation. :+1: IMHO it is still helpful to keep a `pending_block`. I think by having a `pending_block`, the mock impl is closer to a real chain context impl and serves as a reference implementation. This makes it easier to convey the expectation that the chain context is in a state where it is ready to execute a pending block and also allows us to use the `ClientReader::host_timestamp()` provided trait method (just as it is expected to be used by a light client impl).

hu55a1n1: I realized that we don't treat the mock context as a ref impl, so your suggestions make sense. 👍 https://github.com/informalsystems/ibc-rs/pull/1771/commits/3b93cb813fa4edb0eeaf7e19dfaa0462a62fc8a8

modules/src/mock/context.rs
hu55a1n1: Done! 👍 https://github.com/informalsystems/ibc-rs/pull/1771/commits/4e4847eb7bb26b7f53b9934591ef7fb8a4d2d5bc

modules/src/mock/context.rs
hu55a1n1: Done! 👍 I also noticed that `latest_height` wasn't required since we can get the latest height from the history, and we expect the history to be non-empty -> https://github.com/informalsystems/ibc-rs/pull/1771/files#diff-a0183026217183d9147ad2c8fc0827f60c1e4354e9a2bed9da05f5f20fc501b6L842

soareschen (approved):
LGTM.

PR #1791, opened 2022-01-19
Run Security Audit job nightly only

romac:
## Description

Let's run the security audit job daily rather than on every PR and on master. We'll still get alerted of security issues via an issue and will recover the CI status to mean whether or not the various test suite have passed or not.

@mzabaluev What do you think?

### DIFF
--- a/.github/workflows/audit.yaml
+++ b/.github/workflows/audit.yaml
@@ -1,10 +1,5 @@
 name: Security Audit
 on:
-  pull_request:
-    paths: Cargo.lock
-  push:
-    branches: master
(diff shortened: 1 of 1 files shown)


mzabaluev (approved):
I think it's a good idea.
The only downside is that we won't learn immediately when a PR introduces a compromised dependency. But this currently suffers from the broken window syndrome due to #1478, so everyone's been conditioned to pretty much ignore audit failures anyway.

romac:
I agree with your assessment, let's go ahead with this then and maybe revisit this at a later `time` (sic)

Issue #836, opened 2021-04-20
Allow custom options when creating IBC clients


ancazamfir:
## Crate
relayer

## Summary
Current `hermes create client` CLI creates a client using the per chain configuration. We should provide a way for the user to specify the security parameters for the client (either described via CLI options or maybe a `.json` or other type of file including the parameters)

## Acceptance Criteria

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1807, opened 2022-01-24
cli: ADR-006 options for the create client command

mzabaluev:
Closes: #836

## Description

Add flags to the `tx raw create-client` CLI command (aliased as `create client`) to customize some of the parameters as described in [ADR-006](https://github.com/informalsystems/ibc-rs/blob/master/docs/architecture/adr-006-hermes-v0.2-usecases.md#decision).

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/relayer-cli/src/commands/tx/client.rs
+++ b/relayer-cli/src/commands/tx/client.rs
@@ -259,9 +259,9 @@ impl TxUpgradeClientsCmd {
 }
 
 fn parse_trust_threshold(input: &str) -> Result<TrustThreshold, Error> {
-    let (num_part, denom_part) = input
-        .split_once('/')
-        .ok_or_else(|| Error::cli_arg("expected a fraction argument separated by '/'".into()))?;
+    let (num_part, denom_part) = input.split_once('/').ok_or_else(|| {
+        Error::cli_arg("expected a fractional argument, two numbers separated by '/'".into())
+    })?;
     let numerator = num_part
         .trim()
         .parse()
--- a/relayer-cli/src/commands/tx/client.rs
+++ b/relayer-cli/src/commands/tx/client.rs
@@ -346,3 +346,23 @@ impl OutputBuffer {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::parse_trust_threshold;
+
+    #[test]
+    fn test_parse_trust_threshold() {
+        let threshold = parse_trust_threshold("3/5").unwrap();
+        assert_eq!(threshold.numerator(), 3);
+        assert_eq!(threshold.denominator(), 5);
+
+        let threshold = parse_trust_threshold("3 / 5").unwrap();
+        assert_eq!(threshold.numerator(), 3);
+        assert_eq!(threshold.denominator(), 5);
+
+        let threshold = parse_trust_threshold("\t3 / 5  ").unwrap();
+        assert_eq!(threshold.numerator(), 3);
+        assert_eq!(threshold.denominator(), 5);
+    }
+}
--- a/relayer-cli/src/commands/tx/client.rs
+++ b/relayer-cli/src/commands/tx/client.rs
@@ -9,7 +9,8 @@ use ibc::events::IbcEvent;
 use ibc_proto::ibc::core::client::v1::QueryClientStatesRequest;
 use ibc_relayer::chain::handle::ChainHandle;
 use ibc_relayer::config::Config;
-use ibc_relayer::foreign_client::ForeignClient;
+use ibc_relayer::foreign_client::{CreateParams, ForeignClient};
+use tendermint_light_client_verifier::types::TrustThreshold;
 
 use crate::application::app_config;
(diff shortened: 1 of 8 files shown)


relayer-cli/src/commands/tx/client.rs
```diff
@@ -23,6 +24,18 @@ pub struct TxCreateClientCmd {
 
     #[clap(required = true, help = "identifier of the source chain")]
     src_chain_id: ChainId,
+
+    // FIXME: provide option doc
```
romac, line 28: There are a couple FIXMEs to address before merging.

romac (changes requested):
Looks great! Let's address the remaining FIXMEs and we're good to go.

relayer-cli/src/commands/tx/client.rs
romac, line 28:
```suggestion
    /// Override the default clock drift specified in the configuration.
    /// The clock drift is a correction parameter dealing with only approximately synchronized clocks.
    /// The local clock should always be ahead of timestamps from the blockchain; this
    /// is the maximum amount that the local clock may drift behind a timestamp from the
    /// blockchain.
```

relayer-cli/src/commands/tx/client.rs
```diff
@@ -23,6 +24,18 @@ pub struct TxCreateClientCmd {
...
+    #[clap(short = 'd', long)]
+    clock_drift: Option<humantime::Duration>,
+
+    // FIXME: provide option doc
```
romac, line 32:
```suggestion
    /// Override the trusting period specified in the config.
    /// The trusting period specifies how long a validator set is trusted for (must be shorter than the chain's unbonding period)
```

relayer-cli/src/commands/tx/client.rs
```diff
@@ -23,6 +24,18 @@ pub struct TxCreateClientCmd {
...
+    #[clap(short = 'p', long)]
+    trusting_period: Option<humantime::Duration>,
+
+    // FIXME: provide option doc
```
romac, line 36:
```suggestion
    /// Override the trust threshold specified in the configuration.
    /// The trust threshold defines what fraction of the total voting power of a known
    /// and trusted validator set is sufficient for a commit to be accepted going forward.
```

relayer-cli/src/commands/tx/client.rs
romac: Right, let's do that!

mzabaluev: The long docs are applied in 6197b2dbc1007b35c789554306d783530faa1693, so you can check how it looks like with `--help`.

relayer-cli/src/commands/tx/client.rs
mzabaluev: 6197b2dbc1007b35c789554306d783530faa1693

relayer-cli/src/commands/tx/client.rs
mzabaluev: 6197b2dbc1007b35c789554306d783530faa1693

romac (approved):

ancazamfir:
I know this is not spec-ed in the ADR but we should probably also allow for:
```
    /// This flag, when set to true, will allow governance to recover a client
    /// which has expired
    allow_update_after_expiry: bool,
    /// This flag, when set to true, will allow governance to unfreeze a client
    /// whose chain has experienced a misbehaviour event
    allow_update_after_misbehaviour: bool,
```
We currently hardcode these to `true` when we create the client. This can be left for a different low prio PR.
Then the "weird" ones left will be: `proof_specs` (currently in the config but looking very strange there imo) and `upgrade_path`

relayer-cli/src/commands/tx/client.rs
```diff
@@ -23,6 +24,30 @@ pub struct TxCreateClientCmd {
 
     #[clap(required = true, help = "identifier of the source chain")]
     src_chain_id: ChainId,
+
+    /// Override the default clock drift specified in the configuration.
+    ///
+    /// The clock drift is a correction parameter dealing with only approximately synchronized clocks.
+    /// The local clock should always be ahead of timestamps from the blockchain; this
+    /// is the maximum amount that the local clock may drift behind a timestamp from the
+    /// blockchain.
```
adizere, line 33:
```suggestion
    /// The clock drift is a correction parameter. It helps deal with clocks that are only approximately synchronized between the source and destination chains of this client.
    /// The destination chain for this client uses the clock drift parameter when deciding to accept (or reject) a new header (originating from the source chain) for this client.
```

relayer-cli/src/commands/tx/client.rs
```diff
@@ -23,6 +24,30 @@ pub struct TxCreateClientCmd {
...
+    #[clap(short = 'd', long)]
+    clock_drift: Option<humantime::Duration>,
+
+    /// Override the trusting period specified in the config.
+    ///
+    /// The trusting period specifies how long a validator set is trusted for
+    /// (must be shorter than the chain's unbonding period).
+
```
adizere, line 41:
```suggestion
```

Issue #1797, opened 2022-01-21
Proto files upgrade to Cosmos SDK v0.45 & IBC-Go v3.0.0


adizere:
## Summary

We want to upgrade the proto files that `ibc` modules & Hermes uses to match the latest available interfaces from the Cosmos SDK and ibc-go library.

## Acceptance Criteria

- [x] Upgrade the proto files to align with [SDK v0.45](https://github.com/cosmos/cosmos-sdk/releases/tag/v0.45.0)
- [x] Upgrade the proto files to align with [ibc-go v3 line](https://github.com/cosmos/ibc-go/releases/tag/v3.0.0-alpha2)
- [x] Clean-up the use of `NegotiateAppVersion` from the channel handshake logic
    - Ref: https://github.com/cosmos/ibc-go/pull/646

Note that #1168 depends on the present issue.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #1825, opened 2022-01-31
Update protos to IBC v3.0.0-rc.0 and Cosmos SDK v0.45.1

romac:
Closes: #1797

## Description

- [x]  Update Protocol Buffers definitons to Cosmos SDK v0.45.1 et IBC v3.0.0-rc.0
- [x] Remove channel version negotiation step during channel handshake (done in #1840)

     > In IBC v2, relayers would query the `NegotiateAppVersion` endpoint before calling `ChanOpenTry`.
     > Applications would then need to verify that the passed in version was correct.
     > As of IBC v3, applications will perform this version negotiation during the channel handshake, thus removing the need for the relayer to query the version.

- [x] Handle non-standard ports in channel handshake (done in #1840)

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


proto/src/prost/COSMOS_IBC_COMMIT
```diff
@@ -1 +1 @@
-4e9752d10cecb9c39067b156e14b3797d0523afe
+9d79e86da4a8efa2c8bc3af7621adf7aa2c6525a
```
adizere, line 1: Hmm do we need to keep under version control 3 files for each of SDK and IBC module? I think we have

- proto/src/prost/COSMOS_IBC_COMMIT
- proto/src/prost/no_std/COSMOS_IBC_COMMIT
- proto/src/prost/std/COSMOS_IBC_COMMIT

(and similar for SDK commits)

relayer/src/channel/version.rs
```diff
@@ -4,83 +4,16 @@
...
-            // The destination chain may or may not implement `QueryAppVersionRequest`,
-            // which is only available from ibc-go v2.0.0 or newer versions.
-            // We detect if it is implemented by calling `is_unimplemented_port_query`.
-            // If unimplemented, we can recover by simply using the default version.
-            // This ensures compatibility with ibc-go v2.0.0+ and pre-v2.0.0.
-            match dst_app_version(channel) {
-                Ok(v) => Ok(v),
-                Err(e) => {
-                    if let ErrorDetail::GrpcStatus(s) = e.detail() {
-                        if s.is_unimplemented_port_query() {
```
adizere, line 67: I think we can also delete the method `is_unimplemented_port_query` from `error.rs`.

relayer/src/channel/version.rs
```diff
@@ -4,83 +4,16 @@
...
-                            // Ensure compatibility & recover from the error,
-                            // by using the default version.
-                            debug!(chain_id = %dst_chain_id, "resorting to default version because destination chain does not expose app version gRPC endpoint");
-                            return default_by_port(port_id);
-                        }
-                    }
-                    // Propagate the error, this is not a recoverable problem.
-                    Err(e).map_err(|e| ChannelError::query_app_version(channel.dst_chain().id(), e))
```
adizere, line 75: Also delete the error type `QueryAppVersion` from `relayer/src/channel/error.rs`.

adizere: Also cleanup the error type `AppVersion` from `relayer/src/error.rs`

adizere (changes requested):
Super! Left a couple of suggestions for further cleanup, but I think we're pretty much ready.

We should also bump up the `IBC_GO_MODULE_VERSION_REQ` to match ibc-go v3

proto/src/prost/COSMOS_IBC_COMMIT
romac: Yeah I am not sure why we have these in `no_std` and `std` where the top-level one would be enough. I guess it's a byproduct or running `proto-compiler` in both the `std` and `no_std` folders. The top-level one instructs the `sync-protobuf.sh` as to which commits to checkout. I'll see if I can clean this up.

romac:
> We should also bump up the `IBC_GO_MODULE_VERSION_REQ` to match ibc-go v3

b0b29edc

relayer/src/channel/version.rs
romac: 81c41b85

relayer/src/channel/version.rs
romac: 5d1ec0a2

adizere (approved):
Looks good, awesome work Romain!

I think we still have 3 sets of `COSMOS_IBC|SDK_COMMIT` files, would be cool to reduce the redundancy there.

If we decide to do a 0.11.1 release then maybe merge this PR only _after_ the release.

romac:
> I think we still have 3 sets of `COSMOS_IBC|SDK_COMMIT` files, would be cool to reduce the redundancy there.

Done :)

proto/src/prost/COSMOS_IBC_COMMIT
romac: https://github.com/informalsystems/ibc-rs/pull/1825/commits/c26cdf5b9c63b4218088b15d3f87f6d8d9c74a63

adizere:
We might have moved a bit too fast with upgrading the files: https://github.com/cosmos/ibc-go/releases/tag/v3.0.0-beta1. I think we should regenerate the files from v3-beta1, or maybe even better, wait it out until v3 comes out (towards end of Febr)?


>At a high level, when comparing to v3.0.0-alpha2 this release:
> - addresses all issues raised by Trail of Bits during an audit of ICS27
> - simplifies controller chain portID to only contain a prefix and the owner address
> - introduces JSON encoded metadata to ICS27 to store the version, connection sequences, encoding/txtype, and the interchain account address
> - the ics27 host chain will include the MsgResponse in the packet acknowledgement upon successful transaction execution. Please see the auth module documentation for more information.

romac:
Sure, that was expected. We don't want to merge this before v3 is out anyway :)

PR #1826, opened 2022-01-31
Implement `Capability{Reader,Keeper}` traits

hu55a1n1:
Closes: #1769 

- [x] Add `CapabilityReader/Keeper` traits
- [x] Fix isolated crate build
- [x] Remove `ics26::dispatch` ctx clone. 
- [x] Derive `Display` for `ics24::Path` variant types. 
______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


modules/src/core/ics05_port/context.rs
```diff
@@ -1,11 +1,68 @@
...
+    /// Check if the specified port_id is already bounded
+    fn is_bound(&self, port_id: PortId) -> bool {
+        self.get_capability(format!("{}", PortsPath(port_id)))
+            .is_ok()
+    }
+
+    /// Authenticate a capability key against a port_id by checking if the capability was previously
+    /// generated and bound to the specified port
+    fn authenticate(&self, port_id: PortId, capability: &Capability) -> bool {
+        self.authenticate_capability(format!("{}", PortsPath(port_id)), capability)
```
romac, line 28: Can we just do `to_string()` here instead?

```suggestion
        self.authenticate_capability(PortsPath(port_id).to_string(), capability)
```

modules/src/core/ics05_port/context.rs
```diff
@@ -1,11 +1,68 @@
...
+            self.new_capability(format!("{}", PortsPath(port_id)))
+        }
+    }
 }
 
-//  Result<Capability, Error>//return Ok(Capability::new());
+pub trait CapabilityKeeper {
+    /// Create a new capability with the given name.
+    /// Return an error if the capability was already taken.
+    fn new_capability(&mut self, name: impl AsRef<str>) -> Result<Capability, Error>;
```
romac, line 47: Is `AsRef<str>` really the best representation we can find for capabilities names? What are the operations on the capability name that are actually required? Perhaps we can define a trait for that and use `impl CapabilityName` instead?

modules/src/core/ics05_port/error.rs
```diff
@@ -8,6 +8,10 @@ define_error! {
             { port_id: PortId }
             | e | { format_args!("Port {0} is unknown", e.port_id) },
```
romac, line 9:
```suggestion
            | e | { format_args!("port '{0}' is unknown", e.port_id) },
```

modules/src/core/ics05_port/error.rs
```diff
@@ -8,6 +8,10 @@ define_error! {
...
 
+        PortAlreadyBound
+            { port_id: PortId }
+            | e | { format_args!("Port {0} is already bound", e.port_id) },
```
romac, line 13:
```suggestion
            | e | { format_args!("port '{0}' is already bound", e.port_id) },
```

romac (changes requested):

modules/src/core/ics05_port/context.rs
hu55a1n1: I just looked up the SDK impl and it seems capability names are expected to be non-empty strings (i.e. `strings.TrimSpace(name) != ""`) -> https://github.com/cosmos/cosmos-sdk/blob/master/x/capability/keeper/keeper.go#L230 , so I think it is better to use a newtype.

romac: Sweet!

romac (approved):
Awesome work! Just needs a merge from `master` and we're good to go :)

Issue #1716, opened 2021-12-25
User-facing CLI to clear packets


adizere:
## Crate

`ibc-relayer-cli`

## Problem Definition

Currently, Hermes operators need to specify two separate `tx raw` commands to achieve complete packet clearing of a given path, specifically:

- `hermes tx raw packet-recv <OPTIONS>`
- `hermes tx raw packet-ack <OPTIONS>`

This is clumsy for two reasons: (1) because `tx raw` is not meant to be user-facing, yet operators have no other way to achieve packet clearing, and (2) because operators need two separate commands both of which are notoriously complex due to the many positional arguments that these commands expect.

## Proposal

> This is a relayer operator suggestion.

We should provide a single CLI that unifies the two `raw packet-recv` + `raw packet-ack` commands.

Just like we have a ["packet setup" user-facing](https://hermes.informal.systems/commands/path-setup/index.html) commands, we should also have "packet clearing" commands.

Concretely: add a CLI

- `hermes clear packets`

with the following specification --

```
POSITIONAL ARGUMENTS:
    dst_chain_id              identifier of the destination chain
    src_chain_id              identifier of the source chain
    src_port_id               identifier of the source port
    src_channel_id            identifier of the source channel
``` 

The new CLI should act similarly to `packet-recv` and `packet-ack` commands, and should perform packet clearing in **both directions** of the given channel.

## Acceptance Criteria

- [ ] one new user-facing CLI, `hermes clear packets` with the semantics described above

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

Issue #1559, opened 2021-11-10
`hermes tx raw` shows all the data bytes in a column


mpoke:
## Summary of Bug

`hermes tx raw ft-transfer` shows all the data bytes in a column, which makes it hard to look at the other attributes. It would be more appropriate to compact the bytes output.

The same issue is with all messages that contain data, e.g., `packet-ack`.

## Version

v0.8.0

## Acceptance Criteria

- [ ] consider fixing via the suggestion in #175 

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

romac:
Thanks for opening this issue! What output format would you think is most appropriate here? Byte vector on a single line, hex, base64, something else?

mpoke:
> Thanks for opening this issue! What output format would you think is most appropriate here? Byte vector on a single line, hex, base64, something else?

I think a byte vector would be better. I guess since it's a tool for testing/dev it's useful to see the data bytes.

PR #1834, opened 2022-02-01
Implement `clear packets` command

mzabaluev:
Closes: #1716

## Description

Implement `clear packets` command as a combination of
the effects of `tx raw packet-recv` and `tx raw packet-ack`

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/relayer-cli/src/commands.rs
+++ b/relayer-cli/src/commands.rs
@@ -65,7 +65,7 @@ pub enum CliCmd {
     #[clap(subcommand)]
     Upgrade(UpgradeCmds),
 
-    /// Clear entities (e.g. packets) on chains
+    /// Clear objects, such as outstanding packets on a channel.
     #[clap(subcommand)]
     Clear(ClearCmds),
 
--- a/relayer-cli/src/commands.rs
+++ b/relayer-cli/src/commands.rs
@@ -5,6 +5,7 @@
 //! See the `impl Configurable` below for how to specify the path to the
 //! application's configuration file.
 
+mod clear;
 mod completions;
 mod config;
 mod create;
@@ -20,9 +21,10 @@ mod upgrade;
 mod version;
 
 use self::{
-    completions::CompletionsCmd, config::ConfigCmd, create::CreateCmds, health::HealthCheckCmd,
-    keys::KeysCmd, listen::ListenCmd, misbehaviour::MisbehaviourCmd, query::QueryCmd,
-    start::StartCmd, tx::TxCmd, update::UpdateCmds, upgrade::UpgradeCmds, version::VersionCmd,
+    clear::ClearCmds, completions::CompletionsCmd, config::ConfigCmd, create::CreateCmds,
+    health::HealthCheckCmd, keys::KeysCmd, listen::ListenCmd, misbehaviour::MisbehaviourCmd,
+    query::QueryCmd, start::StartCmd, tx::TxCmd, update::UpdateCmds, upgrade::UpgradeCmds,
+    version::VersionCmd,
 };
 
 use core::time::Duration;
@@ -62,6 +64,10 @@ pub enum CliCmd {
     /// Upgrade objects (clients) after chain upgrade
     #[clap(subcommand)]
     Upgrade(UpgradeCmds),
+
+    /// Clear entities (e.g. packets) on chains
+    #[clap(subcommand)]
+    Clear(ClearCmds),
 
     /// Start the relayer in multi-chain mode.
     ///
--- a/relayer-cli/src/commands/clear.rs
+++ b/relayer-cli/src/commands/clear.rs
@@ -3,11 +3,12 @@ use abscissa_core::{Command, Runnable};
 
 use ibc::core::ics24_host::identifier::{ChainId, ChannelId, PortId};
 use ibc::events::IbcEvent;
+use ibc_relayer::chain::handle::ProdChainHandle;
 use ibc_relayer::link::error::LinkError;
 use ibc_relayer::link::{Link, LinkParameters};
 
 use crate::application::app_config;
-use crate::cli_utils::ChainHandlePair;
+use crate::cli_utils::spawn_chain_counterparty;
 use crate::conclude::Output;
 use crate::error::Error;
 
@@ -21,9 +22,6 @@ pub enum ClearCmds {
 
 #[derive(Debug, Parser)]
 pub struct ClearPacketsCmd {
-    #[clap(required = true, help = "identifier of the destination chain")]
-    counterparty_chain_id: ChainId,
-
     #[clap(required = true, help = "identifier of the source chain")]
     chain_id: ChainId,
 
@@ -38,11 +36,16 @@ impl Runnable for ClearPacketsCmd {
     fn run(&self) {
         let config = app_config();
 
-        let chains =
-            match ChainHandlePair::spawn(&config, &self.chain_id, &self.counterparty_chain_id) {
-                Ok(chains) => chains,
-                Err(e) => Output::error(format!("{}", e)).exit(),
-            };
+
+        let chains = match spawn_chain_counterparty::<ProdChainHandle>(
+            &config,
+            &self.chain_id,
+            &self.port_id,
+            &self.channel_id,
+        ) {
+            Ok((chains, _)) => chains,
+            Err(e) => Output::error(format!("{}", e)).exit(),
+        };
 
         let mut ev_list = vec![];
 
--- a/relayer-cli/src/commands/clear.rs
+++ b/relayer-cli/src/commands/clear.rs
@@ -45,35 +45,43 @@ impl Runnable for ClearPacketsCmd {
 
         let mut ev_list = vec![];
 
-        // Clear packets in the src -> dst direction
-
+        // Construct links in both directions.
+        // Some of the checks in the two Link constructor calls may be redundant;
+        // going slowly, but reliably.
         let opts = LinkParameters {
             src_port_id: self.src_port_id.clone(),
             src_channel_id: self.src_channel_id.clone(),
         };
-        let link = match Link::new_from_opts(chains.src.clone(), chains.dst.clone(), opts, false) {
-            Ok(link) => link,
-            Err(e) => Output::error(format!("{}", e)).exit(),
-        };
-
-        run_and_collect_events(&mut ev_list, || link.build_and_send_recv_packet_messages());
-        run_and_collect_events(&mut ev_list, || link.build_and_send_ack_packet_messages());
-
-        // Clear packets in the dst -> src direction.
-        // Some of the checks in the Link constructor may be redundant;
-        // going slowly, but reliably.
-
+        let fwd_link =
+            match Link::new_from_opts(chains.src.clone(), chains.dst.clone(), opts, false) {
+                Ok(link) => link,
+                Err(e) => Output::error(format!("{}", e)).exit(),
+            };
         let opts = LinkParameters {
-            src_port_id: link.a_to_b.dst_port_id().clone(),
-            src_channel_id: link.a_to_b.dst_channel_id().clone(),
+            src_port_id: fwd_link.a_to_b.dst_port_id().clone(),
+            src_channel_id: fwd_link.a_to_b.dst_channel_id().clone(),
         };
-        let link = match Link::new_from_opts(chains.dst, chains.src, opts, false) {
+        let rev_link = match Link::new_from_opts(chains.dst, chains.src, opts, false) {
             Ok(link) => link,
             Err(e) => Output::error(format!("{}", e)).exit(),
         };
 
-        run_and_collect_events(&mut ev_list, || link.build_and_send_recv_packet_messages());
-        run_and_collect_events(&mut ev_list, || link.build_and_send_ack_packet_messages());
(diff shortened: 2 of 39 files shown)


relayer-cli/src/commands.rs
```diff
@@ -63,6 +65,10 @@ pub enum CliCmd {
     #[clap(subcommand)]
     Upgrade(UpgradeCmds),
 
+    /// Clear entities (e.g. packets) on chains
```
mzabaluev, line 68: This help text could be improved, please make your suggestions.

relayer-cli/src/commands/clear.rs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            Ok(ev) => ev,
+            Err(e) => return Output::error(format!("{}", e)).exit(),
+        };
+
+        match link
+            .build_and_send_ack_packet_messages()
+            .map_err(Error::link)
+        {
+            Ok(mut ev1) => ev.append(&mut ev1),
```
mzabaluev, line 64: I'm not sure this is the right thing to do, it just seemed the most logical without getting into the details.

relayer-cli/src/commands.rs
adizere: One suggestion
> Clear outstanding packets (i.e., packet-recv and packet-ack) on a given channel.

relayer/src/link.rs
```diff
@@ -163,7 +163,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Link<ChainA, ChainB> {
     }
 
     /// Implements the `packet-recv` CLI
-    pub fn build_and_send_recv_packet_messages(&mut self) -> Result<Vec<IbcEvent>, LinkError> {
```
adizere, line 166: Neat!

relayer-cli/src/commands/clear.rs
adizere: At first glance this looks solid.

I would like to test it on a dev environment and see how it behaves, however, then I'll do another review.

adizere (commented):
Thanks Mikhail! I did some testing and I have 3 suggestions:

1. This was not clear enough in the issue description (#1716), but it would be ideal for `clear packets` to handle clearing in _both directions_, which means to:  a. clear packets from source towards destination chain;  b. clear packets from destination towards source chain.
    - Currently we only handle case a. I believe.
    - To reproduce, do the following:
    - Setup a single channel with id `channel-0` between `ibc-0` and `ibc-1`
    - Submit a packet at chain `ibc-0`
    - > $ hermes --json tx raw ft-transfer ibc-1 ibc-0 transfer channel-0 0184 -d samoleans -t 1800 -n 1
    - Try to clear packets in direction `ibc-1 (source) -> ibc-0 (dest)`, it will not find anything
    - > $ hermes clear packets ibc-0 ibc-1 transfer channel-0
Success: []

2. Given my suggestion above to handle clearing in both directions, I think a reasonable way to do it as follows:
    1. `build_and_send_recv_packet_messages` on direction source -> dest
    2. `build_and_send_recv_packet_messages` on direction dest -> source
    3. `build_and_send_ack_packet_messages` on direction source -> dest
    4. `build_and_send_ack_packet_messages` on direction dest -> source

In the present implementation, I think we do 1 & 3.

This means we'll need to setup the `Link` objects for both directions, which means we need to find out the port+channel identifier on the destination chain (we only know the port+channel-id on the source chain). In other words, we'll need to query the channel port+channel end on the source chain and extract the [counterparty](https://github.com/informalsystems/ibc-rs/blob/276b46bad18955e4c0aff68366c7e4dad1cdaa49/modules/src/core/ics04_channel/channel.rs#L245) info from there.

3. Testing an debugging this PR is actually difficult because the output is very noisy. It may be a good time to try and tackle issue #1559 as part of this PR, namely: to make the output for `SendPacket`, `WriteAcknowledgement` and `AcknowledgePacket` more concise. Just to emphasize: this scope creep is optional, I'm just suggesting what I would do to make debugging more ergonomic here.

relayer-cli/src/commands.rs
mzabaluev: This text is for the entire `clean` top-level command.

mzabaluev:
> extract the [counterparty](https://github.com/informalsystems/ibc-rs/blob/276b46bad18955e4c0aff68366c7e4dad1cdaa49/modules/src/core/ics04_channel/channel.rs#L245) info from there.

Would something like [this](https://github.com/informalsystems/ibc-rs/blob/242a9ff6def2bdf11c06af7c8bc40d73c575fcfc/relayer-cli/src/commands/query/channel.rs#L51-L52) be a good way to do it?

adizere:
> > extract the [counterparty](https://github.com/informalsystems/ibc-rs/blob/276b46bad18955e4c0aff68366c7e4dad1cdaa49/modules/src/core/ics04_channel/channel.rs#L245) info from there.
> 
> Would something like [this](https://github.com/informalsystems/ibc-rs/blob/242a9ff6def2bdf11c06af7c8bc40d73c575fcfc/relayer-cli/src/commands/query/channel.rs#L51-L52) be a good way to do it?

Yep, we need to do a `chain.query_channel` and from the result of this query we can extract the counterparty info.

relayer-cli/src/commands.rs
adizere: Good point, I didn't realize that! I would simply replace `entities` by `objects`.

mzabaluev:
@adizere I have tested 4bb294c54cc7d9cc9bbaccb2aaf312d19ac31077 with the [repro sequence](https://github.com/informalsystems/ibc-rs/pull/1834#pullrequestreview-870337578) you have provided, and it looks to be working.

mzabaluev:
Adding an e2e test is a challenge, though: I need a clue on what the expected results of the command should be in the Python test classes. These are different for `TxPacketRecv` and `TxPacketAck`, and I'm not sure which data to extract from the combined result.

relayer-cli/src/commands/clear.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    #[clap(required = true, help = "identifier of the source chain")]
+    src_chain_id: ChainId,
+
+    #[clap(required = true, help = "identifier of the source port")]
+    src_port_id: PortId,
+
+    #[clap(required = true, help = "identifier of the source channel")]
+    src_channel_id: ChannelId,
+}
```
ancazamfir, line 35: Since the goal is to clean both directions there shouldn't be a ``src` and `dst`. Maybe `a` and `b` but i think we can just specify the ID-s on one chain.
```suggestion
pub struct ClearPacketsCmd {
    #[clap(required = true, help = "chain identifier")]
    chain_id: ChainId,

    #[clap(required = true, help = "identifier of the port")]
    port_id: PortId,

    #[clap(required = true, help = "identifier of the channel")]
    channel_id: ChannelId,
}
```

relayer-cli/src/commands/clear.rs
```diff
@@ -0,0 +1,90 @@
...
+
+impl Runnable for ClearPacketsCmd {
+    fn run(&self) {
+        let config = app_config();
+
+        let chains = match ChainHandlePair::spawn(&config, &self.src_chain_id, &self.dst_chain_id) {
+            Ok(chains) => chains,
+            Err(e) => return Output::error(format!("{}", e)).exit(),
```
ancazamfir, line 43: Could you fix clippy here?

relayer-cli/src/commands/clear.rs
```diff
@@ -0,0 +1,90 @@
...
+            Ok(link) => link,
+            Err(e) => return Output::error(format!("{}", e)).exit(),
+        };
+
+        run_and_collect_events(&mut ev_list, || link.build_and_send_recv_packet_messages());
+        run_and_collect_events(&mut ev_list, || link.build_and_send_ack_packet_messages());
+
+        // Clear packets in the dst -> src direction.
+        // Some of the checks in the Link constructor may be redundant;
+        // going slowly, but reliably.
```
ancazamfir, line 64: Not sure the order is correct here, in the sense that it will leave pending packets. I believe we need to create the two links (one for A->B and the other for B->A), then `build_and_send_recv_packet_messages` on the two links and after `build_and_send_ack_packet_messages` on the two links. This is because ``build_and_send_recv_packet_messages`  may create pending acks. 

Also I believe that these methods end up waiting for commit response from the chain, which is what we want here. But could you please double check?

relayer-cli/src/commands/clear.rs
mzabaluev: We still need to identify the other chain. What would be a good name for the flag currently named `dst_chain_id`?
These are positional parameters, so these names do not factor in the CLI, but I agree that the names and descriptions should be accurate.

relayer-cli/src/commands/clear.rs
mzabaluev: Thank you @ancazamfir for bringing this up. I'm still learning about these things. I guess the intent is to wait for commit in both directions, which I believe what the low-level `tx raw` commands also achieve. I will take care about interleaving the recv and ack phases as suggested.

relayer-cli/src/commands/clear.rs
mzabaluev: Or, is there a way to determine all counterparty IDs from the channel identified on "the chain"? The `Link` API does not seem to provide that; it also requires the type of the destination chain to be instantiated.

relayer-cli/src/commands/clear.rs
```diff
@@ -0,0 +1,98 @@
...
+                Err(e) => Output::error(format!("{}", e)).exit(),
+            };
+        let opts = LinkParameters {
+            src_port_id: fwd_link.a_to_b.dst_port_id().clone(),
+            src_channel_id: fwd_link.a_to_b.dst_channel_id().clone(),
+        };
+        let rev_link = match Link::new_from_opts(chains.dst, chains.src, opts, false) {
+            Ok(link) => link,
+            Err(e) => Output::error(format!("{}", e)).exit(),
+        };
```
mzabaluev, line 67: This may be the case for a `Link::reverse` method, which would do this and maybe skip on some unnecessary checks.

relayer-cli/src/commands/clear.rs
ancazamfir: There are other commands that do this, for example `QueryUnreceivedAcknowledgementCmd`.

mzabaluev: @ancazamfir With 248940dece0c7412ec221e557cc3232deff9c30c, the usage looks like this:

```
hermes clear packets <COUNTERPARTY_CHAIN_ID> <CHAIN_ID> <PORT_ID> <CHANNEL_ID>
```

ancazamfir: can you check 98b2a5cfdb924ce6d26f7097566eb6c50a2a6865 where i removed the counterparty chain parameter?

mzabaluev: Thanks @ancazamfir, it works with the three parameters. I have adjusted the descriptions for the parameters and the command to follow this change.

relayer/src/chain/cosmos.rs
```diff
@@ -2026,13 +2026,29 @@ impl ChainEndpoint for CosmosSdkChain {
...
-            PacketMsgType::TimeoutUnordered => ReceiptsPath {
-                port_id,
-                channel_id,
-                sequence,
-            }
-            .into(),
-            PacketMsgType::TimeoutOrdered => SeqRecvsPath(port_id, channel_id).into(),
+            PacketMsgType::TimeoutUnordered => {
+                debug!(
+                    "XXXX getting UNORDERED timout proof with SeqRecvsPath {}, {}, {}",
```
mzabaluev: Cleaned this up a bit using structured logging in 8ab1a72a11e371114c64f24eca48f6ad8711e6cc.

mzabaluev:
As reported by @ancazamfir, this only works correctly on unordered channels.

relayer/src/chain/cosmos.rs
ancazamfir: yes, sorry this was not supposed to be there. part of a an in-progress debug session that i was doing on that branch.

ancazamfir (approved):
Great stuff!

relayer-cli/src/commands/clear.rs
romac: Sounds like a nice addition, would you mind doing that as part of this PR?

relayer-cli/src/commands/clear.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        // Schedule RecvPacket messages for pending packets in both directions.
+        // This may produce pending acks which will be processed in the next phase.
+        run_and_collect_events(&mut ev_list, || {
```
romac, line 75: Very nice!

ancazamfir:
> As reported by @ancazamfir, this only works correctly on unordered channels.

Actually this works well for ordered channels when tested with gaia v6.0.0 and ibc-go v2.0.0 so I think it's all good (we won't support ordered channels in older versions)

adizere:
Tested & looks good, cool work here Mikhail, I will merge it ASAP!

adizere (approved):

Issue #1488, opened 2021-10-25
Confusing debug fmt string for PendingTx 


ancazamfir:
## Crate
relayer

## Summary of Bug
The check for verifying pending transactions displays a flipped relay path making things confusing. When timeouts are involved it is even more confusing.

And example log:
```
...
Oct 25 16:26:50.334  INFO ThreadId(83) [injective-1:transfer/channel-2 -> band-laozi-testnet4] relay op. data of 7 msgs(s) to Destination (height 1-1639), delayed by: 4.637976501s [try 1/5]
...
Oct 25 16:26:51.682 DEBUG ThreadId(25) [band-laozi-testnet4] send_tx: broadcast_tx_sync: Response { code: Ok, data: Data([]), log: Log("[]"), hash: transaction::Hash(7098354FD825350999B8EA1C322A0AF3CB7BD7EDF0E38A83AA970D448D34A147) }
....
Oct 25 16:26:52.329 DEBUG ThreadId(25) [band-laozi-testnet4] send_tx: broadcast_tx_sync: Response { code: Ok, data: Data([]), log: Log("[]"), hash: transaction::Hash(83BCA91B8176BA16055E34757AD6689966587DD137CE0096F62C6DB0B05B1415) }
...
Oct 25 16:26:52.329  INFO ThreadId(83) [injective-1:transfer/channel-2 -> band-laozi-testnet4] success
Oct 25 16:26:52.329 TRACE ThreadId(83) [band-laozi-testnet4:transfer/channel-90 -> injective-1] trying to confirm TxHashes: count=2; 7098354FD825350999B8EA1C322A0AF3CB7BD7EDF0E38A83AA970D448D34A147; 83BCA91B8176BA16055E34757AD6689966587DD137CE0096F62C6DB0B05B1415 
Oct 25 16:26:52.425 TRACE ThreadId(83) [band-laozi-testnet4:transfer/channel-90 -> injective-1] transaction is not yet committed: TxHashes: count=2; 7098354FD825350999B8EA1C322A0AF3CB7BD7EDF0E38A83AA970D448D34A147; 83BCA91B8176BA16055E34757AD6689966587DD137CE0096F62C6DB0B05B1415 
Oct 25 16:26:52.425 TRACE ThreadId(83) [band-laozi-testnet4:transfer/channel-90 -> injective-1] total pending transactions left: 1
...
```

## Version
master

## Steps to Reproduce
Run hermes with `tx_confirmation = true` and check the logs.

[...]

PR #1843, opened 2022-02-04
Reverse PendingTx chain id args and remove its Display impl

seanchen1991:
Closes: #1488

## Description

Reverses the order of the `chain_id` and `counterparty_chain_id` arguments when displaying a `PendingTx`. `PendingTx`'s `Display` impl was also removed as per the recent [log reformatting](https://github.com/informalsystems/ibc-rs/pull/1491/commits/d189d2cb8f7f3f9709a113cc762fc483ad2a7f7a) work.

I'm not sure I went about this in the best way though. Is there a better way to handle constructing the relay path? Also, why is that for `PendingTx`'s, the order of the chain IDs need to be reversed in the first place? Is this actually correct? 
______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/relayer/src/link/pending.rs
+++ b/relayer/src/link/pending.rs
@@ -1,4 +1,3 @@
-use core::fmt;
 use core::iter::Iterator;
 use core::time::Duration;
 use std::time::Instant;
@@ -70,6 +69,7 @@ impl<Chain: ChainHandle> PendingTxs<Chain> {
     pub fn insert_new_pending_tx(&self, r: AsyncReply, od: OperationalData) {
         let mut tx_hashes = Vec::new();
         let mut error_events = Vec::new();
+        let relay_path = self.relay_path();
 
         for response in r.responses.into_iter() {
             if response.code.is_err() {
@@ -82,7 +82,7 @@ impl<Chain: ChainHandle> PendingTxs<Chain> {
                 // behavior of handling the OperationalData.
                 trace!(
                     "[{}] putting error response in error event queue: {:?} ",
-                    self,
+                    &relay_path,
                     response
                 );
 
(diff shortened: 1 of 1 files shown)


relayer/src/link/pending.rs
```diff
@@ -82,7 +82,7 @@ impl<Chain: ChainHandle> PendingTxs<Chain> {
                 // behavior of handling the OperationalData.
                 trace!(
                     "[{}] putting error response in error event queue: {:?} ",
-                    self,
+                    &relay_path,
```
mzabaluev, line 85: This is better, but I'd rather rework all these `[{}]` constructs into span-wise KVs whenever the opportunity arises, as was partially done in #1491.

relayer/src/link/pending.rs
```diff
@@ -124,12 +124,24 @@ impl<Chain: ChainHandle> PendingTxs<Chain> {
...
 
-    // Try and process one pending transaction if available.
+    fn relay_path(&self) -> String {
+        format!(
+            "{}:{}/{} -> {}",
+            self.counterparty_chain_id,
+            self.port_id,
+            self.channel_id,
+            self.chain_id()
+        )
```
mzabaluev, line 134: All these could rather be in KV pairs to potentially aid log filtering.

romac: I guess they could but they're also useful when combing through the logs "by hand". But perhaps we can have both?

mzabaluev (approved):
Looks good inasmuch as it resolves the representation problem in logs.

relayer/src/link/pending.rs
seanchen1991: > rework all these [{}] constructs into span-wise KVs
I'm not entirely clear what this means. What will the resulting log lines look like?

relayer/src/link/pending.rs
seanchen1991: > But perhaps we can have both?

I'm not sure what's the best way to go about achieving this.

relayer/src/link/pending.rs
mzabaluev: > I'm not entirely clear what this means. What will the resulting log lines look like?

That will depend on the tracing backend. One of the benefits of using KVs is that they are converted into similar structural fields in logging backends supporting this (e.g. journald). In the terminal output subscriber that is used by Hermes now, a tracing event in a span created with `span!(tracing::Level::ERROR, "some_op", chain="foo", connection="bar")` will look something like this (with some bold and italic font styling added on an ANSI terminal):

```
some_op{chain=foo connection=bar}: here comes the event message
```

So, the information will still be flattened in this simple backend, but it will be easy to decipher and search for (if a bit more verbose).

romac: Let's perhaps do this here:

```rust
trace!(
             chain = %relay_path.chain_id()
            counterparty_chain = %relay_path.counterparty_chain_id,
             port = %relay_path.port_id,
             channel = %relay_path.channel_id,
             "[{}] putting error response in error event queue: {:?} ",
             &relay_path
)
```

romac:
@adizere What do you think?

adizere:
I think the new style is better! Will try to give a more thorough review
but briefly I think we’re good to go!

adizere (approved):
LGTM, neat work here! Glad we're slowly getting rid of `fmt::Display`s.

adizere:
I went ahead and merged, though I overlooked theres was no changelog entry. Cc @seanchen1991 to remember the changelogs in future PRs.

Issue #522, opened 2021-01-14
Module scoping in the libraries


adizere:
## Crate

`ibc` and `relayer`

## Summary

We should consider exporting modules with more restrictive scoping, e.g., `pub(crate)` instead of `pub`.

## Original discussion

I actually wanted to ask this before, why don't we use the crate scope more often. I believe in this case for example all mods, except application, can be public at crate level.

_Originally posted by @ancazamfir in https://github.com/informalsystems/ibc-rs/pull/518#discussion_r557330247_

I don't thin it matters too much for the `relayer-cli` crate, but yeah in general it wouldn't hurt either to make more modules crate scoped, to limit the surface area of our libs.

_Originally posted by @romac in https://github.com/informalsystems/ibc-rs/pull/518#discussion_r557340030_

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

romac:
Let's keep exporting most things publicly until we get a better idea of what is the surface area needed by downstream users of the `ibc` and `ibc-relayer` crate.

PR #1853, opened 2022-02-07
Restrictive scoping in modules [relayer]

Wizdave97:
Fixes https://github.com/informalsystems/ibc-rs/issues/522


romac:
Thanks for the PR @Wizdave97! Unfortunately I won't be able to merge it because this issue is a bit stale, for the reason outlined in this comment I just made in https://github.com/informalsystems/ibc-rs/issues/522#issuecomment-1032327943.

Issue #718, opened 2021-03-01
Rework `ibc::ics04_channel::events` module


romac:
## Crate

`ibc`

## Summary

The various channel events defined in the `ibc::ics04_channel::events` module are all defined as newtypes over the same `Attributes` which has some optional fields which are expected to always be defined for specific events. As such, this forces us to `unwrap` these values in [some places](https://github.com/informalsystems/ibc-rs/pull/692#discussion_r583791447). We should refactor this module to perform this check when parsing the source event, and perhaps refactor the parsing code a little bit while we're at it.
____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

Wizdave97:
@romac  I've made a draft implementation here https://github.com/Wizdave97/ibc-rs/blob/david/rework-channel-events/modules/src/core/ics04_channel/events.rs, I want to know if that's on track, especially the use of `TryFrom<Attributes>` instead of `From<Attributes>` since the other events still use optional `channel_id` values,  or do I enforce existence of `channel_id` on all other events?

PR #1859, opened 2022-02-09
Rework channel events

Wizdave97:
Closes: https://github.com/informalsystems/ibc-rs/issues/718

## Description
Refactor events channel events 

### PR author checklist:
- [x] Linked to GitHub issue.
### Reviewer checklist:
- [ ] Reviewed Files changed in the GitHub PR explorer.

### DIFF
--- a/modules/src/core/ics04_channel/events.rs
+++ b/modules/src/core/ics04_channel/events.rs
@@ -366,7 +366,7 @@ pub struct OpenInit {
 impl OpenInit {
     pub fn attributes(&self) -> Attributes {
         Attributes {
-            height: self.height.clone(),
+            height: self.height,
             port_id: self.port_id.clone(),
             channel_id: self.channel_id.clone(),
             connection_id: self.connection_id.clone(),
@@ -413,7 +413,7 @@ pub struct OpenTry {
 impl OpenTry {
     pub fn attributes(&self) -> Attributes {
         Attributes {
-            height: self.height.clone(),
+            height: self.height,
             port_id: self.port_id.clone(),
             channel_id: self.channel_id.clone(),
             connection_id: self.connection_id.clone(),
@@ -460,7 +460,7 @@ pub struct OpenAck {
 impl OpenAck {
     pub fn attributes(&self) -> Attributes {
         Attributes {
-            height: self.height.clone(),
+            height: self.height,
             port_id: self.port_id.clone(),
             channel_id: self.channel_id.clone(),
             connection_id: self.connection_id.clone(),
@@ -511,7 +511,7 @@ pub struct OpenConfirm {
 impl OpenConfirm {
     pub fn attributes(&self) -> Attributes {
         Attributes {
-            height: self.height.clone(),
+            height: self.height,
             port_id: self.port_id.clone(),
             channel_id: self.channel_id.clone(),
             connection_id: self.connection_id.clone(),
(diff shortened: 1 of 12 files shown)


modules/src/core/ics04_channel/events.rs
```diff
@@ -508,41 +546,66 @@ impl EventType for OpenConfirm {
...
+    pub height: Height,
+    pub port_id: PortId,
+    pub channel_id: ChannelId,
+    pub connection_id: ConnectionId,
+    pub counterparty_port_id: PortId,
+    pub counterparty_channel_id: Option<ChannelId>,
+}
 
 impl CloseInit {
+    pub fn attributes(&self) -> Attributes {
```
romac, line 559: imho this method should take ownership of `self` rather than perform a bunch of clones hidden to the caller.

romac: Alternatively, we could implement `From<CloseInit> for Attributes` for these events and remove this method.

This would make a nice parallel to the `Protobuf` instances we have all over the codebase where we have: `SomeDomainType: TryFrom<RawProtobufType>` and `RawProtobufType: From<SomeDomainType>`.

What do you think @hu55a1n1?

hu55a1n1: I think that's a great idea. 👍 I just looked up usages of the `attributes()` method and it seems we usually don't use the event after calling this function, so it's good to consume the event.

Wizdave97: Okay, I'll do that.

modules/src/events.rs
```diff
@@ -423,12 +423,12 @@ impl IbcEvent {
         }
     }
 
-    pub fn channel_attributes(&self) -> Option<&ChannelAttributes> {
-        match self {
-            IbcEvent::OpenInitChannel(ev) => Some(ev.attributes()),
-            IbcEvent::OpenTryChannel(ev) => Some(ev.attributes()),
-            IbcEvent::OpenAckChannel(ev) => Some(ev.attributes()),
-            IbcEvent::OpenConfirmChannel(ev) => Some(ev.attributes()),
+    pub fn channel_attributes(&self) -> Option<ChannelAttributes> {
```
romac, line 426:
```suggestion
    pub fn channel_attributes(self) -> Option<ChannelAttributes> {
```

romac (changes requested):

modules/src/core/ics04_channel/events.rs
```diff
@@ -1006,28 +1118,40 @@ mod tests {
             counterparty_channel_id: Some("counterparty_test_channel".parse().unwrap()),
         };
         let mut abci_events = vec![];
-        let open_init = OpenInit::from(attributes.clone());
+        let open_init = OpenInit::try_from(attributes.clone()).unwrap();
```
mzabaluev, line 1121: I wonder if this could be refactored into extracting values from an `&Attributes` reference rather than cloning all of the attributes and then picking some of the values from the clone.

Wizdave97: Only one field in `Atributes` is movable so using a reference won't make much difference since will still have to clone all the other fields to use them

romac (approved):
Great work @Wizdave97, thank you :)

romac:
Let's merge this for 0.12.1, ie. after #1870 has been merged and released.

Issue #1864, opened 2022-02-10
`max_expected_time_per_block` in `ChannelReader` returns `Result`


yito88:
## Crate
`ibc`

## Summary
`max_expected_time_per_block` should return `Result<u64, Ics04Error>` in `ChannelReader`

## Problem Definition
Getting the parameter could fail in some implementation

## Proposal
`max_expected_time_per_block` and `block_delay` return `Result`

```rust
    /// Returns the maximum expected time per block
    fn max_expected_time_per_block(&self) -> Result<Duration, Error>;

    fn block_delay(&self, delay_period_time: Duration) -> Result<u64, Error> {
        let expected_time_per_block = self.max_expected_time_per_block()?;
        ...
```

## Acceptance Criteria
`max_expected_time_per_block` and `block_delay` return `Result`

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

romac:
Can you please explain further why this method would possibly fail on some implementations? Would it be possible for the implementation to return a default/made-up value in cases where we cannot return a max_expected_time_per_block?

yito88:
Thank you for the question.
It happens in our case where the parameter is stored into the persistent storage and the read failed due to IO failure.
It's possible to return a default value. However, other reader's functions can return `Result` for the read failure with `implementation_specific` error. That's why the function should returns `Result` as other functions.

hu55a1n1:
I think our expectation with `max_expected_time_per_block` was that it was a user-defined constant that would be part of the binary or read from a config file on startup. That is the reason reading it cannot fail. (IIUC, this is also [how ibc-go implements it](https://github.com/cosmos/ibc-go/blob/c27d5b5f26b55d07489cc02fd0b8ce9fa8218e9a/modules/core/03-connection/keeper/params.go#L10).)

yito88:
I understand that. We should return the default value in cases where some failure happens.
Thanks!

PR #1865, opened 2022-02-10
`max_expected_time_per_block` in `ChannelReader` returns `Result`

yito88:
Closes: #1864

## Description
`max_expected_time_per_block` and `block_delay` return `Result`

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/improvements/ibc/1864-delay-height-error.md
+++ b/.changelog/unreleased/improvements/ibc/1864-delay-height-error.md
@@ -1,0 +1,2 @@
+- max_expected_time_per_block in ChannelReader returns Result
+  ([#1864](https://github.com/informalsystems/ibc-rs/issues/1864))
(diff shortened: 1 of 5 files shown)


romac:
Thanks @yito88, can you explain the rationale behind this change?

romac:
@hu55a1n1 How would you suggest we proceed here?

hu55a1n1:
@yito88 [previously said](https://github.com/informalsystems/ibc-rs/issues/1864#issuecomment-1035856352) that it is possible for them to use default values -> 
> It happens in our case where the parameter is stored into the persistent storage and the read failed due to IO failure.
> It's possible to return a default value.

AFAIK, IBC doesn't require this parameter to be stored on persistent storage as it is immutable/constant and doesn't need to be proven. 
If it is possible for @yito88 to somehow hard-code this parameter, use defaults or load it in memory on program startup, then I would suggest we leave things as they are.

romac:
> AFAIK, IBC doesn't require this parameter to be stored on persistent storage as it is immutable/constant and doesn't need to be proven.
> If it is possible for @yito88 to somehow hard-code this parameter, use defaults or load it in memory on program startup, then I would suggest we leave things as they are.

@yito88 Does @hu55a1n1's suggestion above make sense to you?

yito88:
Thank you for your explanation. It makes sense.

PR #1873, opened 2022-02-14
Additional unit tests for Client State

Wizdave97:
Partial fix for: https://github.com/informalsystems/ibc-rs/issues/1706

## Description

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/bug-fixes/ibc/1706-fix-formatting-for-some-tendermint-errors.md
+++ b/.changelog/unreleased/bug-fixes/ibc/1706-fix-formatting-for-some-tendermint-errors.md
@@ -1,0 +1,3 @@
+- Fixed the formatting of NotEnoughTimeElapsed and NotEnoughBlocksElapsed
+  in tendermint errors ([#1706](https://github.com/informalsystems/ibc-
+  rs/issues/1706))
--- a/.changelog/unreleased/improvements/ibc/1706-add-client-state-tests.md
+++ b/.changelog/unreleased/improvements/ibc/1706-add-client-state-tests.md
@@ -1,0 +1,2 @@
+- Added more unit tests to verify Tendermint ClientState
+  ([#1706](https://github.com/informalsystems/ibc-rs/issues/1706))


modules/src/clients/ics07_tendermint/client_state.rs
```diff
@@ -454,6 +479,164 @@ mod tests {
...
+                    current_time: (now + Duration::from_nanos(2000)).unwrap(),
+                    current_height: Height::new(0, 5),
+                    processed_time: (now + Duration::from_nanos(1000)).unwrap(),
+                    processed_height: Height::new(0, 4),
+                    delay_period_time: Duration::from_nanos(500),
+                    delay_period_blocks: 2,
+                },
+                want_pass: false,
+            },
+        ];
```
adizere, line 539: I was playing with these tests and wanted to provoke an error to see how it looks. I used this new test, which should not pass (but I set it to `want_pass: false,` so that errors get displayed).

```rust
            Test {
                name: "Error".to_string(),
                params: Params {
                    current_time: now,
                    current_height: Height::new(0, 5),
                    processed_time: (now + Duration::from_secs(10)).unwrap(),
                    processed_height: Height::new(0, 4),
                    delay_period_time: Duration::from_secs(50),
                    delay_period_blocks: 2,
                },
                want_pass: true,
            },
```

When running the test I found was that our error definition is incomplete. Specifically, we output this error:

> ClientState::verify_delay_passed() failed for test Error, 
msgParams ... with error Some(not enough time elapsed, current timestamp {0} is still less than earliest acceptable timestamp {1}

Note the `{0}` and `{1}` which should be an interpolation but it is not. All the user sees is this raw information, which is not very useful.

I propose the diff below on how we can change the error type for `NotEnoughTimeElapsed` to provide correct information.

```diff
diff --git a/modules/src/clients/ics07_tendermint/error.rs b/modules/src/clients/ics07_tendermint/error.rs
index 54966531..ad46795d 100644
--- a/modules/src/clients/ics07_tendermint/error.rs
+++ b/modules/src/clients/ics07_tendermint/error.rs
@@ -160,7 +160,7 @@ define_error! {
                 current_time: Timestamp,
                 earliest_time: Timestamp,
             }
-            |_| { "not enough time elapsed, current timestamp {0} is still less than earliest acceptable timestamp {1}" },
+            | e | { format_args!("not enough time elapsed, current timestamp {0} is still less than earliest acceptable timestamp {1}", e.current_time, e.earliest_time) },
```

With this change, the error should look like this
> ... error Some(not enough time elapsed, current timestamp Timestamp(2022-02-15T08:49:13.695546Z) is still less than earliest acceptable timestamp Timestamp(2022-02-15T08:50:13.695546Z)

What I would suggest:

- @Wizdave97 can you please try to recreate my test above and change the definition of error type `NotEnoughTimeElapsed` in this PR so that it provides complete error details? Note that the error definition for `NotEnoughBlocksElapsed` is also incomplete and needs to be fixed.

- @soareschen please let us know if we're using `format_args` in flex-error correctly here.

adizere (changes requested):
Thanks for your contribution David! The tests added here helped me catch a small problem we have with our error types. I left a suggestion for fixing that.

adizere (approved):
Good work!

Will let @hu55a1n1 have a final look & merge this if it looks good to him.

adizere:
One more thing actually: it would be good to add a changelog entry. Documentation on how to do that is

- here: https://github.com/informalsystems/unclog#adding-a-new-unreleased-entry
- and our contributing doc: https://github.com/informalsystems/ibc-rs/blob/master/CONTRIBUTING.md#changelog

It may be a good idea to read our contributing doc in its entirety.

Thank you @Wizdave97!

hu55a1n1 (approved):
Nicely done!🌷 Your tests led me to realise that a few validation checks are not implemented. Preapproved, with some observations that are unrelated to this PR (will let @adizere weigh in here).

adizere:
Nice findings @hu55a1n1. Documented them in #1893.

Again -- great work David!

Issue #1872, opened 2022-02-14
Hermes `clear_on_start` toggle does not work


adizere:
## Crate

`ibc-relayer` and `ibc-relayer-cli`

## Summary of Bug

It seems that configuring `clear_on_start = false` does not work. When this is set to `false`, Hermes should __not__ try to clear packets at startup. But this is not the case. Instead, we're hardcoding that the clearing should always happen

https://github.com/informalsystems/ibc-rs/blob/aef0e6a92736239d50048e35c73c5eca8c9b7814/relayer/src/link/relay_path.rs#L125

The problem was reported by a relayer operator on Bandchain network. They are probably the only ones using `clear_on_start = false` (I'm not aware of other operators using `false` here, and we don't have tests covering this feature, hence this bug went unnoticed for quite some time).

Logs folded here, but not sure they are truly useful. The more important part is the "steps to reproduce" described below.

<details><summary>Logs</summary>

```
Feb 11 02:46:48 ip-172-31-33-244 systemd[1]: Started hermes.
Feb 11 02:46:48 ip-172-31-33-244 bash[2935599]: 2022-02-11T02:46:48.204342Z  INFO ThreadId(01) using default configuration from '/home/ubuntu/.hermes/config.toml'
Feb 11 02:46:48 ip-172-31-33-244 bash[2935599]: 2022-02-11T02:46:48.205405Z  INFO ThreadId(01) [rest] address not configured, REST server disabled
Feb 11 02:46:48 ip-172-31-33-244 bash[2935599]: 2022-02-11T02:46:48.225481Z TRACE ThreadId(01) [injective-1] spawned chain runtime
Feb 11 02:46:48 ip-172-31-33-244 bash[2935599]: 2022-02-11T02:46:48.236287Z TRACE ThreadId(07) parsed version specification for <appd> @ -> SDK=0.44.5; Ibc-Go status=Some(Version { major: 2, minor: 0, patch: 2 })
Feb 11 02:46:48 ip-172-31-33-244 bash[2935599]: 2022-02-11T02:46:48.236338Z DEBUG ThreadId(07) running diagnostic on version info Specs { sdk_version: Version { major: 0, minor: 44, patch: 5 }, ibc_go_version: Some(Version { major: 2, minor: 0, patch: 2 }) }
Feb 11 02:46:48 ip-172-31-33-244 bash[2935599]: 2022-02-11T02:46:48.243593Z  INFO ThreadId(01) [injective-1] chain is healthy
[...]

PR #1879, opened 2022-02-15
Clean up packets clearing logic in `RelayPath`

adizere:
Closes: #1872 

## Description

The bug reported in #1872 exposed a regression: the a bug is that the `clear_on_start` option from Hermes config is ignored, and Hermes always clears packets when it starts up.

The bug appeared because we introduced duplicate logic for keeping track of the `clear_on_start` toggle:

1. we kept track of this toggle in the RelayPath, which was _incorrectly_ initialized to `true` https://github.com/informalsystems/ibc-rs/blob/1a3c1d2c50a8f207c93cce4b19b267ca6d7f3cfb/relayer/src/link/relay_path.rs#L125

2. we kept track of this toggle additionally in the packet command worker task, which was _correctly_ initialized with the `clear_on_start` value configured in the Hermes config.toml
https://github.com/informalsystems/ibc-rs/blob/b65baa9456b25500c49bf34896e79d2939c2ef65/relayer/src/worker.rs#L114-L121

The fix in this PR is to remove the logic from (1) RelayPath, and proposes to keep only the logic from (2) packet command worker. The nice benefit of adopting this fix is that it also simplifies slightly the code in RelayPath (which is one of the most complex parts of the relayer).

### Reproduce the problem
First, please follow the "Steps to Reproduce" from the original issue #1872 to confirm you can reproduce the problem on your local setup.

#### Test that disabling `clear_on_start` works

1. Setup a two chain network using `gm`. Suppose the two networks and `ibc-0` and `ibc-1`.
2. Setup a `channel-0` between the two networks.
3. Trigger and `ft-transfer` on `channel-0` of `ibc-0`
    - >  ./target/debug/hermes --json tx raw ft-transfer ibc-1 ibc-0 transfer channel-0 100 -d samoleans -t 1800 -n 1
    - __take note of the packet sequence number provided in the output__ for example:
    - > "sequence":9
4. Configure `clear_on_start` to be `false` in the config.toml of Hermes.
    - Make sure your log level is set to `log_level = 'debug'`.
    - Also set `clear_interval = 0` to completely disable periodic packet clearing.
5. Do `hermes start`.
   - Notice the log output and it should be visible that Hermes will __not__ clear the outstanding packets.
   - Hermes should not have any activity.
6. In a parallel terminal, trigger and `ft-transfer` using a second Hermes instance (similar to step 3)
    - >  ./target/debug/hermes --json tx raw ft-transfer ibc-1 ibc-0 transfer channel-0 100 -d samoleans -t 1800 -n 1
    - __take note of the packet sequence number provided in the output__:
    - > "sequence":10
7. In the first terminal where Hermes `start` is running, observe the log output, we should now notice some Hermes activity
    - the important part is we should see that the second packet (sequence number 10) is being relayed.
    - the first packet (sequence number 9 triggered at step 3) should remain unrelayed.
    - example log output
    - > 2022-02-15T10:20:02.962841Z DEBUG ThreadId(22) packet_cmd{src_chain=ibc-0 src_port=transfer src_channel=channel-0 dst_chain=ibc-1}:generate{id=tiLSMMUOjC}: /ibc.core.channel.v1.MsgRecvPacket from SendPacketEv(SendPacket - h:0-30551, seq:10, path:channel-0/transfer->channel-0/transfer, toh:0-0, tos:Timestamp(2022-02-15T10:49:52.478069Z)))

#### Test that enabling `clear_on_start` works

Redo the test from above, with the following adjustments:

- at step 4: Configure `clear_on_start` to be `true` (not `false`)
- at step 5: You should see that Hermes _has_ activity and should be relaying the old packet (triggered at step 3).

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/bug-fixes/relayer/1872-clear-packets.md
+++ b/.changelog/unreleased/bug-fixes/relayer/1872-clear-packets.md
@@ -1,1 +1,2 @@
-- Fixed clear_on_start bug ([#1872](https://github.com/informalsystems/ibc-rs/issues/1872))
+- Fixed bug where Hermes cleared packets at startup, despite
+ `clear_on_start = false` ([#1872](https://github.com/informalsystems/ibc-rs/issues/1872))
--- a/.changelog/unreleased/bug-fixes/relayer/1872-clear-packets.md
+++ b/.changelog/unreleased/bug-fixes/relayer/1872-clear-packets.md
@@ -1,0 +1,1 @@
+- Fixed clear_on_start bug ([#1872](https://github.com/informalsystems/ibc-rs/issues/1872))
--- a/relayer/src/link/relay_path.rs
+++ b/relayer/src/link/relay_path.rs
@@ -1,6 +1,5 @@
 use alloc::collections::BTreeMap as HashMap;
 use alloc::collections::VecDeque;
-use std::sync::{Arc, RwLock};
 use std::thread;
 use std::time::Instant;
 
@@ -51,7 +50,6 @@ use crate::link::pending::PendingTxs;
 use crate::link::relay_sender::{AsyncReply, SubmitReply};
 use crate::link::relay_summary::RelaySummary;
 use crate::link::{pending, relay_sender};
-use crate::util::lock::LockExt;
 use crate::util::queue::Queue;
 
 const MAX_RETRIES: usize = 5;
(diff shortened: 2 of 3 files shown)


mzabaluev (approved):
Oh great, one less `should_clear_packets`.

seanchen1991:
The code changes look good to me and my local testing exhibits the expected results 👍

seanchen1991 (approved):
The code changes look good to me and testing locally exhibits the expected behavior 👍

.changelog/unreleased/bug-fixes/relayer/1872-clear-packets.md
```diff
@@ -0,0 +1 @@
+- Fixed clear_on_start bug ([#1872](https://github.com/informalsystems/ibc-rs/issues/1872))
```
romac, line 1:
```suggestion
- Fixed bug where packets were being cleared at startup time, even with `clear_on_start = false`  ([#1872](https://github.com/informalsystems/ibc-rs/issues/1872))
```

Issue #1835, opened 2022-02-02
Test packet relaying on ordered channels


adizere:
## Summary

In preparation for releasing ICA on the hub, we would like to do some testing that channel ordering works as expected in Hermes.

## Proposal

- set up ordered channels between 2 chains
- provoke out-of-order packet relaying by either manually instrumenting the relayer code to drop the first event out of a batch, or by evicting transactions from thee node's mempool before they commit to the blockchain

## Acceptance Criteria

- capture a packet ordering corner-case in an automated test using the Rust framework
  - ensure that Hermes enforces ordering by looking at packet sequence numbers on an ordered channel

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

mzabaluev:
Testing this in a real setup appears to require a modified Gaia with the transfer module hacked to use ordered channels.

I guess we should go for a test case simulating out-or-order packet relaying in the test framework.

soareschen:
I have setup a Gaia branch that support ordered transfer channel at https://github.com/informalsystems/gaia/tree/v6.0.1-ordered. 

Currently the Nix build at Cosmos.nix is not working due to the `replace` directive being ignored by the Nix Go builder: https://github.com/informalsystems/cosmos.nix/pull/63

You can still build the Go project manually, and expose the custom `gaiad` in `$PATH`. Then replace the [bootstrap channel code](https://github.com/informalsystems/ibc-rs/blob/202ae4eb7c0c61363fb0bb9e6a09e6286f23831d/tools/integration-test/src/bootstrap/binary/channel.rs#L85) from `Order::Unordered` to `Order::Ordered`.

After this you should see the some tests like `test_ibc_transfer` still passing, but tests like `test_clear_packet` failing. In particular, `test_clear_packet` would keep retrying because [`worker_stubborn_strategy` is being used here](https://github.com/informalsystems/ibc-rs/blob/202ae4eb7c0c61363fb0bb9e6a09e6286f23831d/relayer/src/worker/packet.rs#L106).

adizere:
Great progress!

[...]

PR #1887, opened 2022-02-17
Fix relayer behavior on ordered channels

soareschen:
Closes: #1835

## Description

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


relayer/src/worker/packet.rs
```diff
@@ -11,11 +11,27 @@ use crate::object::Packet;
...
+    use retry::delay::Fibonacci;
+
+    use crate::util::retry::clamp_total;
+
+    // Default parameters for the retrying mechanism
+    const MAX_DELAY: Duration = Duration::from_secs(2); // 1 minute
+    const MAX_TOTAL_DELAY: Duration = Duration::from_secs(5); // 10 minutes
+    const INITIAL_DELAY: Duration = Duration::from_secs(1); // 1 second
+
+    pub fn default() -> impl Iterator<Item = Duration> {
```
seanchen1991, line 30: Could we add some documentation for this to briefly explain the logic of the default retry strategy? Putting the logic in a `default` method doesn't give any indication via the naming what the strategy is.

soareschen: Yeah I don't really like the `retry` crate we are currently using. It is overly complicated and do not blend well with the idiomatic use of `Result`. It is worth refactoring to simplify the logic.

For this case, I'm just copying the boilerplate from other code, because the current strategies are not very customizable. We'll figure out a proper strategy in a proper PR. Here we just want to identify the place that needs to be modified for the relayer to fail gracefully.

seanchen1991: Ok, should probably add an issue or TODO to catalog this 🙂

adizere:
Ready for review. 

@crodriguezvega & team: let's use this PR for testing ICA.

relayer/src/worker/packet.rs
adizere: I am reverting these changes as it is unclear if they are necessary. I agree this is the right place to customize the retry strategy in case we need to adapt the packet command worker retrying behavior (but that's not clear to me we need to do). 

8eea79fd

adizere:
@soareschen we have a [hiccup with gaiad v6.0.2 tests](https://github.com/informalsystems/ibc-rs/runs/5301135913?check_suite_focus=true):

>error: flake 'github:informalsystems/cosmos.nix/gaia-ordered' does not provide attribute 'packages.x86_64-linux.gaia6_0_2', 'legacyPackages.x86_64-linux.gaia6_0_2' or 'gaia6_0_2'
Error: Process completed with exit code 1.

soareschen:
Oh yeah, I need to create a new branch in Cosmos.nix to have the ordered channel updated with changes from master.

soareschen:
Now it's fixed.

adizere (approved):

Issue #1885, opened 2022-02-17
Config reload results in panic


adizere:
## Summary of Bug

It seems that the configuration reload functionality in Hermes does not work. If triggered, it results into a panic and the process quits.

## Version

hermes 0.11.1+5ddec6d2

## Steps to Reproduce

1. do `hermes start`, make sure you have at least 2 chains
    - in my setup, a few tasks related to client refresh & misbehavior detection are starting up (see logs below)
    - let hermes start up completely, until " Hermes has started" log appears
2. in a separate terminal, do 'kill -SIGHUP `pgrep hermes`'
    - no need to actually change anything in the configuration file to provoke the panic

```
2022-02-17T10:01:36.162478Z  INFO ThreadId(01) Hermes has started
2022-02-17T10:01:50.999784Z  INFO ThreadId(1123) reloading configuration (triggered by SIGHUP)
2022-02-17T10:01:51.002623Z  INFO ThreadId(1123) configuration successfully reloaded
2022-02-17T10:01:51.275033Z  INFO ThreadId(1122) cmd: updating existing chain chain=ibc-0
2022-02-17T10:01:51.275410Z  INFO ThreadId(1122) cmd: removing existing chain chain=ibc-0
2022-02-17T10:01:51.275808Z DEBUG ThreadId(1122) cmd: shutting down workers chain=ibc-0
2022-02-17T10:01:51.674491Z DEBUG ThreadId(51) DetectMisbehaviorWorker{client=07-tendermint-0 src_chain=ibc-1 dst_chain=ibc-0}: task terminated
2022-02-17T10:01:52.289031Z DEBUG ThreadId(20) refresh{client=07-tendermint-0 src_chain=ibc-1 dst_chain=ibc-0}: task terminated
2022-02-17T10:01:52.480667Z DEBUG ThreadId(269) DetectMisbehaviorWorker{client=07-tendermint-1 src_chain=ibc-1 dst_chain=ibc-0}: task terminated
2022-02-17T10:01:52.759909Z DEBUG ThreadId(52) refresh{client=07-tendermint-1 src_chain=ibc-1 dst_chain=ibc-0}: task terminated
2022-02-17T10:01:52.822388Z DEBUG ThreadId(577) DetectMisbehaviorWorker{client=07-tendermint-2 src_chain=ibc-1 dst_chain=ibc-0}: task terminated
[...]

PR #1894, opened 2022-02-18
Disable config reload and throw a warning on receiving SIGHUP

seanchen1991:
Closes: #1885

## Description

Disables configuration reloading upon receiving a SIGHUP signal and throws a warning notifying the user of this fact and that the feature will be deprecated in the future. 

When a SIGHUP signal is sent to Hermes, the logs now look like this
```
2022-02-18T16:04:12.715032Z  INFO ThreadId(01) Hermes has started
2022-02-18T16:04:24.764642Z  WARN ThreadId(25) configuration reloading via SIGHUP has been disabled and will be deprecated in the future
```
______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/relayer-cli/src/commands/start.rs
+++ b/relayer-cli/src/commands/start.rs
@@ -71,7 +71,7 @@ fn register_signals(tx_cmd: Sender<SupervisorCmd>) -> Result<(), io::Error> {
     std::thread::spawn(move || {
         for signal in &mut signals {
             match signal {
-                SIGHUP => warn!("configuration reloading via SIGHUP has been disabled and will be deprecated in the future"),
+                SIGHUP => warn!("configuration reloading via SIGHUP has been disabled. The signal handler will be removed in the future"),
                 SIGUSR1 => {
                     info!("Dumping state (triggered by SIGUSR1)");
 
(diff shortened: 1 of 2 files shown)


relayer-cli/src/commands/start.rs
```diff
@@ -59,9 +59,9 @@ impl Runnable for StartCmd {
 }
 
 /// Register the SIGHUP and SIGUSR1 signals, and notify the supervisor.
-/// - SIGHUP: Trigger a reload of the configuration.
+/// - [DEPRECATED] SIGHUP: Trigger a reload of the configuration.
 /// - SIGUSR1: Ask the supervisor to dump its state and print it to the console.
-fn register_signals(reload: ConfigReload, tx_cmd: Sender<SupervisorCmd>) -> Result<(), io::Error> {
+fn register_signals(_reload: ConfigReload, tx_cmd: Sender<SupervisorCmd>) -> Result<(), io::Error> {
```
adizere, line 64: Should we also mark the `ConfigReload` struct or its `reload` method as `#[deprecated]` for extra caution around that module?

ref: https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute

adizere (approved):
Pre-approving, thanks for the quick fix Sean!

Will let @romac have also a look since he knows the reloading module best.

relayer-cli/src/commands/start.rs
romac: I don't think it's needed since we don't expect anyone to rely on that programatically.

romac:
LGTM modulo a nitpick. I'll take care of removing the config reload code in a follow-up PR.

romac (approved):

Issue #1895, opened 2022-02-21
Allow overriding the tracing filter with the `RUST_LOG` environment variable


romac:
## Crate

`ibc-relayer-cli`

## Summary

The current tracing filter is [hardcoded](https://github.com/informalsystems/ibc-rs/blob/b420a12d677df2c489a002be4cf88839270e48ad/relayer-cli/src/components.rs#L75-L82) and cannot be overridden, eg. to enable debug output for the `tendermint-rpc` crate as one would otherwise do with:

```shell
RUST_LOG=tendermint_rpc=debug hermes start
```

## Problem Definition

For debugging purposes, we would like to enable tracing for dependencies of the relayer, eg. the `tendermint-rpc` crate, but this is not possible at the moment.

## Proposal

Update the tracing filter builder to use the tracing filter specified via the `RUST_LOG` environment variable.

## Acceptance Criteria

```shell
RUST_LOG=tendermint_rpc=debug hermes start
```

shows the debug output of the `tendermint-rpc` crate.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1896, opened 2022-02-21
Allow overriding tracing filter with `RUST_LOG` env var

romac:
Closes: #1895

## Usage

With this PR, one can set the `RUST_LOG` variable to `tendermint_rpc=debug,ibc_relayer=info,ibc_relayer_cli=info` in order to enable debug output for `tendermint-rpc` and info output for the relayer.

```shell
RUST_LOG=tendermint_rpc=debug,ibc_relayer=info,ibc_relayer_cli=info hermes start
```

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/relayer-cli/src/commands/start.rs
+++ b/relayer-cli/src/commands/start.rs
@@ -33,7 +33,7 @@ impl Runnable for StartCmd {
         let config = (*app_config()).clone();
         let config = Arc::new(RwLock::new(config));
 
-        let supervisor_handle = make_supervisor::<ProdChainHandle>(config.clone(), self.full_scan)
+        let supervisor_handle = make_supervisor::<ProdChainHandle>(config, self.full_scan)
             .unwrap_or_else(|e| {
                 Output::error(format!("Hermes failed to start, last error: {}", e)).exit()
             });
--- a/relayer-cli/src/components.rs
+++ b/relayer-cli/src/components.rs
@@ -75,7 +75,7 @@ pub fn enable_ansi() -> bool {
 /// Builds a tracing filter based on the input `log_level`.
 /// Enables tracing exclusively for the relayer crates.
 /// Returns error if the filter failed to build.
-fn build_tracing_filter(log_level: LogLevel) -> Result<EnvFilter, FrameworkError> {
+fn build_tracing_filter(default_level: LogLevel) -> Result<EnvFilter, FrameworkError> {
     let directive = std::env::var(HERMES_LOG_VAR).unwrap_or_else(|_| {
         let target_crates = ["ibc_relayer", "ibc_relayer_cli"];
 
--- a/relayer-cli/src/components.rs
+++ b/relayer-cli/src/components.rs
@@ -92,7 +92,7 @@ fn build_tracing_filter(default_level: LogLevel) -> Result<EnvFilter, FrameworkE
         Ok(out) => Ok(out),
         Err(e) => {
             eprintln!(
-                "Unable to initialize Hermes from filter directive {:?}: {}",
+                "ERROR: unable to initialize Hermes with log filtering directive {:?}: {}",
                 directive, e
             );
 
--- a/relayer-cli/src/components.rs
+++ b/relayer-cli/src/components.rs
@@ -72,20 +72,26 @@ pub fn enable_ansi() -> bool {
     atty::is(atty::Stream::Stdout) && atty::is(atty::Stream::Stderr)
 }
 
+/// The relayer crates targeted by the default log level.
+const TARGET_CRATES: [&str; 2] = ["ibc_relayer", "ibc_relayer_cli"];
+
+/// Build a tracing directive setting the log level for the relayer crates to the
+/// given `log_level`.
+fn default_directive(log_level: LogLevel) -> String {
+    use itertools::Itertools;
+
+    TARGET_CRATES
+        .iter()
+        .map(|&c| format!("{}={}", c, log_level))
+        .join(",")
+}
+
 /// Builds a tracing filter based on the input `log_level`.
 /// Enables tracing exclusively for the relayer crates.
 /// Returns error if the filter failed to build.
 fn build_tracing_filter(default_level: LogLevel) -> Result<EnvFilter, FrameworkError> {
-    let directive = std::env::var(HERMES_LOG_VAR).unwrap_or_else(|_| {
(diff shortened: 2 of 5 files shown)


relayer-cli/src/components.rs
```diff
@@ -71,27 +75,29 @@ pub fn enable_ansi() -> bool {
...
-
-    // SAFETY: unwrap() below works as long as `target_crates` is not empty.
-    let directive_raw = target_crates
-        .iter()
-        .map(|&c| format!("{}={}", c, log_level))
-        .reduce(|a, b| format!("{},{}", a, b))
-        .unwrap();
+fn build_tracing_filter(log_level: LogLevel) -> Result<EnvFilter, FrameworkError> {
+    let directive = std::env::var(HERMES_LOG_VAR).unwrap_or_else(|_| {
+        let target_crates = ["ibc_relayer", "ibc_relayer_cli"];
```
adizere, line 80: I see here we're using `RUST_LOG` not only to replace the log level, but also the crates/modules. Can we give an example on how we're supposed to use `RUST_LOG` to overwrite the tracing parameters?

relayer-cli/src/components.rs
```diff
@@ -1,10 +1,14 @@
...
 use tracing_subscriber::{filter::EnvFilter, util::SubscriberInitExt, FmtSubscriber};
 
-use ibc_relayer::config::GlobalConfig;
+use ibc_relayer::config::{GlobalConfig, LogLevel};
 
 use crate::config::Error;
 
+/// The name of the environment variable through which one can override
+/// the tracing filter built in [`build_tracing_filter`].
+const HERMES_LOG_VAR: &str = "RUST_LOG";
```
adizere, line 10: I recall the integration tests framework making use of `RUST_LOG`, eg in `cargo test`. Would just like to flag that -- is there any interference between this approach and the pre-existing support for `RUST_LOG` in cargo?

adizere (commented):
Left 2 questions, nothing serious.

relayer-cli/src/commands/start.rs
```diff
@@ -33,7 +33,7 @@ impl Runnable for StartCmd {
         let config = (*app_config()).clone();
         let config = Arc::new(RwLock::new(config));
 
-        let supervisor_handle = make_supervisor::<ProdChainHandle>(config.clone(), self.full_scan)
+        let supervisor_handle = make_supervisor::<ProdChainHandle>(config, self.full_scan)
```
mzabaluev, line 36: Less cloning of massive data structures, good.

relayer-cli/src/components.rs
```diff
@@ -71,27 +75,29 @@ pub fn enable_ansi() -> bool {
...
-fn build_tracing_filter(log_level: String) -> Result<EnvFilter, FrameworkError> {
-    let target_crates = ["ibc_relayer", "ibc_relayer_cli"];
-
-    // SAFETY: unwrap() below works as long as `target_crates` is not empty.
-    let directive_raw = target_crates
-        .iter()
-        .map(|&c| format!("{}={}", c, log_level))
-        .reduce(|a, b| format!("{},{}", a, b))
-        .unwrap();
+fn build_tracing_filter(log_level: LogLevel) -> Result<EnvFilter, FrameworkError> {
```
mzabaluev, line 78:
```suggestion
fn build_tracing_filter(default_level: LogLevel) -> Result<EnvFilter, FrameworkError> {
```
Might be useful to rename the argument for clarity, because it is ignored when the `RUST_LOG` environment variable is set.

relayer-cli/src/components.rs
```diff
@@ -71,27 +75,29 @@ pub fn enable_ansi() -> bool {
...
+            .expect("non empty list of target crates")
+    });
 
     // Build the filter directive
-    match EnvFilter::try_new(directive_raw.clone()) {
+    match EnvFilter::try_new(&directive) {
         Ok(out) => Ok(out),
         Err(e) => {
             eprintln!(
                 "Unable to initialize Hermes from filter directive {:?}: {}",
```
mzabaluev, line 95:
```suggestion
                "Unable to initialize Hermes with log filtering directive {:?}: {}",
```

relayer-cli/src/components.rs
romac: Yes, using `RUST_LOG` will override everything. I'll add an example, perhaps something like:

```RUST_LOG=tendermint_rpc=debug,ibc_relayer=info,ibc_relayer_cli=info hermes start```

relayer-cli/src/components.rs
romac: The logs from the integration tests look fine to me: https://github.com/informalsystems/ibc-rs/runs/5272869928?check_suite_focus=true (modulo some doc tests failures, unrelated to this PR and best fixed elsewhere)

romac: By the way, just to clarify, `cargo test` does not care itself about `RUST_LOG` but the integration tests themselves do, cf. https://github.com/informalsystems/ibc-rs/blob/8329d856844df30651f68a25ca5c5db4addbaa2b/tools/integration-test/src/bootstrap/single.rs#L75

romac: What the test suite does is look at the `RUST_LOG` env variable, and then sets the `log_level` property in `config.toml` to that. If the test suite is only ever run with `RUST_LOG=info/trace/debug/error/warn` then everything should work as expected with no change in the output.

adizere: > By the way, just to clarify, `cargo test` does not care itself about `RUST_LOG` but the integration tests themselves do, cf.

Thanks for clarifying that! This part wasn't obvious to me.

adizere (approved):
Approved, but I still don't see any documentation on how to use `RUST_LOG`. Please remember to add that either in guide (release PR) or here.

relayer-cli/src/commands/start.rs
seanchen1991: Looks like it was cloning an `Arc` though, so supposedly it should have just consisted of incrementing a reference count.

mzabaluev: It likely does not matter in the startup code, but in general dereferencing an `Arc` is still cheaper than atomically incrementing a ref count.

Issue #1288, opened 2021-08-16
`tx raw upgrade-chain` gives 'Failed Tx: no confirmation' error


hu55a1n1:
## Crate
relayer

## Summary of Bug
Step 3 of the Hermes guide's [testing client upgrade](https://hermes.informal.systems/commands/upgrade/test.html) procedure gives a `Failed Tx: no confirmation` error, although the tx itself succeeds -> 
```
Aug 13 21:06:49.881 DEBUG [ibc-0] waiting for commit of block(s) 3EB2DF97AB3D2FC94918F5FC7969286E357722DECDDD65A24DDEAEE970447E43
Error: 
   0: upgrade chain error
   1: failed while submitting the Transfer message to chain ibc-0
   2: Failed Tx: no confirmation

Location:
   /home/hussaini/.cargo/registry/src/github.com-1ecc6299db9ec823/flex-error-0.4.2/src/tracer_impl/eyre.rs:10

Backtrace omitted.
```
Apparently, this is because the transaction doesn't result in the creation of events and Hermes interprets this as a 'no confirmation' error.

## Version
v0.6.2

## Steps to Reproduce
Run steps 1 to 3 of the Hermes guide's [testing client upgrade](https://hermes.informal.systems/commands/upgrade/test.html).

## Acceptance Criteria
Hermes is able to detect that the transaction succeeded and prints a success message.
____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
Also note that governane proposals belong to the SDK, not the IBC module, cf a remark by @hu55a1n1.

GovernanceProposals have a content field. 
https://github.com/informalsystems/ibc-rs/blob/ba43f88d6813fb3fe8a60617d797de238df21955/relayer/src/upgrade_chain.rs#L131

For chain upgrades gov proposals that affect the IBC state, this `content` field is an IBC type from `ibc.core.client.v1`
https://github.com/informalsystems/ibc-rs/blob/773e50c31fb17e1b6655a1a1d5b44883c679f229/proto/src/prost/std/ibc.core.client.v1.rs#L57

[...]

Issue #1845, opened 2022-02-05
`height.increment()` not needed for Beefy


livelybug:
Hello,

When implementing Beefy protocol, based on Grandpa consensus, there are 3 kinds of verifications in a hierarchical structure.
* The top tier verification is MMR root verification, by collecting sufficient validators' valid signatures. 
* Based on a valid MMR root, the 2nd tier is [block header verification](https://github.com/informalsystems/ibc-rs/issues/1775#issuecomment-1016043500).
* Based on a valid block header,  the 3rd tier is [on-chain storage verification](https://github.com/octopus-network/substrate/blob/33c518ebbe43d38228ac47e793e4d1c76738a56d/primitives/state-machine/src/lib.rs?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L1078), which calculates the correct on-chain storage by the state root, the storage proof, and the storage key. A successful verification needs the state root and storage proof to be from the same block. However, as the `height` is increased by one when composing proofs of [connection](https://github.com/informalsystems/ibc-rs/blob/e16d5f64fdf7bbd6e0fa9449a567ad9ee1b3c180/relayer/src/chain.rs#L407),  [channel](https://github.com/informalsystems/ibc-rs/blob/e16d5f64fdf7bbd6e0fa9449a567ad9ee1b3c180/relayer/src/chain.rs#L425), and [packet](https://github.com/informalsystems/ibc-rs/blob/e16d5f64fdf7bbd6e0fa9449a567ad9ee1b3c180/relayer/src/chain.rs#L457), the corresponding state root always belong to the header of block height `N + 1`, while the storage proof belongs to block height `N`, resulting in a failed verification

Understand the `height.increment()` may be for Tendermint consensus, is it possible to compose a trait for it, and allow consensus-specific implementation?

Thank you

lyqingye:
I guess it's because the consensusstate contains the app hash (merkleroot) of the previous block, so PacketProof Height  = ConsensusState Height + 1

adizere:
Hey @livelybug and @lyqingye, many thanks for documenting this!

[...]

Issue #1580, opened 2021-11-19
Split Cosmos-specific code out into its own crate


romac:
## Crate

`ibc-relayer`

## Summary

Split out the `CosmosSdk` struct, its `ChainEndpoint` implementation, and all related Cosmos-specific code into a `ibc-relayer-cosmos` crate.

## Problem Definition

While the relayer was developed against and currently only supports Cosmos chains, this will have to change in the future with non-Cosmos chains connecting via IBC. Because it is likely that each and every non-Cosmos chain will require specific support and tailored code in the relayer, the `ibc-relayer` library will grow in size and in scope way past a healthy point for a library.

## Proposal

Split out all the Cosmos-specific code out of the `ibc-relayer` crate and into a new `ibc-relayer-cosmos` crate.

### Code to split out
> TBD

## Acceptance Criteria

The `ibc-relayer` crate does not contain any Cosmos-specific code and a new `ibc-relayer-cosmos` crate is added to the workspace.

## Potential concerns

- Cyclic dependencies between `ibc-relayer` and `ibc-relayer-cosmos`, may require introducing a `ibc-relayer-base` crate or similar to host the traits to be implemented in the `-cosmos` crate.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

mzabaluev:
> Cyclic dependencies between `ibc-relayer` and `ibc-relayer-cosmos`

This can be avoided if the entry API for setting up a Cosmos relayer is made public in `ibc-relayer-cosmos`, and `ibs-relayer` is a base library agnostic of specific chains.

PR #1909, opened 2022-02-22
Extract tx confirmation event on `tx raw upgrade-chain` cmd

mzabaluev:
Closes: #1288

## Description

Process the confirmation to the `UpgradeProposal` message as an extended event outside of the `IbcEvent` API.

* Add the `chain::cosmos::events` module to the `relayer` crate.
* In the module, provide a `CosmosEvent` enum for events specific to Cosmos SDK-based chain API.
* Add the `SubmitProposal` variant to `CosmosEvent`, with `SubmitProposal` struct as the value carrying proposal event data.
* Add an inherent method `send_message_and_wait_commit_cosmos` for `CosmosSdkChain`, as a way to wait for an expected Cosmos-specific confirmation event to a message.
* Use this method in the implementation of the `tx raw upgrade-chain` command.
* The internal method `wait_for_block_commits` is modified to try to extract Cosmos-specific events from an ABCI event list, in addition to any IBC events.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


modules/src/events.rs
```diff
@@ -285,6 +292,8 @@ impl fmt::Display for IbcEvent {
             IbcEvent::TimeoutPacket(ev) => write!(f, "TimeoutPacketEv({})", ev),
             IbcEvent::TimeoutOnClosePacket(ev) => write!(f, "TimeoutOnClosePacketEv({})", ev),
 
+            IbcEvent::SubmitProposal(ev) => write!(f, "SubmitProposalEv({})", ev),
```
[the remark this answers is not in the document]
romac: No idea, let's remove the `Ev` suffix.

modules/src/events.rs
```diff
@@ -322,15 +333,18 @@ impl TryFrom<IbcEvent> for AbciEvent {
...
-    } else if let Some(mut conn_res) = ConnectionEvents::try_from_tx(event) {
+    } else if let Some(mut conn_res) = connection_events::try_from_tx(event) {
         conn_res.set_height(height);
         Some(conn_res)
-    } else if let Some(mut chan_res) = ChannelEvents::try_from_tx(event) {
+    } else if let Some(mut chan_res) = channel_events::try_from_tx(event) {
         chan_res.set_height(height);
         Some(chan_res)
+    } else if let Some(mut gov_res) = gov_events::try_from_tx(event) {
+        gov_res.set_height(height);
```
mzabaluev, line 346: I have cargo-culted this from the other event types; not sure if it makes sense in this case.

adizere: I have a slight preference to leave `ibc::events::from_tx_response_event` untouched, to remain an IBC-module-specific method, with IBC specific functionality.

Cosmos SDK-specific stuff can live in the `ibc-relayer` crate, since only the relayer needs it (not the IBC module itself). So the suggestion would be to:

- undo the changes in `from_tx_response_event`
- move `modules/src/cosmos/gov/events.rs` -> `relayer/src/chain/cosmos/events.rs`
- modify method [all_ibc_events_from_tx_search_response](https://github.com/informalsystems/ibc-rs/blob/b3f9e4ce39cd3133ae9d88de1541f9f795e9ddc9/relayer/src/chain/cosmos.rs#L2229) to handle two types of events:
    - ibc events, which [it does already](https://github.com/informalsystems/ibc-rs/blob/b3f9e4ce39cd3133ae9d88de1541f9f795e9ddc9/relayer/src/chain/cosmos.rs#L2244-L2247) via method `from_tx_response_event`
    - sdk events, which it can do via the new module `cosmos::events::gov_events::try_from_tx(..)`
- we should also rename 

The solution could look as follows:

```diff
diff --git a/relayer/src/chain/cosmos.rs b/relayer/src/chain/cosmos.rs
index e07f0a5d..f817b167 100644
--- a/relayer/src/chain/cosmos.rs
+++ b/relayer/src/chain/cosmos.rs
@@ -2226,7 +2226,7 @@ fn update_client_from_tx_search_response(
         .map(IbcEvent::UpdateClient)
 }

-fn all_ibc_events_from_tx_search_response(chain_id: &ChainId, response: ResultTx) -> Vec<IbcEvent> {
+fn all_events_from_tx_search_response(chain_id: &ChainId, response: ResultTx) -> Vec<IbcEvent> {
     let height = ICSHeight::new(chain_id.version(), u64::from(response.height));
     let deliver_tx_result = response.tx_result;
     if deliver_tx_result.code.is_err() {
@@ -2245,6 +2245,10 @@ fn all_ibc_events_from_tx_search_response(chain_id: &ChainId, response: ResultTx
             trace!("got IBC event: {:#?}", ibc_ev);
             result.push(ibc_ev);
         }
+        if let Some(sdk_ev) = cosmos::events::sdk_events::from_tx_response_event(height, &event) {
+            trace!("got SDK event: {:#?}", ibc_ev);
+            result.push(sdk_ev);
+        }
     }
```

adizere (changes requested):
Have not tested yet. Left an idea for a possible improvement to make the separation IBC <> SDK cleaner. The idea is to keep the `ibc` crate purely about IBC. It's ok to have SDK-related logic in the `ibc-relayer`, so that's where the new module `gov/events.rs` should land.

romac:
Maybe we should start looking into https://github.com/informalsystems/ibc-rs/issues/1580?

adizere:
> Maybe we should start looking into #1580?

Indeed. That issue may be a big lift. We'll need it for a stable v1 release. I think PRs such as the present one and #1845 are nice sub-problems of #1580, so its a good exercise to fix them before that one.

modules/src/events.rs
mzabaluev: The difficulties I've run into with the recommended approach is, the result of `ChainHandle::query_txs` et al. is currently `Vec<IbcEvent>`, and remaking all of these uses to return a new combined event type is quite a slog. @adizere do you think we should commit to such a redesign, or could there be another way to solve this?

I've tried to come up with a simpler solution, special-casing the SDK event handling aside from the generic `ChainHandle` API and adding a cosmos-specific method to call directly, but such a refactor looks hackish and I feel that we should not do this for every code path where an SDK-specific event might be used to confirm a transaction.

mzabaluev:
This is turning into a bigger rework, with the following options:

| Solution | Advantages | Problems |
| ----------- | --------------- | ------------- |
| Extend `IbcEvent` with Cosmos-specific variants | Already implemented in the branch | Breaks modularity, adds Cosmos specific data to `ibc` |
| Enum unifying `IbcEvent` with `SdkEvent` in `relayer` code | Modularity preserved | Lots of tedious changes to code that only really needs `IbcEvent` at this point |
| Extend `IbcEvent` with a dynamically encoded variant to pass custom events back to CLI | Modularity preserved | Soft-typed data added to the API |

adizere:
Thank you for the breakdown Mikhail! This is helpful. I started out assuming option 2 is ideal, but it may be more wise to proceed with 1 (or even 3) and commit to a refactor to `IbcEvent` later.

romac:
Agreed, I vote for option 1. Once we get to add proper support for non-Cosmos chains (together with https://github.com/informalsystems/ibc-rs/issues/1580), we can move the SDK-specific event out of the `ibc` crate. Let's just make sure to document that this is temporary and that non-Cosmos chains which implement IBC using the `ibc` crate need not care about this event.

romac:
You mean #1979, right?

mzabaluev:
I have reworked the solution, taking none of the approaches discussed in https://github.com/informalsystems/ibc-rs/pull/1909#issuecomment-1065095205, but instead adding an entirely different path for receiving Cosmos-specific confirmation events.

modules/src/events.rs
```diff
@@ -261,32 +250,31 @@ impl fmt::Display for IbcEvent {
...
+            IbcEvent::OpenConfirmChannel(ev) => write!(f, "OpenConfirmChannel({:?})", ev),
+            IbcEvent::CloseInitChannel(ev) => write!(f, "CloseInitChannel({})", ev),
+            IbcEvent::CloseConfirmChannel(ev) => write!(f, "CloseConfirmChannel({:?})", ev),
+
+            IbcEvent::SendPacket(ev) => write!(f, "SendPacket({})", ev),
+            IbcEvent::ReceivePacket(ev) => write!(f, "ReceivePacket({})", ev),
+            IbcEvent::WriteAcknowledgement(ev) => write!(f, "WriteAcknowledgement({})", ev),
+            IbcEvent::AcknowledgePacket(ev) => write!(f, "AcknowledgePacket({})", ev),
+            IbcEvent::TimeoutPacket(ev) => write!(f, "TimeoutPacket({})", ev),
+            IbcEvent::TimeoutOnClosePacket(ev) => write!(f, "TimeoutOnClosePacket({})", ev),
```
adizere, line 275: Intersting. I wonder why are these lines marked as changed since I don't think there should be any diff.

adizere (commented):
I went through this quickly and I have some minor remarks only.

Still testing.

> I have reworked the solution, taking none of the approaches discussed in https://github.com/informalsystems/ibc-rs/pull/1909#issuecomment-1065095205, but instead adding an entirely different path for receiving Cosmos-specific confirmation events.

Considering this new approach, I’m wondering what is the best way to choose a winner between #1979 and #1909. Any ideas or recommendations?

mzabaluev:
> Considering this new approach, I’m wondering what is the best way to choose a winner between https://github.com/informalsystems/ibc-rs/pull/1979 and https://github.com/informalsystems/ibc-rs/pull/1909. Any ideas or recommendations?

This one (#1909) is meant to "get the job done" as per the original issue: wait for a confirmation event and print it out.

There are still issues with it, though: as `query_txs` has been left unmodified, the `query tx events` command would not print the extra events.

mzabaluev:
> There are still issues with it, though: as `query_txs` has been left unmodified, the `query tx events` command would not print the extra events.

This can also can be modified to use the `query_extended_events_for_tx`: https://github.com/informalsystems/ibc-rs/blob/1ca72521d3243de7af8d809b87f3c5dfe625cb7c/relayer-cli/src/commands/query/tx/events.rs#L55
But I'm not sure about how to modify the output: a tuple of two event arrays? Or a struct-like output with `ibc_events` and `cosmos_events` fields? Maybe some flattened event array (partially going back to the option 2 from https://github.com/informalsystems/ibc-rs/pull/1909#issuecomment-1065095205)?

relayer/src/chain/cosmos.rs
```diff
@@ -873,6 +878,12 @@ impl CosmosSdkChain {
                             if !events_per_tx.is_empty() {
                                 *events = events_per_tx;
                             }
+                        } else if let Ok(cosmos_events_per_tx) =
+                            self.query_cosmos_events_for_tx(QueryTxHash(response.hash))
+                        {
+                            if !cosmos_events_per_tx.is_empty() {
```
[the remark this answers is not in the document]
mzabaluev: I have reworked this to only perform a single query and extract both kinds of events from the response.
Updated the comments accordingly.

modules/src/events.rs
mzabaluev: This removes the unnecessary "Ev" suffixes from the text repr, as discussed in https://github.com/informalsystems/ibc-rs/pull/1909#discussion_r812036743

relayer/src/chain/cosmos.rs
```diff
@@ -2371,8 +2458,10 @@ fn client_id_suffix(client_id: &ClientId) -> Option<u64> {
 pub struct TxSyncResult {
     // the broadcast_tx_sync response
     response: Response,
-    // the events generated by a Tx once executed
+    // the IBC events generated by a Tx once executed
     events: Vec<IbcEvent>,
```
seanchen1991, line 2462: Now that there's a `cosmos_events` field, this field should also probably be named more explicitly. 
```suggestion
    ibc_events: Vec<IbcEvent>,
```

relayer/src/chain/cosmos/events.rs
```diff
@@ -0,0 +1,57 @@
+//! Extension events specific to chains followed via Cosmos SDK.
```
seanchen1991, line 1: I'm not quite sure what "followed via Cosmos SDK" means. Would "chains that adhere to the Cosmos SDK" be more accurate?

relayer/src/chain/cosmos/events.rs
```diff
@@ -0,0 +1,57 @@
...
+
+use tendermint::abci::tag::Tag;
+use tendermint::abci::Event as AbciEvent;
+
+use ibc::Height;
+
+// Event type strings
+const SUBMIT_PROPOSAL_EVENT: &str = "submit_proposal";
+
+#[derive(Debug, Serialize)]
```
seanchen1991, line 13:
```suggestion
/// Variants that encode the possible message types sent via Cosmos SDK modules
#[derive(Debug, Serialize)]
```

relayer/src/chain/cosmos/events.rs
```diff
@@ -0,0 +1,57 @@
...
+pub enum CosmosEvent {
+    SubmitProposal(SubmitProposal),
```
seanchen1991, line 15:
```suggestion
    /// An event sent via the governance module indicating a new governance proposal was submitted
    SubmitProposal(SubmitProposal),
```

seanchen1991 (changes requested):

relayer/src/chain/cosmos.rs
```diff
@@ -930,16 +939,69 @@ impl CosmosSdkChain {
...
+                1,
+                1, // get only the first Tx matching the query
+                Order::Ascending,
+            ))
+            .map_err(|e| Error::rpc(self.config.rpc_addr.clone(), e))?;
+
+        if response.txs.is_empty() {
+            Ok((vec![], vec![]))
+        } else {
+            let tx = response.txs.remove(0);
```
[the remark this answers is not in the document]
mzabaluev: Good point, though I suspect the win will be negligible in most cases.

relayer/src/chain/cosmos/events.rs
romac: _chains built on the Cosmos SDK_ ?

relayer/src/chain/cosmos/events.rs
```diff
@@ -0,0 +1,59 @@
...
+            }
+            _ => None,
+        }
+    }
+}
+
+/// Data of a "submit_proposal" event carrying a "proposal_id" tag.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub struct SubmitProposal {
+    pub proposal_id: u64,
```
romac, line 37: Let's introduce a new type for this?

mzabaluev: I'd rather not overdo it at this point, since we don't have any use of the field besides dumping it in the command output.

mzabaluev:
The clippy lints are unrelated, will be fixed by #2079.

relayer/src/chain/cosmos.rs
```diff
@@ -2371,8 +2453,10 @@ fn client_id_suffix(client_id: &ClientId) -> Option<u64> {
 pub struct TxSyncResult {
     // the broadcast_tx_sync response
     response: Response,
-    // the events generated by a Tx once executed
-    events: Vec<IbcEvent>,
+    // the IBC events generated by a Tx once executed
```
seanchen1991, line 2456:
```suggestion
    /// The IBC events generated by a Tx once executed
```

relayer/src/chain/cosmos.rs
```diff
@@ -2371,8 +2453,10 @@ fn client_id_suffix(client_id: &ClientId) -> Option<u64> {
...
+    ibc_events: Vec<IbcEvent>,
+    // Cosmos-specific events generated by a Tx once executed
```
seanchen1991, line 2458:
```suggestion
    /// Cosmos-specific events generated by a Tx once executed
```

relayer/src/chain/cosmos.rs
```diff
@@ -2371,8 +2453,10 @@ fn client_id_suffix(client_id: &ClientId) -> Option<u64> {
 pub struct TxSyncResult {
     // the broadcast_tx_sync response
```
seanchen1991, line 2454:
```suggestion
    /// The broadcast_tx_sync response
```

seanchen1991 (approved):
Just a few nits. Otherwise, looks good to me 🙂

relayer/src/chain/cosmos.rs
```diff
@@ -930,16 +939,69 @@ impl CosmosSdkChain {
...
+                let cosmos_events = all_cosmos_events_from_tx_search_response(self.id(), &tx);
+                Ok((ibc_events, cosmos_events))
+            }
+        }
+    }
+
+    /// A Cosmos-specific variant of `send_messages_and_wait_commit`,
+    /// expecting a Cosmos event in response to the message committed as a
+    /// transaction.
+    pub fn send_message_and_wait_commit_cosmos(
```
adizere, line 975: Do I understand correctly that the only difference between this method and `send_messages_and_wait_commit` is that the former accepts a single `msg: Any`, whereas the latter accepts as parameter a `msgs: TrackedMsgs` (which it splits across multiple tx-es).

The comment and method name here implies that is some more significant difference to the other `send_messages_and_wait_*` method.

adizere (commented):
I left a couple of questions that should be addressed or clarified before merging.

Given the complexity of changes in the present PR, the fix suggested in #1979 is simpler and does a good enough job to accept it as a solution for #1288. We should probably accept the current PR if it gives us insights, if it offers a useful refactoring, or if it adds support that might be needed in the future, eg non-SDK support or helps in a way with integration tests; I don't understand if any of these are true at the moment. That's my current assessment.

I think Romain & Soares are in the best position to choose the right fix between PRs #1979 and #1909, so I will defer to them. (Again, both PRs "solve the problem" from a user-level perspective).

relayer/src/chain/cosmos.rs
```diff
@@ -864,15 +869,19 @@ impl CosmosSdkChain {
                             ))];
 
                             // Otherwise, try to resolve transaction hash to the corresponding events.
-                        } else if let Ok(events_per_tx) =
-                            self.query_txs(QueryTxRequest::Transaction(QueryTxHash(response.hash)))
+                        } else if let Ok((events_per_tx, cosmos_events_per_tx)) =
+                            self.query_extended_events_for_tx(QueryTxHash(response.hash))
```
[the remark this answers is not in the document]
mzabaluev: >     1. Does this mean that _all_ Hermes CLIs will do `query_extended_events_for_tx` as of this PR?

Yes. I chose this as a way to avoid copying `wait_for_block_commits` to do almost exactly the same job, while not exposing these internal changes in the generic `send_messages_and_wait_commit` API.

>     2. Does this have any performance implications or does it affect the return value for any other Hermes CLI command (aside from `tx raw upgrade chain)?

I expect the performance implications to be negligible: Cosmos and IBC events are [parsed out of the same list](https://github.com/informalsystems/ibc-rs/blob/fd3d09e717df8f7d480b11baba25fe73d9974318/relayer/src/chain/cosmos.rs#L965) of ABCI events, which is still retrieved in a single query (in an intermediate change it was not so, but I have reworked it to avoid this exact performance hit).

As for the return value, the only way it would change the output of any other CLI command or other relayer logic using `send_messages_and_wait_commit`, is if the caller expected an IBC event in response to the message, but got no IBC events and some recognized Cosmos event(s) instead. In this case, the return value would be an empty list (as the Cosmos events get discarded in the generic API). I expect there are no such cases at the moment, and if any appear, that would indicate something wrong in our expectations about the ABCI chain.

relayer/src/chain/cosmos.rs
mzabaluev: I did not want to replicate the message batching code, since the only current use of this method sends a single message.
Maybe this needs to be clarified in the documentation.

adizere:
Closing this in favor of the simpler solution offered in #1979.

We will revive this code as the need arises in the future. In particular:

- the method `query_extended_events_for_tx` would come in handy when we will require waiting on non-IBC events;
- the method `send_message_and_wait_commit_cosmos` would come in handy when we will require a separate path for submitting transactions that is SDK-specific, which short-circuits the `ChainHandle` interface and could also be useful in testing frameworks.

PR #1912, opened 2022-02-23
Fix up app.toml configs for gaiad 6.x in dev-env scripts

mzabaluev:
Disable the grpc-web server which is neither configured by the scripts nor used by the CLI,
so that the port used by the default configuration is not attempted to be bound by multiple gaiad processes.

If we ever have a need to use grpc-web in the dev environment (and not Docker, nix, etc.), the scripts will need to be adjusted
in a separate change to also override the grpc-web ports.

______

### PR author checklist:
- ~Added tests: integration (for Hermes) or unit/mock tests (for modules).~
- [ ] Linked to GitHub issue.

### Reviewer checklist:


adizere (approved):
Amazing!

By the way, I did a quick check and it seems that `gm` supports gaia v6; we just need to be careful to
- install the correct `gaiad` version on $PATH
- erase the chain data directories (in `~/.gm/$CHAINID` before we `gm start` chains with gaiad v6

hu55a1n1 (approved):
💯👌

Issue #1913, opened 2022-02-23
Support CosmWasm environment for `ibc-proto`


romac:
## Crate

`ibc-proto`

## Summary

> The reason for this PR is to have the ibc-proto crate being reusable in the context of a cosmwasm light client, a new type of [IBC client](https://github.com/cosmos/ibc/tree/master/spec/client/ics-008-wasm-client).
> this PR includes a non-exhaustive set of changes required to have the ibc-proto compilable and runnable in a cosmwasm environment. 
> The
> >  adding feature client to opt out client implementations as tonic dependency must be removed when compiling within 
> 
> change is necessary because in cosmwasm we don't have access to the ``getrandom`` crate, so I needed to remove dependency on tonic.
> 
> The
> > adding #[derive(::schemars::JsonSchema)] attribute for structs that might be of value within an IBC-related cosmwasm contract
> 
> extension is needed if we want to pass such structures in and out of a cosmwasm smart contract (serde alone is not enough)
> I think the following one explain themselves
> > making no_std a feature, so above derive macro can work 
> 
> This one is actually important outside the scope of cosmwasm as well, as you already added ``serde::Serialize`` and ``serde::Deserialize`` support. 
> > adding #[serde(default)] field attribute to fields that might be omitted by Golang omitempty directive.
> 
> _Originally posted by @lightyear15 in https://github.com/informalsystems/ibc-rs/issues/1628#issuecomment-1047927312_

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #1914, opened 2022-02-23
Support CosmWasm in ibc-proto

romac:
Closes: #1913 

## Description

- Add a new `client` feature to gate the `tonic` client code, implies `std`.
- Add a new `json-schema` feature to derive `schemars::JsonSchema` on some proto types, implies `std`.
- Add `#[serde(default)]` to fields that might be omitted by Golang `omitempty` directive.
- Change serialization of byte arrays to Base64 for compatibility with Go.
- Remove distinction between std and no_std for generated code.

______

### PR author checklist:
- [x] Passing existing no_std checks. 
- [x] Linked to GitHub issue.

### Reviewer checklist:


adizere:
>Remove distinction between std and no_std for generated code.

💘 

Will review soon, but if @lightyear15 is good with the changes we can LGTM.

soareschen (approved):

proto/Cargo.toml
```diff
@@ -27,11 +27,15 @@ prost-types = { version = "0.9", default-features = false }
...
 version = "=0.23.5"
 default-features = false
 
 [features]
-default = ["std"]
-std = ["tonic", "tonic/codegen", "tonic/transport", "tonic/prost"]
+default = ["std", "client"]
+client = ["std", "tonic", "tonic/codegen", "tonic/transport", "tonic/prost"]
+json-schema = ["std", "schemars"]
+std = ["base64"]
```
lightyear15, line 41: Shouldn't this be the opposite?
```
base64 = ["std"]
```

proto/src/lib.rs
```diff
@@ -218,3 +205,19 @@ pub mod ibc {
 pub mod ics23 {
     include_proto!("ics23.rs");
 }
+
+#[cfg(feature = "std")]
```
lightyear15, line 209: In my view, there is no use case where this crate does not need to be compatible with Go, so the ``base64`` encoding/decoding should not be optional.

I don't know what use case you guys have for this crate to support ``no_std``, but you can try to explore [``encode_config_slice``](https://docs.rs/base64/latest/base64/fn.encode_config_slice.html) and [``decode_config_slice``](https://docs.rs/base64/latest/base64/fn.decode_config_slice.html) with ``default-features=false`` for [``base64``](https://docs.rs/crate/base64/latest/source/Cargo.toml) crate

proto/Cargo.toml
romac: Here `std` is the feature and `base64` the optional crate that is enabled when the `std` feature is enabled. That said, I agree with your comment below (https://github.com/informalsystems/ibc-rs/pull/1914#discussion_r814571968) and will enable the base64 encoding (and therefore Go compat) for both std and no_std.

proto/src/lib.rs
romac: https://github.com/informalsystems/ibc-rs/pull/1914/commits/245c1e761d2d47c25a1327976f775e046119f1bb

romac:
@lightyear15 Does this work for you?

lightyear15 (approved):
sorry for the delay.
I am fairly confident that it works. I can compile my small cosmwasm contract and run some unit tests using the cosmwasm vm
:+1: 
Thanks

PR #1915, opened 2022-02-23
Run link checker nightly instead of on every PR

romac:
The link checker yields false positives way too often and therefore breaks the CI status for no reason.

Let's move it to nightly and check on it from time to time.

### DIFF
--- a/.github/workflows/markdown-link-check.yml
+++ b/.github/workflows/markdown-link-check.yml
@@ -1,10 +1,8 @@
 name: Check Markdown links
 
 on:
-  push:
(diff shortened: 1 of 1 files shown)


mzabaluev (approved):
The workflow looks good.

A disadvantage I see is that it does not catch a newly introduced broken link (e.g. a typo in a changelog entry) immediately. Then it becomes "nobody's problem" that does not necessarily get fixed quickly.

romac:
True, but I am hoping someone (myself included ofc) will rise to the occasion from time to time :)

Issue #1772, opened 2022-01-14
Connection delay period is checked against relayer's local time 


hu55a1n1:
## Crate
relayer

## Summary of Bug
The relayer logic that implements connection delay support checks packets against the relayer's local time. This is inconsistent with the IBC handlers implementation that checks against the block time of the latest on-chain block. 

## Version
v0.10.0

## Acceptance Criteria
The connection delay logic uses the target chain's latest block time. 
____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #1868, opened 2022-02-14
Failed packet verification: packet-specified delay period has not been reached


lyqingye:
## Crate
hermes

## Summary of Bug
```shell
error logs: 

2022-02-13T19:38:23.630978Z  INFO ThreadId(23) [chain:port/channel-0 -> target] clearing pending packets height=Some(Height { revision: 0, height: 1435559 })
2022-02-13T19:38:24.052961Z  INFO ThreadId(23) [chain:port/channel-0 -> target] found unprocessed SendPacket events for [Sequence(25246)] (first 10 shown here; total=1)
2022-02-13T19:38:24.052984Z  INFO ThreadId(23) [chain:port/channel-0 -> target] generate messages from batch with 1 events
2022-02-13T19:38:24.062747Z  INFO ThreadId(23) [chain:port/channel-0 -> target] scheduling op. data with 1 msg(s) for Destination (height 0-1435560)
2022-02-13T19:38:24.875063Z  INFO ThreadId(23) [chain:port/channel-0 -> target] finished scheduling pending packets clearing height=Some(Height { revision: 0, height: 1435559 })
2022-02-13T19:38:25.062889Z  INFO ThreadId(23) [chain:port/channel-0 -> target] relay op. data of 1 msgs(s) to Destination (height 0-1435560), delayed by: 1.00012235s [try 1/5]
2022-02-13T19:38:25.062913Z  INFO ThreadId(23) [chain:port/channel-0 -> target] prepending Destination client update @ height 0-1435560
2022-02-13T19:38:25.102220Z  INFO ThreadId(23) [chain:port/channel-0 -> target] assembled batch of 2 message(s)
2022-02-13T19:38:25.109056Z ERROR ThreadId(12) [target] estimate_gas: failed to simulate tx. propagating error to caller: 
gRPC call failed with status: status: Unknown, 
message: "failed to execute message; message index: 1: 
receive packet verification failed: couldn't verify counterparty packet commitment: failed packet commitment verification for client (07-tendermint-0): cannot verify packet until time: 1644781101580283164, current time: 1644781099180419908: packet-specified delay period has not been reached", 
details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "x-cosmos-block-height": "891340"} }
```
log analysis: 
1. latest height = 1435560, clearing pending packet height = latest height - 1 = 1435559
2. query unprocessed events at height = 1435559
3. build packet proof at height = 1435559
4. build update client at height = 1435560

[...]

Issue #1773, opened 2022-01-14
Relayer logic for connection delay doesn't account for block delay


hu55a1n1:
## Crate
relayer

## Summary of Bug
Relayer logic for connection delay doesn't account for block delay. More specifically we're missing the equivalent of the following handler check (which is analogous to the delay time check) -> 
```
assert(currentHeight() >= processedHeight + delayPeriodBlocks)
// where `delayPeriodBlocks` is calculated as ->
delayPeriodBlocks = ceil(delayPeriodTime/max_expected_time_per_block)
// and `max_expected_time_per_block` is defined in genesis state
```

## Version
v0.10.0

## Acceptance Criteria
The relayer logic adds the aforementioned check.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
Wasn't this closed with https://github.com/informalsystems/ibc-rs/pull/1917 ?

adizere:
Thanks for confirming Shoaib!

PR #1917, opened 2022-02-23
Fix the connection delay logic to use the header update block time

hu55a1n1:
Closes: #1772

## Description
Fix the connection delay logic to use the target chain's latest block time and header update block time.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


relayer/src/link/relay_path.rs
```diff
@@ -696,10 +698,15 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
 
     /// Returns `true` if the delay for this relaying path is zero.
     /// Conversely, returns `false` if the delay is non-zero.
-    pub fn zero_delay(&self) -> bool {
+    fn zero_delay(&self) -> bool {
```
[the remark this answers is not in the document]
hu55a1n1: Good point! 👌 Done -> eb1914caad0432f246e07eaf2cb92c106dbaa49e

relayer/src/link/operational_data.rs
```diff
@@ -178,6 +178,19 @@ impl OperationalData {
...
+        self.batch.iter().any(|msg| {
+            matches!(
+                msg.event,
+                IbcEvent::ReceivePacket(_)
+                    | IbcEvent::WriteAcknowledgement(_)
+                    | IbcEvent::AcknowledgePacket(_)
+                    | IbcEvent::TimeoutPacket(_)
+                    | IbcEvent::TimeoutOnClosePacket(_)
+            )
+        })
```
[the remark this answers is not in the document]
hu55a1n1: There's a `IbcEvent::packet(&self) -> Option<&Packet>` function that we could reuse here -> https://github.com/informalsystems/ibc-rs/blob/master/modules/src/events.rs#L446
```rust
pub fn has_packet_msgs(&self) -> bool {
    self.batch.iter().any(|msg| msg.event.packet().is_some())
}
```
But I'm a little confused if we should because I don't think we should take `SendPacket` events into account here. 🤔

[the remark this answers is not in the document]
hu55a1n1: I think I was mistaken about not checking for `SendPacket` events. Just updated the code to use the `msg.event.packet().is_some()` I described above, although I am still not completely sure if this is full-proof or if we need to match by `msg.msg.type_url` to be certain if we're dealing with packet related messages.

relayer/src/link/relay_path.rs
```diff
@@ -1387,22 +1425,36 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+                    self.dst_chain()
+                        .query_host_consensus_state(update_height)
+                        .map_err(LinkError::relayer)?
+                        .timestamp()
                 }
             };
+            let duration = Timestamp::now()
+                .duration_since(&update_time)
+                .unwrap_or(Duration::ZERO);
+            Instant::now().sub(duration)
```
hu55a1n1, line 1450: IMHO, `Instant::now()` is a better choice v/s the chain's time (i.e. latest block's timestamp), this is because the chain handler will ultimately check this logic against the timestamp of the next block (which will contain this tx/message). 
One problem with this choice is that it might lead to false positives (with the right timing and drift). (I believe https://github.com/informalsystems/ibc-rs/issues/1868 is one such case)

relayer/src/chain/cosmos.rs
```diff
@@ -1946,6 +1946,15 @@ impl ChainEndpoint for CosmosSdkChain {
         }
     }
 
+    fn query_host_consensus_state(&self, height: ICSHeight) -> Result<Self::ConsensusState, Error> {
+        let height = Height::try_from(height.revision_height).map_err(Error::invalid_height)?;
+
+        let response = self
+            .block_on(self.rpc_client.block(height))
```
hu55a1n1, line 1953: The plan is to replace this with the `/header` query. (I'm working on adding it to tendermint-rs -> https://github.com/informalsystems/tendermint-rs/issues/832)

hu55a1n1: https://github.com/informalsystems/tendermint-rs/pull/1101

relayer/src/link/relay_path.rs
```diff
@@ -748,25 +780,28 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
         dst_chain_height: Height,
         tracking_id: &str,
-    ) -> Result<(), LinkError> {
-        if self
-            .src_chain()
-            .proven_client_consensus(self.src_client_id(), dst_chain_height, Height::zero())
-            .is_ok()
-        {
-            return Ok(());
+    ) -> Result<Height, LinkError> {
```
soareschen, line 788: The retry logic inside for loop is very difficult to understand and verified to work correctly. Can we split it into a recursive function that does the retry? Something like:

```rust
    /// Handles updating the client on the source chain
    /// Returns the height at which the client update was processed
    fn update_client_src(
        &self,
        dst_chain_height: Height,
        tracking_id: &str,
    ) -> Result<Height, LinkError> {
        if let Some(update_height) = Self::update_height(
            self.src_chain(),
            self.src_client_id().clone(),
            dst_chain_height,
        ) {
            Ok(update_height)
        } else {
            self.do_update_client_src(
                dst_chain_height,
                tracking_id,
                MAX_RETRIES,
            )
        }
    }

    /// Perform actual update_client_src with retries.
    ///
    /// Note that the retry is only performed in the case when there
    /// is a ChainError event. It would return error immediately if
    /// there are other errors returned from calls such as
    /// build_update_client_on_src.
    fn do_update_client_src(
        &self,
        dst_chain_height: Height,
        tracking_id: &str,
        retries_left: usize
    ) -> Result<Height, LinkError> {
        info!(
            "sending update_client to client hosted on source chain for height {} (retries left: {})",
            dst_chain_height,
            retries_left
        );

        let src_update = self.build_update_client_on_src(dst_chain_height)?;
        let tm = TrackedMsgs::new(src_update, tracking_id);

        let src_tx_events = self
            .src_chain()
            .send_messages_and_wait_commit(tm)
            .map_err(LinkError::relayer)?;

        info!("result: {}", PrettyEvents(&src_tx_events));

        for event in src_tx_events {
            match event {
                err_event @ IbcEvent::ChainError(_) => {
                    if retries_left == 0 {
                        return Err(LinkError::client(ForeignClientError::chain_error_event(
                            self.src_chain().id(),
                            err_event,
                        )))
                    } else {
                        return self.do_update_client_src(
                            dst_chain_height,
                            tracking_id,
                            retries_left - 1,
                        )
                    }
                }
                IbcEvent::UpdateClient(ev) => return Ok(ev.height()),
                _ => {}
            }
        }

        if retries_left == 0 {
            Err(LinkError::update_client_failed())
        } else {
            self.do_update_client_src(
                dst_chain_height,
                tracking_id,
                retries_left - 1,
            )
        }
    }
```

soareschen (changes requested):
Can we have an integration test to ensure that the connection delay logic is working correctly?

relayer/src/link/relay_path.rs
hu55a1n1: Thanks for the suggestion, Soares! Done! 👌

hu55a1n1: I wonder how we should handle the case where the `send_messages_and_wait_commit()` succeeds but the response contains neither an `UpdateClient` event nor a chain error. With this impl we'd be retrying and return a new kind of error if none of the retries work, but IIUC, we previously assumed that this was unlikely and called `src_err_ev.unwrap()` on the event.

relayer/src/link/relay_path.rs
```diff
@@ -748,48 +780,70 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+                        )));
+                    } else {
+                        return self.do_update_client_src(
+                            dst_chain_height,
+                            tracking_id,
+                            retries_left - 1,
+                        );
+                    }
+                }
+                IbcEvent::UpdateClient(ev) => return Ok(ev.height()),
```
hu55a1n1, line 838: I think we need to make sure that this `UpdateClient` belongs to this particular client. (not sure if we have a check elsewhere for this)

relayer/src/link/relay_path.rs
```diff
@@ -748,48 +780,70 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+        let tm = TrackedMsgs::new(src_update, tracking_id);
+        let src_tx_events = self
+            .src_chain()
+            .send_messages_and_wait_commit(tm)
+            .map_err(LinkError::relayer)?;
+        info!("result: {}", PrettyEvents(&src_tx_events));
+
+        for event in src_tx_events {
+            match event {
+                err_event @ IbcEvent::ChainError(_) => {
```
hu55a1n1, line 824: Also, we're first matching against the chain error here which may have been emitted by a different message. Maybe it's better to do this in two passes - first do a `src_tx_events.iter().find()` for `UpdateClient` and then do another one for `ChainError`.

relayer/src/link/relay_path.rs
hu55a1n1: Seems like the tx can contain only `MsgUpdateClient`s for this client, so this shouldn't be a problem.

relayer/src/link/relay_path.rs
hu55a1n1: Looks like the order is correct because we want to retry until all `MsgUpdateClient`s succeed, but it would be nice to retry only the ones that failed. So I think it will be better to not return on the first `UpdateClient`. Also, I don't think we're interested in the `UpdateClientEvent` height here, we actually need the height at which this event was emitted. 
There is one other case that isn't handled here which is when there is a `ClientMisbehaviourEvent`. 
I'll try to fix these issues in the next commits.

relayer/src/link/relay_path.rs
hu55a1n1: Seems like it is indeed possible that the returned events contain neither an `UpdateClient` event nor a chain error, which is when there is a `ClientMisbehaviour` event. So, I believe this implementation is doing the right thing by returning an `Err(LinkError::update_client_failed())`.

soareschen:
Hmm interestingly, having a 30 seconds connection delay when basing only on my nary-test branch, but fails after merging with your branch.

hu55a1n1:
> I opened https://github.com/informalsystems/ibc-rs/pull/1950 to test this branch.
Thanks @soareschen! Awesome stuff! 💯 

> Hmm interestingly, having a 30 seconds connection delay when basing only on my nary-test branch, but fails after merging with your branch.
That's strange. 🤔 I identified some improvements in the code so I moved it back to draft state. I'll test again with your branch once it's ready and let you know how it goes.

relayer/src/link/relay_path.rs
ancazamfir: I think this change is ok but not clear to me it is needed to solve the actual issue. Aren't we supposed to change this line further down in `try_fetch_scheduled_operational_data`:
`op.scheduled_time.elapsed() > connection_delay`

and get the time since the last host consensus state instead of `elapsed()`?

hu55a1n1: The idea was to set the `scheduled_time` to a time in the past when the header was updated (i.e. `Instant::now().sub(duration)` where duration is the timestamp of the block that includes the header update). So, `op.scheduled_time.elapsed()` returns the time elapsed since the header update. Also, I thought the chain handler will check this against the block's timestamp which contains this packet, so `now()` is a better choice than the latest host timestamp. Not sure if my reasoning is correct. 🤔

relayer/src/link/relay_path.rs
hu55a1n1: https://github.com/informalsystems/ibc-rs/pull/1917/commits/9192eecaa2b5582d0277179b7dda984d27d83210

relayer/src/link/relay_path.rs
ancazamfir: I think this part is correct. But #1757 is about this:
```
The relayer logic that implements connection delay support checks packets against the relayer's local time. This is inconsistent with the IBC handlers implementation that checks against the block time of the latest on-chain block.
```

It refers to the packet verification potentially failing. If the update client was included in block `h` and the last block is `H`, the packet handler for a Tx to be included in `H+1` will check that `H.time - h.time` is bigger than connection delay. But in `try_fetch_scheduled_operational_data()` we basically check against `Instant::now() - h.time`. So we may send the packet too early.

We have fixed a similar issue when deciding if a packet has timed out https://github.com/informalsystems/ibc-rs/blob/f11415afbf583124d1a3ae52c0a972a7680bd716/modules/src/core/ics04_channel/packet.rs#L167

Note: To complicate things, there is an additional check in the packet handler that asserts that a minimum number `n` of blocks must have been created since the update client block. `n` is computed as `ceil(connection_delay/max_expected_time_per_block)`. We should then do some basic check along these lines in the relayer code. `max_expected_time_per_block` can be found in the genesis info:
```
{
  ...
  "app_state": {
    ...
    "ibc": {
      ...
      "connection_genesis": {
        ...
        "params": {
          "max_expected_time_per_block": "30000000000"
        }
...
```

ancazamfir: > Also, thanks for reminding me about the block delay 👌, I had opened issue #1773 for that. Should we address that as well (as part of this PR)?

Ideally here unless is a complex change.

ancazamfir: > I believe #1868 is one such case

Looked through that but it looks like 0 connection delay there. So probably it's a different issue.

hu55a1n1: I think we were seeing a similar issue in the integration test PR #1950. The fix was to increase the `connection_delay` by `avg_block_time` of the host so that we are sure that the packet verification succeeds -> https://github.com/informalsystems/ibc-rs/pull/1950/commits/19fd37f05c07faf216c6b25b6ed1680c212c4b97
Is there a way to get a chain's avg block time (using RPC)?

soareschen:
I have added a connection delay test to test on this PR directly. You can run it with:

```bash
cargo test -p ibc-integration-test -- test_connection_delay
```

The test is currently failing and needs to be fixed. We can move the fix from #1950 to here.

relayer/src/link/relay_path.rs
ancazamfir: > > My understanding was that the handler does this instead -> `((H + 1).time - h.time) > conn_delay`
> 
> Ah yes, you are correct! forgot context is set in `BeginBlock` (so before `deliverTx()`)

So actually this is true for `deliverTx()` handling. However when we do the Tx simulation, the checks run on the simulate state copy at latest height `H` which may have been created on chain before the time we checked for elapse of the delay. This is why test fails. 

Your changes in 19fd37f05c07faf216c6b25b6ed1680c212c4b97 are then fine, we just need to get `max_expected_time_per_block` from the genesis parameters instead of the constant `AVG_BLOCK_TIME_SECS`. 
Given that this is a chain parameter I am wondering if we can get/cache it when we create the chain handler.

We are probably just lucky that the test passes in master as we mainly test with small block times (1sec) and because we use `Instant::now()` for `scheduled_time` which is more recent than the block containing the client update...so we end up delaying more.

ancazamfir: Thinking about this more, maybe it's still better to go with https://github.com/informalsystems/ibc-rs/pull/1917#discussion_r825734002. So we get the `last_header_time` (using the same `query_host_consensus_state()`) and then something like:
`last_header_time.sub(op.scheduled_time) > connection_delay`

with the weird way of moving between `Timestamp` and `Instant` :( Maybe have some wrappers for these 

Dealing with block times only will avoid additional failure scenarios caused by clock accuracy differences between hermes and the "chain" (header BFT time or proposer time in the future).

hu55a1n1: > However when we do the Tx simulation, the checks run on the simulate state copy at latest height H which may have been created on chain before the time we checked for elapse of the delay. This is why test fails.

Wow! That's a tricky one, nice catch! 👍 

> Your changes in https://github.com/informalsystems/ibc-rs/commit/19fd37f05c07faf216c6b25b6ed1680c212c4b97 are then fine, we just need to get max_expected_time_per_block from the genesis parameters instead of the constant AVG_BLOCK_TIME_SECS.

I checked the ibc-go codebase and they seem to view the `max_expected_time_per_block` differently. The default value is 30 seconds and they view it as the max time a chain might take to mint a block. (https://github.com/cosmos/ibc-go/pull/171#discussion_r630932520) Should we still use it? Or is there an average block time counterpart?

ancazamfir: There is no average block time and even if it were it wouldn't always work. Using max also delays relaying quite a bit. To avoid this we can just use the latest block time as I mentioned in the last comment.

hu55a1n1: Great idea! I tried to do this with https://github.com/informalsystems/ibc-rs/pull/1917/commits/4df249a22f9754a343293df805d3804853da814d.

ancazamfir: Maybe we can cache this at some point.

relayer/src/link/relay_path.rs
```diff
@@ -1438,62 +1664,88 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+                        delay_time,
                         odata.batch.len(),
                         odata.target,
                     );
 
                     // Wait until the delay period passes
-                    thread::sleep(delay_left);
+                    thread::sleep(delay_time);
                 }
+                (None, Some(_blocks)) => {}
```
hu55a1n1, line 1743: It isn't clear to me how to handle cases where we have a remaining block delay.

tools/integration-test/src/tests/connection_delay.rs
```diff
@@ -0,0 +1,106 @@
...
+const CONNECTION_DELAY: Duration = Duration::from_secs(10);
+
+#[test]
+fn test_connection_delay() -> Result<(), Error> {
+    run_binary_channel_test(&ConnectionDelayTest)
+}
+
+pub struct ConnectionDelayTest;
+
+impl TestOverrides for ConnectionDelayTest {
```
hu55a1n1, line 17: Seems like that `max_delay_block` in the integration test defaults to `0` instead of `10s`.

relayer/src/link/relay_path.rs
```diff
@@ -693,103 +734,281 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
-            let dst_update = self.build_update_client_on_dst(src_chain_height)?;
-            info!(
-                "sending updateClient to client hosted on destination chain for height {} [try {}/{}]",
-                src_chain_height,
-                i + 1, MAX_RETRIES,
-            );
-
-            let tm = TrackedMsgs::new(dst_update, tracking_id);
+    #[inline]
+    fn conn_block_delay_elapsed(
```
hu55a1n1, line 773: This method of conn delay calculation is actually Tendermint specific according to the spec. So IIUC, chains can decide to ignore conn-delay completely or use an impl of their choice. This is another thing we might want to consider when we're adding support for non-SDK chains.

relayer/src/link/operational_data.rs
```diff
@@ -89,23 +91,28 @@ pub struct OperationalData {
     pub proofs_height: Height,
     pub batch: Vec<TransitMessage>,
     pub target: OperationalDataTarget,
-    /// Stores the time when the clients on the target chain has been updated, i.e., when this data
-    /// was scheduled. Necessary for packet delays.
-    pub scheduled_time: Instant,
     pub tracking_id: String,
+    // Stores `Some(non-zero connection-delay)` or `None`
```
seanchen1991, line 95: Can we keep this as a doc comment instead of changing it to a regular comment?

relayer/src/link/relay_path.rs
```diff
@@ -1438,62 +1598,90 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+        self.src_operational_data.replace(unelapsed_src_ods);
         self.dst_operational_data.replace(unelapsed_dst_ods);
-
-        (elapsed_src_ods, elapsed_dst_ods)
+        Ok((elapsed_src_ods, elapsed_dst_ods))
     }
 
     /// Fetches an operational data that has fulfilled its predefined delay period. May _block_
     /// waiting for the delay period to pass.
     /// Returns `None` if there is no operational data scheduled.
```
seanchen1991, line 1630: This doc comment needs to be updated to reflect this function's new return type.

relayer/src/link/relay_path.rs
seanchen1991: What situations would lead to there being remaining block delay?

relayer/src/link/relay_path.rs
hu55a1n1: https://github.com/informalsystems/ibc-rs/pull/1917/commits/d1b712b9dbcc847bbd10133bef0064653a5f3617

relayer/src/link/operational_data.rs
hu55a1n1: Done! 👍 https://github.com/informalsystems/ibc-rs/pull/1917/commits/d1b712b9dbcc847bbd10133bef0064653a5f3617
I previously changed it to a regular comment because the `connection_delay` field is private, but I think your suggestion makes sense also because cargo explicitly supports `--document-private-items`.

relayer/src/link/relay_path.rs
hu55a1n1: The block delay is calculated as ->
```
block_delay = time_delay / max_expected_time_per_block
```
So the `block_delay` is basically calculating the (minimum) number of blocks a chain would create in the `time_delay` duration. 
And the default value for `max_expected_time_per_block` in ibc-go is 30s. The rationale for choosing this value is described in detail in this comment -> https://github.com/cosmos/ibc-go/pull/171#discussion_r633870345 
> I see the block delay as handling the edge case where the receiving chain slows down unexpectedly (...)

So, `max_expected_time_per_block` is not the same as average block-time of the chain and is usually larger. This means that the calculated `block_delay` is usually much smaller than the average number of blocks the chain would be able to create in that (delay) duration. So, IMHO it is quite unlikely that there would be a remaining `block_delay` (if everything is configured correctly) and it should be sufficient if the relayer simply handles the `time_delay`.

modules/src/core/ics04_channel/context.rs
```diff
@@ -279,3 +273,15 @@ pub trait ChannelKeeper {
     /// Should never fail.
     fn increase_channel_counter(&mut self);
 }
+
+pub fn calculate_block_delay(
+    delay_period_time: Duration,
+    max_expected_time_per_block: Duration,
+) -> u64 {
+    let expected_time_per_block = max_expected_time_per_block;
```
ancazamfir, line 281: Is `expected_time_per_block` needed?

relayer/src/link/relay_path.rs
```diff
@@ -1438,62 +1598,90 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
             (true_res, false_res)
         }
 
-        let connection_delay = self.channel.connection_delay;
+        let src_chain_time = self.src_time_latest()?;
+        let src_max_block_time = self.src_max_block_time()?;
+        let src_latest_height = self.src_latest_height()?;
```
ancazamfir, line 1603: I am a bit concerned that we add the overhead of getting this data from chain for 0-delay paths which are the overwhelming majority of cases...and these calls are not needed there. Could we skip these for the `self.channel.connection_delay.is_zero()` case:
```suggestion
        let (elapsed_src_ods, unelapsed_src_ods) = if self.channel.connection_delay.is_zero() {
            (self.src_operational_data.take(), VecDeque::new())
        } else {
            let src_chain_time = self.src_time_latest()?;
            let src_max_block_time = self.src_max_block_time()?;
            let src_latest_height = self.src_latest_height()?;
            ...
```

relayer/src/link/relay_path.rs
```diff
@@ -1438,62 +1598,90 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+                    && op
+                        .conn_block_delay_elapsed(src_max_block_time, src_latest_height)
+                        .is_ok()
             });
 
-        self.src_operational_data.replace(unelapsed_src_ods);
-
+        let dst_chain_time = self.dst_time_latest()?;
+        let dst_max_block_time = self.dst_max_block_time()?;
+        let dst_latest_height = self.dst_latest_height()?;
```
ancazamfir, line 1614: same comment as above.

relayer/src/link/relay_path.rs
```diff
@@ -1438,62 +1598,90 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+                let src_chain_time = self.src_time_latest()?;
+                let src_block_time = self.src_max_block_time()?;
+                let src_latest_height = self.src_latest_height()?;
+                (
+                    odata.conn_time_delay_elapsed(src_chain_time).err(),
+                    odata
+                        .conn_block_delay_elapsed(src_block_time, src_latest_height)
+                        .err(),
+                    Some(odata),
+                )
```
ancazamfir, line 1645: Here is not as important as this is called from the CLI but could be something like:
```suggestion
            if let Some(odata) = self.src_operational_data.pop_front() {
                if self.channel.connection_delay.is_zero() {
                    (None, None, Some(odata))
                } else {
                    let src_chain_time = self.src_time_latest()?;
                    let src_block_time = self.src_max_block_time()?;
                    let src_latest_height = self.src_latest_height()?;
                    (
                        odata.conn_time_delay_elapsed(src_chain_time).err(),
                        odata
                            .conn_block_delay_elapsed(src_block_time, src_latest_height)
                            .err(),
                        Some(odata),
                    )
                }
```

relayer/src/link/relay_path.rs
ancazamfir: > So, IMHO it is quite unlikely that there would be a remaining block_delay (if everything is configured correctly) and it should be sufficient if the relayer simply handles the time_delay.

We should probably wait `blocks * max_expected_time_per_block`. 

I think the reasoning behind this is to allow a minimum number of blocks to pass between the client update and the recv or timeout packets so client misbehavior is detected, evidence submitted, client is frozen. And as you pointed out in the ibc-go issue, it should be possible to handle also the case where chain slows down significantly. For example if we expect the  block time to be at most 30sec, with a connection delay of 2 min the block check requires 4 blocks between the update client and the packet. Even if the block after the update is > 2min. Without this check it would be possible to process the `recv_packet` before the evidence of client misbehavior (that freezes the client).

relayer/src/link/relay_path.rs
```diff
@@ -1438,62 +1598,90 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
                         "waiting ({:?} left) for a scheduled op. data with batch of size {} targeting {}",
-                        delay_left,
+                        delay_time,
                         odata.batch.len(),
                         odata.target,
                     );
 
                     // Wait until the delay period passes
-                    thread::sleep(delay_left);
+                    thread::sleep(delay_time);
```
ancazamfir, line 1677: Should we add some warning message to the user if `delay_time != 0` so it is clear that the CLI is not broken but needs to wait for the delay_time?

modules/src/core/ics04_channel/context.rs
hu55a1n1: Removed 👍 https://github.com/informalsystems/ibc-rs/pull/1917/commits/ef95dd2b5350ddd88f192c7932f133ddbf23c817

relayer/src/link/relay_path.rs
hu55a1n1: Thanks for noticing this! I took a slightly different approach from the one you suggested here. I used closures to evaluate the queries lazily and only if required. There are two cases where we don't need to handle the connection delay - one when it is zero and one where it is non-zero, but the `OperationalData` doesn't have packet messages. So IMHO the closure approach is better because it has the added benefit that it handles all these cases with the same code, but please let me know if you think otherwise and I will change it to handle these cases separately.  https://github.com/informalsystems/ibc-rs/pull/1917/files#diff-2a5a700c0b1975ad9d64941997aa53e0b6e26e577fa8f1e9a2aba4abe7cf1f95R1604

relayer/src/link/relay_path.rs
hu55a1n1: Done https://github.com/informalsystems/ibc-rs/pull/1917/files#diff-2a5a700c0b1975ad9d64941997aa53e0b6e26e577fa8f1e9a2aba4abe7cf1f95R1625 👍

relayer/src/link/relay_path.rs
hu55a1n1: I noticed that there is an `info!()` log [above that code](https://github.com/informalsystems/ibc-rs/pull/1917/files#diff-2a5a700c0b1975ad9d64941997aa53e0b6e26e577fa8f1e9a2aba4abe7cf1f95R1664-R1665), do you mean we should change it to `warn!()`? Or should we add a CLI output line to the packet CLI commands?

soareschen (changes requested):

relayer/src/link/relay_path.rs
```diff
@@ -1413,87 +1575,135 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+                    latest_height,
+                )
+            }
+        }
+    }
+
+    /// Fetches an operational data that has fulfilled its predefined delay period. May _block_
+    /// waiting for the delay period to pass.
+    /// Returns `Ok(None)` if there is no operational data scheduled.
+    pub(crate) fn fetch_scheduled_operational_data(
```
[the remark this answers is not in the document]
hu55a1n1: I think this is a valid concern, although the code in question predates this PR. I just tried to extract this code (i.e. `fetch_scheduled_operational_data()`, `build_and_send_ack_packet_messages()`, etc.) into `relayer-cli/src/commands.rs` but it seems like it requires a lot of visibility changes which include making the `relay_path` mod public and exposing relayer internals like `SyncSender`, so I am a little reluctant to implement this. This would have been easier if there was a `pub(workspace)`.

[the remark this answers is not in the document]
soareschen: Alternatively can we move them into separate files in the same crate, i.e. `relayer/src/link/relay_path/cli.rs`? It is allowed to have multiple `impl RelayPath {}` blocks within the same crate.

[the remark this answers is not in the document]
hu55a1n1: Nice idea! https://github.com/informalsystems/ibc-rs/pull/1917/commits/e0583586d24dfca4e2da9aeb5ced05d653843ecf

relayer/src/link/relay_path.rs
```diff
@@ -693,103 +733,212 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+                    }
+                    _ => Err(LinkError::update_client_failed()),
+                }
+            }
+            (_, _, event) => {
+                if !matches!(event, Some(IbcEvent::ClientMisbehaviour(_))) && retries_left > 0 {
+                    self.do_update_client_dst(src_chain_height, tracking_id, retries_left - 1)
+                } else {
+                    Err(LinkError::update_client_failed())
+                }
```
soareschen, line 870:
```suggestion
            (_, _, Some(IbcEvent::ClientMisbehaviour(_))) if retries_left > 0 => {
                self.do_update_client_dst(src_chain_height, tracking_id, retries_left - 1)
            }
            _ => {
                Err(LinkError::update_client_failed())
            }
```

relayer/src/link/relay_path.rs
```diff
@@ -693,103 +733,212 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+        let mut error = None;
+        let mut update = None;
+        let mut other = None;
+
+        while let Some(event) = tx_events.pop() {
+            match event {
+                IbcEvent::ChainError(_) => error = Some(event),
+                IbcEvent::UpdateClient(event) => update = Some(event),
+                _ => other = Some(event),
+            }
```
soareschen, line 783:
```suggestion
        Option<ibc::core::ics02_client::events::ClientMisbehaviour>,
    ) {
        let mut error = None;
        let mut update = None;
        let mut other = None;

        while let Some(event) = tx_events.pop() {
            match event {
                IbcEvent::ChainError(_) => error = Some(event),
                IbcEvent::UpdateClient(event) => update = Some(event),
                IbcEvent::ClientMisbehaviour(event) => other = Some(event),
                _ => {}
            }
        }
```

It looks like here we are only interested in the `ClientMisbehaviour` for the `other` field. So probably it is clearer to specify that explicitly.

Otherwise, it is also possible to have events that are not `ClientMisbehaviour` overriding `other` and make that information lost.

On the other hand, I am not very sure why you are specifically matching `None` for `other` at the call site. Are those implying that there were no `ClientMisbehaviour` event, or that there were no other events other than `ChainError`, `UpdateClient`?

soareschen (changes requested):

relayer/src/link/relay_path.rs
```diff
@@ -693,103 +733,212 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+        let tm = TrackedMsgs::new(dst_update, tracking_id);
+        let dst_tx_events = self
+            .dst_chain()
+            .send_messages_and_wait_commit(tm)
+            .map_err(LinkError::relayer)?;
+        info!("result: {}", PrettyEvents(&dst_tx_events));
+
+        let (error, update, other) = Self::event_per_type(dst_tx_events);
+        match (error, update, other) {
+            (None, Some(update_event), None) => Ok(update_event.height()),
```
soareschen, line 840: Would be helpful to write out detailed comments on what it really means when each of these branches are triggered.

relayer/src/link/operational_data.rs
```diff
@@ -178,6 +200,155 @@ impl OperationalData {
...
+
+    /// Sets the update height that is used for connection-delay calculations
+    pub fn set_update_height(&mut self, update_height: Height) {
+        if let Some(mut delay) = self.connection_delay.as_mut() {
+            delay.update_height = Some(update_height);
+        }
+    }
+
+    /// Returns `Ok(remaining-delay)` on success or `LinkError` if the input closure fails.
+    pub fn conn_time_delay_remaining<ChainTime>(
```
soareschen, line 238: This and `conn_block_delay_remaining` can be made private.

I think of the privacy keywords more as a way to signal where a function is being used. If we see a private function, we know we only need to look through one file to find out where that function is used. While for public functions we have to search through the entire code base to figure out each of the use case.

relayer/src/link/relay_path.rs
hu55a1n1: So the problem was that I wasn't able to find a spec that guarantees that a client update will only emit these 3 types of events and that there will be only one such event per type. And then we're not sure how to handle those other events. Looking at ibc-go, it does seem like only these 3 types of events are possible but then again that is SDK specific code. I like your idea of changing `event_per_type()` to return only `ClientMisbehaviour` events. With that we can then match for the `ChainError` and `ClientMisbehaviour` events first, but we still need to make sure that all update messages succeeded. The good thing is that ibc-go emits `UpdateClient` events even for redundant updates so we will always have one event per update message. And I think it is reasonable to assume that if there are no `ChainError`s and `ClientMisbehaviour`s, then all updates succeeded. WDYT?

hu55a1n1: https://github.com/informalsystems/ibc-rs/pull/1917/commits/14dcb2ae44e47c17a0b03a56aa32d1a74840ee3e

relayer/src/link/relay_path.rs
hu55a1n1: https://github.com/informalsystems/ibc-rs/pull/1917/commits/14dcb2ae44e47c17a0b03a56aa32d1a74840ee3e

relayer/src/link/relay_path.rs
soareschen: TBH I feel that the current approach is pretty ad hoc and unclear at handling the logic, but I also don't know any way of improving it without significant rewriting of the code. So I think we can leave that as is now.

If necessary, we can investigate how the ibc-go really works later on and improve the logic.

soareschen (approved):

relayer/src/link/relay_path.rs
ancazamfir: Ah, didn't realize we have an `info`. I think it's fine.

relayer/src/link/relay_path.rs
ancazamfir: Yes, that works better!

ancazamfir (approved):
Looks good!! Thanks @hu55a1n1 !

PR #1919, opened 2022-02-25
Fix new clippy warnings in Rust 1.59

romac:


adizere (approved):
Looks slightly cleaner and simpler now, thanks Romain!

Approving, but before merging, is the integration test is spurious or needs attention?

PR #1942, opened 2022-03-07
Cache module documentation

seanchen1991:
## Description

Adds documentation for the `relayer/src/cache.rs` module. 
______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/relayer/src/cache.rs
+++ b/relayer/src/cache.rs
@@ -1,3 +1,7 @@
+//! Module to provide caching support for the relayer.
+//!
+//! Utilizes the [`moka`](https://docs.rs/moka) crate, which provides full
+//! concurrency of retrievals and a high expected concurrency for updates.
 use core::fmt::Formatter;
 use std::fmt;
 use std::time::Duration;
@@ -15,11 +19,21 @@ const CONNECTION_CACHE_TTL: Duration = Duration::from_secs(10 * 60);
 const CLIENT_STATE_CACHE_TTL: Duration = Duration::from_millis(500);
 const LATEST_HEIGHT_CACHE_TTL: Duration = Duration::from_millis(200);
 
+/// The main cache data structure, which comprises multiple sub-caches for caching
+/// different chain components, each with different time-to-live values.
+///
+/// There should be one `Cache` instantiated per every chain runtime.
 #[derive(Clone)]
 pub struct Cache {
+    /// Cache storing [`ChannelEnd`]s keyed by their [`PortChannelId`]s.
     channels: MokaCache<PortChannelId, ChannelEnd>,
+    /// Cache storing [`ConnectionEnd`]s keyed by their [`ConnectionId`]s.
     connections: MokaCache<ConnectionId, ConnectionEnd>,
+    /// Cache storing [`AnyClientState`]s keyed by their [`ClientId`]s.
     client_states: MokaCache<ClientId, AnyClientState>,
+    /// The latest `Height` associated with the chain runtime this `Cache` is associated with.
+    // TODO: Could `Height` be stored with a TTL without having to instantiate a separate
+    // `MokaCache` instance?
     latest_height: MokaCache<(), Height>,
 }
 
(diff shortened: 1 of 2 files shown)


relayer/src/cache.rs
```diff
@@ -110,6 +131,8 @@ impl Cache {
         }
     }
 
+    /// Attempts to fetch the latest [`Height`] value. If a cache miss is encountered,
+    /// this method attempts to insert the height into the cache.
```
soareschen, line 135: Should note that technically, a chain's latest height is always increasing. We cache the latest height with a small TTL so that the latest height query returns the same height if the same query is repeated within a small time frame.

We rely on the assumption that the relayer will retry query on the latest height if it needs the chain to progress to the next height. However this assumption may be broken in some part of our code. Notably, the self-connected transfer test fails if the TTL is set to too high. I'll create a separate issue to track this.

relayer/src/cache.rs
```diff
@@ -74,6 +90,9 @@ impl Cache {
         }
     }
 
+    /// Attempts to fetch a [`ConnectionEnd`] via its [`ConnectionId`]. If a cache miss
+    /// is encountered, this method attempts to insert the connection into the cache,
+    /// assuming it is open.
```
soareschen, line 95: We only cache the connection and channel if the returned state is open.

soareschen (changes requested):

relayer/src/cache.rs
soareschen: Submitted #1947.

relayer/src/cache.rs
```diff
@@ -63,6 +76,9 @@ impl Cache {
         }
     }
 
+    /// Attempts to fetch a [`ChannelEnd`] via its [`PortChannelId`]. If a cache miss
+    /// is encountered, this method attempts to insert the channel into the cache,
+    /// assuming the returned channel state is open.
```
soareschen, line 81: It is more like: return a cached result in case there is a cache hit for the given `PortChannelId`. Otherwise attempt to fetch the `ChannelEnd` using the provided fetcher function `f`. If the fetcher returns an `Ok` result with the channel end in open state, store a copy of it in the cache and then return the result.

Similar applies to the other functions.

soareschen (changes requested):

relayer/src/cache.rs
adizere: Nitpicking, but agree that this comment could use some rephrasing. Some questions to understand possible confusions that could arise:

- "Attempts to fetch" is ambiguous: fetch from cache or fetch from chain via `f`?
    - could restate as "Attempts to read from cache" or Soares' proposal
- "attempts to insert the channel into the cache": maybe document that this method resorts to an expensive call to the chain (via `f`) upon cache miss.
- "assuming the returned channel state is open": it is unclear where is the channel returned from? the channel returned from `f` or from this method?

relayer/src/cache.rs
```diff
@@ -120,6 +141,11 @@ impl Cache {
         }
     }
 
+    /// Attempts to fetch the latest [`Height`] value. This value is cached with a small
```
adizere, line 144:
```suggestion
    /// Returns the latest [`Height`] value, if present in the cache. This value is cached with a small
```

relayer/src/cache.rs
```diff
@@ -84,6 +100,9 @@ impl Cache {
         }
     }
 
+    /// Attempts to fetch a [`ConnectionEnd`] via its [`ConnectionId`]. If a cache miss
```
adizere, line 103:
```suggestion
    /// Returns the cached [`ConnectionEnd`] via its [`ConnectionId`], if present in the cache. If a cache miss
```

relayer/src/cache.rs
```diff
@@ -103,6 +122,8 @@ impl Cache {
         }
     }
 
+    /// Attempts to fetch an [`AnyClientState`] via its [`ClientId`]. If a cache miss
```
adizere, line 125:
```suggestion
    /// Returns the [`AnyClientState`] via its [`ClientId`], if present in the cache. If a cache miss
```

adizere (commented):
Thanks Sean for the neat docs! We were speeding to merge the caching PR during the hackaton and overlooked the comments & docs.

Left some suggestion to improve the wording.

soareschen (approved):

PR #1944, opened 2022-03-07
Packet data now prints on one line as a byte vector

plafer:
Closes: #1559 

## Description

`Vec<T>` pretty prints as one item per line (using `"{:#?}"`). To get around that and have the pretty-print debug output have every element on one line, I wrap `Vec<T>` in a `newtype` and implement the `Debug` trait manually on it, which lets me control the pretty-printed debug output.

## Questions
- This PR introduces breaking changes to the public interface. How do we typically manage such changes?
-  Is this a sufficiently big change to warrant a changelog entry?
______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/modules/src/core/ics04_channel/packet.rs
+++ b/modules/src/core/ics04_channel/packet.rs
@@ -115,7 +115,7 @@ pub struct Packet {
 struct PacketData<'a>(&'a [u8]);
 
 impl<'a> ::core::fmt::Debug for PacketData<'a> {
-    fn fmt(&self, formatter: &mut ::core::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+    fn fmt(&self, formatter: &mut ::core::fmt::Formatter<'_>) -> Result<(), ::core::fmt::Error> {
         write!(formatter, "{:?}", self.0)
     }
 }
--- a/modules/src/core/ics04_channel/packet.rs
+++ b/modules/src/core/ics04_channel/packet.rs
@@ -97,7 +97,9 @@ impl core::fmt::Display for Sequence {
     }
 }
 
-#[derive(Clone, Debug, Default, Hash, PartialEq, Eq, Deserialize, Serialize)]
+// Remember: when the contents of the struct are changed, the `Debug`
+// implementation must be adjusted to reflect the changes.
+#[derive(Clone, Default, Hash, PartialEq, Eq, Deserialize, Serialize)]
 pub struct Packet {
     pub sequence: Sequence,
     pub source_port: PortId,
@@ -109,6 +111,33 @@ pub struct Packet {
     pub timeout_height: Height,
     pub timeout_timestamp: Timestamp,
 }
+
+struct PacketData<'a>(&'a [u8]);
+
+impl<'a> ::core::fmt::Debug for PacketData<'a> {
+    fn fmt(&self, formatter: &mut ::core::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        write!(formatter, "{:?}", self.0)
+    }
+}
+
+impl ::core::fmt::Debug for Packet {
+    fn fmt(&self, formatter: &mut ::core::fmt::Formatter<'_>) -> Result<(), ::core::fmt::Error> {
+        let data_wrapper = PacketData(&self.data);
+
+        formatter
+            .debug_struct("Packet")
+            .field("sequence", &self.sequence)
+            .field("source_port", &self.source_port)
+            .field("source_channel", &self.source_channel)
+            .field("destination_port", &self.destination_port)
+            .field("destination_channel", &self.destination_channel)
+            .field("data", &data_wrapper)
+            .field("timeout_height", &self.timeout_height)
+            .field("timeout_timestamp", &self.timeout_timestamp)
+            .finish()
+    }
+}
+
 
 impl Packet {
     /// Checks whether a packet from a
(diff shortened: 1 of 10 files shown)


modules/src/core/ics04_channel/packet.rs
```diff
@@ -105,7 +123,7 @@ pub struct Packet {
     pub destination_port: PortId,
     pub destination_channel: ChannelId,
     #[serde(serialize_with = "crate::serializers::ser_hex_upper")]
-    pub data: Vec<u8>,
+    pub data: PacketData,
```
mzabaluev, line 126: Could the `Debug` impl of `Packet` be hand-implemented (with all the help you can get from `DebugStruct` and a temporary wrapper struct) rather than introducing a newtype for the member?

plafer: Good idea, I didn't know about `DebugStruct`. The only downside that I see to that approach would be that when we modify `Packet`, we need to remember to go add that field to the debug output. I still think it's a cleaner approach, thanks for the suggestion!

It's probably easier to close this PR and open a different one, instead of reverting all the changes I made.

plafer:
The new implementation requires us to be careful when we update the `Packet` struct: we must manually update the `Debug` output as well. I added a comment over Packet to help us remember.

I explored writing a `derive` macro to generate the `Debug` output automatically, but it quickly became a lot of bloat code just for print output. Let me know if you can think of a better way to do it.

modules/src/core/ics04_channel/packet.rs
```diff
@@ -110,6 +112,32 @@ pub struct Packet {
...
+
+impl<'a> ::core::fmt::Debug for PacketData<'a> {
+    fn fmt(&self, formatter: &mut ::core::fmt::Formatter<'_>) -> Result<(), ::core::fmt::Error> {
+        write!(formatter, "{:?}", self.0)
+    }
+}
+
+impl ::core::fmt::Debug for Packet {
+    fn fmt(&self, formatter: &mut ::core::fmt::Formatter<'_>) -> Result<(), ::core::fmt::Error> {
+        let data_wrapper = PacketData(&self.data);
```
hu55a1n1, line 125: Thanks, @plafer! Nicely done! 👌 
Might be helpful to destructure the packet here so that the build fails when we forget to update the debug impl after we modify the `Packet` struct. And maybe add a comment here to remind us to update the code below when that happens.
```suggestion
        let Packet {
            sequence,
            source_port,
            source_channel,
            destination_port,
            destination_channel,
            data,
            timeout_height,
            timeout_timestamp,
        } = self;
        let data_wrapper = PacketData(data);
```

hu55a1n1, line 125: Thanks, @plafer! Nicely done! 👌 
Might be helpful to destructure the packet here so that the build fails when we forget to update the debug impl after we modify the `Packet` struct. And maybe add a comment here to remind us to update the code below when that happens.
```suggestion
        let Packet {
            sequence,
            source_port,
            source_channel,
            destination_port,
            destination_channel,
            data,
            timeout_height,
            timeout_timestamp,
        } = self;
        let data_wrapper = PacketData(data);
```

plafer: Great idea, I didn't think of that!

plafer:
@hu55a1n1 Ready for review again

hu55a1n1 (approved):
🌷

modules/src/core/ics04_channel/packet.rs
```diff
@@ -110,6 +110,46 @@ pub struct Packet {
...
+        let Packet {
+            sequence: _,
+            source_port: _,
+            source_channel: _,
+            destination_port: _,
+            destination_channel: _,
+            data,
+            timeout_height: _,
+            timeout_timestamp: _,
+        } = self;
```
mzabaluev, line 136: Good, I didn't think of using this guard statement!

modules/src/core/ics04_channel/packet.rs
```diff
@@ -110,6 +110,46 @@ pub struct Packet {
     pub timeout_timestamp: Timestamp,
 }
 
+struct PacketData<'a>(&'a [u8]);
+
+impl<'a> ::core::fmt::Debug for PacketData<'a> {
+    fn fmt(&self, formatter: &mut ::core::fmt::Formatter<'_>) -> Result<(), ::core::fmt::Error> {
```
mzabaluev, line 116: Minor nit: it would be better to use `core::fmt` at the top to reduce namespacing cruft here.

mzabaluev (approved):

modules/src/core/ics04_channel/packet.rs
plafer: I usually like to be explicit about path resolution; this version makes its clear when quickly glancing at it that I'm referring to the `core` crate rather than a potential `core` submodule (which in the general case I have to go check if such a module exists). However it's very minor indeed, and removing it has the benefit of conforming to the style of the rest of the repo.

Come to think of it, it's also by and large mitigated by being careful not to name any of our modules the name of an external crate.

modules/src/core/ics04_channel/packet.rs
seanchen1991: If all you want is the `data` field, this can be written more concisely as `let Packet { data, .. } = self;`

plafer: Here, the "verbosity" is intentional: See [@hu55a1n1 's suggestion](https://github.com/informalsystems/ibc-rs/pull/1944#discussion_r823489356).

Basically we want the compiler to yell if we change any field in `Packet`, so that we can remember to update the debug output.

seanchen1991: That makes sense. And a pretty clever way to achieve that 😄

PR #1976, opened 2022-03-17
Re-introduce relayer supervisor layer in test framework

soareschen:
## Description

This PR re-introduces the relayer supervisor test layer to allow the relayer supervisor to be spawned automatically by the test framework.

A new `RunWithSupervisor` struct is introduced to wrap any test cases at the chain layer or higher. The runner will spawn the supervisor before the main test is executed.

A new `SupervisorOverride` trait is introduced to override whether `RunWithSupervisor` should spawn the supervisor. Automatic spawning of supervisor can be disabled by implementing `should_spawn_supervisor()` to return `false` in `TestOverrides`.

The `RunWithSupervisor` wrapper is applied by test runner functions at the highest level before test execution. For example, `run_binary_chain_test` spawns the supervisor right after the chains are bootstrapped, but `run_binary_channel_test` only spawns the supervisor after the channels have been bootstrapped.

When manually composing `RunWithSupervisor` with other runner structs, one should make sure that `RunWithSupervisor` is only present once inside the compositions. Otherwise multiple relayer supervisors may be spawned and running concurrently during the test.

More explanation and documentation will be added in this PR.

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


romac (commented):
Looks good! Do we need to do anything for the ICA filter test or can it stay as is?

soareschen:
> Looks good! Do we need to do anything for the ICA filter test or can it stay as is?

I think no need, since it is using `run_binary_connection_test` already automatically spawns the supervisor. Unless the original test should _not_ spawn the supervisor?

romac (approved):

PR #1979, opened 2022-03-17
Skip waiting for confirmation events on tx raw upgrade-chain

mzabaluev:
Closes: #1288

## Description

An alternative to #1909 simplified as per suggestions in https://github.com/informalsystems/ibc-rs/issues/1288#issuecomment-1066675759, with the `tx raw upgrade-chain` command modified to only submit the transaction to the node and print out the hash.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


adizere (commented):
I think this solution is a good compromise, but does feel like a compromise.

I had a look at our [guide for chain upgrade](https://hermes.informal.systems/commands/upgrade/test.html) and noticed that step 3 (which we're reworking here) is following by a step 4 where we manually investigate the success of the `upgrade-chain` tx. From this perspective, the approach in this PR doesn't even feel like a compromise, because the user is expected to anyway investigate the success of the upgrade themselves.

Thanks Mikhail for the patience with this thorny issue, your diligence in exploring solutions here is impressive. 💯 

Looking forward to the comparison with #1909.

adizere (approved):

Issue #1759, opened 2022-01-11
Complete ICS20 ft-transfer implementation


hu55a1n1:
## Crate
ibc

## Summary
This is an attempt to gather all requirements for a complete ICS20 implementation. 

## Problem Definition
- [x] #1758 
- [ ] ICS20 Keeper - this would require access to an `AccountReader/Keeper` and a `BankReader/Keeper`, similar to the ibc-go's `Keeper` (see https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/keeper/keeper.go) ->
```go
type Keeper struct {
        // ...
	ics4Wrapper   types.ICS4Wrapper
	channelKeeper types.ChannelKeeper
	portKeeper    types.PortKeeper
	authKeeper    types.AccountKeeper
	bankKeeper    types.BankKeeper
}
```
The `Keeper` must additionally provide helper methods for port-binding, denom-traces, etc.
- [ ] ICS26 callbacks (See https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/ibc_module.go)
- [ ] Relay code (See https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/keeper/relay.go)

## Proposal
I propose we design an API that doesn't make assumptions on the SDK stack. A detailed proposal can be discussed via a draft PR. 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Wizdave97:
@hu55a1n1  @adizere  put up a draft PR for the ICS 20 Keeper, please give a review https://github.com/informalsystems/ibc-rs/pull/1884

PR #1884, opened 2022-02-17
ICS20 ADR Proposal

Wizdave97:
Closes: https://github.com/informalsystems/ibc-rs/issues/2090

## Description

______

### PR author checklist:
- [ ] Added changelog entry, using [`unclog`](https://github.com/informalsystems/unclog).
- [ ] Added tests: integration (for Hermes) or unit/mock tests (for modules). 
- [ ] Linked to GitHub issue.
- [ ] Updated code comments and documentation (e.g., `docs/`).

### Reviewer checklist:

- [ ] Reviewed `Files changed` in the GitHub PR explorer.
- [ ] Manually tested (in case integration/unit/mock tests are absent).

PR #1989, opened 2022-03-22
Complete ICS20 Implementation

hu55a1n1:
Closes: #1759 
References: #1884 
______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


plafer:
#1884 was designed so that `ICS-20` is a separate library, independent of ibc-rs. We should either pull this into a separate library, or change the ADR

modules/src/core/ics26_routing/handler.rs
```diff
@@ -386,12 +370,6 @@ mod tests {
                 msg: Ics26Envelope::Ics4ChannelMsg(ChannelMsg::ChannelOpenAck(msg_chan_ack)),
                 want_pass: true,
             },
-            //ICS20-04-packet
-            Test {
```
adizere, line 390: These tests are disabled due to the new way we call into `ics20_msg_dispatcher`, which is done via `ics4_packet_callback`.

hu55a1n1:
The `core::ics26_routing::handler::tests::routing_module_and_keepers` test is failing because `Ics26Context` no longer has a `Ics20Context` supertrait. 
https://github.com/informalsystems/ibc-rs/pull/1989/files#diff-d5bb0fe91166c315b1d6483a67f4659e6e5f97d0c3423c3daee772a59929deb5L32-L35

This is because the `core` module is expected to implement `Ics26Context` and routing, while the `transfer` app module is expected to implement `Ics20Context`. It appears that the current design of the `MockContext` doesn't allow for such a separation of concerns, because the `MockContext` itself implements the `Ics26Context`, and stores all IBC related data paths; so it isn't clear how application modules stored by the `MockRouter` can access IBC data. 

Here are two solutions that I could think of ->
1. Move all IBC related data fields into a separate type (`IbcStore`?) and put that behind an `Arc<Mutex<IbcStore>>` or something. 
2. Use `mpsc::channel` and the app-driver pattern - so that app modules are able to access the data via the channel that is created as part of the `MockContext` ctor. This would also require us to provide a `MockContext::run()` method that would run on a separate thread to service requests on the channel.

modules/src/applications/ics20_fungible_token_transfer/msgs/transfer.rs
```diff
@@ -23,9 +25,9 @@ pub struct MsgTransfer {
     /// the channel by which the packet will be sent
     pub source_channel: ChannelId,
     /// the tokens to be transferred
-    pub token: Option<ibc_proto::cosmos::base::v1beta1::Coin>,
+    pub token: IbcCoin,
     /// the sender address
-    pub sender: Signer,
+    pub sender: Address,
     /// the recipient address on the destination chain
     pub receiver: Signer,
```
DaviRain-Su, line 32: Why this receiver Type is not Address?

hu55a1n1: The `Address` is a valid `bech32` encoded address. I think the handlers should treat it as opaque data and leave such validation up to the host chain, so technically both `sender` and `receiver` should be `Signer` types and I am currently changing them both to be `Signer`. But we need to communicate that the address needs validation through the API and document it. I think the best place to document this would be the `FromStr` impl for `AccountId` ->
https://github.com/informalsystems/ibc-rs/pull/1989/files#diff-8f9bde3d68bd2354e1fa18e322a80e0c6d0b595d6b281db282116f948b44b8d5R137

modules/src/applications/ics20_fungible_token_transfer/events.rs
```diff
@@ -116,6 +117,22 @@ impl From<DenomTraceEvent> for ModuleEvent {
     }
 }
 
+pub struct TransferEvent {
```
DaviRain-Su, line 120: Maybe we need add transfer denom and amount.

hu55a1n1: I instead chose to log this to stay consistent with ibc-go -> https://github.com/informalsystems/ibc-rs/pull/1989/files#diff-3dee6bfbafa486af143dfe453c7cc29776865fc1da85b431cbf5fb57eaa637b9R114-R117

modules/src/applications/ics20_fungible_token_transfer/mod.rs
```diff
@@ -1,14 +1,21 @@
...
 
-mod denom;
 pub use denom::*;
 
+/// Module identifier for the ICS20 application.
+pub const MODULE_ID_STR: &str = "transfer";
+
 /// The port identifier that the ICS20 applications
 /// typically bind with.
 pub const PORT_ID: &str = "transfer";
```
plafer, line 21: can we make this `PORT_ID_STR`?

modules/src/applications/ics20_fungible_token_transfer/context.rs
```diff
@@ -1,5 +1,326 @@
...
+        let contents = format!("{}/{}", port_id, channel_id);
+
+        let mut hasher = Sha256::new();
+        hasher.update(VERSION.as_bytes());
+        hasher.update(b"0");
+        hasher.update(contents.as_bytes());
+        let hash: Vec<u8> = hasher.finalize().to_vec().drain(20..).collect();
+
+        String::from_utf8(hex::encode_upper(hash))
+            .expect("hex encoded bytes are not valid UTF8")
```
plafer, line 63: Could we give a more meaningful error *in case* this ever happens e.g. something of the kind `ibc-rs internal error. Please submit a bug report containing the stack trace.`

modules/src/applications/ics20_fungible_token_transfer/context.rs
```diff
@@ -1,5 +1,326 @@
...
+        channel_id: ChannelId,
+    ) -> Result<<Self as Ics20Reader>::AccountId, Ics20Error> {
+        // https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-028-public-key-addresses.md
+        let contents = format!("{}/{}", port_id, channel_id);
+
+        let mut hasher = Sha256::new();
+        hasher.update(VERSION.as_bytes());
+        hasher.update(b"0");
+        hasher.update(contents.as_bytes());
+        let hash: Vec<u8> = hasher.finalize().to_vec().drain(20..).collect();
```
plafer, line 60: The way `drain()` is used here, we're keeping everything *but the first 20 bytes*. That is, similar to `vec[20:]` in Python.

```suggestion
        let hasher_result = hasher.finalize();
        let hash: &[u8] = hasher_result.chunks(20).next().unwrap();
```

modules/src/applications/ics20_fungible_token_transfer/context.rs
```diff
@@ -1,5 +1,326 @@
...
+        &self,
+        port_id: &PortId,
+        channel_id: ChannelId,
+    ) -> Result<<Self as Ics20Reader>::AccountId, Ics20Error> {
+        // https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-028-public-key-addresses.md
+        let contents = format!("{}/{}", port_id, channel_id);
+
+        let mut hasher = Sha256::new();
+        hasher.update(VERSION.as_bytes());
+        hasher.update(b"0");
```
plafer, line 58: `b"0"` actually evaluates to byte `48` (from ascii)

```suggestion
        hasher.update([0]);
```

modules/src/applications/ics20_fungible_token_transfer/context.rs
```diff
@@ -1,5 +1,326 @@
...
+    + ChannelReader
+    + PortReader
+{
+    type AccountId: Into<String> + FromStr;
+
+    /// get_port returns the portID for the transfer module.
+    fn get_port(&self) -> Result<PortId, Ics20Error>;
+
+    /// Returns the escrow account id for a port and channel combination
+    fn get_channel_escrow_address(
```
plafer, line 48: We should add a few tests to make sure that we generate the same address as `ibc-go` given the same input port/channel.

plafer: For example, [this](https://www.mintscan.io/cosmos/account/cosmos1x54ltnyg88k0ejmk8ytwrhd3ltm84xehrnlslf) is the account on cosmos hub between cosmos hub and osmosis (channel `channel-141` and port `transfer`). We should make sure that we always get this address out given these inputs (and probably a few others).

modules/src/applications/transfer/context.rs
```diff
@@ -0,0 +1,348 @@
...
+    fn set_denom_trace(&mut self, denom_trace: &DenomTrace) -> Result<(), Ics20Error>;
+}
+
+pub trait Ics20Reader:
+    BankReader<AccountId = <Self as Ics20Reader>::AccountId>
+    + AccountReader<AccountId = <Self as Ics20Reader>::AccountId>
+    + ChannelReader
+    + PortReader
+{
+    type AccountId: ToString + TryFrom<Signer>;
```
plafer, line 38: 1. Why do we need the `ToString`?
2. Why `TryFrom<Signer>` as opposed to `From<Signer>`? Isn't `Signer` supposed to represent an account as well?
3. `Ics20Keeper::AccountId` only requires `ToString`. Is this intended?

modules/src/applications/transfer/packet.rs
```diff
@@ -0,0 +1,43 @@
...
+
+use ibc_proto::ibc::applications::transfer::v2::FungibleTokenPacketData as RawPacketData;
+use serde::{Deserialize, Serialize};
+
+use super::error::Error;
+use super::{Amount, DenomTrace, PrefixedCoin};
+use crate::signer::Signer;
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct PacketData {
```
plafer, line 13: How about we name this the more descriptive `FungibleTokenPacketData`?

modules/src/applications/transfer/relay/on_recv_packet.rs
```diff
@@ -0,0 +1,86 @@
...
+    ctx: &Ctx,
+    output: &mut ModuleOutputBuilder,
+    packet: &Packet,
+    data: PacketData,
+) -> Result<Box<WriteFn>, Ics20Error> {
+    if !ctx.is_receive_enabled() {
+        return Err(Ics20Error::receive_disabled());
+    }
+
+    let receiver = data
```
plafer, line 20: Can we name this `receiver_account`? "Receiver" and "Sender" can stand for a chain or an account, so it would help readability to make explicit which we're referring to.

modules/src/applications/transfer/context.rs
```diff
@@ -0,0 +1,348 @@
...
+
+    /// get_port returns the portID for the transfer module.
+    fn get_port(&self) -> Result<PortId, Ics20Error>;
+
+    /// Returns the escrow account id for a port and channel combination
+    fn get_channel_escrow_address(
+        &self,
+        port_id: &PortId,
```
plafer, line 46: ICS-20 indexes the escrow addresses only by channel id, whereas ibc-go uses channel+port ids. Since channel ids are already assumed to be unique, isn't the `port_id` redundant here? I think we should do as ICS-20 does and index only by channel id

modules/src/applications/transfer/context.rs
```diff
@@ -0,0 +1,348 @@
...
+    if channel_id.sequence() > (u32::MAX as u64) {
+        return Err(Ics20Error::chan_seq_exceeds_limit(channel_id.sequence()));
+    }
+
+    if order != Order::Unordered {
+        return Err(Ics20Error::channel_not_unordered(order));
+    }
+
+    let bound_port = ctx.get_port()?;
+    if port_id != &bound_port {
```
plafer, line 157: ICS-20 doesn't have this check (only ibc-go does). Is there a good reason to force chains to only bind the ICS-20 module to a predefined port?

modules/src/applications/transfer/context.rs
```diff
@@ -0,0 +1,348 @@
...
+        module: &Self::AccountId,
+        amt: &IbcCoin,
+    ) -> Result<(), Ics20Error>;
+}
+
+pub trait BankReader {
+    type AccountId: ToString + TryFrom<Signer>;
+
+    /// Returns true if the specified account is not allowed to receive funds and false otherwise.
+    fn is_blocked_account(&self, account: &Self::AccountId) -> bool;
```
plafer, line 119: Do we really need to expose `is_blocked_account()` and run the check ourselves in `process_recv_packet()`? The `BankKeeper::send_coins()` function is allowed to fail, so we can leave it to the implementations to check if the receiver address is blocked before transferring the funds (which they should do anyways).

modules/src/applications/transfer/context.rs
```diff
@@ -0,0 +1,348 @@
...
+
+    /// get_transfer_account returns the ICS20 - transfers AccountId.
+    fn get_transfer_account(&self) -> Self::AccountId;
+}
+
+pub trait AccountReader {
+    type AccountId: ToString;
```
plafer, line 126: What are the benefits of using `AccountId` to identify accounts as opposed to using `Address` directly? One consequence of the current interface is that an account can have multiple addresses. Is this how we intend to think about what an "account" is?

Note: the SDK's [`AccountI`](https://github.com/cosmos/cosmos-sdk/blob/450cd7fc8708ccb0fa21f05e251d9804a2063b79/x/auth/types/account.go#L307) models an account as having only one associated address.

modules/src/applications/ics20_fungible_token_transfer/mod.rs
hu55a1n1: Done! eff458c2c03a7ac803f941968be521397f5bfe20

modules/src/applications/ics20_fungible_token_transfer/context.rs
hu55a1n1: Oops sorry, my bad. Nice catch btw! Fixed in ea0d62234ef387f27a8198128c0b1457f31d2bbf.

modules/src/applications/ics20_fungible_token_transfer/context.rs
hu55a1n1: 0d349d23d9a024e2d3766e3b86fff939fa917479

modules/src/applications/transfer/relay/on_recv_packet.rs
hu55a1n1: d6ec510bd7e5583b3739ff094d6fd3ca2194c445

modules/src/applications/ics20_fungible_token_transfer/context.rs
```diff
@@ -1,5 +1,326 @@
...
+    }
+
+    /// Returns true iff send is enabled.
+    fn is_send_enabled(&self) -> bool;
+
+    /// Returns true iff receive is enabled.
+    fn is_receive_enabled(&self) -> bool;
+
+    /// Returns true iff the store contains a `DenomTrace` entry for the specified `HashedDenom`.
+    fn has_denom_trace(&self, hashed_denom: &HashedDenom) -> bool {
```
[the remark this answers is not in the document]
hu55a1n1: IMHO, it's okay to have default implementations for convenience and some standard lib traits (eg. `std::iter`) also use this style. But I'm okay with removing this because it's only used at one place in the codebase. 👍 (823e7dd7637057e5bee0907b6ca2132212a076db)

modules/src/applications/transfer/context.rs
hu55a1n1: Great idea! Removed `is_blocked_account()` -> https://github.com/informalsystems/ibc-rs/pull/1989/commits/2c61fc296fdc6f3258793ed7809a9b46ae922d2f

modules/src/applications/transfer/context.rs
hu55a1n1: 1. Nice catch 👌, turns out we don't need `ToString`. (I thought we might need it for logging somewhere but that isn't the case.)
2. An `AccountId` is a valid account type on the host chain while a `Signer` is just a non-empty string with very generic/minimal validation. 
3. Just removed the `ToString` bound all together (a1cdafa27034b0f92e38804239f0a47c9a08426b). The bounds are based on what the lib code uses and since `Ics20Keeper` and its supertraits don't use `TryFrom<Signer>` I didn't add that.

modules/src/applications/transfer/context.rs
hu55a1n1: Seems like this is related to a vulnerability in the spec in the case of 'crossing-hellos' identified during an audit (see point 2 here https://github.com/cosmos/ibc-go/issues/65). There's an ibc-go PR for removing 'crossing-hellos' (https://github.com/cosmos/ibc-go/issues/1311), so it might be nice to raise this concern there and ask if there are any plans to backport this to ibc-go v2.

modules/src/applications/transfer/relay/on_recv_packet.rs
```diff
@@ -0,0 +1,86 @@
...
+                let hashed_denom = coin.denom.hashed();
+                if ctx.has_denom_trace(&hashed_denom) {
+                    ctx.set_denom_trace(&coin.denom)
+                        .map_err(|e| e.to_string())?;
+                }
+
+                let amount = IbcCoin::from(coin);
+                ctx.mint_coins(&ctx.get_transfer_account(), &amount)
+                    .map_err(|e| e.to_string())?;
+                ctx.send_coins_from_module_to_account(
```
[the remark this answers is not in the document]
hu55a1n1: IIUC, the SDK only allows module accounts to mint coins to restrict the ability to mint coins to certain 'privileged' accounts. On our side, both module accounts and user accounts are identified using `AccountId`s, so I think it makes sense to have different functions to send coins from user accounts to other user accounts and module accounts to user accounts. We can change the naming of functions to not use 'module' accounts and use privileged accounts or something or we can also use a different type to represent these accounts instead of `AccountId`.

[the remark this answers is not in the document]
plafer: But that's an implementation detail for the SDK, no? Our trait could ask the chain "mint coins to this user address", and the chain can do so however it wishes. In the case of the SDK, it would mint coins at the `transfer` module address, and send the coins from there to the user account (as we do here).

The chain can use whatever security policy for minting coins as it wishes. If it doesn't like the address that we request the `mint` to, then it can fail the request. We don't need to know why and how it got to that conclusion.

In other words, I could currently add a method `mint_coins_to_user_addr()` to the trait, and implement as we do here (mint to module account, then send to user addr), without changing the SDK's policy that coins can only be minted at module addresses.

modules/src/applications/transfer/context.rs
hu55a1n1: Just realized that this code is no longer used after recent refactoring, so removed it as well. https://github.com/informalsystems/ibc-rs/pull/1989/commits/2d3a224d304780110290761878f4ce5246db4311

modules/src/applications/transfer/denom.rs
```diff
@@ -0,0 +1,628 @@
...
+impl<'a> TryFrom<Vec<&'a str>> for TracePath {
+    type Error = Error;
+
+    fn try_from(v: Vec<&'a str>) -> Result<Self, Self::Error> {
+        if v.len() % 2 != 0 {
+            return Err(Error::invalid_trace_length(v.len()));
+        }
+
+        let mut trace = vec![];
+        let id_pairs = v.windows(2).map(|paths| (paths[0], paths[1]));
```
plafer, line 80: We can use `chunks_exact(2)` here instead of `windows(2).step_by(2)`

modules/src/applications/transfer/denom.rs
```diff
@@ -0,0 +1,628 @@
...
+        for (pos, (port_id, channel_id)) in id_pairs.step_by(2).enumerate() {
+            let port_id =
+                PortId::from_str(port_id).map_err(|e| Error::invalid_trace_port_id(pos, e))?;
```
plafer, line 83: The current error message prints out the `TracePrefix` index. It would be better & sufficient to print out the failing port, perhaps with its channel too?

Same goes for when the channel id parse fails.

modules/src/applications/transfer/denom.rs
```diff
@@ -0,0 +1,628 @@
...
+        Ok(trace.into())
+    }
+}
+
+impl FromStr for TracePath {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let parts = {
+            let parts: Vec<&str> = s.split('/').collect();
```
plafer, line 101: Do we want to support base denoms with slashes? Note that I submitted an [issue](https://github.com/cosmos/ibc/issues/737) in the ICS repo so that the standard gets updated to match ibc-go, but it's unclear if the standard will get updated.

modules/src/applications/transfer/denom.rs
```diff
@@ -0,0 +1,628 @@
...
+            .collect::<Vec<String>>()
+            .join("/");
+        write!(f, "{}", path)
+    }
+}
+
+/// Indicates whether the sender chain is acting as a source or sink. Each send to any chain other
+/// than the one it was previously received from is a movement forwards in the token's timeline. In
+/// these instances the sender chain is acting as the source zone.
+pub enum Source {
```
plafer, line 127: Would it be better to rename as
+ `Source` -> `TokenSource`
+ `Sender` -> `SenderChain`
+ `Receiver` -> `ReceiverChain`

?

modules/src/applications/transfer/denom.rs
```diff
@@ -0,0 +1,628 @@
...
+
+impl DenomTrace {
+    /// Returns a coin denomination for an ICS20 fungible token in the format
+    /// 'ibc/{Hash(trace_path/base_denom)}'.
+    pub fn hashed(&self) -> HashedDenom {
+        HashedDenom::from(self)
+    }
+
+    /// Returns true iff this path has the specified prefix
+    pub fn has_prefix(&self, prefix: &TracePrefix) -> bool {
```
plafer, line 150: Can we rename to `starts_with()` or `trace_starts_with()`? 

The current name to me is ambiguous as whether the method checks 
1. for the `prefix` simply being present in the trace, or 
2. for the first prefix of the trace to be equal to the input `prefix`.

modules/src/applications/transfer/denom.rs
```diff
@@ -0,0 +1,628 @@
...
+            .0
+            .first()
+            .map(|p| p == prefix)
+            .unwrap_or(false)
+    }
+
+    /// Removes the specified prefix from the `trace_path` if there is a match.
+    pub fn remove_prefix(&mut self, prefix: &TracePrefix) {
+        if self.has_prefix(prefix) {
+            self.trace_path.0.drain(..0);
```
plafer, line 161: Can we avoid using `drain()` since the interface is confusing, as previously discussed? Maybe we can change the order of the prefixes in `TracePath` such that removing a prefix can be achieved using `pop()`?

modules/src/applications/transfer/denom.rs
```diff
@@ -0,0 +1,628 @@
...
+        }
+    }
+
+    /// Adds the specified prefix to the `trace_path`.
+    pub fn add_prefix(&mut self, prefix: TracePrefix) {
+        self.trace_path.0.push(prefix)
```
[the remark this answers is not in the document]
hu55a1n1: Great catch again! 💯 https://github.com/informalsystems/ibc-rs/pull/1989/commits/a7622c9b295599ee130aeb9e2a16fcc4122bef95 & https://github.com/informalsystems/ibc-rs/pull/1989/commits/76b1a2709d32682fc6b14c0c49f4fc3a64952e0b
I also added a test for this and moved those methods to `TracePath` just as you suggested.

modules/src/applications/transfer/denom.rs
hu55a1n1: Done! https://github.com/informalsystems/ibc-rs/pull/1989/commits/9194f01a86830618b87f9d5cd86737f2b478fe86

modules/src/applications/transfer/denom.rs
```diff
@@ -0,0 +1,628 @@
...
+/// than the one it was previously received from is a movement forwards in the token's timeline. In
+/// these instances the sender chain is acting as the source zone.
+pub enum Source {
+    Sender,
+    Receiver,
+}
+
+/// A type that contains the base denomination for ICS20 and the source tracing information path.
+#[derive(Clone, Debug, Eq, PartialEq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct DenomTrace {
```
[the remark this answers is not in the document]
hu55a1n1: I agree that `DenomTrace` is more than just a trace as it also contains the base denom. IMHO, I think it would be better to rename `DenomTrace` to `PrefixedDenom`. Then `Denom` could be left as it is or we could also rename that to `BaseDenom`.  WDYT?

modules/src/applications/transfer/denom.rs
hu55a1n1: Didn't know about `chunks_exact()` but looks like exactly what we need. 👍 https://github.com/informalsystems/ibc-rs/pull/1989/commits/df94b922e89c9ab9d7781c9eb6f7c8bea22e42aa

modules/src/applications/transfer/relay/on_recv_packet.rs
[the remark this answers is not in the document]
hu55a1n1: You're right 👍 -> https://github.com/informalsystems/ibc-rs/pull/1989/commits/a75453db9074da904655b3142caaaca37bb57a41

modules/src/applications/transfer/context.rs
hu55a1n1: I think this is checking if the port-id specified in channel messages matches the port that the ICS-20 module is bound to. AFAIK the core IBC router should only call this module callback if it is registered against this port - that means the `port_id` in the above code is guaranteed to be the port that this module is bound to, and that's a guarantee that core IBC provides. It is however possible that `get_port()` returns a port that is not the one this module is bound to, possibly due to a app dev error.

modules/src/applications/transfer/denom.rs
hu55a1n1: I think the `FromStr` impls of both `PortId` and `ChannelId` already return errors that print the identifiers along with a reason for parse failure.

modules/src/applications/transfer/denom.rs
```diff
@@ -0,0 +1,653 @@
...
+    }
+
+    /// Adds the specified prefix to the trace path.
+    pub fn add_trace_prefix(&mut self, prefix: TracePrefix) {
+        self.trace_path.add_prefix(prefix)
+    }
+
+    /// Returns `Source::Receiver` if the denomination originally came from the receiving chain and
+    /// `Source::Sender` otherwise.
+    pub fn source_chain(&self, prefix: &TracePrefix) -> Source {
```
plafer, line 189: This method relies on the caller to build the prefix in the appropriate way for its semantics to be correct. In other words, the caller must construct `prefix` from a packet's `source_port` and `source_channel` for the return value of this function to be correct. This is an indication that this function doesn't belong in the `DenomTrace` type. Also, intuitively, a denom trace has no notion of a "source chain", let alone a chain.

We should create an `Ics20Packet` struct, which contains the deserialized `PacketData` as a field. We can then put `source_chain(&self)` as a method of `Ics20Packet`. IMO this is cleaner, as an ICS20 packet has all the information necessary to make that decision.

Bonus: with this proposed change we get the benefit that `process_recv_packet()` doesn't take a `Packet` and `PacketData`, where `PacketData` is contained within `Packet` in an serialized form. It can now take an `Ics20Packet`. That was bothering me :).

modules/src/applications/transfer/denom.rs
```diff
@@ -0,0 +1,653 @@
...
+        write!(f, "{}", path)
+    }
+}
+
+/// Indicates whether the sender chain is acting as a source or sink. Each send to any chain other
+/// than the one it was previously received from is a movement forwards in the token's timeline. In
+/// these instances the sender chain is acting as the source zone.
+pub enum Source {
+    Sender,
+    Receiver,
```
plafer, line 157: We should add another field: `Neither`. If Bob sends Charlie Alice's token, then neither Bob nor Charlie is the source.

modules/src/applications/transfer/denom.rs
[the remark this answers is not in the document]
plafer: I agree; let's rename `Denom` -> `BaseDenom` too.

The standard uses `denom` to signify a "prefixed denom" as well as a "base denom", so not using `Denom` in our code eliminates that source of confusion.

modules/src/applications/transfer/denom.rs
```diff
@@ -0,0 +1,653 @@
...
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_denom_trace() -> Result<(), Error> {
+        assert_eq!(
+            DenomTrace::from_str("transfer/channel-0/uatom")?,
+            DenomTrace {
+                trace_path: "transfer/channel-0".parse()?,
```
plafer, line 471: Can we write the expected values explicitly (e.g. `vec![TracePath(vec![TracePrefix::new(...)])]`) as opposed to using `parse()`? In the current form, it's hard for me to tell whether or not we're comparing `DenomTrace::from_str` with itself (or almost).

This applies to all other expected values in the tests

modules/src/applications/transfer/context.rs
```diff
@@ -0,0 +1,342 @@
...
+fn cosmos_adr028_escrow_address(port_id: &PortId, channel_id: ChannelId) -> Vec<u8> {
+    let contents = format!("{}/{}", port_id, channel_id);
+
+    let mut hasher = Sha256::new();
+    hasher.update(VERSION.as_bytes());
+    hasher.update([0]);
+    hasher.update(contents.as_bytes());
+
+    let mut hash = hasher.finalize().to_vec();
+    hash.drain(20..);
```
plafer, line 73: Similarly here, can we use `chunks_exact()` instead of `drain()`?

modules/src/applications/transfer/context.rs
adizere: So this check guards against an application development error? If so, makes sense to keep it.

modules/src/applications/transfer/denom.rs
adizere: Let's not block on this question.

modules/src/applications/transfer/denom.rs
adizere: I have trouble understanding this enum, so a renaming here would be warranted. Additionally, "source" is very overloaded term.

modules/src/applications/transfer/denom.rs
Wizdave97: I don't think that would work, there should ideally only be two cases, when the host chain is acting as a source or sink, might be better to rename the enum and it's  variants to better express this. 
```rust 
     pub enum Zone {
         Source,
         Sink
     }

Comment (plafer) [reply modules/src/applications/transfer/denom.rs:157]: Assuming that we move the `source_chain()` to a new `Ics20Packet` struct as proposed [here](https://github.com/informalsystems/ibc-rs/pull/1989#discussion_r879758017), then returning a `Zone` as you suggest here would force us to add an extra argument to `source_chain()` as so:
```rust
fn source_chain(&self, HostIsReceiving | HostIsSending) -> Zone;
```
This is because a packet has all the information needed to decide whether the sender, receiver or neither is the source of the token being sent (by comparing the denom prefix with the packet's source channel/port), but a packet alone can't tell a given chain if it's source/sink. You need to add the information of if you're sending or receiving the packet. We would then need yet another `enum` for `HostIsReceiving` and `HostIsSending`. I find it more simple to simply have that information directly in the match statements (i.e. return that the source is sender/receiver/neither, and handle what to do in each case differently based on if you're the receiving or sending chain).

In other words, we would still have 2 branches in our match statements. For example, the `match` statement in `process_recv_packet()` would now look like (without including the `Ics20Packet` suggested refactoring):
```rust
match data.token.denom.source_chain(&prefix) {
    Source::Receiver | Source::Neither => { ... },
    Source::Sender => { ... }
}
```
The change here simply properly models the fact that there are 3 cases when asking "who is the source of the tokens in this token transfer" without changing the handling logic. For readability, really.

Does that make sense?

modules/src/applications/transfer/context.rs
plafer: I think I was more referring to the fact that modules can bind multiple ports in general, and so theoretically ICS-20 module could be bound to `"transfer"` and `"other_port"`. However, this check prevents us from doing that. I don't know if a chain will ever need that though? The upside of this tradeoff is indeed that it guards against dev error. I don't have a strong preference for keeping/removing it!

modules/src/applications/transfer/denom.rs
plafer: Oh, and `flex-error` would stack all these errors so that what the user ultimately sees is both the port/channel id, as well as this index message? If so, then yep let's leave it as is.

modules/src/applications/transfer/denom.rs
Wizdave97: Yeah, that makes sense.

modules/src/applications/transfer/denom.rs
```diff
@@ -0,0 +1,653 @@
...
+        if self.trace_path.0.is_empty() {
+            write!(f, "{}", self.base_denom)
+        } else {
+            write!(f, "{}/{}", self.trace_path, self.base_denom)
+        }
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, PartialOrd, Ord, From)]
+pub struct HashedDenom(Vec<u8>);
```
plafer, line 264: Should we even have a `HashedDenom`? I find that it adds a lot of complexity to this library (e.g. `IbcCoin` only exists because we have `HashedDenom`), which I believe should only contain the core token transfer logic. IMO, whether or not to hash denoms before storing them is an implementation detail of how chains store denoms, let alone the fact that we're currently forcing chains to deal with hashed denoms. In my view, this fits best in some `utils` library that chains can use if they so choose. 

The one reason I can see that we might want to keep it is so that we can emit the same events as ibc-go, and ibc-go currently emits events containing trace hashes. However, we could also simply hash the denom internally to create the event, and throw the hash away after.

modules/src/applications/transfer/relay/on_ack_packet.rs
```diff
@@ -0,0 +1,19 @@
+use crate::applications::transfer::acknowledgement::Acknowledgement;
+use crate::applications::transfer::context::Ics20Context;
+use crate::applications::transfer::error::Error as Ics20Error;
+use crate::applications::transfer::packet::PacketData;
+use crate::applications::transfer::relay::refund_packet_token;
+use crate::core::ics04_channel::packet::Packet;
+
+pub fn process_ack_packet(
```
plafer, line 8: This is a fairly short function called only once, we could just inline the implementation in `on_acknowledgement_packet()`?

modules/src/applications/transfer/relay/on_timeout_packet.rs
```diff
@@ -0,0 +1,13 @@
+use crate::applications::transfer::context::Ics20Context;
+use crate::applications::transfer::error::Error as Ics20Error;
+use crate::applications::transfer::packet::PacketData;
+use crate::applications::transfer::relay::refund_packet_token;
+use crate::core::ics04_channel::packet::Packet;
+
+pub fn process_timeout_packet(
```
plafer, line 7: Similarly, can we inline this function where it is called in `on_timeout_packet()`?

modules/src/applications/transfer/denom.rs
```diff
@@ -0,0 +1,653 @@
...
+        }
+
+        let bytes = hex::decode_upper(parts[1]).map_err(Error::parse_hex)?;
+        Ok(Self(bytes))
+    }
+}
+
+/// A decimal type for representing token transfer amounts.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord, Display, From)]
+pub struct Amount(U256);
```
plafer, line 300: How are users expected to use this struct? For example, how do you implement `BankKeeper::burn_coins()` when I don't have a subtract operation, and that it's not possible for me to get to the underlying `U256`?

modules/src/applications/transfer/denom.rs
```diff
@@ -0,0 +1,653 @@
...
+}
+
+impl From<u64> for Amount {
+    fn from(v: u64) -> Self {
+        Self(v.into())
+    }
+}
+
+// We only provide an `Add<Decimal>` implementation which always panics on overflow.
+impl Add<Self> for Amount {
```
plafer, line 318: Seems like a vector of attack to crash a chain. I don't think a library should ever crash on error. We could instead expose `checked_add()` and more, and let the chain decide how it wants to handle overflows.

modules/src/applications/transfer/msgs/transfer.rs
```diff
@@ -23,7 +24,7 @@ pub struct MsgTransfer {
     /// the channel by which the packet will be sent
     pub source_channel: ChannelId,
     /// the tokens to be transferred
-    pub token: Option<ibc_proto::cosmos::base::v1beta1::Coin>,
+    pub token: IbcCoin,
```
plafer, line 27: I don't think denoms of the form `"ibc/{hash}"` are ever supposed to be sent over the wire, right? They're only an ibc-go internal detail IIUC.

modules/src/applications/transfer/relay/send_transfer.rs
```diff
@@ -0,0 +1,116 @@
...
+        data,
+        timeout_height: msg.timeout_height,
+        timeout_timestamp: msg.timeout_timestamp,
+    };
+
+    let HandlerOutput {
+        result,
+        log,
+        events,
+    } = send_packet(ctx, packet).map_err(Error::ics04_channel)?;
```
plafer, line 92: The assumption is that the chain will revert any changes made on error, right? Otherwise, if `send_packet()` fails, then we will have escrowed/burned the tokens without actually sending them.

modules/src/applications/transfer/context.rs
```diff
@@ -0,0 +1,342 @@
...
+    hash
+}
+
+pub trait BankKeeper {
+    type AccountId;
+
+    /// This function should enable sending ibc fungible tokens from one account to another
+    fn send_coins(
```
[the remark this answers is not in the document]
hu55a1n1: Good point! https://github.com/informalsystems/ibc-rs/pull/1989/commits/49ad00cd1500719e00425a0f12b7439e1c0675b9

modules/src/applications/transfer/denom.rs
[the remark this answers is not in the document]
hu55a1n1: https://github.com/informalsystems/ibc-rs/pull/1989/commits/b386e86173e83644dae0502a3f4b07e446b707da

modules/src/applications/transfer/relay/on_recv_packet.rs
```diff
@@ -0,0 +1,76 @@
...
+            let denom_trace_event = DenomTraceEvent {
+                trace_hash: coin.denom.hashed(),
+                denom: coin.denom.clone(),
+            };
+            output.emit(denom_trace_event.into());
+
+            Ok(Box::new(move |ctx| {
+                let ctx = ctx.downcast_mut::<Ctx>().unwrap();
+                let hashed_denom = coin.denom.hashed();
+                if ctx.get_denom_trace(&hashed_denom).is_some() {
```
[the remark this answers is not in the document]
hu55a1n1: eb5128aef3c72cb6882838ab76230fc33557208a

modules/src/applications/transfer/context.rs
hu55a1n1: I think `truncate` is a better choice. 81a98043c9a3ec8bd2c82be6d5f4731b4a58fedc

modules/src/applications/transfer/denom.rs
hu55a1n1: > We should create an Ics20Packet struct, which contains the deserialized PacketData as a field. 

This method is also called with the port and channel fields of `MsgTransfer` in `send_transfer()`. 
IMHO, we would be better off avoiding avoiding the complexity and following ibc-go's API - free-standing functions that return `bool`s as simple as -> 
```rust
fn is_sender_chain_source(source_port: &PortId, source_channel: ChannelId, denom: &PrefixedDenom) -> bool {
    !is_receiver_chain_source(source_port, source_channel, denom)
}

fn is_receiver_chain_source(source_port: &PortId, source_channel: ChannelId, denom: &PrefixedDenom) -> bool {
    // ...
}
```

modules/src/applications/transfer/context.rs
plafer: Absolutely!

modules/src/applications/transfer/context.rs
DaviRain-Su: seem the function `get_channel_escrow_address` for one chain have oneself implement.

modules/src/applications/transfer/denom.rs
hu55a1n1: Fixed by implementing `Into<U256>` and exposing `checked_add/sub()`. https://github.com/informalsystems/ibc-rs/pull/1989/commits/e16bccabaf4ce32bccdcbad4bf38625fb9572825

modules/src/applications/transfer/denom.rs
hu55a1n1: Will implement using freestanding funcs as discussed.

modules/src/applications/transfer/relay/send_transfer.rs
hu55a1n1: Right, I documented this in a comment -> 27a5d74733da424bc470d75024afec54b0900529

Issue #1921, opened 2022-02-25
Client `create` parameters are ignored


adizere:
## Summary of Bug

When creating a client with `hermes create client SRC DST` we can pass three parameters:

1. `-d, --clock-drift <CLOCK_DRIFT>`
    - Override the default clock drift specified in the configuration
2. `-p, --trusting-period <TRUSTING_PERIOD>`
    - Override the trusting period specified in the config
3. `-t, --trust-threshold <TRUST_THRESHOLD>`
    - Override the trust threshold specified in the configuration

It seems that these parameters are just ignored by Hermes.

## Version

I think this was added recently (#836), but I reproduced on current master

hermes 0.12.0+ffb2c522

ffb2c522

## Steps to Reproduce

We know the following: In config.toml we set for chain `ibc-1`:
- `trusting_period = '1days'`
-  `clock_drift = '500s'` as the basis for [computing the clock drift](https://github.com/informalsystems/ibc-rs/blob/0edb6b764ea07988e999982dbad9be1ad68f364d/relayer/src/chain/cosmos.rs#L2594)
- `trust_threshold = { numerator = '1', denominator = '3' }`

These are the default we would expect to be able to overwrite whenever we create a client for source chain `ibc-1`.

### For trusting period parameter `-p`

Steps:
1. Create a client for source chain `ibc-1` and destination `ibc-0` overwritting the trusting period
2. hermes create client ibc-0 ibc-1 -p 3days

```
Running `target/debug/hermes create client ibc-0 ibc-1 -p 3days`
2022-02-25T14:18:39.335177Z  INFO ThreadId(21) send_tx{id=ibc-0}: refresh: retrieved account sequence=21 number=1
Success: CreateClient(
...
            client_id: ClientId(
                "07-tendermint-3",
```

3. Query the newly-created client: `hermes query client state ibc-0 07-tendermint-3 | grep trusting`

```
trusting_period: 86400s,
```

It is still using the default 1day instead of 3days.

### For trust threshold `-t`

1. `hermes create client ibc-0 ibc-1 -t 2/3`

```
"07-tendermint-4",
```

2. `hermes query client state ibc-0 07-tendermint-3 | grep -A3 trust_level`

```
trust_level: TrustThreshold {
            numerator: 1,
            denominator: 3,
        },
```

### For max clock drift `-d`

1. `hermes create client ibc-0 ibc-1 -d 1s`
[...]

PR #2004, opened 2022-03-23
Use custom options of the `create client` command

mzabaluev:
Closes: #1921

## Description

Rework settings passed to `build_client_state` to make the custom parameters of `create client` command work.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


relayer/src/foreign_client.rs
```diff
@@ -277,6 +279,46 @@ impl HasExpiredOrFrozenError for ForeignClientError {
...
+}
+
+impl CreateOptions {
+    fn build_client_settings(
+        self,
+        src_chain_config: &ChainConfig,
+        dst_chain_config: &ChainConfig,
+    ) -> ClientSettings {
+        let max_clock_drift = self.max_clock_drift.unwrap_or_else(|| {
+            src_chain_config.clock_drift
```
romac, line 306: Can we keep `calculate_client_state_drift` and use it here to make the code a bit more self-documenting?

adizere (changes requested):
I did some tests and the CLI works as expected when customizing trusting period and trust threshold.

For customizing clock drift, I realize it's not clear to me what the expected behavior should be.

Example:

> hermes create client ibc-0 ibc-1 -t 1/2 -p 1200s -d 2s

Do we want `2s` to be either:
- option A: the final clock drift used in this client state? Or

- option B: do we want to use `2s` to compute the client's clock drift (like we used to do [here](https://github.com/informalsystems/ibc-rs/blob/43e5ccf44945c6bf989236d709a051dc5e997f80/relayer/src/chain/cosmos.rs#L2560)).

In this PR, we implement ~~option B~~ option A. I agree with this. So the CLI above gives

>...
        trust_level: TrustThreshold {
            numerator: 1,
            denominator: 2,
        },
        trusting_period: 1200s,
        unbonding_period: 1814400s,
        max_clock_drift: 2s,

Note that the `max_clock_drift` is set to `2s`, as I specified in the create command CLI. This means we should update the CLI help command, which is not accurate at the moment, because it says the following:
>    -d, --clock-drift <CLOCK_DRIFT>
            Override the default clock drift specified in the configuration.

With `-d` we're not exactly overriding the configuration value. Instead, we are providing the exact `clock_drift` to be used in the created client state.

relayer/src/foreign_client.rs
ancazamfir: imho we should keep as much of this in `cosmos.rs`. While `max_clock_drift` is a chain parameter and expected to be configured for all chains, the `trusting_period` and `trust_threshold` are tendermint specific and we should avoid having them in clear in this file. This problem exists in master and not introduced by this PR but we shouldn't make things worse. Ideally we should check in the CLI that the parameters apply for the source chain type and in this file `CreateOptions` should be serialized, later decoded in `cosmos.rs`. 

Here is the client state for the solomachine:
https://github.com/informalsystems/ibc-rs/blob/06cff5763459dd8ebf054131cc146a1b9a441358/proto/src/prost/ibc.lightclients.solomachine.v2.rs#L4-L17

ancazamfir:
> Do we want `2s` to be either:
> 
> * option A: the final clock drift used in this client state? Or
> * option B: do we want to use `2s` to compute the client's clock drift (like we used to do [here](https://github.com/informalsystems/ibc-rs/blob/43e5ccf44945c6bf989236d709a051dc5e997f80/relayer/src/chain/cosmos.rs#L2560)).
> 
> In this PR, we implement option B.

I think we implement option A, right?

relayer/src/foreign_client.rs
mzabaluev: I agree @ancazamfir, but I think `CreateOptions` could be an enum, with `CreateOptions::Tendermint` as the only currently defined variant carrying the tendermint-specific options.

ancazamfir:
One other thought is to warn the user if the configured client `max_clock_drift` is smaller than the destination chain `max_block_time`. In this case there is a high chance of "header in the future" chain errors.

tools/test-framework/src/bootstrap/binary/chain.rs
```diff
@@ -160,7 +160,7 @@ pub fn bootstrap_foreign_client<ChainA: ChainHandle, ChainB: ChainHandle>(
     let foreign_client =
         ForeignClient::restore(ClientId::default(), chain_b.clone(), chain_a.clone());
 
-    let event = foreign_client.build_create_client_and_send(&Default::default())?;
```
adizere, line 163: Can we add a new test? Would like to assert that:

1. default settings specified in the config.toml for source chain is respected when we create a new client

2. we correctly override the default settings with custom ones.

tools/integration-test/src/tests/client_settings.rs
```diff
@@ -0,0 +1,58 @@
...
+impl BinaryChainTest for ClientSettingsTest {
+    fn run<ChainA: ChainHandle, ChainB: ChainHandle>(
+        &self,
+        _config: &TestConfig,
+        relayer: RelayerDriver,
+        chains: ConnectedChains<ChainA, ChainB>,
+    ) -> Result<(), Error> {
+        let _refresh_tasks = spawn_refresh_client_tasks(&chains.foreign_clients)?;
+        let _supervisor = relayer.spawn_supervisor()?;
+        todo!("verify that the clients are created with the expected settings")
```
mzabaluev, line 48: @adizere @soareschen Here's the requested test using non-default settings, but I need help with properly querying the client configuration.

tools/integration-test/src/tests/client_settings.rs
```diff
@@ -0,0 +1,125 @@
...
+        config.chains[0].trust_threshold = TrustThreshold::new(13, 23).unwrap().try_into().unwrap();
+
+        config.chains[1].clock_drift = Duration::from_secs(6);
+        config.chains[1].max_block_time = Duration::from_secs(15);
+        config.chains[1].trusting_period = Some(Duration::from_secs(340_000));
+        config.chains[1].trust_threshold = TrustThreshold::TWO_THIRDS.try_into().unwrap();
+    }
+}
+
+impl BinaryChainTest for ClientDefaultsTest {
```
soareschen, line 45: I think it would be clearer to have a `BinaryNodeTest` for this case, with the foreign clients being created manually with the custom client settings you want. Basically the test should call `ForeignClient::new()` or `ForeignClient::restore()` directly, and then call `ForeignClient::build_create_client_and_send()` with the custom client settings, because this is the actual method we want to test here.

For this particular test, it is more of testing that the test framework is correctly initializing the foreign clients with the wanted settings. We could still leave the test here for testing the correctness of the test framework itself.

I think it is also worth writing some simple tests to demonstrate why we need a `SettingsTestOverrides`. For example, is having the ability to override the `max_clock_drift` setting helps us write new integration tests that we couldn't before? If the tests are too complicated to write in this PR, then we should file a new issues to follow up on writing these tests.

tools/integration-test/src/tests/client_settings.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+struct ClientDefaultsTest;
+
+struct ClientSettingsTest;
+
+struct SettingsTestOverrides;
+
+impl TestOverrides for ClientDefaultsTest {
+    fn modify_relayer_config(&self, config: &mut Config) {
```
soareschen, line 32: What is the purpose of this config override? When I comment out this override, the test still passes.

tools/integration-test/src/tests/client_settings.rs
mzabaluev: Good points, the tests could exercise the API directly indeed rather than deferring it to the framework.

As for `ClientSettingsOverride`, I think the ability to tweak the pre-created clients could be useful, and it does not get in the way of writing any tests that don't need this feature. The long trait bound lists in the framework code itself could become unwieldy, but they could be made more manageable with helper traits if needed. One thing this override trait cannot address is n-ary tests, but the same problem exists for the other override traits.

tools/integration-test/src/tests/client_settings.rs
mzabaluev: Really? I get an assertion failure as expected:

```
Message:  assertion failed: `(left == right)`
  left: `10s`,
 right: `24s`
Location: tools/integration-test/src/tests/client_settings.rs:55
```

soareschen: Ah I missed the fact that there are actually two tests in the file. Perhaps we should consider whether to group multiple independent tests in the same file inside nested `mod {}` for easier identification.

tools/integration-test/src/tests/client_settings.rs
```diff
@@ -0,0 +1,125 @@
...
+        let client_id = chains.foreign_clients.client_b_to_a.id();
+        let state = query_client_state(chains.handle_a, client_id)?;
+        assert_eq!(state.max_clock_drift, Duration::from_secs(14));
+        assert_eq!(state.trusting_period, Duration::from_secs(340_000));
+        assert_eq!(state.trust_level, TrustThreshold::TWO_THIRDS);
+        Ok(())
+    }
+}
+
+impl TestOverrides for SettingsTestOverrides {
```
soareschen, line 67: We can remove `SettingsTestOverrides` and just write `impl TestOverrides for ClientSettingsTest`, since the overrides are not shared across multiple tests.

mzabaluev: Correct, I have cargo-culted the separate overrides object approach from another test before noticing.

tools/integration-test/src/tests/client_settings.rs
soareschen: Ok, so do you think it is unnecessary to write another test using `BinaryNodeTest` to test `ForeignClient::build_create_client_and_send()` directly?

> One thing this override trait cannot address is n-ary tests, but the same problem exists for the other override traits.

For the case of ports override, we use nested arrays to provide a matrix of overrides. But I think we can just omit that for now and only add them when it is needed.

relayer/src/chain/client.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use crate::chain::cosmos;
+use crate::config::ChainConfig;
+
+/// Client parameters for the `build_create_client` operation.
+///
+/// The parameters are specialized for each supported chain type.
+#[derive(Clone, Debug)]
+pub enum ClientSettings {
+    Cosmos(cosmos::client::Settings),
```
adizere, line 11: Neat!

adizere (approved):
I only did a cursory review of the code, looks good, I have no suggestions there, cool work Mikhail esp. with integration test and modular support for capturing client state setting!

The CLI behaves as expected.

Will let @ancazamfir and @soareschen weigh in on the details of the implementation.

adizere:
Added the "needs guide update" label as it would be worthwhile to document the extended support for `-t` `-d` and `-p` options in the `create client` guide entry: https://hermes.informal.systems/commands/path-setup/clients.html#create-client

tools/integration-test/src/tests/client_settings.rs
mzabaluev: @romac says custom client settings could be useful to test misbehavior detection.

relayer/src/chain/client.rs
mzabaluev: I'm not sure if `Cosmos` or `Tendermint` is the right name to use here.
`AnyClientState` uses `Tendermint`, maybe we should use it here too for consistency?

adizere: This is a very good point. `Tendermint` is more appropriate than `Cosmos`.

relayer/src/chain/cosmos.rs
```diff
@@ -2077,17 +2078,25 @@ impl ChainEndpoint for CosmosSdkChain {
...
+            .expect("`fill_in_from_chain_configs` should have been called");
         let unbonding_period = self.unbonding_period()?;
-
-        let max_clock_drift = calculate_client_state_drift(self.config(), &dst_config);
+        let trusting_period = settings
+            .trusting_period
+            .unwrap_or_else(|| self.trusting_period(unbonding_period));
+        let trust_threshold = settings
+            .trust_threshold
+            .unwrap_or(self.config.trust_threshold.into());
```
soareschen, line 2093: There is an error here with `cargo clippy`.

```
error: use of `unwrap_or` followed by a function call
    --> relayer/src/chain/cosmos.rs:2093:14
     |
2093 |             .unwrap_or(self.config.trust_threshold.into());
     |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `unwrap_or_else(|| self.config.trust_threshold.into())`
     |
```

soareschen (approved):

tools/integration-test/src/tests/client_settings.rs
soareschen: Actually I meant writing an additional test to test the function directly. But anyway it's a nit and doesn't matter too much.

relayer/src/chain/client.rs
mzabaluev: OTOH the module hierarchy under `chain` uses `cosmos`. This is a relayer API anyway, so we don't have to be super-diligent.

PR #2006, opened 2022-03-23
Add metric for query cache hits

romac:
Closes: #2036

- [x] Add a `query_cache_hits` metric, indexed by chain and query type.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/features/relayer/2036-caching-metrics.md
+++ b/.changelog/unreleased/features/relayer/2036-caching-metrics.md
@@ -1,0 +1,2 @@
+- Added cache hits/miss metrics in telemetry
+  ([#2036](https://github.com/informalsystems/ibc-rs/issues/2036))
--- a/.changelog/unreleased/features/relayer/2036-caching-metrics.md
+++ b/.changelog/unreleased/features/relayer/2036-caching-metrics.md
@@ -1,2 +1,2 @@
-- Added cache hits/miss metrics in telemetry
+- Add a metric for query cache hits
   ([#2036](https://github.com/informalsystems/ibc-rs/issues/2036))
--- a/relayer/src/cache.rs
+++ b/relayer/src/cache.rs
@@ -23,6 +23,16 @@ const CHANNEL_CACHE_CAPACITY: u64 = 10_000;
 const CONNECTION_CACHE_CAPACITY: u64 = 10_000;
 const CLIENT_STATE_CACHE_CAPACITY: u64 = 10_000;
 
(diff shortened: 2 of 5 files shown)


relayer/src/cache.rs
```diff
@@ -84,20 +94,20 @@ impl Cache {
...
-            Ok(chan)
+            Ok((chan, CacheStatus::Hit))
         } else {
             // Only cache a channel end if the channel is open.
             let chan = f()?;
             if chan.state().is_open() {
                 self.channels.insert(id.clone(), chan.clone());
             }
-            Ok(chan)
+            Ok((chan, CacheStatus::Miss))
```
adizere, line 110: Hmm I imagine it might be cleaner to add calls to `telemetry!` directly in the cache layer (in these methods here), instead of doing it at the caller (in the `relayer/src/chain/handle/cache.rs`). What do you think?

relayer/src/chain/handle/cache.rs
```diff
@@ -131,8 +132,15 @@ impl<Handle: ChainHandle> ChainHandle for CachingChainHandle<Handle> {
...
     fn query_latest_height(&self) -> Result<Height, Error> {
         let handle = self.inner();
-        self.cache
-            .get_or_try_update_latest_height_with(|| handle.query_latest_height())
+        let (result, in_cache) = self
+            .cache
+            .get_or_try_update_latest_height_with(|| handle.query_latest_height())?;
+
+        if in_cache == CacheStatus::Hit {
+            telemetry!(query_cache_hit, &self.id(), "query_latest_height");
```
adizere, line 140: I guess the difficulty of pushing `telemetry!` calls inside the `cache` is that we don't have access to the chain identifier (`self.id()`). But maybe we can store the chain identifier as a field in the cache layer to fix this difficulty.

relayer/src/cache.rs
romac: In the cache layer we don't know what is the exact type of query that has been performed, hence why I chose to add this code here. Also, it keeps a nice separation between the cache and the queries+metrics.

adizere (approved):

.changelog/unreleased/features/relayer/2036-caching-metrics.md
```diff
@@ -0,0 +1,2 @@
+- Added cache hits/miss metrics in telemetry
```
romac, line 1:
```suggestion
- Add a metric for query cache hits
```

Issue #1978, opened 2022-03-17
Relaying of packets slows down exponentially if many channels are created after the relayer


soareschen:
## Summary of Bug

If many new channels are created after the relayer supervisor is spawned, the relaying of packets seem to slow down exponentially. This does not applies if the channels are created _before_ the supervisor is spawned.

## Steps to Reproduce

PR #1977 provides a test case `test_slow_ternary_ibc_transfer` that runs very slowly with the supervisor spawned before the channels are created. The original test `test_ternary_ibc_transfer` demonstrates that the test finishes much faster if the supervisor is spawned after the channels are created.

## Acceptance Criteria

The test cases `test_slow_ternary_ibc_transfer` and `test_ternary_ibc_transfer` in #1977 should take roughly the same time to finish.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

romac:
@ancazamfir managed to reproduced this with 3 chains, whether or not channels are spawned before or after the supervisor.  `git bisect` points to the first commit in #1656. Still unclear what exactly is going on.

romac:
We are both seeing that the SendPacket event is coming super late through the WebSocket, everything after that is super quick. Not clear what this has to do with #1656 since it does not even touch the event monitor.

soareschen:
It could be because #1656 does the refactoring of the original code, so you kind of track it manually before that commit.

I suspect it might be misuse of retry logic somewhere with an exponential back off mechanism. That might be why the events are processed slower and slower as the number of channels increases, even though the CPU is mostly idle.

romac:
The cause of slowness was that the height of the events we were pulling from the subscriptions would drift exponentially from the real latest height, because we were not pulling them from the event monitor stream fast enough.

We were trying to get an event from the stream of events (with `try_recv_multiple`) every 500ms.

So with two chains we have:
- call try_recv_multiple and get a NewBlock from chain A
- wait 500ms
- call try_recv_multiple and get a NewBlock from chain B
- wait 500ms
- call try_recv_multiple and get a NewBlock from chain A
etc.

So we get an event per chain roughly once every second.


With three chains:

- call try_recv_multiple and get a NewBlock from chain A
- wait 500ms
- call try_recv_multiple and get a NewBlock from chain B
- wait 500ms
[...]

PR #1977, opened 2022-03-17
Reproduce slowness of relayer when spawned before channels are bootstrapped

soareschen:
Closes #1978.

This PR is based on #1976.

## Description

This PR reproduces the relayer runnning very slowly in the test `test_ternary_ibc_transfer` when the relayer is spawned _first_ before the channels are being bootstrapped. The test can be reproduced manually by running:

```bash
cargo test -p ibc-integration-test --features manual -- test_slow_ternary_ibc_transfer
```

The ternary IBC transfer test performs IBC transfer between 3 fully connected chains. i.e. there are IBC channels established from chains A to A, A to B, A to C, B to B, B to C, C to C. As compared to the binary transfer tests, due to the increased number of chains and channels, the relaying seems to slow down exponentially as a result.

In the original test, the relaying starts pretty much immediately after an IBC transfer is initiated. But in this modified test, relaying the first IBC transfer takes 41 seconds, and subsequently 71 seconds, 75 seconds, and 118 (!) seconds for each transfer to be relayed. This is a rather significant delay that should not be ignored.

There don't seem to be any significant slow down if the relayer supervisor is spawned first in the case of binary transfer test, which can be reproduced manually with:

```bash
cargo test -p ibc-integration-test --features manual -- test_slow_ibc_transfer
```

More investigation is needed to determine why the relaying slows down exponentially as more channels are added after the relayer starts, but not if the channels are added _before_ the relayer is started.
______

### PR author checklist:
- [ ] Added changelog entry, using [`unclog`](https://github.com/informalsystems/unclog).
- [ ] Added tests: integration (for Hermes) or unit/mock tests (for modules). 
- [ ] Linked to GitHub issue.
- [ ] Updated code comments and documentation (e.g., `docs/`).

### Reviewer checklist:

- [ ] Reviewed `Files changed` in the GitHub PR explorer.
- [ ] Manually tested (in case integration/unit/mock tests are absent).

PR #2007, opened 2022-03-24
Fix slowness of relayer with 3+ chains and low average block time

romac:
Closes: https://github.com/informalsystems/ibc-rs/issues/2008
Closes: #1978 
______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


relayer/src/util/recv_multiple.rs
```diff
@@ -1,21 +1,49 @@
...
-    // Receive the message
-    let result = oper.recv(r).ok()?;
+        // Receive the message
+        let result = oper.recv(r).ok()?;
+
+        Some((k, result))
+    }
 
-    Some((k, result))
+    pub fn try_recv_multiple(&mut self) -> Option<(&K, T)> {
```
adizere, line 36: Should we still keep this?

romac: No I guess we can remove it.

ancazamfir:
Should we try also with https://github.com/informalsystems/ibc-rs/pull/1977?

adizere (approved):
> > Should we try also with #1977?
> 
> @soareschen is on it :)


Great find Romain, amazing work team! Approving preemptively here.

soareschen:
I have verified that #1977 has been fixed with this PR.

I have also put a fix to correct the use of `spawn_background_task`. In short, the step runner given to `spawn_background_task` is not supposed to block forever. This is because when the `TaskHandle` is dropped, we want to send the shutdown signal to the task runner for it to exit the loop and terminate the task cleanly. If the step runner is blocking forever, then the shutdown signal is never handled, and so the thread join handle never return and cause a deadlock.

Issue #2013, opened 2022-03-25
Incomplete object-capability system 


hu55a1n1:
## Summary of Bug
The object-cap system in the IBC module is incomplete and needs improvement. The idea is to have the handlers receive the capabilities that are obtained using the `lookup_module_by_{port,channel}()` methods. 

## Version
master 

## Acceptance Criteria
- [ ] Handlers must be able to authenticate obj-caps.
- [ ] Make it easier for users to get type-safe obj-caps from the capability system. without having to deal with capability identifiers.  
- [ ] Update mock tests
____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

hu55a1n1:
Here's a minimal impl that leverages Rust's type system and visibility. The two most important properties of object capabilities are non-forgeability and ability to uniquely associate/identify a resource. ->
```rust
pub mod ibc {
    // Use atomics for static mut cap index
    use core::sync::atomic::AtomicUsize;
    static CAP_INDEX: AtomicUsize = AtomicUsize::new(0);

    // Define capability trait that cannot be implemented by types outside this module
    pub trait Capability: private::Sealed {
        fn index(&self) -> usize;
    }

    mod private {
        pub trait Sealed {}
    }

    // blanket impls so we don't have to impl Capability for ref types
    impl<T: Capability> Capability for &'_ T {
        fn index(&self) -> usize {
            (**self).index()
        }
    }

    impl<T: Capability> private::Sealed for &'_ T {}

    // A resource that the capability provides access to. eg. Channel/Port path
	// This must be serializable to a string so it can be persisted on storage.
    pub trait Resource: ToString {}

    #[derive(Debug)]
    pub struct Error;

    pub trait CapabilityKeeper {
        /// Store a new capability with the given name.
        /// Return an error if the capability was already taken.
        fn new_capability<C: Capability>(
            &mut self,
            _name: impl Resource,
            capability: Option<C>,
        ) -> Result<C, Error> {
            Ok(capability.unwrap())
        }

        /// Claim the specified capability using the specified name.
        /// Return an error if the capability was already taken.
        fn claim_capability(&mut self, _name: impl Resource, _capability: &impl Capability) {}

        /// Release a previously claimed or created capability
[...]

PR #2018, opened 2022-03-27
Fix obj-caps facility

hu55a1n1:
Closes: #2013

## Description
______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/modules/src/core/ics04_channel/context.rs
+++ b/modules/src/core/ics04_channel/context.rs
@@ -12,7 +12,7 @@ use crate::core::ics04_channel::handler::recv_packet::RecvPacketResult;
 use crate::core::ics04_channel::handler::{ChannelIdState, ChannelResult};
 use crate::core::ics04_channel::{error::Error, packet::Receipt};
 use crate::core::ics05_port::capabilities::{CapabilityName, ChannelCapability};
-use crate::core::ics05_port::context::CapabilityReader;
+use crate::core::ics05_port::context::{CapabilityKeeper, CapabilityReader};
 use crate::core::ics24_host::identifier::{ChannelId, ClientId, ConnectionId, PortId};
 use crate::core::ics24_host::path::ChannelCapabilityPath;
 use crate::core::ics26_routing::context::ModuleId;
@@ -41,8 +41,6 @@ pub trait ChannelReader: CapabilityReader {
         client_id: &ClientId,
         height: Height,
     ) -> Result<AnyConsensusState, Error>;
-
-    fn authenticated_capability(&self, port_id: &PortId) -> Result<ChannelCapability, Error>;
 
     fn get_next_sequence_send(
         &self,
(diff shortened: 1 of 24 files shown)


modules/src/core/ics05_port/context.rs
```diff
@@ -1,70 +1,114 @@
...
-    fn release_capability(&mut self, name: CapabilityName, capability: Capability);
+    fn release_capability(
+        &mut self,
+        name: CapabilityName,
+        capability: Self::Capability,
+    ) -> Result<(), Error>;
 }
 
-pub trait CapabilityReader {
+pub trait CapabilityReader<M: Into<ModuleId>> {
```
hu55a1n1, line 84: This is the main change in this PR. `CapabilityReader` and `CapabilityKeeper` are now generic over `ModuleId`. This follows from the intuition that they are always scoped to a module and the library never makes calls to the underlying host system's (unscoped) `CapabilityReader`. 
I experimented with two other designs - one with a composition style approach (see https://github.com/informalsystems/ibc-rs/pull/2018/commits/ec20b5afcb967e181bb3da7d7e65c34bc07774f2) and another with a freestanding `CapabilityReader` with a blanket scoped implementation, but both those designs proved to be unwieldy. The benefit of this approach is that it makes the least assumptions on the host's capability system and allows users to write a blanket implementation for all modules.

modules/src/core/ics05_port/context.rs
```diff
@@ -1,70 +1,114 @@
...
+        &self,
+        port_id: PortId,
+        capability: &Self::Capability,
+    ) -> Result<(), Error> {
+        self.authenticate_capability(&port_capability_name(port_id), capability)
     }
 }
 
-pub trait PortKeeper: CapabilityKeeper + PortReader {
+pub trait PortCapabilityKeeper<M: Into<ModuleId>>:
```
hu55a1n1, line 40: `PortCapabilityKeeper` and `ChannelCapabilityKeeper` have been extracted out of `Port/ChannelKeeper`. These traits only have provided methods are provided for convenience.

modules/src/core/ics24_host/path.rs
```diff
@@ -37,6 +37,7 @@ pub enum Path {
     Connections(ConnectionsPath),
     Ports(PortsPath),
     ChannelEnds(ChannelEndsPath),
+    ChannelCapability(ChannelCapabilityPath),
```
hu55a1n1, line 40: This is a new `Path` that defines a key for the channel capability in the persistent store.

modules/src/core/ics26_routing/context.rs
```diff
@@ -31,9 +44,12 @@ pub trait Ics26Context:
     + ConnectionKeeper
     + ChannelKeeper
     + ChannelReader
-    + PortReader
+    + ChannelCapabilityKeeper<CoreModuleId, Capability = <Self as Ics26Context>::Capability>
```
hu55a1n1, line 47: All `Capability` related keepers that the `Ics26Context` depends on, are scoped to the `CoreModule` and all associated `Capability` types are bound to be the same.

modules/src/core/ics05_port/capabilities.rs
```diff
@@ -2,80 +2,10 @@
...
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
-impl From<u64> for Capability {
-    fn from(index: u64) -> Self {
-        Self { index }
-    }
+pub trait Capability {
```
hu55a1n1, line 7: `Capability` is a trait now so the host system is free to define it's own non-forgeable and secure `Capability` type.

modules/src/core/ics04_channel/handler/acknowledgement.rs
```diff
@@ -19,10 +22,15 @@ pub struct AckPacketResult {
     pub seq_number: Option<Sequence>,
 }
 
-pub fn process(
-    ctx: &dyn ChannelReader,
+pub(crate) fn process<Ctx, Cap>(
```
hu55a1n1, line 25: All ICS04 handlers are now generic over `Capability` and authenticate the provided capability before continuing with actual handler logic.

modules/src/core/ics04_channel/context.rs
```diff
@@ -280,6 +272,71 @@ pub trait ChannelKeeper {
...
+        &self,
+        port_id: PortId,
+        channel_id: ChannelId,
+        capability: &Self::Capability,
+    ) -> Result<(), Error> {
+        self.authenticate_capability(&channel_capability_name(port_id, channel_id), capability)
+            .map_err(Error::ics05_port)
+    }
+
+    fn create_channel_capability(
```
hu55a1n1, line 309: This is an important deviation from the SDK's capability design. We require implementations to be able to create capabilities without mutating the state. The capability returned from this call used then stored to the state by calling `ChannelCapabilityKeeper::store_channel_capability()`. This is required because our handlers do not mutate state, but still need to be able to create capabilities.

modules/src/core/ics04_channel/handler/chan_open_try.rs
```diff
@@ -125,14 +136,18 @@ pub(crate) fn process(
     // Transition the channel end to the new state & pick a version.
     new_channel_end.set_state(State::TryOpen);
 
+    let (channel_id_state, channel_cap) = if msg.previous_channel_id.is_none() {
```
hu55a1n1, line 139: We only create a new capability only if a `previous_channel_id` isn't specified.

modules/src/mock/context.rs
```diff
@@ -124,19 +128,60 @@ pub struct MockContext {
...
 impl Default for MockContext {
     fn default() -> Self {
+        let ocap = Arc::new(Mutex::new(MockOCap::default()));
         Self::new(
             ChainId::new("mockgaia".to_string(), 0),
             HostType::Mock,
             5,
             Height::new(0, 5),
         )
+        .with_router(dummy_router(ocap.clone()))
```
hu55a1n1, line 148: The default implementation now comes with a `dummy_router()` and an ocap system where the default port is already owned by the `DummyModule`. This setup simplifies handler tests.

modules/src/mock/context.rs
```diff
@@ -124,19 +128,60 @@ pub struct MockContext {
...
+        .with_ocap(ocap)
+        .with_port_capability(PortId::default(), dummy_module_id())
+    }
+}
+
+impl Clone for MockContext {
```
hu55a1n1, line 154: We manually implement `Clone` because `ocap` is a pointer. This implementation sets up the ocap pointer to it's initial state as per the `Default` impl.

modules/src/mock/context.rs
```diff
@@ -365,7 +424,27 @@ impl MockContext {
         channel_end: ChannelEnd,
     ) -> Self {
         let mut channels = self.channels.clone();
-        channels.insert((port_id, chan_id), channel_end);
+        channels.insert((port_id.clone(), chan_id), channel_end);
+        let (module_id, _) = self
```
hu55a1n1, line 428: Adding a channel automatically creates a new associated capability and is claimed by the module that owns the `port` capability.

modules/src/mock/context.rs
```diff
@@ -124,19 +128,60 @@ pub struct MockContext {
 
     /// ICS26 router impl
     router: MockRouter,
+
+    /// Object capabilites impl
+    ocap: Arc<Mutex<MockOCap>>,
```
hu55a1n1, line 133: This must be shared across modules, hence the `Arc<Mutex<>>`

hu55a1n1 (commented):
Adding some comments to help describe the main changes in this PR and the rationale for some design choices.

modules/src/core/ics04_channel/handler/send_packet.rs
```diff
@@ -32,7 +32,8 @@ pub fn send_packet(ctx: &dyn ChannelReader, packet: Packet) -> HandlerResult<Pac
         return Err(Error::channel_closed(packet.source_channel));
     }
 
-    let _channel_cap = ctx.authenticated_capability(&packet.source_port)?;
+    // Fixme(hu55a1n1)
```
hu55a1n1, line 35: To fix this, we would need `Ics20Context` to have access to a scoped `CapabilityReader` and since there's already a PR for ICS20 impl, I updated the PR author checklist to include this -> https://github.com/informalsystems/ibc-rs/pull/1989.

romac:
@hu55a1n1 Shall we close this now that we've decided not go forward with the OCap model for the time being?

Issue #1970, opened 2022-03-16
gRPC connection query at latest blockchain height may fail


ancazamfir:
## Summary of Bug
@mircea-c reported this problem:
Hi team! I'm running into a strange issue. Trying to run Hermes between gravity bridge and cosmoshub and Hermes v0.12.0 shows:
```
2022-03-03T19:40:34.805800Z ERROR ThreadId(01) skipped workers for connection connection-20 on chain gravity-bridge-3, reason: relayer error: connection not found: connection-464
2022-03-03T19:40:34.805834Z  INFO ThreadId(01) channel channel-17 on chain gravity-bridge-3 is: OPEN; state on dest. chain (cosmoshub-4) is: UNINITIALIZED
```
When I run it manually I get the same issue, but if I run it again immediately after then it seems to work:

```
mircea@sysmon:/workbench/cephalopodequipment/config/hermes$ /workbench/current/hermes/v0.12.0/hermes -c manual.toml query channel ends gravity-bridge-3 transfer channel-17
2022-03-03T19:40:15.461529Z DEBUG ThreadId(01) Options: QueryChannelEndsCmd { chain_id: ChainId { id: "gravity-bridge-3", version: 3 }, port_id: PortId("transfer"), channel_id: ChannelId("channel-17"), height: None, verbose: false }
Error: connection not found: connection-464
mircea@sysmon:/workbench/cephalopodequipment/config/hermes$ /workbench/current/hermes/v0.12.0/hermes -c manual.toml query channel ends gravity-bridge-3 transfer channel-17
2022-03-03T19:40:17.417384Z DEBUG ThreadId(01) Options: QueryChannelEndsCmd { chain_id: ChainId { id: "gravity-bridge-3", version: 3 }, port_id: PortId("transfer"), channel_id: ChannelId("channel-17"), height: None, verbose: false }
Success: ChannelEndsSummary {
    chain_id: ChainId {
        id: "gravity-bridge-3",
        version: 3,
    },
    client_id: ClientId(
        "07-tendermint-21",
    ),
    connection_id: ConnectionId(
        "connection-20",
    ),
    channel_id: ChannelId(
        "channel-17",
    ),
    port_id: PortId(
        "transfer",
    ),
    counterparty_chain_id: ChainId {
        id: "cosmoshub-4",
        version: 4,
    },
    counterparty_client_id: ClientId(
        "07-tendermint-582",
    ),
    counterparty_connection_id: ConnectionId(
[...]

PR #2021, opened 2022-03-28
Anca/app latest height

ancazamfir:
Closes: #1970

## Description

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


relayer/src/chain/cosmos.rs
```diff
@@ -1231,17 +1231,35 @@ impl ChainEndpoint for CosmosSdkChain {
...
+            // yes, use the chain latest block time
+            status.sync_info.latest_block_time
+        } else {
+            // no, retrieve the time of the header at application latest height
+            self.block_on(self.rpc_client.commit(abci_status.last_block_height))
+                .map_err(|e| Error::rpc(self.config.rpc_addr.clone(), e))?
+                .signed_header
+                .header
+                .time
+        };
```
ancazamfir, line 1258: This is the main change for this PR. It requires 3 RPCs instead of one. And this happens for handling of every IBC send packet event with impact on relayer performance and extra churn on the full node. So I am a bit hesitant to move this forward until performance tests are done and/or get all info in one RPC.
I opened https://github.com/tendermint/tendermint/issues/8248 to get input from tendermint team.

romac: I am also wary of increasing the pressure on the node, especially for every `SendPacket` event.

Are we seeing this only when issuing queries manually or is it also happening during relaying? If only the former, I would vote for waiting until we can get away with a single RPC call or until something [like this](https://github.com/tendermint/tendermint/issues/8248#issuecomment-1087480473) is implemented.

ancazamfir: This is seen during relaying and can cause failures for any query. In one extreme case reported by Mircea (see issue) hermes was not relaying at all on a channel as it was not able to spawn workers due to `connection not found`.
This was fixed with a partial fix, similar to the one in this PR but without the `time` fix (i.e. we would only fix the height to be the application height). When I did some tests with custom chains (with increased block execution times) the lack of `time` fix would sometime cause `header in the future` chain error for client updates.

adizere:
I did some basic benchmarking with two RPC nodes:
1. ecostake: https://rpc-cosmoshub.ecostake.com/ 
2. pupmos: https://rpc.cosmoshub.pupmos.network

Both found from here: https://chain-registry.netlify.app

I looked at these 5 endpoints:
1. /status
2. /abci_info
4. /commit
5. /header?height=X
6. /blockchain?minHeight=X&maxHeight=X

These numbers are preliminary, and may not reflect other operator's endpoints performance.

Basic takeaway: If we can, we should avoid the `/commit` endpoint, which can be ~30% slower than the others. `/abci_info` can also be problematic.

<details><summary>Data</summary>

The numbers are in milliseconds.

|          | /status | /abci_info | /commit | /header | /blockchain |
|----------|---------|------------|---------|---------|-------------|
| pupmos   | 108     | 105        | 134     | 101     | 106         |
| ecostake | 107     | 152        | 143     | 98      | 99          |

</details>

ancazamfir:
> I did some basic benchmarking with two RPC nodes:

Thanks @adizere for this! For some reason I don't see the `header` RPC. What are the exact commands you used? And with which tendermint version?

adizere:
> For some reason I don't see the `header` RPC. What is the exact command you used for this?

Good point. The `header` endpoint is still a draft PR in tendermint-rs: https://github.com/informalsystems/tendermint-rs/pull/1101.
The data I extracted above was by querying via `time curl http:$RPC/$ENDPOINT`

relayer/src/chain/cosmos.rs
```diff
@@ -638,16 +638,44 @@ impl ChainEndpoint for CosmosSdkChain {
...
+            .map_err(|e| Error::rpc(self.config.rpc_addr.clone(), e))?;
+
+        // Query `/blockchain` endpoint to pull the block metadata corresponding to
+        // the latest block that the application committed.
+        // TODO: Replace this query with `/header`, once it's available.
+        //  https://github.com/informalsystems/tendermint-rs/pull/1101
+        let blocks = self
+            .block_on(
+                self.rpc_client
+                    .blockchain(abci_info.last_block_height, abci_info.last_block_height),
```
adizere, line 660: Note to reviewers:
- we query `/blockchain?minHeight=last_block_height&maxHeight=last_block_height` to fetch the metadata for a single block

relayer/src/chain/cosmos.rs
```diff
@@ -1516,13 +1544,20 @@ fn do_health_check(chain: &CosmosSdkChain) -> Result<(), Error> {
...
 
-    let status = chain.status()?;
+    let status = chain.chain_status()?;
 
     // Check that transaction indexing is enabled
     if status.node_info.other.tx_index != TxIndexStatus::On {
         return Err(Error::tx_indexing_disabled(chain_id.clone()));
     }
 
+    // Check that the chain identifier matches the network name
```
adizere, line 1554: Note: This query has not much to do with the present PR. It's just an additional non-critical health check that can reveal misconfigurations of config.toml.

adizere (commented):
We need to reproduce the problem reliably. I haven't yet checked Anca's instructions from [here](https://github.com/informalsystems/ibc-rs/issues/1970#issuecomment-1069143962), but that's probably the best place to start from.

romac:
I believe I am still seeing the same issue, when running the script against Gaia v6.0.3 using [a patched Tendermint](https://github.com/informalsystems/ibc-rs/issues/1970#issuecomment-1069143962):


```
Query Chain ID: ibc-0, Conn ID: connection-0, Height: 62, ABCI_Height: 61
2022-04-27T07:44:42.335551Z DEBUG ThreadId(01) Options: QueryConnectionEndCmd { chain_id: ChainId { id: "ibc-0", version: 0 }, connection_id: ConnectionId("connection-0"), height: Some(62) }
Error:
   0: connection not found: connection-0



Query Chain ID: ibc-0, Conn ID: connection-0, Height: 62, ABCI_Height: 62
2022-04-27T07:44:45.386849Z DEBUG ThreadId(01) Options: QueryConnectionEndCmd { chain_id: ChainId { id: "ibc-0", version: 0 }, connection_id: ConnectionId("connection-0"), height: Some(62) }
Success: ConnectionEnd {
    state: Open,
    client_id: ClientId(
        "07-tendermint-0",
    ),
    counterparty: Counterparty {
        client_id: ClientId(
            "07-tendermint-0",
        ),
        connection_id: Some(
            ConnectionId(
                "connection-0",
            ),
        ),
        prefix: ibc,
    },
    versions: [
        Version {
            identifier: "1",
            features: [
                "ORDER_ORDERED",
                "ORDER_UNORDERED",
            ],
        },
    ],
    delay_period: 0ns,
}
```

romac:
Detailed instructions for reproducing the issue and results:

1. cd /tmp
1. mkdir gaia-test
1. cd gaia-test
1. git clone https://github.com/tendermint/tendermint
1. cd tendermint
1. git switch v0.34.14
1. apply

    ```
    diff --git a/state/execution.go b/state/execution.go
    index 2cc3db6e7..16f150702 100644
    --- a/state/execution.go
    +++ b/state/execution.go
    @@ -323,6 +323,8 @@ func execBlockOnProxyApp(
            }
        }

    +  time.Sleep(time.Millisecond * 500)
    +
        // End block.
        abciResponses.EndBlock, err = proxyAppConn.EndBlockSync(abci.RequestEndBlock{Height: block.Height})
        if err != nil {
    ```

1. cd ..
1. git clone https://github.com/cosmos/gaia
1. cd gaia
1. apply

    ```diff
    diff --git a/go.mod b/go.mod
    index 7465a0be..e3a508c5 100644
    --- a/go.mod
    +++ b/go.mod
    @@ -124,4 +124,5 @@ replace (
        github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.44.2
        github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
        google.golang.org/grpc => google.golang.org/grpc v1.33.2
    +   github.com/tendermint/tendermint => /tmp/gaia-test/tendermint
     )
    ```

1. `make build`
1. `vim ~/.gm/gm.toml`

    ```toml
    [global]
    add_to_hermes        = false
    auto_maintain_config = true
    extra_wallets        = 0
    gaiad_binary         = "/private/tmp/gaia-test/gaia/build/gaiad"
    hdpath               = ""
    home_dir             = "$HOME/.gm"
    ports_start_at       = 27040

    [global.hermes]
    binary            = "$HOME/Informal/Code/Current/ibc-rs/target/debug/build/hermes"
    config            = "$HOME/.hermes/config.toml"
    log_level         = "info"
    telemetry_enabled = true
    telemetry_host    = "127.0.0.1"
    telemetry_port    = 3001

    [ibc-0]
    ports_start_at = 27000
    add_to_hermes  = true

    [ibc-1]
    ports_start_at = 27010
    add_to_hermes  = true
    ```

1. `gm reset`
1. `gm start`
1. `gm hermes config`
1. remove `filter = false` from `~/.hermes/config.toml`
1. Go to `ibc-rs` working directory
1. `git fetch origin`
1. `git switch anca/app_latest_height`
1. `git pull`
1. `cargo build`
1. `gm hermes keys`
1. `cargo run -- create channel ibc-0 -c ibc-1 --port-a transfer --port-b transfer --new-client-connection`
1. paste this in `query_conn.sh`

    ```shell
    #!/bin/bash

    addr=$1
    chain=$2
    conn=$3
    for i in {1..100}
    do
        read -r height < <(
            curl -s "$addr"/status | jq -r '.result.sync_info.latest_block_height')
        read -r abci_height < <(
            curl -s "$addr"/abci_info | jq -r '.result.response.last_block_height')
        echo "Query Chain ID: ${chain}, Conn ID: ${conn}, Height: ${height}, ABCI_Height: ${abci_height}"
        ./target/debug/hermes query connection end $chain $conn -H $height
        sleep 1
    done
    ```

1. `bash query_conn.sh 'http://localhost:27000' ibc-0 connection-0`

adizere:
Great instruction for testing Romain, thank you for sharing that!

I was able to see the same issue as you. E.g.:
>Query Chain ID: ibc-0, Conn ID: connection-0, Height: 101, ABCI_Height: 100
2022-04-27T09:15:28.911674Z  INFO ThreadId(01) using default configuration from '/Users/adi/.hermes/config.toml'
Error: connection not found: connection-0

---

After digging into it, I realize that `hermes query connection end` which we use in this `query_conn.sh` testing script does not invoke `query_application_status`, which is what this PR solves. So I think it would be expected that we encounter this problem when running `query_conn.sh` script.

To summarize, I think the purpose of `query_conn.sh` is to simulate the problem that Hermes encounters when querying with blockchain height (`-H $height`) instead of querying with application height (`-H $abci_height`). I don't understand how this problem relates to the original issue (#1970), however, because I don't see where exactly does Hermes rely on blockchain height in the supervisor or scan part as these logs show:

> 2022-03-03T19:40:34.805800Z ERROR ThreadId(01) skipped workers for connection connection-20 on chain gravity-bridge-3, reason: relayer error: connection not found: connection-464
2022-03-03T19:40:34.805834Z  INFO ThreadId(01) channel channel-17 on chain gravity-bridge-3 is: OPEN; state on dest. chain (cosmoshub-4) is: UNINITIALIZED


That being said, this PR only replaces the use of blockchain height with application height specifically for `query_application_status`. I'm not convinced this is enough to solve the error reported in the issue (#1970).

ancazamfir:
The script retrieves the blockchain height then queries the connection with that height. If the height is higher than the application query state height you see that error. I used this script because in the original issue the command used was `query channel ends` which queries the connection with the blockchain latest height.
Also in hermes we sometimes query the latest height and then we do subsequent queries and gather proofs with that fixed height. 
You can try to run tests  (e.g. e2e) from master with the repro setup and you will see quite a few failures.

romac:
> You can try to run tests (e.g. e2e) from master with the repro setup and you will see quite a few failures.

Alright so I've done just that a couple times, both against `master` and against this branch, and I did indeed see a bunch of errors with `master` when querying channel and connection ends but not a single one with this branch 🎉 

To me this is good to merge :)

- [Logs for latest run against `master`](https://gist.github.com/romac/e19b79cb0c7ce13f844ab9c598ee1390#file-e2e-master-txt)
- [Logs for latest run against this branch](https://gist.github.com/romac/e19b79cb0c7ce13f844ab9c598ee1390#file-e2e-pr-txt)

adizere (approved):
Thank you Anca for kicking off the fix and your guidance! Amazing work Romain with testing! LGTM 🚀

Issue #1998, opened 2022-03-22
Config max_block_time does not match chain's block time in integration tests


hu55a1n1:
## Summary
Currently, `max_block_time` defaults to `Duration::default()`, but since #1917 uses this parameter to calculate block delay, it is important that it matches the max block time of the chain we're testing against. 
https://github.com/informalsystems/ibc-rs/blob/master/tools/test-framework/src/types/single/node.rs#L143-L144

## Acceptance Criteria
- [x] ~~The default value matches gaiad's default (?)~~
- [X] ~~The value matches the block time for the chain being tested against.~~
- [X] ~~Addition of a failing integration test where max_block_time doesn't match the chain's block time.~~
      - moved to https://github.com/informalsystems/ibc-rs/pull/2032
- [x] Change default max_block_time to 30s

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #1445, opened 2021-10-11
Update client may cause "new header has a time from the future" chain error


ancazamfir:
## Crate
relayer

## Summary of Bug
@mircea-c has reported that the following errors are seen once in a while when updating the `sifchain` client (`07-tendermint-395`) on `cosmoshub`:
```
Oct 11 09:13:03.057 DEBUG ThreadId(42) [cosmoshub-4] simulated gas: Err(
    0: #033[91mGRPC call return error status status: InvalidArgument, message: "failed to execute message; message index: 0: cannot update client with ID 07-tendermint-395: failed to verify header: inval
id header: new header has a time from the future 2021-10-11 13:12:56.673817244 +0000 UTC (now: 2021-10-11 13:12:51.047099028 +0000 UTC; max clock drift: 5s): invalid request", details: [], metadata: Meta
dataMap { headers: {"content-type": "application/grpc"} }#033[0m
```
After looking at the sif chain configuration (client created with `clock_drift = 5sec`) and comparing block times (5-6 sec on sif, 7-13 sec on hub), the following seems to be the issue.
Here is an example (`-` is 1 sec, `b` are blocks on sif, `B` are blocks on the hub).
```
	   1:ev 2:clUpdate
           |    |  |
           |    | 3: fetch header (h)
           v    v  |
b----b-----b----b-----b----b      (sifchain 5/6 sec blocktimes)
                |  | 4: simulate/ send Tx
                |  v
B-------B------------B------B     (cosmoshub 7-13 sec)
        ^       |
        | 5: create context with latest header (H)
        |       |
        v       v
        <-------> 6: check that h.timestamp - drift < H.timestamp (aka now)
```
The steps are:
1 - an IBC event that requires a clientUpdate
2 - hermes starts to build the client update message and
3 - fetches the header `h` from sifchain (this happens for the header at event height + 1)
4 - sends a simulate request to the hub (can happen on send_tx as well)
5 - the chain receives the request and creates a context with the latest header `H` on the hub
6 - the client handler is called with that context and the message. Among other things it [checks](https://github.com/tendermint/tendermint/blob/ded310093e0d771c9ed27f296921cb6b23d99f29/light/verifier.go#L253-L256) that the `h.timestamp` is not "in the future", i.e. `H.timestamp + client_state.clock_drift > h.timestamp`

(one 13 sec example on the hub:
h:7969646 - ts:16:31:24
h:7969647 - ts:16:31:37
)

[...]

PR #2032, opened 2022-03-29
Change default max_block_time to 30s

hu55a1n1:
Closes: #1998
______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/bug-fixes/relayer/1998-default-max-block-time.md
+++ b/.changelog/unreleased/bug-fixes/relayer/1998-default-max-block-time.md
@@ -1,0 +1,2 @@
+- Adjusted max_block_time default value to 30s
+  ([#1998](https://github.com/informalsystems/ibc-rs/issues/1998))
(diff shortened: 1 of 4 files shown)


hu55a1n1:
@ancazamfir, I noticed that we're using the `max_block_delay` from the `ChainConfig` to also fix https://github.com/informalsystems/ibc-rs/issues/1445 (Update client may cause "new header has a time from the future" chain error). My understanding is that that value (and it's default) is the same as `maxExpectedTimePerBlock` which is defined in [ibc-go here](https://github.com/cosmos/ibc-go/blob/main/modules/core/03-connection/types/params.go#L11), which is also used in the connection block-delay calculation in hermes in #1917. Can you confirm if this is correct? And if it is okay to set the default to `30s`?

adizere:
AZ: We shouldn't keep `max_block_time` as a config param in Hermes. Instead, programmaticaly fetch it from the chain (genesis param) and cache it in the relayer.

hu55a1n1:
> AZ: We shouldn't keep max_block_time as a config param in Hermes. Instead, programmaticaly fetch it from the chain (genesis param) and cache it in the relayer.

I opened https://github.com/informalsystems/ibc-rs/issues/2037 to address this.

Issue #2040, opened 2022-03-31
Allow instantiating relayer with different wallets in integration test


soareschen:
## Summary

Make it simple to spawn relayer supervisor and `ChainHandle`s with different wallet provided in the config.

## Problem Definition

A key challenge with the current design of `ChainHandle` is that the wallet is configured pseudo-globally and cannot be changed dynamically or overridden when a transaction is submitted. This present challenges when writing integration tests, because if the same wallet is used concurrently in different operations, the account sequence mismatch errors will happen and cause the tests to potentially fail.

This presents problem if we want to spawn multiple relayers to run in parallel, as in the case of #2024. To do so, we need to have ways to instantiate multiple relayer configurations with different wallets, and specify which relayer or `ChainHandle` we want to choose from.

One complication for the design is that we cannot easily swap the wallets within a relayer's `Config` value. This is because the wallets are different for each chain, so we have to effectively change a whole set of wallets depending on how many chains are present in the config.

## Proposal

#2039 is a work in progress to figure out how we can solve this problem without too much boilerplate. It uses const generics with a hardcoded array of relayer wallets, so that tests can for example get 3 sets of relayers to be used.
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #2012, opened 2022-03-25
`ChainDriver::transfer_token()` does not perform subsequent IBC transfer when called multiple times


soareschen:
## Summary of Bug

Currently the `ChainDriver::transfer_token()` method calls the Gaia command `gaiad tx ibc-transfer transfer` to perform the IBC transfer. However when submitting the IBC transfer one right after another, only the first call of it would pass.

This might be a limitation of `gaiad` that we have to investigate. Otherwise, we would have to reimplement `ChainDriver::transfer_token()` to perform the IBC transfer using the `ChainHandle` method.

## Steps to Reproduce

#2011 modifies the IBC transfer test to perform multiple IBC transfers at once. The result is that the test fails unless `TRANSFER_COUNT` is set to `1`.

## Acceptance Criteria

Multiple calls to `ChainDriver::transfer_token()` should succeed in submitting IBC transfer.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

soareschen:
A temporary workaround in #2024 is to wait for 1 second before performing another `transfer_token()`. This is probably due to gaiad only supporting a single transaction to be submitted per wallet per block. So consecutive calls within the same block period will fail.

An alternative workaround is to poll the sender wallet balance before and after performing the transaction, and only return after the wallet balance has been reduced. This will require the least amount of work as compared to other kinds of refactoring.

Issue #375, opened 2020-11-06
Clippy practices


adizere:
## Crate

`ibc`

## Summary

We have inconsistent practices regarding the use of clippy exceptions, in particular to `too_many_arguments` and `large_enum_variant`. We should decide on a standard.

## Problem Definition

We sometimes annotate enums with `#[allow(clippy::large_enum_variant)]` to avoid clippy warnings:

https://github.com/informalsystems/ibc-rs/blob/12cb1d6187e878eee0e40295fe0b804ddef0b846/modules/src/ics02_client/msgs.rs#L13-L18

And sometimes we box large enum variants for the same reason (to avoid clippy warnings):

https://github.com/informalsystems/ibc-rs/blob/12cb1d6187e878eee0e40295fe0b804ddef0b846/modules/src/ics03_connection/msgs.rs#L29-L34

The `too_many_arguments` check also poses issues, for example this method has 11 arguments, but clippy issues warnings for anything larger than 7:

https://github.com/informalsystems/ibc-rs/blob/12cb1d6187e878eee0e40295fe0b804ddef0b846/modules/src/ics07_tendermint/client_state.rs#L34-L47

While we're at it, we should also consider doing a sweeping through the repo and clean other clippy exception, for example:

https://github.com/informalsystems/ibc-rs/blob/c32393a4f344877e34957d13a768e6623486c587/modules/src/tx_msg.rs#L15

## Improvement proposal:

- [ ] perhaps we can eliminate methods that have excessive number of arguments
- [ ] use a repo-wide clippy configuration with larger thresholds than the defaults

## Original discussion

I wonder if we can have a clippy [config to parametrize](https://rust-lang.github.io/rust-clippy/master/index.html#too_many_arguments) the "too_many_arguments" threshold (default is 7) for this crate, since we're using this _allow_ quite often. We are also cutting some corners with allowing `large_enum_variant`.

_Originally posted by @adizere in https://github.com/informalsystems/ibc-rs/pull/355#discussion_r516654739_

In future, I think this deserves a discussion about clippy in general, if we want to have a repo-wide [configuration](https://github.com/rust-lang/rust-clippy#configuration) where we adjust the clippy parameters ... so we can remove the allow/clippy annotations altogether. I can open an issue to track this if that seems interesting.

_Originally posted by @adizere in https://github.com/informalsystems/ibc-rs/pull/355#discussion_r516832986_

____

#### For Admin Use

- [X] Not duplicate issue
[...]

PR #2044, opened 2022-04-01
Refactor `send_tx` methods into plain functions

### LINKED ISSUE #1975: Refactor `send_tx`-related methods into plain functions
## Summary

The method `CosmosSdkChain::send_tx()` method is tightly coupled with the `CosmosSdkChain` struct, with deeply nested method calls. This makes it challenging to use the functionality without spawning the entire relayer. This is because the construction of `CosmosSdkChain` is not straightforward, and involves a lot of indirection through other structures such as `ChainHandle` and `ChainRuntime`.

This complication is the main reason why `ibc-test-framework` currently performs token transfer [by calling the `gaiad` command](https://github.com/informalsystems/ibc-rs/blob/ac9cf078523aed08b442251423f27a7232bd3964/tools/test-framework/src/chain/driver/transfer.rs#L53), because the current design of `send_tx()` does not allow such simple functionality without a `ChainHandle`. However since `ChainHandle` is configured with a global wallet that cannot be changed, this presents challenges of reusing the same `ChainHandle` to perform transfer for different wallets.

Moving forward, if we want to improve the relayer performance by parallelizing the query operations and send_tx operations, it is also necessary to do this refactoring so that they do not end up in the same struct `CosmosSdkChain`.

____

#### For Admin Use

- [ ] Not duplicate issue

soareschen:
Closes #1975.

## Description

This PR attempts to rework the core logic of `send_tx` methods in `CosmosSdkChain` into plain functions. This is essential in helping us to solve the root problems in issues such as #2012, #2040, and supporting sending of transactions with fees in ICS29.

At this stage, I have managed to refactor `send_tx_with_account_sequence_retry` and its inner calls into plain functions, and simplify the data access a little.

Since this is a refactoring PR, no logic should be changed. The integration tests should help to ensure that nothing breaks with the refactoring.

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


relayer/src/chain/cosmos/types/tx.rs
```diff
@@ -0,0 +1,18 @@
+use ibc::events::IbcEvent;
+use ibc_proto::cosmos::tx::v1beta1::{AuthInfo, TxBody};
+use tendermint_rpc::endpoint::broadcast::tx_sync::Response;
+
+pub struct SignedTx {
```
soareschen, line 5: This is a common struct for both `Tx` and `TxRaw` so that there is no need to have different versions of functions returning slightly different results.

relayer/src/chain/cosmos/estimate.rs
```diff
@@ -0,0 +1,147 @@
...
+        auth_info: Some(signed_tx.auth_info),
+        signatures: signed_tx.signatures,
+    };
+
+    let estimated_fee = estimate_fee_with_tx(&gas_config, grpc_address, &config.id, tx).await?;
+
+    Ok(estimated_fee)
+}
+
+async fn estimate_fee_with_tx(
```
soareschen, line 52: This is refactored from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/chain/cosmos.rs#L332-L348

relayer/src/chain/cosmos/estimate.rs
```diff
@@ -0,0 +1,147 @@
...
+use crate::chain::cosmos::gas::{gas_amount_to_fees, PrettyFee};
+use crate::chain::cosmos::simulate::send_tx_simulate;
+use crate::chain::cosmos::types::account::Account;
+use crate::chain::cosmos::types::gas::GasConfig;
+use crate::config::types::Memo;
+use crate::config::ChainConfig;
+use crate::error::Error;
+use crate::keyring::KeyEntry;
+
+pub async fn estimate_tx_fees(
```
soareschen, line 17: This is refactored from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/chain/cosmos.rs#L309-L355

relayer/src/chain/cosmos/batch.rs
```diff
@@ -0,0 +1,174 @@
...
+        )
+        .await?;
+
+        responses.push(response);
+    }
+
+    Ok(responses)
+}
+
+async fn send_messages_as_batches(
```
soareschen, line 94: This extracts the code from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/chain/cosmos.rs#L1095-L1100

relayer/src/chain/cosmos/encode.rs
```diff
@@ -0,0 +1,179 @@
...
+
+use crate::chain::cosmos::types::account::{Account, AccountNumber, AccountSequence};
+use crate::chain::cosmos::types::tx::SignedTx;
+use crate::config::types::Memo;
+use crate::config::AddressType;
+use crate::config::ChainConfig;
+use crate::error::Error;
+use crate::keyring::{sign_message, KeyEntry};
+
+pub fn sign_and_encode_tx(
```
soareschen, line 17: This is refactored from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/chain/cosmos.rs#L357-L369

relayer/src/chain/cosmos/encode.rs
```diff
@@ -0,0 +1,179 @@
...
+    // A protobuf serialization of a SignDoc
+    let mut signdoc_buf = Vec::new();
+    prost::Message::encode(&sign_doc, &mut signdoc_buf).unwrap();
+
+    let signed = sign_message(key, signdoc_buf, address_type).map_err(Error::key_base)?;
+
+    Ok(signed)
+}
+
+fn encode_signer_info(
```
soareschen, line 103: This is refactored from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/chain/cosmos.rs#L744-L762

relayer/src/chain/cosmos/encode.rs
```diff
@@ -0,0 +1,179 @@
...
+    Ok(SignedTx {
+        body,
+        body_bytes,
+        auth_info,
+        auth_info_bytes,
+        signatures: vec![signed_doc],
+    })
+}
+
+fn encode_key_bytes(key: &KeyEntry) -> Result<Vec<u8>, Error> {
```
soareschen, line 70: This is refactored from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/chain/cosmos.rs#L690-L695

relayer/src/chain/cosmos/encode.rs
```diff
@@ -0,0 +1,179 @@
...
+    let mut pk_buf = Vec::new();
+
+    prost::Message::encode(&key.public_key.public_key.to_bytes(), &mut pk_buf)
+        .map_err(|e| Error::protobuf_encode("PublicKey".into(), e))?;
+
+    Ok(pk_buf)
+}
+
+fn encode_sign_doc(
```
soareschen, line 79: This is refactored from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/chain/cosmos.rs#L785-L790

relayer/src/keyring.rs
```diff
@@ -389,6 +371,33 @@ impl KeyRing {
     }
 }
 
+/// Sign a message
+pub fn sign_message(
+    key: &KeyEntry,
+    msg: Vec<u8>,
+    address_type: &AddressType,
+) -> Result<Vec<u8>, Error> {
```
soareschen, line 379: This is refactored from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/keyring.rs#L355-L360

relayer/src/chain/cosmos/wait.rs
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+}
+
+async fn update_tx_sync_result(
+    chain_id: &ChainId,
+    rpc_client: &HttpClient,
+    rpc_address: &Url,
+    tx_sync_result: &mut TxSyncResult,
+) -> Result<(), Error> {
```
soareschen, line 71: This is refactored from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/chain/cosmos.rs#L855-L878

relayer/src/chain/cosmos/types/gas.rs
```diff
@@ -0,0 +1,71 @@
...
+    config.max_gas.unwrap_or(DEFAULT_MAX_GAS)
+}
+
+fn gas_adjustment_from_config(config: &ChainConfig) -> f64 {
+    config
+        .gas_adjustment
+        .unwrap_or(DEFAULT_GAS_PRICE_ADJUSTMENT)
+}
+
+fn fee_granter_from_config(config: &ChainConfig) -> String {
```
soareschen, line 52: This is refactored from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/chain/cosmos.rs#L545-L550

relayer/src/chain/cosmos/types/gas.rs
```diff
@@ -0,0 +1,71 @@
...
+    config
+        .default_gas
+        .unwrap_or_else(|| max_gas_from_config(config))
+}
+
+pub fn max_gas_from_config(config: &ChainConfig) -> u64 {
+    config.max_gas.unwrap_or(DEFAULT_MAX_GAS)
+}
+
+fn gas_adjustment_from_config(config: &ChainConfig) -> f64 {
```
soareschen, line 46: This is refactored from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/chain/cosmos.rs#L558-L562

relayer/src/chain/cosmos/types/gas.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+}
+
+pub fn default_gas_from_config(config: &ChainConfig) -> u64 {
+    config
+        .default_gas
+        .unwrap_or_else(|| max_gas_from_config(config))
+}
+
+pub fn max_gas_from_config(config: &ChainConfig) -> u64 {
```
soareschen, line 42: This is refactored from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/chain/cosmos.rs#L540-L542

relayer/src/chain/cosmos/types/gas.rs
```diff
@@ -0,0 +1,71 @@
...
+            max_gas: max_gas_from_config(config),
+            gas_adjustment: gas_adjustment_from_config(config),
+            gas_price: config.gas_price.clone(),
+            max_fee: max_fee_from_config(config),
+            fee_granter: fee_granter_from_config(config),
+        }
+    }
+}
+
+pub fn default_gas_from_config(config: &ChainConfig) -> u64 {
```
soareschen, line 36: This is refactored from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/chain/cosmos.rs#L535-L536

relayer/src/chain/cosmos/types/gas.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub default_gas: u64,
+    pub max_gas: u64,
+    pub gas_adjustment: f64,
+    pub gas_price: GasPrice,
+    pub max_fee: Fee,
+    pub fee_granter: String,
+}
+
+impl GasConfig {
+    pub fn from_chain_config(config: &ChainConfig) -> GasConfig {
```
soareschen, line 24: This is newly added to pre-compute all gas parameters so that we don't need to re-compute it over and over again.

relayer/src/chain/cosmos/types/account.rs
```diff
@@ -0,0 +1,70 @@
+use core::fmt;
+use ibc_proto::cosmos::auth::v1beta1::BaseAccount;
+
+/// Wrapper for account number and sequence number.
+///
+/// More fields may be added later.
+#[derive(Clone, Debug, PartialEq)]
+pub struct Account {
```
soareschen, line 8: This file is moved from `chain::cosmos::acccount` to `chain::cosmos::types::acccount`.

relayer/src/chain/cosmos/tx.rs
```diff
@@ -0,0 +1,68 @@
...
+
+async fn send_tx_with_fee(
+    config: &ChainConfig,
+    rpc_client: &HttpClient,
+    key_entry: &KeyEntry,
+    account: &Account,
+    tx_memo: &Memo,
+    messages: Vec<Any>,
+    fee: &Fee,
+) -> Result<Response, Error> {
```
soareschen, line 48: This is refactored from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/chain/cosmos.rs#L357-L375

relayer/src/chain/cosmos/tx.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    Ok(response)
+}
+
+/// Perform a `broadcast_tx_sync`, and return the corresponding deserialized response data.
+async fn broadcast_tx_sync(
+    rpc_client: &HttpClient,
+    rpc_address: &Url,
+    data: Vec<u8>,
+) -> Result<Response, Error> {
```
soareschen, line 61: This is moved from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/chain/cosmos.rs#L2337-L2348

relayer/src/chain/cosmos/tx.rs
```diff
@@ -0,0 +1,68 @@
...
+
+pub async fn estimate_fee_and_send_tx(
+    config: &ChainConfig,
+    rpc_client: &HttpClient,
+    grpc_address: &Uri,
+    key_entry: &KeyEntry,
+    account: &Account,
+    tx_memo: &Memo,
+    messages: Vec<Any>,
+) -> Result<Response, Error> {
```
soareschen, line 23: This is refactored from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/chain/cosmos.rs#L304-L308

relayer/src/chain/cosmos/gas.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    let a = BigInt::from(a);
+    let f = BigRational::from_float(f).expect("f is finite");
+    (f * a).ceil().to_integer()
+}
+
+/// Adjusts the fee based on the configured `gas_adjustment` to prevent out of gas errors.
+/// The actual gas cost, when a transaction is executed, may be slightly higher than the
+/// one returned by the simulation.
+fn adjust_gas_with_simulated_fees(config: &GasConfig, gas_amount: u64) -> u64 {
```
soareschen, line 47: This is refactored from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/chain/cosmos.rs#L567-L577

relayer/src/transfer.rs
```diff
@@ -71,6 +75,57 @@ impl FromStr for Amount {
...
+    Timestamp(Timestamp),
+    HeightAndTimestamp(Height, Timestamp),
+}
+
+impl TransferTimeout {
+    pub fn new(
+        timeout_height_offset: u64,
+        timeout: Duration,
+        destination_chain_status: &ChainStatus,
+    ) -> Result<Self, TransferError> {
```
soareschen, line 90: This is refactored from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/transfer.rs#L98-L111

relayer/src/transfer.rs
```diff
@@ -71,6 +75,57 @@ impl FromStr for Amount {
...
+        let offset_is_zero = timeout_height_offset == 0;
+        let timeout_is_zero = timeout == Duration::ZERO;
+
+        let timeout_height = destination_chain_status.height.add(timeout_height_offset);
+
+        let timeout_timestamp = (destination_chain_status.timestamp + timeout)
```
soareschen, line 96: Here I updated the timeout timestamp to be computed from the destination chain's timestamp rather than the relayer's local time.

I believe I haven't mistaken the ordering, and it should be the destination chain, not the source chain's timestamp, correct?

relayer/src/chain/cosmos/query/status.rs
```diff
@@ -0,0 +1,36 @@
...
+use tendermint_rpc::{Client, HttpClient, Url};
+
+use crate::chain::ChainStatus;
+use crate::error::Error;
+
+pub async fn query_status(
+    chain_id: &ChainId,
+    rpc_client: &HttpClient,
+    rpc_address: &Url,
+) -> Result<ChainStatus, Error> {
```
soareschen, line 12: This is refactored from:

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/chain/cosmos.rs#L907-L916

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer/src/transfer.rs#L98-L111

relayer/src/chain/cosmos/batch.rs
```diff
@@ -0,0 +1,174 @@
...
+            events: events_per_tx,
+        };
+
+        tx_sync_results.push(tx_sync_result);
+    }
+
+    Ok(tx_sync_results)
+}
+
+fn batch_messages(config: &ChainConfig, messages: Vec<Any>) -> Result<Vec<Vec<Any>>, Error> {
```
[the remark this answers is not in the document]
romac: Much cleaner :)

relayer/src/chain/cosmos/batch.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    if !current_batch.is_empty() {
+        batches.push(current_batch);
+    }
+
+    Ok(batches)
+}
+
+fn message_size(message: &Any) -> Result<usize, Error> {
```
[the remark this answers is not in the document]
romac: While we're at it, I think we should be able to use `prost::Message::encoded_len` here instead of actually performing the serialization: https://docs.rs/prost/latest/prost/trait.Message.html#tymethod.encoded_len

relayer/src/chain/cosmos/estimate.rs
romac: Much nicer!

relayer/src/chain/cosmos/gas.rs
romac: Much better name, and thanks for the doc comments here and everywhere else!

relayer/src/chain/cosmos/query.rs
```diff
@@ -0,0 +1,149 @@
...
+use crate::chain::cosmos::version::Specs;
+use crate::chain::QueryResponse;
+use crate::error::Error;
+
+pub mod account;
+pub mod status;
+pub mod tx;
+
+pub fn packet_query(request: &QueryPacketEventDataRequest, seq: Sequence) -> Query {
+    tendermint_rpc::query::Query::eq(
```
romac, line 24:
```suggestion
    Query::eq(
```

relayer/src/chain/cosmos/query/tx.rs
```diff
@@ -0,0 +1,250 @@
...
+use ibc::query::QueryTxRequest;
+use ibc::Height as ICSHeight;
+use tendermint::abci::Event;
+use tendermint_rpc::endpoint::tx::Response as ResultTx;
+use tendermint_rpc::{Client, HttpClient, Order, Url};
+
+use crate::chain::cosmos::query::{header_query, packet_query, tx_hash_query};
+use crate::error::Error;
+
+/// This function queries transactions for events matching certain criteria.
```
romac, line 17: Great work with the documentation, here and everywhere else :)

relayer/src/chain/cosmos/retry.rs
```diff
@@ -0,0 +1,186 @@
...
+fn do_send_tx_with_account_sequence_retry<'a>(
+    config: &'a ChainConfig,
+    rpc_client: &'a HttpClient,
+    grpc_address: &'a Uri,
+    key_entry: &'a KeyEntry,
+    account: &'a mut Account,
+    tx_memo: &'a Memo,
+    messages: Vec<Any>,
+    retry_counter: u64,
+) -> Pin<Box<dyn Future<Output = Result<Response, Error>> + 'a>> {
```
romac, line 81: Do we need this explicit result type instead of an `async fn` because the compiler cannot infer the proper lifetime bound?

relayer/src/chain/cosmos/types/gas.rs
romac: Great idea!

relayer/src/transfer.rs
romac: cc @hu55a1n1 can you take a look at this, please?

tools/test-framework/src/error.rs
```diff
@@ -45,8 +45,8 @@ define_error! {
             [ ConnectionError ]
             | _ | { "connection error"},
 
-        Packet
-            [ PacketError ]
+        Transfer
+            [ TransferError ]
             | _ | { "packet error"},
```
romac, line 50:
```suggestion
            | _ | { "transfer error"},
```

relayer/src/chain/cosmos/batch.rs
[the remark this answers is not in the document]
soareschen: Great to know, fixed!

relayer-cli/src/error.rs
```diff
@@ -69,8 +69,8 @@ define_error! {
             [ ConnectionError ]
             |_| { "connection error" },
 
-        Packet
-            [ PacketError ]
+        Transfer
+            [ TransferError ]
             |_| { "packet error" },
```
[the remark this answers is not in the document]
soareschen: Fixed.

relayer/src/chain/cosmos/query.rs
soareschen: Fixed.

relayer/src/chain/cosmos/query/tx.rs
soareschen: Thanks, I think I just copied from the original comments.

relayer/src/chain/cosmos/retry.rs
soareschen: Yeah Rust do not currently support recursive async functions with the `async fn` sugar. I have added some comments to explain it.

romac: Right, didn't realize it was recursive.

Thanks for the comments. Since afaik this is the only occurrence of such a recursive async fn I don't we need to use https://docs.rs/async-recursion/latest/async_recursion/ but I am leaving it here just in case.

soareschen: Yeah probably don't need that for now. Actually aside from the boilerplate, using `Pin<Box<dyn Future>>` is not that difficult or intimidating once you wrote it a few times.

relayer/src/transfer.rs
```diff
@@ -71,6 +75,57 @@ impl FromStr for Amount {
...
+        timeout_height_offset: u64,
+        timeout_duration: Duration,
+        destination_chain_status: &ChainStatus,
+    ) -> Result<Self, TransferError> {
+        let offset_is_zero = timeout_height_offset == 0;
+        let timeout_is_zero = timeout_duration == Duration::ZERO;
+
+        let timeout_height = destination_chain_status.height.add(timeout_height_offset);
+
+        let timeout_timestamp = (destination_chain_status.timestamp + timeout_duration)
```
hu55a1n1, line 96: I think we're correctly using the `destination_chain_status.timestamp` (i.e. the receiving chain's timestamp) here. 👍

relayer/src/transfer.rs
```diff
@@ -71,6 +75,57 @@ impl FromStr for Amount {
...
+            .map_err(TransferError::timestamp_overflow)?;
+
+        if offset_is_zero && timeout_is_zero {
+            Err(TransferError::zero_timeout())
```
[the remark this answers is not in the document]
soareschen: Hmm I was just following the original logic of the ft-transfer command here: 

https://github.com/informalsystems/ibc-rs/blob/3e34b4b916eb91f680412e5fdd3ea045a99526d2/relayer-cli/src/commands/tx/transfer.rs#L129-L135

But we can relax that constraint if it makes sense. Then it is also simpler to just skip the checking and allow any height and timestamp. What do you think @romac?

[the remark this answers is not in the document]
romac: Sounds good to me. Let's perhaps also double check that the ft-transfer's behavior matches what will now be allowed here.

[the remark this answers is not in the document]
soareschen: Done.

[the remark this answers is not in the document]
soareschen: Oops, looks like that broke the Python E2E tests.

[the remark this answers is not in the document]
soareschen: Oh yeah, that makes sense. Nice catch! Fixed in [225f241](https://github.com/informalsystems/ibc-rs/pull/2044/commits/225f2417239e81bd01265f937262759540340454).

relayer/src/chain/cosmos/batch.rs
```diff
@@ -0,0 +1,166 @@
...
+
+pub async fn send_batched_messages_and_wait_commit(
+    config: &ChainConfig,
+    rpc_client: &HttpClient,
+    rpc_address: &Url,
+    grpc_address: &Uri,
+    key_entry: &KeyEntry,
+    account: &mut Account,
+    tx_memo: &Memo,
+    messages: Vec<Any>,
```
mzabaluev, line 25: This turns a two-argument function into an eight(!)-argument function.
There is a [clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#too_many_arguments) against this, is it overridden somewhere?

Can this be alleviated with struct arguments with largely `Default`-initialized fields, to simplify use of these functions? Something in the spirit of #2078 would be nice.

adizere: Related to https://github.com/informalsystems/ibc-rs/issues/375

romac: Let's address this in a follow-up PR, perhaps even in #2078.

relayer/src/lib.rs
```diff
@@ -7,6 +7,7 @@
     unused_qualifications,
     rust_2018_idioms
 )]
+#![allow(clippy::too_many_arguments)]
```
romac, line 10: @mzabaluev the lint is overridden here

romac (approved):
Amazing stuff, really improves the structure and legibility of the code 🎉

relayer/src/chain/cosmos/batch.rs
soareschen: I can go on a whole day to argue on this. But for time sake, the short summary is that having many function arguments reveals the _symptom_ of a program design. The practice of hiding many arguments behind an object the OOP way is merely uncovering the symptom, and allow the program design to further devolve because it does not look that bad syntactically when new fields are added.

The right way to fix the program design is to identify the symptoms of a program design by revealing how many variables a piece of code depends on, and then refactor on it. A functional design allows the program to be refactored incrementally, i.e. we can reduce the arguments passed into one function from the bottom up without having to touch the other functions. On the other hand, an OOP approach makes it much more challenging to perform any refactoring, because modifying one object field unavoidably affect every method in the monolithic object.

Here we have surfaced up the symptoms of the original program design and shows what need to be fixed next. In future PRs we can then identify proper functional design patterns that can simplify the code by curing the symptoms, instead of hiding the symptoms behind OOP again.

soareschen: Also note that unlike the original OOP methods, the function arguments do not propagate uniformly to all inner functions. There are also difference in reference types, with there being only one `&mut` reference as compared to the viral `&mut self` that allows any variable to be modified.

The OOP design is especially unsuited for Rust when `&mut` references are involved. It overly broadens every field in an object to become mutable, and the mutation infection easily spread until it turns every object method into using `&mut`. Mutable objects are especially difficult to deal with in Rust because of the exclusive reference. Before we know it, we get a fat object hiding behind `Arc<Mutex>` because the use of `&mut self` went out of control.

mzabaluev: I agree the "god object" pattern perpetuated, in part, by `ChainEndpoint` must be refactored. I'm not sure making the code less usable temporarily is a good way to go about it, but I'll remove my objection for the sake of expediency and a start on better code organization.

A vaguely related example where a proliferation of flat, non-composable functions makes the code less maintainable is the various `build_*{,_and_send}` methods for each type of message, with largely duplicated implementation code.

Issue #2068, opened 2022-04-06
Only allow channel-ids of the form `channel-{N}`


hu55a1n1:
## Summary of Bug
The `ChannelId::FromStr` impl does check if the given string is a valid ICS24 identifier but doesn't check if it is of the form `channel-{N}`. 

## Version
master

## Acceptance Criteria
- [ ] It shouldn't be possible to create a channel that isn't of the form `channel-{N}`. 
- [ ] Should we remodel `ChannelId` as `ChannelId(u64)` instead?
____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

romac:
Good catch! Let's model it as a newtype of u64 indeed!

Issue #1696, opened 2021-12-20
Proposal for redesigning parts of the ibc-rs modules' API


hu55a1n1:
## Crate
ibc-rs

## Proposal
Following is a list of ideas for redesigning parts of our ibc-rs modules' API accompanied with some reasoning for why these changes are required. (Some of these have already been implemented in PR #1583.) ->

- [ ] Convert the `Proofs` struct into an enum. (See https://github.com/informalsystems/ibc-rs/pull/1583#discussion_r766923695)
```rust
 // change this ->
 pub struct Proofs {
    object_proof: CommitmentProofBytes,
    client_proof: Option<CommitmentProofBytes>,
    consensus_proof: Option<ConsensusProof>,
    other_proof: Option<CommitmentProofBytes>,
    height: Height,
}

// to something like this -> 
pub enum Proofs {
    ChannelOpenTry(ChannelOpenTryProof), 
    // ...
}

// or maybe this -> 
pub struct SimpleProof(CommitmentProofBytes);
pub struct ComplexProof {
    pub proof: CommitmentProofBytes, 
    pub consensus_proof: ConsensusProof,
    // ...
}

pub enum Proofs {
    ConnectionOpenTry(ComplexProof), 
    ChannelOpenTry(SimpleProof), 
    // ...
}
```
- [ ] `ics24_host::Path` variants as types - Since Rust doesn't treat enum variants as types, it might be a good idea to extract these variants as distinct types and use those types as the actual enum variants. This will allow us to enforce that certain functions accept only a certain type of path. eg. `verify_client_consensus_state()` should only accept a `Path::ClientConsensusState`.
- [ ] Follow [RFC-356-no-module-prefixes](https://github.com/rust-lang/rfcs/blob/master/text/0356-no-module-prefixes.md). This basically means we rename `ibc::core::ics04_channel::msgs::MsgChannelCloseConfirm` to `ibc::core::ics04_channel::msgs::ChannelCloseConfirm`, and either re-export it as `MsgChannelCloseConfirm` in `msgs.rs` or add it to the prelude.
- [ ] Public `struct`s should either have public fields or setters/getters (if required), but not both. e.g.
```rust
pub struct MsgChannelCloseInit {
    pub port_id: PortId, // <- public
    // ...
}

impl MsgChannelCloseInit {
	pub fn port_id(&self) -> &PortId {
        &self.port_id
    } // <- also has a getter
}
```
[...]

PR #2071, opened 2022-04-06
Model `ChannelId` as newtype u64 and improve validation

hu55a1n1:
Closes: #2068 

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/modules/src/core/ics04_channel/events.rs
+++ b/modules/src/core/ics04_channel/events.rs
@@ -612,10 +612,10 @@ impl TryFrom<Attributes> for CloseInit {
             Ok(CloseInit {
                 height: attrs.height,
                 port_id: attrs.port_id.clone(),
-                channel_id: channel_id.clone(),
+                channel_id: *channel_id,
                 connection_id: attrs.connection_id.clone(),
                 counterparty_port_id: attrs.counterparty_port_id.clone(),
-                counterparty_channel_id: attrs.counterparty_channel_id.clone(),
+                counterparty_channel_id: attrs.counterparty_channel_id,
             })
         } else {
             Err(EventError::channel(Error::missing_channel_id()))
--- a/modules/src/core/ics04_channel/events.rs
+++ b/modules/src/core/ics04_channel/events.rs
@@ -1112,10 +1112,10 @@ mod tests {
         let attributes = Attributes {
             height: Height::default(),
             port_id: "test_port".parse().unwrap(),
-            channel_id: Some("test_channel".parse().unwrap()),
+            channel_id: Some("channel-0".parse().unwrap()),
             connection_id: "test_connection".parse().unwrap(),
             counterparty_port_id: "counterparty_test_port".parse().unwrap(),
-            counterparty_channel_id: Some("counterparty_test_channel".parse().unwrap()),
+            counterparty_channel_id: Some("channel-1".parse().unwrap()),
         };
         let mut abci_events = vec![];
         let open_init = OpenInit::try_from(attributes.clone()).unwrap();
@@ -1164,9 +1164,9 @@ mod tests {
         let packet = Packet {
             sequence: Sequence::from(10),
             source_port: "a_test_port".parse().unwrap(),
-            source_channel: "a_test_channel".parse().unwrap(),
+            source_channel: "channel-0".parse().unwrap(),
             destination_port: "b_test_port".parse().unwrap(),
-            destination_channel: "b_test_channel".parse().unwrap(),
+            destination_channel: "channel-1".parse().unwrap(),
             data: "test_data".as_bytes().to_vec(),
             timeout_height: Height::new(1, 10),
             timeout_timestamp: Timestamp::now(),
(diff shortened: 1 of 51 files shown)


modules/src/core/ics24_host/identifier.rs
```diff
@@ -337,7 +337,7 @@ impl Default for PortId {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
-pub struct ChannelId(String);
+pub struct ChannelId(u64);
```
romac, line 340: Let's derive `Copy` here? :)

romac: PSA: If you do the change, note that `cargo clippy --fix --allow-dirty` can remove all the unnecessary clones for you :)

hu55a1n1: > PSA: If you do the change, note that cargo clippy --fix --allow-dirty can remove all the unnecessary clones for you :)

This is awesome! Thank you! 🙏 Here's the diff -> https://github.com/informalsystems/ibc-rs/pull/2071/commits/b83fa5bcefa533a13c6f2c577fd4a6a3376a2759. Saved me a lot of time! 🎉

romac:
Any idea why this test only fails in this branch (because of #2075) and was not caught before?

https://github.com/informalsystems/ibc-rs/blob/1a6b41ad4ec5e2f2183bcdc74e4be99a4713f414/relayer/src/config/filter.rs#L450-L453

hu55a1n1:
> Any idea why this test only fails in this branch (because of https://github.com/informalsystems/ibc-rs/issues/2075) and was not caught before?

@romac and I figured out that this was because the original test used the `['ft-transfer', 'network-1']` combination in tests which didn't hit this bug because the `ChannelId`s didn't match. We only hit the bug once this PR disallowed `ChannelId`s that didn't have the channel-` prefix and we modified the test to `['ft-transfer', 'channel-1']`.

modules/src/core/ics04_channel/events.rs
```diff
@@ -256,7 +256,7 @@ impl From<Attributes> for Vec<Tag> {
         if let Some(channel_id) = a.channel_id {
             let channel_id = Tag {
                 key: CHANNEL_ID_ATTRIBUTE_KEY.parse().unwrap(),
-                value: channel_id.as_str().parse().unwrap(),
```
mzabaluev, line 259: `.to_string().parse()` looks like an unnecessary string conversion roundtrip. But an ABCI tag value is a UTF-8 string super-encoded as base64 for an inexplicable reason. Go figure.

modules/src/core/ics04_channel/handler/chan_open_try.rs
```diff
@@ -22,7 +22,7 @@ pub(crate) fn process(
     // Unwrap the old channel end (if any) and validate it against the message.
     let (mut new_channel_end, channel_id) = match &msg.previous_channel_id {
```
mzabaluev, line 23: If `ChannelId` is `Copy`, would it be nicer to match it by value and avoid the dereferencing cruft in match arms?

hu55a1n1: Do we always want to pass `ChannelId` by value? Halfway through this and the diff is huge, wondering if it's worth it ->
```
$ git diff --shortstat
 49 files changed, 220 insertions(+), 220 deletions(-)
```

modules/src/core/ics04_channel/events.rs
hu55a1n1: Right, it seems unnecessary. But I don't think there's any other public constructor for `Tag::Value`. 🤔 So not sure how to fix this. IIUC, the tag value was originally designed to be opaque bytes and later it was decided that (event) indexing was required and we're now left with this base64 encoded string.

romac: Nothing to fix here, maybe just add a comment?

modules/src/core/ics04_channel/handler/chan_open_try.rs
romac: It's more idiomatic but perhaps not worth it indeed. Let's file it in the backlog :)

hu55a1n1: Added this as an entry to this meta issue https://github.com/informalsystems/ibc-rs/issues/1696

modules/src/core/ics04_channel/events.rs
hu55a1n1: Just noticed that we already have this comment above -> https://github.com/informalsystems/ibc-rs/pull/2071/files/91c4c04dd191753eeac90159d2912262dfee17e0#diff-5d54d6bafcc46a42f95443d7c324b7c79563ad1f089618a686a8eb0d54e8beb0L237-L242

```rust
/// # Note
/// The parsing of `Key`s and `Value`s never fails, because the
/// `FromStr` instance of `tendermint::abci::tag::{Key, Value}`
/// is infallible, even if it is not represented in the error type.
/// Once tendermint-rs improves the API of the `Key` and `Value` types,
/// we will be able to remove the `.parse().unwrap()` calls.
```

romac (approved):

modules/src/core/ics04_channel/events.rs
mzabaluev: Yes, nothing to fix there, it's just an observation (maybe we should have an etiquette about this :blush:)

PR #2080, opened 2022-04-08
Increase timeout for assert_eventual_wallet_amount to reduce CI flakiness

soareschen:
## Description

The tests on CI sometimes have flaky failure from `assert_eventual_wallet_amount`, which returns error if the target wallet do not reach the target amount within 30 seconds.

The time limit might be too low in the CI environment, in particular since we allow multiple tests to be running in parallel, and thus increase the load of concurrent full nodes. We will try and increase the timeout to 60 seconds and see if it reduce the flakiness in the integration tests.

If the issue still persist, we may want to consider reducing the concurrent tests being run, so that each full node have sufficient resources to produce the blocks fast enough.

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -74,7 +74,7 @@ jobs:
         run: |
           nix shell github:informalsystems/cosmos.nix#${{ matrix.gaiad }} -c cargo \
             test -p ibc-integration-test --no-fail-fast -- \
-            --nocapture --test-threads=1
+            --nocapture --test-threads=2
 
   ibc-go-integration-test:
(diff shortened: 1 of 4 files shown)


adizere (approved):

soareschen:
It seems like the timeout still happen with other operations like waiting for tx commit. I try to limit the parallel threads to 2, as it takes about the same time as having no limit. Let's hope this is enough to stop the tests from timing out.

romac (approved):
Sweet, thanks! Let's see if this resolves the flaky tests problem

PR #2082, opened 2022-04-08
Only deploy the guide when a new release is created

romac:
## Description

Only deploy the guide to `gh-pages` when a new release is created.

This should alleviate the lack of versioning of the guide by only updating the published version whenever we create a new release of the project, which happens automatically when we push a `vX` tag.

Only downside is that we would have to do the deploy step manually whenever we want to update the guide (eg. because of a typo) in between releases, but that seems manageable.

### DIFF
--- a/.github/workflows/guide.yml
+++ b/.github/workflows/guide.yml
@@ -9,6 +9,8 @@ on:
   pull_request:
     paths:
       - guide/**
+  release:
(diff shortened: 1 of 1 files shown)


mzabaluev (approved):
Good! I like how the mdbook build is still run on PRs, so this does not remove quality control.

Issue #1862, opened 2022-02-09
Implement `query packet pending` command


mzabaluev:
Add a `hermes query packet` subcommand that shows all pending packets for a channel at both ends.

_Originally posted by @ancazamfir in https://github.com/informalsystems/ibc-rs/issues/1834#issuecomment-1033780559_

mzabaluev:
I need some more clue on how the pending packets should be represented in the output of the command.

If this command is to display packets pending at both ends of a channel, should the output:

1. provide a structured representation of the queues at each end as named fields; or
2. dump all packets in a flat list, but include the information about the chain the packets are pending on?

Cc @mircea-c @ancazamfir

ancazamfir:
I think 1. @mircea-c should have a big say here.
For me, the problem with the current CLIs is that a lot of concentration is required to figure things out.
I usually do:
- query local commitments
- query unreceived packets on counterparty
- query unreceived acks locally

And obtain something like this:
```
pending_local_commitments: [1, 2, 3]
not_received_on_counterparty: [2, 3]
received_on_counterparty_but_not_acked_locally: [1]
```
Then repeat the same for the other side. It would be nice to have ^ with one CLI.

The other problem with the current CLIs is with the formatting, we display one sequence per line. When there are tens/ hundreds of packets the output is pretty much useless without some extra parsing. Something like this would be nice for human eyes:
```
pending_local_commitments: [100-200, 202, 207]
not_received_on_counterparty: [150-200, 202, 207]
received_on_counterparty_but_not_acked_locally: [100-149]
```
But then probably not good for tooling/ automation.

mzabaluev:
@ancazamfir So it seems that in addition to the directionality information, the output would classify pending packets by kind. Should there be any details on the packet items themselves, or just IDs, possibly collated as you suggest?

> But then probably not good for tooling/ automation.

We've got the JSON mode for that, so we can make it nice for the `Debug` impl.

mircea-c:
I think what @ancazamfir suggested is excellent. It gets my vote.

mzabaluev:
> ```
> pending_local_commitments: [1, 2, 3]
> not_received_on_counterparty: [2, 3]
> received_on_counterparty_but_not_acked_locally: [1]
> ```

@ancazamfir As the first set is strictly a union of the other two, do we need to display it? Perhaps in text mode, but for JSON it would be redundant information.

mzabaluev:
[...]

PR #2096, opened 2022-04-14
Implement `query packet pending` command

mzabaluev:
Closes: #1862

## Description

Add a new subcommand `pending` to `query packet`, displaying a summary of pending packet commitment IDs for the identified channel and its reverse counterpart.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/relayer-cli/src/commands/query/packet/acks.rs
+++ b/relayer-cli/src/commands/query/packet/acks.rs
@@ -36,15 +36,16 @@ impl QueryPacketAcknowledgementsCmd {
 
         debug!("Options: {:?}", self);
 
-        let (chains, ccc) = spawn_chain_counterparty::<BaseChainHandle>(
+        let (chains, chan_conn_cli) = spawn_chain_counterparty::<BaseChainHandle>(
             &config,
             &self.chain_id,
             &self.port_id,
             &self.channel_id,
         )?;
 
-        let (seqs, height) = acknowledgements_on_chain(&chains.src, &chains.dst, &ccc.channel)
-            .map_err(Error::supervisor)?;
+        let (seqs, height) =
+            acknowledgements_on_chain(&chains.src, &chains.dst, &chan_conn_cli.channel)
+                .map_err(Error::supervisor)?;
 
         Ok(PacketSeqs { seqs, height })
     }
--- a/relayer-cli/src/commands/query/packet/acks.rs
+++ b/relayer-cli/src/commands/query/packet/acks.rs
@@ -36,14 +36,14 @@ impl QueryPacketAcknowledgementsCmd {
 
         debug!("Options: {:?}", self);
 
-        let (chains, channel) = spawn_chain_counterparty::<BaseChainHandle>(
+        let (chains, ccc) = spawn_chain_counterparty::<BaseChainHandle>(
             &config,
             &self.chain_id,
             &self.port_id,
             &self.channel_id,
         )?;
 
-        let (seqs, height) = acknowledgements_on_chain(&chains.src, &chains.dst, &channel)
+        let (seqs, height) = acknowledgements_on_chain(&chains.src, &chains.dst, &ccc.channel)
             .map_err(Error::supervisor)?;
 
         Ok(PacketSeqs { seqs, height })
--- a/relayer-cli/src/commands/query/packet/pending.rs
+++ b/relayer-cli/src/commands/query/packet/pending.rs
@@ -17,10 +17,10 @@ use crate::prelude::*;
 /// at both ends of a channel.
 #[derive(Debug, Serialize)]
 struct Summary {
-    /// The packets sent on the identified chain.
-    forward: PendingPackets,
+    /// The packets sent on the source chain as identified by the command.
+    src: PendingPackets,
     /// The packets sent on the counterparty chain.
-    reverse: PendingPackets,
+    dst: PendingPackets,
 }
 
 /// This command does the following:
@@ -54,7 +54,7 @@ impl QueryPendingPacketsCmd {
     fn execute(&self) -> Result<Summary, Error> {
         let config = app_config();
 
-        let (chains, ccc) = spawn_chain_counterparty::<BaseChainHandle>(
+        let (chains, chan_conn_cli) = spawn_chain_counterparty::<BaseChainHandle>(
             &config,
             &self.chain_id,
             &self.port_id,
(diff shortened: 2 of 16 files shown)


mzabaluev:
The command name could be subject to bikeshedding. It does not feel like a proper member of the `query packet` family.
Can it be `query packets` or `list packets`? The rationale is, it's a high level command displaying a summary of all packet commitments currently known as being in some pending state on both ends of the channel.

tools/integration-test/src/tests/query_packet.rs
```diff
@@ -0,0 +1,87 @@
...
+        let channel_end = chains.handle_a().query_channel(
+            channel.port_a.as_ref().value(),
+            channel.channel_id_a.as_ref().value(),
+            Height::zero(),
+        )?;
+        let channel_end = IdentifiedChannelEnd::new(
+            channel.port_a.clone().into_value(),
+            channel.channel_id_a.clone().into_value(),
+            channel_end,
+        );
```
mzabaluev, line 73: @soareschen Should there be an easier way to instantiate an `IdentifiedChannelEnd`, maybe as a method of `ConnectedChannel`?

soareschen: Yes, I think we can probably add a method in `ibc_test_framework::relayer::channel::TaggedChannelEndExt`. 

```rust
trait TaggedChannelEndExt<ChainA, ChainB> {
    fn as_identified_channel_end(&self) -> DualTagged<ChainA, ChainB, IdentifiedChannelEnd>;
}
```

Otherwise there is a tagged version of `query_channel` at `ibc_test_framework::relayer::channel::query_channel_end`. We could alternatively add a new function like `query_identified_channel_end` there.

```rust
fn query_identified_channel_end<ChainA: ChainHandle, ChainB>(
    handle: &ChainA,
    channel_id: &TaggedChannelIdRef<ChainA, ChainB>,
    port_id: &TaggedPortIdRef<ChainA, ChainB>,
) -> Result<DualTagged<ChainA, ChainB, IdentifiedChannelEnd>, Error>
```

The tagged version of query channel helps us ensure that we do not mix up the channel identifiers some where along the way.

tools/integration-test/src/tests/query_packet.rs
```diff
@@ -0,0 +1,149 @@
...
+
+#[test]
+fn test_query_packet_pending() -> Result<(), Error> {
+    run_binary_channel_test(&QueryPacketPendingTest)
+}
+
+pub struct QueryPacketPendingTest;
+
+impl TestOverrides for QueryPacketPendingTest {
+    fn modify_relayer_config(&self, config: &mut Config) {
```
seanchen1991, line 17: Just want to clarify: for the purposes of this test, is it fine that pending packets are relayed once the relayer has been started? That's what will happen if `config.mode.tx_confirmation` is `true`, but I don't know what the default value of `tx_confirmation` is. 

If it's `false`, then pending packets won't be cleared due to the interaction between `tx_confirmation` and `clear_interval`. If it's `true`, then they will be cleared on account of the interaction. 

If it doesn't matter that pending packets are cleared after the relayer has been started, then I think it's also fine to ignore this interaction.

tools/integration-test/src/tests/query_packet.rs
```diff
@@ -0,0 +1,149 @@
...
+        config.mode.packets.clear_on_start = false;
+        config.mode.packets.clear_interval = 0;
+    }
+
+    fn should_spawn_supervisor(&self) -> bool {
+        false
+    }
+
+    // Unordered channel: will permit gaps in the sequence of relayed packets
+    fn channel_order(&self) -> Order {
```
seanchen1991, line 29: We can remove this method since the trait implementation defaults to `Order::Unordered`.

tools/integration-test/src/tests/query_packet.rs
```diff
@@ -0,0 +1,149 @@
...
+        chains.node_a.chain_driver().transfer_token(
+            &channel.port_a.as_ref(),
+            &channel.channel_id_a.as_ref(),
+            &wallet_a.address(),
+            &wallet_b.address(),
+            amount1,
+            &denom_a,
+        )?;
+
+        sleep(Duration::from_secs(2));
```
seanchen1991, line 63: I've seen in other integration tests that the process sleeps for 1 second. Is there any particular rationale for having it sleep for 2 in this case?

tools/integration-test/src/tests/query_packet.rs
```diff
@@ -0,0 +1,149 @@
...
+            amount2,
+            &denom_b,
+        )?;
+
+        info!(
+            "Performing IBC transfer with amount {}, which should *not* be relayed",
+            amount2
+        );
+
+        sleep(Duration::from_secs(2));
```
seanchen1991, line 124: Same question here about sleeping for 2 seconds vs 1.

seanchen1991 (dismissed):
Left a few comments and questions. There are few places where the variable name `ccc` is used. Could we replace these with more descriptive variable names?

tools/integration-test/src/tests/query_packet.rs
mzabaluev: I had one sporadically failed run with 1 second sleep here, and none after I bumped it to 2.
It would be good to have some way to wait for confirmation.

relayer-cli/src/commands/query/packet/acks.rs
```diff
@@ -36,14 +36,14 @@ impl QueryPacketAcknowledgementsCmd {
 
         debug!("Options: {:?}", self);
 
-        let (chains, channel) = spawn_chain_counterparty::<BaseChainHandle>(
+        let (chains, ccc) = spawn_chain_counterparty::<BaseChainHandle>(
```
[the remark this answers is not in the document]
mzabaluev: Yeah... The type is a carry bag for returning channel-connection-client, it's never used as a parameter, so it's always clear from use of the members what are we accessing. Will rename for further clarity.

relayer-cli/src/commands/query/packet/pending.rs
```diff
@@ -0,0 +1,89 @@
...
+
+use crate::cli_utils::spawn_chain_counterparty;
+use crate::conclude::Output;
+use crate::error::Error;
+use crate::prelude::*;
+
+/// A structure to display pending packet commitment sequence IDs
+/// at both ends of a channel.
+#[derive(Debug, Serialize)]
+struct Summary {
```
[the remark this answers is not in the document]
mzabaluev: It's important to bikeshed this indeed, because these names show up in serialization and output.
No objections to `src` and `dst` from me.

tools/integration-test/src/tests/query_packet.rs
mzabaluev: The test never runs the supervisor, packets are relayed step by step by invoking methods on a `Link`.
This is to get the packet summaries into an expected state without dependency on timing.

mzabaluev:
The test failure looks like an unrelated flaky test.

ancazamfir:
Ran a few tests, things look good. A bit hard to read with lots of packets. Have you tried to implement the collated output?

ancazamfir:
So the output looks something like this:
```
Success: Summary {
    forward: PendingPackets {
        unreceived: [
            2203,
...
            2212,
        ],
        pending_acks: [
            2183,
...
           2202,
        ],
    },
    reverse: PendingPackets {
        unreceived: [
            14,
...
           23,
        ],
        pending_acks: [
            4,
...
            13,
        ],
    },
}
```
Looking at the field names, since all are unreceived/ pending, maybe `unreceived` -> `pending_recv`?
So we have `pending_recv` and `pending_acks`?
Or have `unreceived_packets` and `unreceived_acks`? Though this looks weird as sometimes we use `packets` to refer to all types (i.e. send, recv, timeout, ack)

mzabaluev:
> Have you tried to implement the collated output?

Not yet. I can do it, or maybe leave this for a later enhancement? The other `query packet` subcommands could benefit from this as well.

romac:
> Looking at the field names, since all are unreceived/ pending, maybe `unreceived` -> `pending_recv`? So we have `pending_recv` and `pending_acks`?

+1 for `pending_recv` and `pending_acks`

romac:
> IMO we can leave it like this for this PR and clean things up in a subsequent PR.

+1

romac:
@mzabaluev is this ready for merge or do you have more pending changes? are we leaving the collated output for a later PR?

mzabaluev:
@romac We can leave the collation for later; I've filed it as #2142.

romac:
Great, then I'll leave it to @seanchen1991 to take another look at this and merge.

seanchen1991 (commented):
It looks like there are still a few lingering clippy lints reporting unnecessary `clone` calls. Are these calls being left in intentionally?

adizere:
> It looks like there are still a few lingering clippy lints reporting unnecessary `clone` calls. Are these calls being left in intentionally?

Good point, we should fix those!

romac (approved):

Issue #2104, opened 2022-04-19
Commitment is not compatible with ibc-go


yito88:
## Summary of Bug
A commitment made by ibc-rs was not compatible with cosmos(ibc-go)

- Make a commitment in ibc-go
https://github.com/cosmos/ibc-go/blob/457095517b7832c42ecf13571fee1e550fec02d0/modules/core/04-channel/types/packet.go#L18-L34

- Make a commitment in ibc-rs
https://github.com/informalsystems/ibc-rs/blob/37dbe8c42ae09403432ebfcfdb2a8a34d86909e9/modules/src/core/ics04_channel/handler/verify.rs#L73-L77

## Version
v0.12.0

## Steps to Reproduce
`core::ics04_channel::handler::verify::verify_packet_recv_proofs()` failed due to the commitment

## Acceptance Criteria
The commitment made by Cosmos can be verified.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2105, opened 2022-04-19
Fix packet commitment computation and incorrect ack verification

yito88:
Closes: #2104 
Closes: #2114  

## Description
Make a commitment as ibc-go

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/modules/src/core/ics02_client/client_def.rs
+++ b/modules/src/core/ics02_client/client_def.rs
@@ -118,7 +118,7 @@ pub trait ClientDef: Clone {
         port_id: &PortId,
         channel_id: &ChannelId,
         sequence: Sequence,
-        commitment: String,
+        commitment: Vec<u8>,
     ) -> Result<(), Error>;
 
     /// Verify a `proof` that a packet has been commited.
@@ -437,7 +437,7 @@ impl ClientDef for AnyClient {
         port_id: &PortId,
         channel_id: &ChannelId,
         sequence: Sequence,
-        commitment: String,
+        commitment: Vec<u8>,
     ) -> Result<(), Error> {
         match self {
             Self::Tendermint(client) => {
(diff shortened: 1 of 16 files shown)


modules/src/mock/context.rs
```diff
@@ -833,9 +849,10 @@ impl ChannelKeeper for MockContext {
         timeout_height: Height,
         data: Vec<u8>,
     ) -> Result<(), Ics04Error> {
-        let input = format!("{:?},{:?},{:?}", timeout_timestamp, timeout_height, data);
-        self.packet_commitment
-            .insert(key, ChannelReader::hash(self, input));
+        self.packet_commitment.insert(
```
hu55a1n1, line 852: I wonder if it would be better to change this trait method (i.e. `store_packet_commitment()` and others like it) so that we give it the precalculated commitment because we already allow users to define how the commitment is calculated. (cc: @romac, @plafer)

hu55a1n1 (commented):
Thanks, @yito88! Nicely done! 👌 Left some suggestions and comments.

modules/src/core/ics02_client/client_def.rs
```diff
@@ -494,7 +493,7 @@ impl ClientDef for AnyClient {
         port_id: &PortId,
         channel_id: &ChannelId,
         sequence: Sequence,
-        ack: Acknowledgement,
```
Skysaysky, line 497: I think the continued use of the Acknowledgement type gives a better indication of what is mean.

yito88: I think the "acknowledgment commitment" should be verified here instead of `Acknowledgement`.

DaviRain-Su: > I think the "acknowledgment commitment" should be verified here instead of `Acknowledgement`.

Maybe give a newtype AcknowledgementCommitment

yito88: I added PacketCommitment and AcknowledgementCommitment

hu55a1n1 (approved):
Thanks for the amazing work @yito88! 🥇❤️

romac:
Can we give this PR a more meaningful name before merging please? :)

PR #2108, opened 2022-04-19
ADR 009: ChainEndpoint/ChainHandle Consolidation

plafer:
## Description

This PR proposes ADR 8, which aims to reduce the code duplication between the `ChainEndpoint` and `ChainHandle` traits.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


plafer:
Please let me know if you can think of better names for `ChainBase` and `unroll_template`.

mzabaluev:
In my understanding, `ChainEndpoint` is the extension point to enable runtime backends for different chains (in the future).
`ChainHandle` is used to build utility on the API client side, such as mocks for testing, statistics middleware, and the like.

One potential concern with this change is that it may introduce unnecessary coupling between the client side and the runtime mechanics. For example, in a future async rewrite we might choose to convert components to async piecemeal or even leave some of them synchronous. Use of a common trait would make that process unnecessarily complicated.

That said, I'm all for reducing the surface area of the traits by deduplicating methods, using auxiliary data types, etc.

adizere:
> In my understanding, `ChainEndpoint` is the extension point to enable runtime backends for different chains (in the future).
> `ChainHandle` is used to build utility on the API client side, such as mocks for testing, statistics middleware, and the like.

This is a very good point. It is highlighted also in our of older ADRs (005). The runtime backends are described [here](https://github.com/informalsystems/ibc-rs/blob/master/docs/architecture/adr-005-relayer-v0-implementation.md#2-the-chain-runtime). The API client side can also be seen as [a "relayer application" which uses the `ChainHandle`](https://github.com/informalsystems/ibc-rs/blob/master/docs/architecture/adr-005-relayer-v0-implementation.md#3-the-relayer-application).

One advantage of this 2-sided architecture is that we can customize handlers (relayer application logic) independently of runtimes. A concrete example is in the caching layer implementation, which was added as a [custom chain handle](https://github.com/informalsystems/ibc-rs/pull/1932/files#diff-19b80b2fef3e4df2d30f9ceb9a799dc2bccc85d3771e30b47fa8ad8122057b06).

adizere:
One more input we should take into account in the redesign of Runtime/Handlers is the need for supporting parallel queries in the runtime. The problem is as follows:

Whenever a relayer object (`ForeignClient`, `Connection`, `RelayPath`, etc) wants to assemble a message (such as a client update message), it needs to do certain queries on both the source and destination chain. Currently, each such query is serialized via the Handle -> crossbeam -> Runtime (blocking) interface. This enforces that all relayer objects must wait on each other when they want to query a certain chain, a limitation that likely affects performance.

Can we fulfil this requirement with the proposed redesign, or can we accommodate it?

plafer:
Good point @mzabaluev. How I see it, nothing needs to change in the ADR other than having both `ChainEndpoint` and `ChainHandle` have some `ChainBase` as supertrait. This will facilitate further changes we expect that would have these traits' interface diverge.

@adizere IIUC, you're referring to the problem that `ChainRuntime::run()` takes one message at a time out of the channel, delegates the network request to the inner `ChainEndpoint`, and waits until it gets the response back until it starts processing the next item in the channel. That to me calls for refactoring the Runtime to use `async` features, and falls out of the scope of this ADR. I will say though that I don't foresee the changes I'm proposing here to make this future `async` refactor more difficult.

mzabaluev:
@adizere When we get an async tendermint-rs, we could start with converting the runtime (and `ChainEndpoint`) to async. If we choose to not do the same for the `ChainHandle` APIs at that point, the common base trait would be a hindrance.

docs/architecture/adr-008-chain-endpoint-handle-consolidation.md
```diff
@@ -0,0 +1,123 @@
...
+    query_height: Height,
+) -> ...;
+
+fn query_consensus_states(
+    &self,
+    request: QueryConsensusStatesRequest,
+) -> ...;
+```
+
+All methods will be refactored to take a request object as argument.
```
romac, line 27: I would rather go the other way and have all methods take multiple parameters. Reason for this is that the `*Request` datatypes are Protobuf definitions which I would rather not expose in the public API directly, as those are merely an internal implementation detail.

adizere:
Meta: removed this PR from milestone 0.14 (we're using milestones for issues, not PRs). It's OK to push this work into next release train (0.15).

docs/architecture/adr-008-chain-endpoint-handle-consolidation.md
```diff
@@ -0,0 +1,123 @@
...
+    &self,
+    request: QueryConsensusStatesRequest,
+) -> Result<Vec<AnyConsensusStateWithHeight>, Error> {
+    self.send(|reply_to| ChainRequest::QueryConsensusStates { request, reply_to })
+}
+```
+
+## Status
+
+Proposed
```
adizere, line 108: Let's also record the status in the table here https://github.com/informalsystems/ibc-rs/tree/master/docs/architecture#table-of-contents

docs/architecture/adr-008-chain-endpoint-handle-consolidation.md
```diff
@@ -0,0 +1,123 @@
...
+## Changelog
+* 2022-04-19: Initial Proposal
+
+## Context
+There is a lot of common methods in the `ChainHandle` and `ChainEndpoint` traits, sometimes with minute differences between one another. This document provides a way to remove the duplication of methods for increased maintainability of the codebase, along with a few suggestions to standardize the method signatures. It also suggests a procedural macro to simplify the currently repetitive implementation of the `query_*` methods, which would be made possible given the proposed standardization of the method signatures.
+
+## Decision
+
+### Query methods parameters
+There are currently discrepancies between how methods take their arguments. Some take a `request` object, and others take fine-grained arguments that will be used to build a `request` object in the implementation of the method. For example, `query_consensus_state()` takes arguments that will be used to build a request object, whereas `query_consensus_states()` takes a request object directly.
```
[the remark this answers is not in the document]
plafer: As discussed with @romac, this ADR would have each method have a domain `Request*` type, irrespective of if it ends up using RPC, gRPC, etc, effectively abstracting that away.

adizere (approved):
Feel free to merge Philippe. Nice consolidation!

PR #2118, opened 2022-04-25
Fix `execute_schedule` method leaking operational data

seanchen1991:
Closes: #1153

## Description

This PR adds an integration test that verifies that the `execute_schedule` method drops operational data when an error is encountered. It also fixes the leaking of data such that the integration test passes. 

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


seanchen1991:
@adizere I implemented a fix for `execute_schedule`'s leaky logic, but interestingly, when I ran the integration test, the final assertion fails because the number of queued up packets == the number of transfers that were performed in the test, not 1 less.  

I.e. If 10 transfers were performed, then currently the length of `dst_operational_data` in the final assertion is 10, _not_ 9, which is what we expected. 

That.. shouldn't be happening, right?

relayer/src/link/operational_data.rs
```diff
@@ -17,9 +17,12 @@ use crate::chain::tracking::TrackingId;
 use crate::link::error::LinkError;
 use crate::link::RelayPath;
 
+/// The chain that a piece of `OperationalData` is bound for.
 #[derive(Clone, Copy, PartialEq)]
 pub enum OperationalDataTarget {
+    /// The source chain.
```
adizere, line 23:
```suggestion
    /// The source chain, i.e., the chain which generated the events batch in the operational data.
```

relayer/src/link/operational_data.rs
```diff
@@ -17,9 +17,12 @@ use crate::chain::tracking::TrackingId;
...
     Source,
+    /// The destination chain.
```
adizere, line 25:
```suggestion
    /// The destination chain, i.e., the chain on the receiving side of packets.
```

relayer/src/link/relay_path.rs
```diff
@@ -98,8 +98,8 @@ pub struct RelayPath<ChainA: ChainHandle, ChainB: ChainHandle> {
     // mostly timeout packet messages.
     // The operational data targeting the destination chain
     // comprises mostly RecvPacket and Ack msgs.
-    pub(crate) src_operational_data: Queue<OperationalData>,
-    pub(crate) dst_operational_data: Queue<OperationalData>,
+    pub src_operational_data: Queue<OperationalData>,
+    pub dst_operational_data: Queue<OperationalData>,
```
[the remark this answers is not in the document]
adizere: Not a big issue from what I can tell. The [`cli.rs`](https://github.com/informalsystems/ibc-rs/blob/6abaa31e909586a625ae4c4ab5fd3540ab3d5c38/relayer/src/link/cli.rs#L31) seems to need acccess, hence the `crate` specifier.

relayer/src/link/relay_path.rs
```diff
@@ -1278,27 +1279,109 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
-        for od in dst_ods {
-            let reply =
-                self.relay_from_operational_data::<relay_sender::AsyncSender>(od.clone())?;
+            match elapsed_result {
+                Ok(elapsed) => {
+                    if elapsed {
+                        match self
+                            .relay_from_operational_data::<relay_sender::AsyncSender>(od.clone())
+                        {
+                            Ok(reply) => self.enqueue_pending_tx(reply, od),
```
adizere, line 1334: The logic here is quite deep (~5 lvls of indentation) and I'm having difficulties understanding it. I appreciate the thorough doc comments for the method, which help, and wondering if we should have a helper inner `fn` here or some other approach to make things more obvious.

relayer/src/link/relay_path.rs
```diff
@@ -1278,27 +1279,109 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+        Ok(unprocessed)
+    }
+
+    /// While there are pending operational data items, this function
+    /// performs the relaying of packets corresponding to those
+    /// operational data items to both the source and destination chains.
+    ///
+    /// Any operational data items that do not get successfully relayed are
+    /// dropped. Subsequent pending operational data items that went unprocessed
+    /// are queued up again for re-submission.
```
adizere, line 1363: Neat!

tools/integration-test/src/tests/execute_schedule.rs
```diff
@@ -0,0 +1,82 @@
...
+            info!("Performing IBC transfer #{} from chain A to chain B", i);
+        }
+
+        assert_eq!(relay_path_a_to_b.dst_operational_data.len(), NUM_TXS);
+
+        chains.node_b.value().kill()?;
+
+        match relay_path_a_to_b.execute_schedule() {
+            Ok(_) => panic!("Expected an error when relaying tx from A to B"),
+            Err(_) => assert_eq!(relay_path_a_to_b.dst_operational_data.len(), NUM_TXS - 1),
```
adizere, line 77: Nit: Can we state the connection between this code and the top-level comment in this file "all of the subsequent .. transactions should exist in the pending queue." ? From what I can tell, `NUM_TXS - 1` represent the subsequent transactions that should not have been dropped, right?

tools/integration-test/src/tests/execute_schedule.rs
```diff
@@ -0,0 +1,82 @@
+//! This test ensures that the `RelayPath::execute_schedule` method does not
+//! drop any pending scheduled operational data when a prior transaction fails
+//! to send. Subsequent pieces of operational data that were scheduled should
+//! be re-queued and not dropped.
+//!
+//! In order to test this behavior, the test manually relays at least 2 IBC transfers
+//! from chain A to chain B. Chain B is then shut down in order to force the transactions
+//! to be queued up again for re-submission. It is expected that the first transfer
+//! be dropped and not be present in the pending queue, but all of the subsequent
```
adizere, line 9: Nit: these are not transactions, but operational data (i.e., batches or messages waiting to be relayed).

tools/integration-test/src/tests/execute_schedule.rs
```diff
@@ -0,0 +1,82 @@
...
+                &channel.channel_id_a.as_ref(),
+                &chains.node_a.wallets().user1(),
+                &chains.node_b.wallets().user1().address(),
+                &chains.node_a.denom(),
+                amount1,
+            )?;
+
+            relay_path_a_to_b.schedule_packet_clearing(None)?;
+
+            info!("Performing IBC transfer #{} from chain A to chain B", i);
```
adizere, line 68:
```suggestion
            info!("Performing IBC send packet with a token transfer #{} from chain A to be received by chain B", i);
```

tools/integration-test/src/tests/execute_schedule.rs
```diff
@@ -0,0 +1,82 @@
...
+            chains.node_a.chain_driver().ibc_transfer_token(
+                &channel.port_a.as_ref(),
+                &channel.channel_id_a.as_ref(),
+                &chains.node_a.wallets().user1(),
+                &chains.node_b.wallets().user1().address(),
+                &chains.node_a.denom(),
+                amount1,
+            )?;
+
+            relay_path_a_to_b.schedule_packet_clearing(None)?;
```
adizere, line 66: What happens in this test is quite subtle, so let's leave some hints/comments? From what I can tell, each send packet (`ibc_transfer_token`) materializes into a separate operational data when we call `schedule_packet_clearing`.

relayer/src/link/operational_data.rs
seanchen1991: The language being proposed for the `Source` variant vs. the `Destination` variant has me wondering what the relationship is between packets and operational data; how are the two related? Why is it that the destination chain receives packets, not operational data?

tools/integration-test/src/tests/execute_schedule.rs
seanchen1991: Yes, that's correct. I'll rename the constant to `BATCH_SIZE`; hopefully that makes it more clear.

relayer/src/link/relay_path.rs
seanchen1991: I couldn't figure out a clean way to encapsulate this logic in a separate function. In some cases this function would need to return a single piece of operational data (if it hasn't elapsed), in other cases it needs to return `(VecDeque<OperationalData>, Option<Error>)`; I don't see a clean way to structure the return type. 

I've opted instead to annotate the nested logic as clearly as possible. Hopefully that addresses your concern?

relayer/src/link/operational_data.rs
adizere: > what the relationship is between packets and operational data; how are the two related?

This is also something I've been struggling a lot with. Hope these notes below help. The important concepts are marked in __bold__:


- one chain (source) __sends__ packets
    - a user (typically) submits transactions with these kind of packets
- the other chain (destination) __receives__ packets
    - a relayer submits these kind of packets to the destination chain
    - if no relayer acts in time, then the relayer submits a __timeout__ packet to the source chain (instead of **receive** packet to destination chain)
- when a chain processes a send packets, that translates into an event
    - when a chain processes a receive packet, that also translates into an event
    - Hermes acts upon these kind of events reactively
- Hermes translates events into operational data (a concept internal to Hermes), which encompasses both an event as well as any resulting -- receive or timeout -- packet that Hermes has to submit based on that event
- so operational data is just an internal representation that Hermes keeps for pairing together an event with the action that that event entails


aside from the main question, but:
- all packets (send, receive, timeout) are IBC __messages__
- beside packets, there are other types of messages: client update messages, notably, and soon relayer fees messages (I think).
- connection and channel handshakes are also types of IBC messages
- Hermes typically batches multiple IBC messages into a __transaction__

so the three primitive data types are:
- packets, messages, and transactions

I'm thinking we should turn this into a diagram and put in the architecture or docs/ somewhere.

seanchen1991: I opted to capture as much of this that is relevant to the `OperationalData` type in its doc comment. Hopefully it helps clear some of this up a bit.

relayer/src/link/operational_data.rs
```diff
@@ -78,18 +78,25 @@ pub struct TransitMessage {
...
+/// an event received from a chain along with any other packets related to the event (i.e.
+/// 'receive' or 'timeout' packets) that the relayer has to submit in response to the event.
 ///
-/// Each `OperationalData` item is uniquely identified by the combination of two attributes:
-///     - `target`: represents the target of the packet messages, either source or destination chain,
-///     - `proofs_height`: represents the height for the proofs in all the messages.
-///       Note: this is the height at which the proofs are queried. A client consensus state at
-///       `proofs_height + 1` must exist on-chain in order to verify the proofs.
+/// Each `OperationalData` item is uniquely identified by the combination of its `target` chain
+/// and its `proofs_height` value.
```
seanchen1991, line 86: With the introduction of the `tracking_id` field, does this still hold true? I would think that the `tracking_id` is now the main source of unique identification for a piece of operational data.

adizere: Context: The `tracking_id` adds the ability to track a set of events from the moment Hermes receives them via websocket until it submits the transactions corresponding to those events. An operational data inherits the `tracking_id` from the events it contains. Each `tracking_id` should be unique, but has no semantic meaning.

> With the introduction of the tracking_id field, does this still hold true?

It's still true, but now I realize it's not a very useful piece of information, it may even be confusing, so cool to remove it!

adizere (commented):
We should fix the comments. Otherwise I tested and looks good!

adizere (approved):
LGTM, great work with this Sean!

PR #2140, opened 2022-04-26
extract ics07 verify_* function to ics02-client/client_def.rs as default impl function

DaviRain-Su:
## Description

Since the overall validation of the data is dependent on ics23, I think it is possible to have these validation functions as the default implementation in ics02.

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/modules/src/core/ics02_client/client_def.rs
+++ b/modules/src/core/ics02_client/client_def.rs
@@ -15,10 +15,18 @@ use crate::core::ics04_channel::packet::Sequence;
 use crate::core::ics23_commitment::commitment::{
     CommitmentPrefix, CommitmentProofBytes, CommitmentRoot,
 };
+use crate::core::ics23_commitment::merkle::{self, apply_prefix};
 use crate::core::ics24_host::identifier::{ChannelId, ClientId, ConnectionId, PortId};
+use crate::core::ics24_host::path::{
+    AcksPath, ChannelEndsPath, ClientConsensusStatePath, ClientStatePath, CommitmentsPath,
+    ConnectionsPath, ReceiptsPath, SeqRecvsPath,
+};
+use crate::core::ics24_host::Path;
 use crate::downcast;
 use crate::prelude::*;
 use crate::Height;
+use prost::Message;
+use tendermint_proto::Protobuf;
 
 #[cfg(any(test, feature = "mocks"))]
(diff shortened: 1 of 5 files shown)


modules/src/core/ics02_client/client_state.rs
```diff
@@ -11,6 +11,7 @@ use crate::clients::ics07_tendermint::client_state;
 use crate::core::ics02_client::client_type::ClientType;
 use crate::core::ics02_client::error::Error;
 use crate::core::ics02_client::trust_threshold::TrustThreshold;
+use crate::core::ics23_commitment::specs::ProofSpecs;
```
hu55a1n1, line 14: I don't think we can have a `proof_spec()` method in `ClientState` because that is Tendermint specific, and it should be possible to create a client that doesn't use ICS23 style proofs. By extension, verification functions that rely on proof specs are also client-specific so IMHO they should not be present in the default implementation or trait API.

modules/src/core/ics02_client/client_def.rs
```diff
@@ -150,7 +236,25 @@ pub trait ClientDef: Clone {
...
+        client_state.verify_height(height)?;
+        self.verify_delay_passed(ctx, height, connection_end)?;
+
+        let mut seq_bytes = Vec::new();
+        u64::from(sequence)
+            .encode(&mut seq_bytes)
+            .expect("buffer size too small");
+
+        let seq_path = SeqRecvsPath(port_id.clone(), *channel_id);
+        self.verify_membership(
```
hu55a1n1, line 249: Note that `verify_membership()` doesn't provide access to the context, so (although this is only a default impl) it might be restrictive for light clients whose verification depends on data other than what we're giving this function here.

hu55a1n1 (commented):
Thanks for the PR @DaviRain-Su! 🙏 And sorry for the delay in the review. The `ClientDef` and `ClientState` traits must be client agnostic. 
> Since the overall validation of the data is dependent on ics23.

I think it should be possible to create clients that don't depend on ICS23 proof verification. So, IMHO the changes in this PR make the API difficult to use for other light client implementations. (see review comments) Are there specific reasons for these changes that you need for your light client implementation? If so, then please let us know and we can try to figure together how best to accommodate your requirements. 
Also tagging @romac for his inputs.

Issue #1910, opened 2022-02-22
Refactor ibc-integration-test to produce test setup similar to dev-env script


soareschen:
## Crate

`ibc-integration-test`

## Summary

Currently the integration test is setup to always generate randomized parameters such as chain ID, RPC ports, and client IDs. This leads to some users still preferring to use the legacy dev-env script, for the convenience of having deterministic config and parameters at the risk of failure caused by port conflict.

It should be pretty straightforward to produce a specialized test setup using the integration test framework with non-randomized parameters and with config that is close to the example config in the Hermes guide. This should eliminate the remaining reason for users to stick with the legacy dev-env script.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2147, opened 2022-04-27
Deterministic integration test environment

soareschen:
Closes: #1910

## Description

This PR modifies the executable `test_setup_with_binary_channels` to produce a test environment that is similar to the legacy `dev-env` script. By running `cargo run --bin test_setup_with_binary_channels`, you'd get a test environment as follows:

- Test store directory at `data/test`, unless overridden by `$TEST_STORE_DIR`. The directory must be empty for the setup to succeed.
- Environment variables loadable via `source data/test/binary-channels.env`.
- Relayer config present at `data/test/relayer-config.toml`.
- Chains setup with the names `ibc-0` and `ibc-1`. (no random suffix)
- Wallets setup with the names `validator`, `relayer`, `user1`, and `user2`. (no random suffix)
- The IBC identifiers follow the chain index, i.e. `07-tendermint-0`, `connection-0`, `channel-0` for `ibc-0`, and `07-tendermint-1`, `connection-1`, `channel-1` for `ibc-1`. (no random IDs)
- Fixed denom `samoleans` on _both_ chains.

The setup is done by refactoring the bootstrap scripts to turn off random ID generation based on `test_config.bootstrap_with_random_ids`.

The main difference of this test setup with `dev-env` is as follows:

- The ports are randomized, but can be accessible via environment variables like `$NODE_A_RPC_ADDR`.

Hopefully this makes it compelling enough to eliminate all remaining reasons to keep using `dev-env` for some developers.

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


soareschen:
I have updated the bootstrap script such that when `bootstrap_with_random_ids` is `false`, the IBC IDs follow the chain IDs, i.e. `07-tendermint-0`, `connection-0`, `channel-0` for `ibc-0`, and `07-tendermint-1`, `connection-1`, `channel-1` for `ibc-1`.

I have also disabled the relayer supervisor, since it is also not enabled by `dev-env`. Instead users can start/stop the relayer at any time by explicitly calling `hermes start`.

@seanchen1991 FYI the Python end-to-end tests pass with the chain setup provided here. @adizere @romac perhaps we can simplify the CI pipeline for running the Python end-to-end tests using `ibc-test-framework` an Cosmos.nix, instead of using Docker and `dev-env`.

soareschen:
I have now also added a ternary test setup to setup 3 chains with 3x3 connected channels. This can be setup by running `cargo run --bin test_setup_with_binary_channel`. The test environment variables can be loaded from `data/test/nary-channels.env`.

Note that in this case, the IBC identifiers are all too confusing to be derived using common sense, so the use of environment variables are mandatory. For example, the variable `$CLIENT_ID_0_to_2` is the client ID on `ibc-2` that is correspond to the client on `ibc-0`. Similarly, `$CHANNEL_ID_0_to_2` is the channel ID on `ibc-2` that is correspond to the channel on `ibc-0`.

With N-ary chains, the connection and channel IDs can no longer be identified with a single chain. Instead, the naming convention follows the contravariant convention in `ForeignClient` and `ClientId`. i.e. the originally named `$CHANNEL_ID_A` in the binary case actually meant `$CHANNEL_ID_B_TO_A`. This is just the naming complexity that we have to deal with in a multi-chain world.

romac (approved):

tools/integration-test/src/tests/python.rs
```diff
@@ -0,0 +1,71 @@
...
+        let base_dir = env::var("PWD").unwrap_or(current_dir);
+
+        let command_args = [&format!("{}/e2e/run.py", base_dir), "-c", config_path];
+
+        let output = Command::new("python3")
+            .args(&command_args)
+            .current_dir(base_dir)
+            .stdout(Stdio::inherit())
+            .stderr(Stdio::inherit())
+            .output()?;
```
soareschen, line 55: > How would you suggest we proceed in order to run the existing e2e Python test suite using the testing framework environment?

@romac turns out this become rather simple: we run the Python E2E tests inside the integration test itself.

romac: Sweet!

flake.lock
```diff
@@ -0,0 +1,801 @@
+{
```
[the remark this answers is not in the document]
adizere: > Soares: This is necessary. Prevent breaking our CI, which could happen if we use the latest version of Cosmos.nix.

PR #2151, opened 2022-04-27
Fix release by removing `with-bitcoin` feature for `hdpath` dependency

romac:
Related: https://github.com/informalsystems/ibc-rs/pull/2148

The release of `bitcoin` v0.28.0 broke our release because `cargo publish` does not use the lock file and `hdpath` depends on `bitcoin >= v0.21` which pulls in v0.28.0 whereas we currently require v0.27.1.

Until we explicitly depend on `bitcoin` v0.28.0, let's not use the `with-bitcoin` feature of `hdpath` and instead inline the `From<StandardHDPath> for DerivationPath` instance in our source code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -97,9 +97,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.56"
+version = "1.0.57"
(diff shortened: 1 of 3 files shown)


hu55a1n1 (approved):
🙏

Issue #2087, opened 2022-04-12
Hermes takes a very long time to clear high number of pending packets


ancazamfir:
## Summary
With thousands of pending packets hermes progress when clearing packets is very very slow. Some steps in a typical flow (e.g. `build_recv_packet_and_timeout_msgs`) are:
- determine the sequences of pending packets
- `query_txs()` for each sequence to find the packet
- generate the messages
- for each chunk of N msgs:
   - create the Tx
   - simulate the Tx 
   - broadcast_tx_sync
   
During testing with thousands of pending packets I noticed that a Tx simulation will fail pretty early. And the clearing process above starts from the beginning. With 5k pending packets on a local (fast) setup just the `query_txs()` takes 2-3 minutes. And it takes 3-5 mins in the end to clear a few hundred packets.
The typical failure for simulation is the account mismatch [error](https://github.com/cosmos/cosmos-sdk/issues/11597). And the recaching of the account sequence on that error causes the second try to fail again immediately. 
Note: ignoring that sequence mismatch and continuing with the default gas improves things significantly.

## Problem Definition
Clearing a high number of pending packets takes a very long time

## Proposal
- limit the number of events generating one operational data. For example for 5k pending packets we could generate 50 operational data elements. This way we minimize the work hermes does until the Tx-es go through or fail.
- probably also redesign the retry in `relay_pending_packets()`.
- (maybe) ignore the sequence mismatch in the Tx simulation. If there is a real mismatch the checkTx will catch that and we can recache there.

## Acceptance Criteria

> Copied from #2077. Please see that issue and comments therein for an extended discussion on acceptance criteria and requirements.

### Discussion
Hermes CLI for packet clearing should provide support for splitting the workload in multiple, smaller batches, to make this command more tractable. This can be done, for instance, with any of the following suggestions:

A. By adding a flag `clear packets ... --limit X` to specify that Hermes should limit itself to fetching only X amount of packets/acks (instead of all pages). The `limit` option should probably be mandatory.

B. Alternatively, we could handle splitting into smaller batches of X messages automatically within the implementation of `clear packets` and have a simpler interface.

The advantage of A. is that the command is a lot more interactive and will finish faster. The problem is that the operators will not know if there will still be packets left to relay, so they will need to repeatedly invoke `clear packets --limit X` until this method returns an empty result (signalling there was nothing to clear).

[...]

Issue #594, opened 2021-02-01
Cleanup relayer CLIs


ancazamfir:
## Crate
realyer-cli

## Summary of Bug
Some of the relayer CLIs have redundant (e.g. connection and channel IDs are not used in `open-init` CLIs) or unneeded parameters (e.g. for `open-ack` and `open-confirm` the client IDs can be obtained by querying the connections). 
All CLIs should be reviewed.

## Steps to Reproduce
Exercise the relayer CLIs

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

romac:
See https://github.com/informalsystems/ibc-rs/commit/9c98b898c2941cbc077c0358658098f95200852b for `tx raw` commands for connections.

romac:
Done for `tx raw chan-open-*`  commands as well

PR #2167, opened 2022-05-02
Support for incremental packet clearing

adizere:
Closes: #2087 

## Basic manual tests

This PR fixes an issue with Hermes taking a very long time, being unresponsive or being unable to clear packets in production. We cannot reproduce the production conditions, so the manual tests will be more rudimental.

### Setup

1. Use `gm` to create and start two networks
    - create a channel between them
    - `create channel ibc-0 -c ibc-1 --port-a transfer --port-b transfer --new-client-connection`
3. `gm hermes config` and `gm hermes keys`
4. Adjust the gas parameters to allow large transactions
    - ```diff
      - gas_price = { price = 0.001, denom = 'stake' }
      + gas_price = { price = 0.01, denom = 'stake' }
      - max_gas = 1000000
      + max_gas = 10000000```
    - do the above for both of your networks

### Test with this branch

1. Create 2000 ft token transfers
    - `hermes tx raw ft-transfer ibc-0 ibc-1 transfer channel-0 9999 -o 1000 -n 2000`
    - this should take 10~20 seconds or so depending on your setup

2. Clear the outstanding packets
    - `hermes clear packets ibc-1 transfer channel-0`
    - every so often, the output will show the progress report
> 2022-05-02T13:41:10.560540Z  INFO ThreadId(01) PacketRecvCmd{src_chain=ibc-1 src_port=transfer src_channel=channel-0 dst_chain=ibc-0}: built packet data for 50 events total=2010 left=1810

After all the event data is pulled from the chain, Hermes proceeds to submit the transactions one by one

> 2022-05-02T13:43:24.955983Z  INFO ThreadId(01) PacketRecvCmd{src_chain=ibc-1 src_port=transfer src_channel=channel-0 dst_chain=ibc-0}:relay{odata=2diAXcipxM ->Destination @1-113; len=50}: success
2022-05-02T13:43:24.956192Z  INFO ThreadId(01) PacketRecvCmd{src_chain=ibc-1 src_port=transfer src_channel=channel-0 dst_chain=ibc-0}: ready to fetch a scheduled op. data with batch of size 50 targeting Destination
2022-05-02T13:43:24.964239Z  INFO ThreadId(01) PacketRecvCmd{src_chain=ibc-1 src_port=transfer src_channel=channel-0 dst_chain=ibc-0}:relay{odata=-_50M9IdL2 ->Destination @1-113; len=50}: optimistically resolved consensus_state_height upper_bound=1-114 result=1-35
2022-05-02T13:43:24.984645Z  INFO ThreadId(01) PacketRecvCmd{src_chain=ibc-1 src_port=transfer src_channel=channel-0 dst_chain=ibc-0}:relay{odata=-_50M9IdL2 ->Destination @1-113; len=50}: assembled batch of 51 message(s)
2022-05-02T13:43:25.048935Z  INFO ThreadId(21) wait_for_block_commits: waiting for commit of tx hashes(s) 7750973471BC89F98D4C70876E15E79A908FEC4B37CA8DDD4E25DF44B6CD360C, 9DB59FD0337166E4EAA87CB2ED7F5BC6CE9B660E167A12D2F718A74AE613D252 id=ibc-0
2022-05-02T13:43:30.081510Z  INFO ThreadId(01) PacketRecvCmd{src_chain=ibc-1 src_port=transfer src_channel=channel-0 dst_chain=ibc-0}:relay{odata=-_50M9IdL2 ->Destination @1-113; len=50}: [Sync->ibc-0] result events: 

### Test with `master` branch

1. Create 2000 ft token transfers
    - `hermes tx raw ft-transfer ibc-0 ibc-1 transfer channel-0 9999 -o 1000 -n 2000`
    - this should take 10~20 seconds or so depending on your setup

2. 2. Clear the outstanding packets
    - `hermes clear packets ibc-1 transfer channel-0`
    - This time, the process will take a long time to build the packet data for _all_ 2000 packets in a single query
    - This will be reflected in a very long initial waiting time. On my setup this takes ~1 minute, but in production it can be on the order of tens of minutes (which can cause subsequent errors)
> 2022-05-02T13:48:46.040035Z  INFO ThreadId(01) using default configuration from '/Users/adi/.hermes/config.toml'
2022-05-02T13:49:32.861348Z  INFO ThreadId(01) PacketRecvCmd{src_chain=ibc-1 src_port=transfer src_channel=channel-0 dst_chain=ibc-0}: found unprocessed SendPacket events for [Sequence(2011), Sequence(2012), Sequence(2013), Sequence(2014), Sequence(2015), Sequence(2016), Sequence(2017), Sequence(2018), Sequence(2019), Sequence(2020)] (first 10 shown here; total=2000)

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


lyqingye:
hi @adizere 

I read your PR, and the main change is that you divide the build packet data into multiple batches, each with a size of 50. I use my setup for testing. 10000 packets are generated at one tx. It can be seen from the log that the average cost of building packet data for 50 events is 2.5s(that is take much time)
When Hermes successfully builds 10000 packet data and fails in the send TX process (account sequence mismatch), Hermes will rebuild the non relayed packets.

lyqingye:
I try to add events store to my  [branch](https://github.com/falcons-x/ibc-rs/commit/7fda39ac9b9422419344dab2beffe1b4cffffb62) , it clear 10000packet, only take 2min

lyqingye:
> Thanks for the feedback!
> 
> > When Hermes successfully builds 10000 packet data and fails in the send TX process (account sequence mismatch), Hermes will rebuild the non relayed packets again.
> 
> It's not clear if this is the expected behavior or it's unexpected (i.e., a problem). Could you clarify?
> 
> > I try to add events store(for improve performance query_txs) to my [branch](https://github.com/falcons-x/ibc-rs/commit/7fda39ac9b9422419344dab2beffe1b4cffffb62) , it clear 10000packet, only take 2min.
> > however, storing events takes a lot of space. It takes 20m for 10000 packets, including send packets and write ack events. so I wrote a command to clear the relayed events!
> 
> This is impressive! Do you use this approach in production? Also: how much space and time does it take exactly?
> 
> * Is it 20Megabytes for 10.000 packets?
> * How much time?

The process of clear packet is as follows:
query_unreceived_ packets -> build_packet_events -> genertate_msgs_from_events -> send_msgs
when a send_msgs failed , the whole process will restart. I don't know why the sequence mismatch error occurs.

Events store takes 20MB to store 10000 packets
It takes about 1 ~ 2s to query 10000 packets from store

relayer-cli/src/conclude.rs
```diff
@@ -136,7 +136,6 @@ impl fmt::Display for Result {
 /// which typically signals a success (UNIX process return code `0`) or an error (code `1`). An
 /// optional `result` can be added to an output.
 ///
-#[derive(Debug)]
```
seanchen1991, line 139: Should there be an explicit `impl fmt::Debug for Output` now that the `derive` has been removed?

relayer/src/chain/counterparty.rs
```diff
@@ -425,26 +426,18 @@ pub fn unreceived_acknowledgements_sequences(
...
-) -> Result<Vec<u64>, Error> {
-    let counterparty = channel.channel_end.counterparty();
-    let counterparty_channel_id = counterparty
-        .channel_id
-        .as_ref()
-        .ok_or_else(Error::missing_counterparty_channel_id)?;
-
-    let (commit_sequences, _) = commitments_on_chain(
+    path: &PathIdentifiers,
+) -> Result<(Vec<u64>, Height), Error> {
```
seanchen1991, line 430: Can we update the doc comment for this method to detail the new return type? It would also be helpful if the doc comment explained why the `Height` is returned in the `Ok` case.

relayer/src/chain/counterparty.rs
```diff
@@ -425,26 +426,18 @@ pub fn unreceived_acknowledgements_sequences(
...
+        &path.counterparty_channel_id,
     )?;
 
-    unreceived_packets_sequences(
-        chain,
-        &channel.port_id,
-        &channel.channel_id,
-        commit_sequences,
-    )
+    let sn =
```
seanchen1991, line 437: Can we make this a more explicit and descriptive variable name?

relayer/src/chain/counterparty.rs
```diff
@@ -505,30 +498,25 @@ pub fn acknowledgements_on_chain(
...
-    let counterparty = channel.channel_end.counterparty();
-    let counterparty_channel_id = counterparty
-        .channel_id
-        .as_ref()
-        .ok_or_else(Error::missing_counterparty_channel_id)?;
-
-    let (commitments_on_src, _) =
-        commitments_on_chain(chain, &channel.port_id, &channel.channel_id)?;
+    path: &PathIdentifiers,
+) -> Result<(Vec<u64>, Height), Error> {
```
seanchen1991, line 502: Can we similarly update the doc comment for this method to reflect the new return type?

relayer/src/foreign_client.rs
```diff
@@ -45,6 +45,13 @@ const MAX_MISBEHAVIOUR_CHECK_DURATION: Duration = Duration::from_secs(120);
 
 const MAX_RETRIES: usize = 5;
 
+/// Optional limit on the amount of results to fetch when querying
+/// the client consensus states.
+/// Rationale: The consensus state query is one of the most expensive
```
seanchen1991, line 50: What happens then when this optimistic query fails, i.e., 10 results turned out to not be enough?

relayer/src/foreign_client.rs
```diff
@@ -1067,17 +1061,32 @@ impl<DstChain: ChainHandle, SrcChain: ChainHandle> ForeignClient<DstChain, SrcCh
     /// Retrieves all consensus states for this client and sorts them in descending height
     /// order. If consensus states are not pruned on chain, then last consensus state is the one
     /// installed by the `CreateClient` operation.
-    fn consensus_states(&self) -> Result<Vec<AnyConsensusStateWithHeight>, ForeignClientError> {
+    ///
+    /// The optional `u64` can be used to limit the number of consensus states that
+    /// are requested.
+    fn consensus_states(
```
seanchen1991, line 1067: What do you think about renaming this method to reflect the fact that it now can fetch the provided number of consensus states? Perhaps `fetch_consensus_states_with_limit` or something like that?

seanchen1991 (changes requested):

relayer/src/chain/counterparty.rs
romac: Perhaps even introduce a struct in place of the tuple, and a newtype for the `u64` while we are at it?

relayer-cli/src/conclude.rs
adizere: Not sure we need the `Debug` impl here. I removed the `derive` since it was not being used for anything.

We can re-add the `derive(Debug)` later on if needed. But I would actually prefer to make `Output` work with `Display` instead of `Debug`. The reason is that with `Debug` the output is too verbose and it's difficult to parse the logs, so we're forced to override of `impl Debug for X` to make output more concise (like we do in this PR and in here https://github.com/informalsystems/ibc-rs/pull/1944). Going with `Display` instead of `Debug` will be better, but that would require some bigger refactoring (which we can do with #594 if we see the need).

lyqingye:
This is the scenario of relaying 2000 packets (using event store). In the generate-opdata phase, check whether the packets are already handled and why they are checked again in the refresh schedule, 

image

more question
spawn_packet_worker and spawn_packet_cmd_worker whether the role of worker is repeated because of spawn_packet_cmd_worker using sync to commit txs, at this time, the transaction has not been confirmed, spawn_packet_worker will get the lock and continue to perform refresh_schedule &execute_schedule .  whether this logic is repeated?

adizere:
> why they are checked again in the refresh schedule,

Currently, the "generate" and the "refresh" phases are completely decoupled. This means there may be several minutes that elapse between the two phases. For this reason, the "refresh" phase double-checks that something was not relayed (by another relayer instance) or the packet did not time-out (since it was generated). 

Does this make sense?

This is a place where we can optimize things and I can imagine we can skip the "refresh" in some cases. But we have not experimented with this idea yet.

relayer/src/foreign_client.rs
adizere: be4e626f

relayer/src/foreign_client.rs
adizere: 8999ff55

relayer/src/chain/counterparty.rs
adizere: 64a4531e

relayer/src/link/cli.rs
```diff
@@ -51,23 +58,48 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Link<ChainA, ChainB> {
...
+        // Relay
+        let mut results = vec![];
+        for events_chunk in query_packet_events_with(
+            &sequences,
+            src_response_height,
+            self.a_to_b.src_chain(),
+            &self.a_to_b.path_id,
+            query_send_packet_events,
+        ) {
+            let mut last_events = self.a_to_b.relay_from_events(events_chunk)?;
```
seanchen1991, line 94: Mentioned during a weekly ibc-rs call, but is it within the scope of this PR to attempt to async-ify the relaying of events here?

ancazamfir (changes requested):
I left some review comments and addressed them in 2775ed06ba694081c0608a2d96a17cb4930807b2 if you want to take a look. I tested quite a bit with this and looks good, encountering still issues with seq mismatch during simulation.

relayer/src/link/cli.rs
adizere: >Not sure. This PR is becoming way too complex. I would concentrate on the minimal fixes for the original issue without doing major rework.

+1

I would like to pursue async-ification when it's off the critical path of fixing bugs or adding important features (as in this PR).

relayer/src/foreign_client.rs
```diff
@@ -1067,17 +1061,32 @@ impl<DstChain: ChainHandle, SrcChain: ChainHandle> ForeignClient<DstChain, SrcCh
...
+    fn consensus_states_with_limit(
+        &self,
+        result_limit: Option<u64>,
+    ) -> Result<Vec<AnyConsensusStateWithHeight>, ForeignClientError> {
+        let pagination = if let Some(limit) = result_limit {
+            ibc_proto::cosmos::base::query::pagination::latest_limited(limit)
+        } else {
+            ibc_proto::cosmos::base::query::pagination::all()
+        };
+
```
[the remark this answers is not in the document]
adizere: Agree with doing the changes in a different PR. We'll later on redo the changes from this commit 0cede457316a5d31ab0c10181ebb193752b015a4

ancazamfir (approved):
lgtm, great stuff @adizere !!!
Tested all scenarios I could think of and things look good for this issue. There are still things to improve but not here.
I opened https://github.com/informalsystems/ibc-rs/issues/2211 for the redundant client updates that cause not only the warning message and more fees to pe paid by hermes.

Also there is the issue of fragmenting in batches even earlier, on packet commitment query as this might also take a long time (i couldn't find the issue for this but i believe there is one?)

relayer/src/chain/counterparty.rs
ancazamfir: > Can we update the doc comment for this method to detail the new return type? It would also be helpful if the doc comment explained why the `Height` is returned in the `Ok` case.

fixed

> Perhaps even introduce a struct in place of the tuple, and a newtype for the `u64` while we are at it?

Maybe we do it in a different PR?

adizere: Agree let's do it in a separate PR.
I started working on this but it wasn't as easy as it sounds. Two complications are that:

1. A struct already exists that could do the job, but that needs to be moved from `relayer-cli` to `relayer` crate and some adaptations are necessary
https://github.com/informalsystems/ibc-rs/blob/f7cbe1cfea7f4b7aeb1bf0e99a1791f46abb2e89/relayer-cli/src/commands/query/packet/acks.rs#L16-L19

2. the method `unreceived_packets` has multiple uses and we have adapt all callers if we change the API.

PR #2171, opened 2022-05-02
Fix guide workflow, take 2

romac:


.github/workflows/guide.yml
```diff
@@ -55,7 +55,7 @@ jobs:
       # Only deploy guide when releasing a new version of Hermes
       - name: Deploy to GitHub Pages
         uses: peaceiris/actions-gh-pages@v3
-        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v0.')
+        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v0')
```
mzabaluev, line 58: Still not v1-ready :grin:

adizere: hehe

adizere (approved):
Pre-approved.

.github/workflows/guide.yml
romac: Hopefully we'll have a better solution for the guide by then, ie. with versioning etc.

.github/workflows/guide.yml
```diff
@@ -5,7 +5,7 @@ on:
     branches:
       - master
     tags:
-      - v0.*
+      - v[0-9]+.*
```
mzabaluev, line 8: The dot in the regex matches any symbol, if you need to match literally the dot, you need to escape it?

.github/workflows/guide.yml
mzabaluev: Maybe if you rely on the workflow trigger being either push to `master` or to a `v[0-9]+\.*` tag, it's enough to discern here that the push was for a tag.

.github/workflows/guide.yml
mzabaluev:
```suggestion
      - v[0-9]+\.*
```

.github/workflows/guide.yml
romac: Good point!

.github/workflows/guide.yml
mzabaluev: Ah yes, I forgot the regex must match the whole version string, and furthermore my suggestion is broken.

mzabaluev (approved):

Issue #2112, opened 2022-04-21
Hermes telemetry requirements, take (2)


adizere:
## Summary

This is a follow-up to #1373.

We should add support for tracking the following metrics as part of Hermes operation:

| Done | Metric name                     | Description                                                                                                                   | Meter type    | 
 | ---- | -----------                     | -----------                                                                                                                   | ----------    | 
 |      | `wallet_balance`     | An indicator with how much balance (coins) there is left in each wallet key that Hermes is using. | Gauge       | 
 |      | `ws_events`     | Indicates how many events did Hermes receive via the websocket subscription, in total since starting up, per chain. | Counter       | 
 |      | `error_websocket_reconnect`     | Number of times Hermes had to reconnect to the WebSocket endpoint                                                             | Counter       | 
 |      | `tx_latency`     | Indicates the latency for all transactions submitted to a specific chain (i.e., difference between the moment when Hermes received an event until the corresponding transaction(s) were confirmed). | ValueRecorder       | 

## Acceptance Criteria

- [ ] Add support for the four metrics above
- [ ] Update the [guide](https://hermes.informal.systems/telemetry.html) to reflect the new metrics that Hermes exposes
____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

PR #2182, opened 2022-05-04
Add new metrics requested by operators

romac:
Closes: #2112

## Description

| Done | Metric name                     | Description                                                                                                                   | Meter type    | 
 |:----:| -----------                     | -----------                                                                                                                   | ----------    | 
 |       | `wallet_balance`     | An indicator with how much balance (coins) there is left in each wallet key that Hermes is using. | Gauge       | 
 |   ☑️  | `ws_events`     | Indicates how many IBC events did Hermes receive via the websocket subscription, in total since starting up, per chain. | Counter       | 
 |   ☑️  | `ws_reconnect`     | Number of times Hermes had to reconnect to the WebSocket endpoint                                                             | Counter       | 
 |       | `tx_latency`     | Indicates the latency for all transactions submitted to a specific chain (i.e., difference between the moment when Hermes received an event until the corresponding transaction(s) were confirmed). Requires `tx_confirmation = true`. | ValueRecorder       | 
 |    ☑️   | `msg_num`     | Indicates how many messages Hermes submitted to a specific chain. | Counter       | 

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/telemetry/src/state.rs
+++ b/telemetry/src/state.rs
@@ -113,6 +113,7 @@ impl TelemetryState {
         self.receive_packets.add(count, labels);
     }
 
+    /// Number of acknowledgment packets relayed, per channel
     pub fn ibc_acknowledgment_packets(
         &self,
         src_chain: &ChainId,
@@ -129,6 +130,7 @@ impl TelemetryState {
         self.acknowledgment_packets.add(count, labels);
     }
 
+    /// Number of timeout packets relayed, per channel
     pub fn ibc_timeout_packets(
         &self,
         src_chain: &ChainId,
@@ -145,6 +147,7 @@ impl TelemetryState {
         self.timeout_packets.add(count, labels);
     }
 
+    /// Number of queries emitted by the relayer, per chain and query type
     pub fn query(&self, chain_id: &ChainId, query_type: &'static str) {
         let labels = &[
             KeyValue::new("chain", chain_id.to_string()),
@@ -154,6 +157,7 @@ impl TelemetryState {
         self.queries.add(1, labels);
     }
 
+    /// Number of cache hits for queries emitted by the relayer, per chain and query type
     pub fn query_cache_hit(&self, chain_id: &ChainId, query_type: &'static str) {
         let labels = &[
             KeyValue::new("chain", chain_id.to_string()),
@@ -163,6 +167,7 @@ impl TelemetryState {
         self.query_cache_hits.add(1, labels);
     }
 
+    /// Number of time the relayer had to reconnect to the WebSocket endpoint, per chain
     pub fn ws_reconnect(&self, chain_id: &ChainId) {
         let labels = &[KeyValue::new("chain", chain_id.to_string())];
 
--- a/telemetry/src/state.rs
+++ b/telemetry/src/state.rs
@@ -56,6 +56,9 @@ pub struct TelemetryState {
 
     /// Number of cache hits for queries emitted by the relayer, per chain and query type
     query_cache_hits: Counter<u64>,
+
+    /// Number of time the relayer had to reconnect to the WebSocket endpoint, per chain
+    ws_reconnect: Counter<u64>,
 }
 
 impl TelemetryState {
@@ -156,6 +159,12 @@ impl TelemetryState {
 
         self.query_cache_hits.add(1, labels);
     }
+
+    pub fn ws_reconnect(&self, chain_id: &ChainId) {
+        let labels = &[KeyValue::new("chain", chain_id.to_string())];
+
+        self.ws_reconnect.add(1, labels);
+    }
 }
 
 impl Default for TelemetryState {
@@ -206,6 +215,11 @@ impl Default for TelemetryState {
             query_cache_hits: meter
                 .u64_counter("cache_hits")
                 .with_description("Number of cache hits for queries emitted by the relayer, per chain and query type")
+                .init(),
+
+            ws_reconnect: meter
+                .u64_counter("ws_reconnect")
+                .with_description("Number of time the relayer had to reconnect to the WebSocket endpoint, per chain")
                 .init(),
         }
     }
--- a/telemetry/src/state.rs
+++ b/telemetry/src/state.rs
@@ -62,6 +62,9 @@ pub struct TelemetryState {
 
     /// How many IBC events did Hermes receive via the WebSocket subscription, per chain
     ws_events: Counter<u64>,
+
+    /// How many messages Hermes submitted to the chain, per chain
+    msg_num: Counter<u64>,
 }
 
 impl TelemetryState {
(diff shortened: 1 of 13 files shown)


romac:
Not sure what's going on here with the `tonic`-related errors.

@mzabaluev any idea?

mzabaluev:
@romac When was `proto/src/prost/cosmos.bank.v1beta1.rs` generated?
It's best to merge from master, then regenerate the proto files with `./scripts/sync-protobuf.sh`, because prost-build has been updated.

romac:
That was it, thanks @mzabaluev!

telemetry/src/state.rs
```diff
@@ -56,6 +58,32 @@ pub struct TelemetryState {
...
+    ws_reconnect: Counter<u64>,
+
+    /// How many IBC events did Hermes receive via the WebSocket subscription, per chain
+    ws_events: Counter<u64>,
+
+    /// How many messages Hermes submitted to the chain, per chain
+    msg_num: Counter<u64>,
+
+    /// The balance in each wallet that Hermes is using, per wallet, denom and chain
+    wallet_balance: ValueRecorder<u64>,
```
lyqingye, line 72: the `ValueRecorder` using `HistogramAggregator` ,   so I don't think it applies to `wallet_balances`

romac: Good point, what metric type would you use instead?

romac: I guess we need a custom `AggregatorSelector` for the Prometheus exporter.

lyqingye: > I guess we need a custom `AggregatorSelector` for the Prometheus exporter.

agree!
https://docs.rs/opentelemetry/latest/src/opentelemetry/sdk/metrics/selectors/simple.rs.html#29-61

lyqingye: > Good point, what metric type would you use instead?

I am also try to using `Prometheus::gauge` for instead, but it is not so convenient. Label must be specified when creating.
code like that: 
```rust
// create 
let registry = prometheus::default_registry();
let opts = Opts::new("signer_balance", "Balance Of Signer")
                .const_label("chain_id", chain_id.to_string())
                .const_label("address", address);
let gauge = Gauge::with_opts(opts).unwrap();
registry.register(Box::new(gauge)).unwrap();

// use 
gauge.set(balance);
```

romac: Done https://github.com/informalsystems/ibc-rs/pull/2182/commits/800f73b79a086da95cf4ec293e774a6b0a28c0c3

seanchen1991 (approved):
Some types need documentation, but other than that, looks good to me 🙂

relayer/src/link/relay_path.rs
```diff
@@ -305,7 +306,11 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
 
     // Determines if the events received are relevant and should be processed.
     // Only events for a port/channel matching one of the channel ends should be processed.
```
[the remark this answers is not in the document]
romac: 2d3a34d2

adizere:
I just now had a chance to play with the new metrics. Awesome work!

I wanted to ask about these three values

> tx_latency_submitted_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="0.5"} 0
tx_latency_submitted_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="0.9"} 0
tx_latency_submitted_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="0.99"} 0

it seems like they're always `0` and not sure how to interpret them. Any thoughts @seanchen1991 or @romac ?

romac:
These are bucket counts, used to compute a histogram but the bucket bounds are wrong. We can just ignore them or adapt them once we have some idea of what the typical latency is in production.

https://opentelemetry.io/docs/reference/specification/metrics/datamodel/#histogram

adizere:
thanks for the doc!

I think we should either:
1. fill in with real numbers
    - I tried with 1s/10s/20s on a local setup and those values seem meaningful for local testnets
    - for production values, this is something we can get help from @mircea-c hopefully. but if we don't before the release I'm not sure we should release it as it is, as they are confusing

2. put the median, first quartile, third quartile, instead of bucket bounds
    - for bucket bounds, not sure what value do they add even if we have production data -> it would be better to have data is not specific to the performance of machines or block frequency because these vary wildly
    - quartiles don't need to be parametrized, so they will make sense regardless of relayer machine configuration or blockchain parameters

3. or keep only the sum and count (which can lead to computing the average hopefully in Grafana)

relayer/src/link/relay_path.rs
```diff
@@ -703,6 +711,18 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
 
         let msgs = odata.assemble_msgs(self)?;
 
+        telemetry!({
+            let (chain, counterparty, channel_id, port_id) = self.target_info(odata.target);
+
+            ibc_telemetry::global().tx_submitted(
```
adizere, line 717: It seems like we're counting submitted transactions here, but this can lead to miscounting in two scenarios:

1. when there is an account sequence mismatch, we may submit the same transaction twice, so we can run into situations where we end up with
> tx_latency_confirmed_count{chain="ibc-1",channel="channel-0",counterparty="ibc-0",port="transfer"} 4
> tx_latency_submitted_count{chain="ibc-1",channel="channel-0",counterparty="ibc-0",port="transfer"} 5

and there will be a permanent offset between "submitted" and "confirmed", which is confusing: it can mean that a transaction was submitted but never confirmed, indicating some transactions maybe leaked


2. in the runtime, we may split this set of `msgs` across multiple transactions, but we only record a single record in the `tx_latency_submitted_count` metric. so the relayer may be submitting 10 separate transactions but only record 1 value here. I'm not sure this is a problem, it depends on users expectations actually.

romac: I wonder if this really matter? We are not counting transactions here but just measuring their latency, whether they're submitted twice or eventually rejected. The count here is only useful for computing the mean imho.

adizere: This is a good point!

The question is if relayer operators use the `tx_latency_submitted_count` (resp. `tx_latency_submitted_count`) as a standalone value -- or do they use it solely for finding the latency mean? I will ask this Q in the next call with operators.


As a side note, while playing with, I think I found a way to fix the problem with a tentative solution here 45e3a94e

romac: The fix looks good, and makes the metric usable both ways so let's roll with that if someone can confirm it works. Won't have time to check myself unfortunately.

Issue #2180, opened 2022-05-04
Use of system clock time to determine client state


faddat:
## Summary of Bug

The osmo client on evmos is live, but hermes uses the time of the computer running it to determine frozen/expiry:

Client state:

```yaml
client_state:
  '@type': /ibc.lightclients.tendermint.v1.ClientState
  allow_update_after_expiry: true
  allow_update_after_misbehaviour: true
  chain_id: osmosis-1
  frozen_height:
    revision_height: "0"
    revision_number: "0"
  latest_height:
    revision_height: "3484087"
    revision_number: "1"
  max_clock_drift: 25s
  proof_specs:
  - inner_spec:
      child_order:
      - 0
      - 1
      child_size: 33
      empty_child: null
      hash: SHA256
      max_prefix_length: 12
      min_prefix_length: 4
    leaf_spec:
      hash: SHA256
      length: VAR_PROTO
      prefix: AA==
      prehash_key: NO_HASH
      prehash_value: SHA256
    max_depth: 0
    min_depth: 0
  - inner_spec:
      child_order:
      - 0
      - 1
      child_size: 32
      empty_child: null
      hash: SHA256
      max_prefix_length: 1
      min_prefix_length: 1
    leaf_spec:
      hash: SHA256
      length: VAR_PROTO
      prefix: AA==
      prehash_key: NO_HASH
      prehash_value: SHA256
    max_depth: 0
    min_depth: 0
  trust_level:
    denominator: "3"
    numerator: "1"
  trusting_period: 864000s
  unbonding_period: 1209600s
```

**when we try to relay**

```
  - Client: 07-tendermint-0
    * Connection: connection-0
      | State: OPEN
      | Counterparty state: OPEN
      + Channel: channel-0
        | Port: transfer
        | State: OPEN
        | Counterparty: channel-204
# Chain: osmosis-1
  - Client: 07-tendermint-1566
  - Client: 07-tendermint-1674
  - Client: 07-tendermint-1840
    * Connection: connection-1440
      | State: OPEN
      | Counterparty state: OPEN
      + Channel: channel-204
        | Port: transfer
        | State: OPEN
        | Counterparty: channel-0
  - Client: 07-tendermint-1924
  - Client: 07-tendermint-1925
```

**when we try to do a very important client update**

```
hermes tx raw update-client evmos_9001-2 07-tendermint-0
[...]

PR #2183, opened 2022-05-04
Avoid using local time in client update

adizere:
Closes: #2180 

## Description

Simple fix: use source network status instead of local time when computing client expiry status.

______

### PR author checklist:
- [ ] ~~Added tests: integration (for Hermes) or unit/mock tests (for modules).~~
- [X] Linked to GitHub issue.
- [] ~~Updated code comments and documentation (e.g., `docs/`).~~

### Reviewer checklist:

### DIFF
--- a/relayer/src/foreign_client.rs
+++ b/relayer/src/foreign_client.rs
@@ -642,8 +642,20 @@ impl<DstChain: ChainHandle, SrcChain: ChainHandle> ForeignClient<DstChain, SrcCh
             .consensus_state(client_state.latest_height())?
             .timestamp();
 
+        let current_src_network_time = self
+            .src_chain
+            .query_application_status()
+            .map_err(|e| {
+                ForeignClientError::client_refresh(
+                    self.id().clone(),
+                    "failed querying the application status of source chain".to_string(),
+                    e,
+                )
+            })?
+            .timestamp;
+
         // Compute the duration since the last update of this client
-        let elapsed = Timestamp::now().duration_since(&last_update_time);
+        let elapsed = current_src_network_time.duration_since(&last_update_time);
 
         if client_state.expired(elapsed.unwrap_or_default()) {
             return Err(ForeignClientError::expired_or_frozen(
--- a/.changelog/unreleased/bug-fixes/relayer/2180-client-expiry-time.md
+++ b/.changelog/unreleased/bug-fixes/relayer/2180-client-expiry-time.md
@@ -1,0 +1,2 @@
+- Fixed client expiry computation to avoid using local time.
+  ([#2180](https://github.com/informalsystems/ibc-rs/issues/2180))


faddat:
Thank you very much

relayer/src/foreign_client.rs
```diff
@@ -642,8 +642,20 @@ impl<DstChain: ChainHandle, SrcChain: ChainHandle> ForeignClient<DstChain, SrcCh
...
+                    self.id().clone(),
+                    "failed querying the application status of source chain".to_string(),
+                    e,
+                )
+            })?
+            .timestamp;
+
         // Compute the duration since the last update of this client
-        let elapsed = Timestamp::now().duration_since(&last_update_time);
+        let elapsed = current_src_network_time.duration_since(&last_update_time);
```
ancazamfir, line 658: I think this change is good. But are we also missing a similar check in `build_update_client_with_trusted()`, specifically for the trusted consensus state? Here we are checking for the latest but we don't check for the trusted we are going to use in the update. 

Maybe we should consider refactoring a bit `build_update_client_with_trusted()`.

ancazamfir:
For config validation and client creation with parameters, maybe we should add a check for `trusting_period > N * block_time`, not sure about the value of `N` but definitely `>=1`. 
For clients that hermes doesn't create it's not much we can do.

relayer/src/foreign_client.rs
adizere: Let me know what you think Anca!

84a6e23

relayer/src/foreign_client.rs
```diff
@@ -875,6 +941,26 @@ impl<DstChain: ChainHandle, SrcChain: ChainHandle> ForeignClient<DstChain, SrcCh
             trusted_height
         };
 
+        // Check if the consensus state at `trusted_height` is within trusting period
+        if let ConsensusStateTrusted::NotTrusted {
+            elapsed,
+            consensus_state_timestmap,
+            network_timestamp,
+        } = self.check_consensus_state_trusting_period(&client_state, &trusted_height)?
```
ancazamfir, line 949: Maybe do this check only if `trusted_height != client_state.latest_height()`? This way for most of the cases we don't `query_application_status()` twice.

relayer/src/foreign_client.rs
```diff
@@ -875,6 +941,26 @@ impl<DstChain: ChainHandle, SrcChain: ChainHandle> ForeignClient<DstChain, SrcCh
...
+        {
+            warn!(
+                trusted_height = %trusted_height,
+                network_timestmap = %network_timestamp,
+                consensus_state_timestamp = %consensus_state_timestmap,
+                elapsed = ?elapsed,
+                "[{}] cannot build client update message because the provided trusted height is outside of trusting period!",
+                self);
```
ancazamfir, line 957: A bit of inconsistency here. If `trusted_height == client_state.latest_height()` we catch the consensus state expiration above in `self.validated_client_state()` but there is no `warn!` there i think. But if  `trusted_height != client_state.latest_height()` we issue the warning here.
Also why is this not `error!`?

relayer/src/foreign_client.rs
ancazamfir: Looks good, added a couple more comments.

relayer/src/foreign_client.rs
adizere: ac3dbc09

relayer/src/foreign_client.rs
adizere: Good catch!

ac3dbc09

seanchen1991 (approved):
:+1:

PR #2161, opened 2022-04-28
Update to prost 0.10 and tonic 0.7

mzabaluev:
## Description

Rolls up a bunch of dependency updates and code regeneration for the gRPC/protobuf stack.

______

### PR author checklist:
- [ ] Added changelog entry, using [`unclog`](https://github.com/informalsystems/unclog).
- [ ] Added tests: integration (for Hermes) or unit/mock tests (for modules). 
- [ ] Linked to GitHub issue.
- [ ] Updated code comments and documentation (e.g., `docs/`).

### Reviewer checklist:

- [ ] Reviewed `Files changed` in the GitHub PR explorer.
- [ ] Manually tested (in case integration/unit/mock tests are absent).

PR #2199, opened 2022-05-09
Generate ibc proto with buf

mzabaluev:
This is an attempt to switch to [buf](https://buf.build/) and [protoc-gen-prost](https://github.com/neoeinstein/protoc-gen-prost) for generating protobuf data structures in `ibc_proto`. As both cosmos-sdk and ibc-go already use buf, it's a way to piggyback on dependency management in the upstream `buf.yaml` files, rather than try to follow the intricacies of third-party vendoring locally.

Includes #2161.

## Unresolved problems

- [ ] Still manually specifying checkouts of cosmos-sdk and ibc-go.
- [ ] ibc.mock falls out because we can't make a dependency on ibc-go,
  as it is not published on BSR. https://github.com/cosmos/ibc-go/issues/1345
- [ ] Can't generate off an arbitrary commit on the proto source repositories,
  buf seems to only look on the default branch?
______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


mzabaluev:
Superseded by #2213. This can be resurrected later when upstream buf modules are available.

PR #1981, opened 2022-03-21
ICS29 Fee Middleware Support

soareschen:
## Description

This is a work in progress to support ICS29 fee middleware. This branch tracks the changes made in https://github.com/cosmos/ibc-go/pull/276, which is not yet merged.

Tasks:

- [x] #1968
- [x] #1964

______

### PR author checklist:
- [ ] Added changelog entry, using [`unclog`](https://github.com/informalsystems/unclog).
- [ ] Added tests: integration (for Hermes) or unit/mock tests (for modules). 
- [ ] Linked to GitHub issue.
- [ ] Updated code comments and documentation (e.g., `docs/`).

### Reviewer checklist:

- [ ] Reviewed `Files changed` in the GitHub PR explorer.
- [ ] Manually tested (in case integration/unit/mock tests are absent).

PR #2205, opened 2022-05-10
Use `send_tx` directly in `ChainDriver` for integration test

soareschen:
Closes: #2012
Prerequisite for: #1981

# Description

This is a follow up on #2044 to use `send_tx` directly in `ChainDriver` for integration test. This allows us to write tests that send arbitrary transactions with any wallet inside the integration test.

Prior to this, we have the limitation that transactions have to be submitted via the `gaiad` command, with quirks like #2012. Otherwise the `ChainHandle` design has the limitation of having a global wallet that is tied to the relayer, which makes it very tricky to submit transactions using multiple non-relayer wallets.

This is required for #1981 so that we can send atomic transaction that include `MsgPayPacketFee`, as well as submitting `MsgPayPacketFeeAsync` using the user wallet. Without this, it will be problematic to do this via `gaiad`, and it is unclear whether the `gaiad` CLI currently supports IBC transfer with packet fee. This PR makes it straightforward for us to submit such transactions directly, as long as there is a protobuf definition that can convert the domain type into `prost::Any`.

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


tools/integration-test/src/tests/client_expiration.rs
```diff
@@ -298,55 +309,14 @@ impl BinaryChainTest for PacketExpirationTest {
                 &denom_a,
             )?;
 
-            {
-                info!("sending first IBC transfer after client is expired. this should cause packet worker to fail");
-
-                chains.node_a.chain_driver().transfer_token(
```
soareschen, line 304: The tests here has to be removed, because since Gaia v7 the transaction will fail. Apparently we didn't catch this error previously, because submitting the transaction via the `gaiad` command do not result in error.

romac (approved):

PR #2213, opened 2022-05-13
Update to cosmos-sdk 0.46 and tendermint 0.35

### LINKED ISSUE #1560: Investigate new event structure of tendermint-go v0.35
## Context

This issue is about a certain breaking change in one of our dependencies.
For a full picture of Hermes dependencies, refer to https://github.com/informalsystems/ibc-rs/discussions/2030.

## Problem Definition

Tendermint-go has changed the structure of the events they emit through the websocket.
Ref: https://github.com/tendermint/tendermint/pull/6634

This feature is released in the tendermint-go 0.35.x line ([changelog](https://github.com/tendermint/tendermint/blob/v0.35.0/CHANGELOG.md#breaking-changes)), which will appear in SDK v0.46.x line ([milestone](https://github.com/cosmos/cosmos-sdk/milestone/46)). This probably means ~~late Q1 2022~~ late Q2 2022.

Further context on the upstream changes, thanks to @creachadair:

>  The Event and EventAttribute structures that Tendermint uses are defined as protobuf messages, e.g., [1]. Those messages can be encoded either in the protobuf wire format (binary) or JSON (text). The binary encoding does _not_ change as a result of the schema change, but the text encoding _does_ change.
[1]: https://github.com/tendermint/spec/blob/master/proto/tendermint/abci/types.proto#L276

> .. if the IBC code uses the websocket, then yes, it will get events in JSON representation. And in that case, it will be affected by the schema change.

More context: the `events` field structure changes from a hash-map to a vector:
https://github.com/informalsystems/tendermint-rs/blob/master/rpc/src/event.rs#L21

## Proposal 

We need to run Hermes against a chain with SDK 0.46 (once that is released) and make sure any breaking changes in the event system of tendermint-go do not affect Hermes. If those changes break event parsing in Hermes, we have to adapt our code to the upstream changes.

Currently, marking this issue as blocked until an SDK 0.46 appears.

## Acceptance Criteria

- [ ] test Hermes against chains built on SDK 0.46.x
- [ ] upgrade Hermes event parsing as necessary to maintain compatibility with tm-go 0.35
- [X] ~~ensure backwards compatibility with SDK pre-0.46.x~~
  - ~~AZ: This should be fine: we should be backwards compat.~~

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

mzabaluev:
Resolves #1560
See also #2276

## Description

Generate `ibc_proto` from a cosmos-sdk 0.46 pre-release revision, to evaluate API changes.
ibc-rs is also checked out from an `update-pr-branch` WIP branch.

Also updates tendermint-rs to 0.24.0-pre.2.

To be able to do this, ibc-proto-compiler is changed to parse the `buf.lock` file in cosmos-sdk and fetch the dependency modules listed there using the `buf export` command. The modules' export directories are added to the include path.

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


mzabaluev:
File under "the things we have to do before https://github.com/cosmos/ibc-go/issues/1345 is resolved".

relayer/src/event/rpc.rs
```diff
@@ -178,78 +175,12 @@ pub fn get_all_events(
...
-        extract_events(height, block_events, "channel_close_init", "channel_id"),
-        height,
-    );
-    append_events::<ChannelEvents::CloseConfirm>(
-        &mut events,
-        extract_events(height, block_events, "channel_close_confirm", "channel_id"),
-        height,
-    );
+    for abci_ev in block_events {
+        if let Some(ev) = channel_events::try_from_tx(abci_ev) {
```
mzabaluev, line 181: Note that this can in theory extract any channel events, not only the ones that the previous code extracted specifically by matching a key name. So a `WriteAcknowlegement`, `AcknowledgePacket`, or a `TimeoutPacket` could get into the returned list where previous code would ignore these events. I don't know if it is a problem in practice.

relayer/src/event/rpc.rs
```diff
@@ -141,31 +138,31 @@ pub fn get_all_events(
...
                         if matches!(chan_event, IbcEvent::SendPacket(_)) {
+                            // We are interested in the tag "hash" of an event with type "tx",
+                            // but this is trace level, so print the whole list of events.
                             // Should be the same as the hash of tx_result.tx?
-                            if let Some(hash) =
-                                events.get("tx.hash").and_then(|values| values.get(0))
-                            {
-                                tracing::trace!(event = "SendPacket", "tx hash: {}", hash);
-                            }
+                            // TODO: use a helper function to filter to just the interesting
```
adizere, line 163: Let's keep the tx hash trace if possible.

ALso update the comments for this method.

mzabaluev:
The integration test failures with gaia-rho (currently merged in the master branch) are due to a discrepancy with JSON serialization: tendermint-go has "omitempty" flags [on fields of](https://github.com/tendermint/tendermint/blob/8682489551b69d6b31947c9253c8c2f86fe4f2c7/abci/types/types.pb.go#L1543) `ResponseInfo`, while the Rust counterpart in tendermint-proto does not have a corresponding `serde(default)` attribute. Somehow the value of the `app_version` field supplied by the RPC server is considered empty and omitted.

modules/Cargo.toml
```diff
@@ -45,19 +45,23 @@ num-traits = { version = "0.2.15", default-features = false }
...
-version = "=0.23.7"
+version = "=0.24.0-pre.2"
 default-features = false
 
 [dependencies.tendermint-proto]
-version = "=0.23.7"
+version = "=0.24.0-pre.2"
+default-features = false
+
+[dependencies.tendermint-rpc]
```
mzabaluev, line 55: Here's another gotcha: `ibc` cannot begin to depend on `tendermint-rpc` because it must be no_std compatible...

modules/src/core/ics02_client/events.rs
```diff
@@ -1,8 +1,8 @@
 //! Types for the IBC events emitted from Tendermint Websocket by the client module.
 
 use serde_derive::{Deserialize, Serialize};
-use tendermint::abci::tag::Tag;
-use tendermint::abci::Event as AbciEvent;
+use tendermint_rpc::abci::tag::Tag;
+use tendermint_rpc::abci::Event as AbciEvent;
```
mzabaluev, line 5: ...but the RPC ABCI event types have migrated to `tendermint-rpc` in 0.24.

Either this change can be reverted in tendermint-rs, or we should move the event parsing infrastructure to the `relayer` crate, or rework the whole subscription mechanism for 0.24.

adizere:
Can we update the
- name
- description
- issue

attached to this PR? Would be great to make it clear to external parties what the work here is for.

As I understand it, the present PR provides compatibility with SDK 0.46 (but is not compat. with pre-0.46).

mzabaluev:
Closing this due to changing situation with cosmos-sdk dependencies. Some of the changes will be repurposed later.

Issue #2168, opened 2022-05-02
Unable to create a new connection in the presence of concurrent relayers


arirubinstein:
## Summary of Bug
On a network with other relayers running, it's not possible to create a new channel as hermes fails to create the connection since another relayer opens it `connection state is not INIT or TRYOPEN (got STATE_OPEN)`.

full logs: https://gist.github.com/arirubinstein/a37aa0441617e539546d5fc77dbcef0b

## Version
0.13.0

## Steps to Reproduce

1. Run a hermes relayer on a network
2. on another hermes instance, attempt to create a new channel with the same config

## Acceptance Criteria

Hermes should be able to reliably create new channels and be tolerant of other relayers changing state

- [ ] `create connection` CLI should be able to tolerate cases when the connection handshake is finished by other relayer processes; in this case, the CLIs should return success instead of retrying to resubmit useless transactions and fail
- [ ] integration test
- [ ] #2254
____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2224, opened 2022-05-19
Fix connection handshake retry bugs

ancazamfir:
Closes: #2168

## Description
- [x] changed the retry mechanism to constant retry delay 
- [x] ~~added check for connection IDs at the two connection ends (mismatch happens due to crossing handshake messages causing proof failures), exit retry loop on this check failing~~
- [x] recover from mismatch ID case by updating the `b_side` connection ID and continue retry loop instead of erring
- [x] ~~make the similar changes in channel handshake if needed~~ tracked separately in #2254 
- [X] ~~add integration tests~~
    - Adi: Looked a bit into it and unclear how we can do it. We need support to instantiate two relayer instances (a supervisor doing `hermes start` plus an instance doing `hermes create connection`).

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


ancazamfir:
Same problems exist at channel level. Let's review the connection solution and I will make the similar changes in the channel handshake.

relayer/src/connection.rs
```diff
@@ -500,114 +410,231 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Connection<ChainA, ChainB> {
...
-                    self.a_side.connection_id = Some(extract_connection_id(&result)?.clone());
-                    println!("🥂  {} => {:#?}\n", self.a_side.chain.id(), result);
-                    break;
-                }
-            }
+    /// Queries the chains for latest connection end information. It verifies the relayer connection
+    /// IDs and updates them if needed.
+    /// Returns the states of the two connection ends.
+    ///
+    /// The relayer connection stores the connection identifiers on the two chains a and b.
```
seanchen1991, line 417: Subtle question here: are the connection identifiers stored _on_ the chains A and B, or are they stored on the relayer and _pertain_ to chains A and B?

I ask because the next sentence seems to indicate that there are corresponding on-chain identifiers, which makes me think that the latter statement is correct.

seanchen1991 (commented):
I appreciate the effort it must have taken to diagram the `update_connection_and_query_states` function, however I find it confusing. I'm not easily able to match up what is said in the descriptions with what the diagram is showing. Hopefully that can be cleared up.

relayer/src/connection.rs
ancazamfir: Not sure I understand the question. The IDs of all IBC objects (clients, connections, channels) are assigned by the on-chain handlers at creation time and stored on-chain. The relayer gets them from the response to the message that created them. In the connection handshake case, the relayer saves them in the `Connection` object under `side_a` or `side_b`.

relayer/src/connection.rs
```diff
@@ -500,114 +410,231 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Connection<ChainA, ChainB> {
...
+    ///  | <-- Init (r1)                                        |
+    ///  | a_id = 1                                             |
+    ///  |                                         Try (r2) --> |
+    ///  |                    b_id = 100, b_counterparty_id = 1 |
+    ///  |                                         Try (r1) --> |
+    ///  |                    b_id = 101, b_counterparty_id = 1 |
+    ///  | <-- Ack (r2)
+    ///  | a_id = 1, a_counterparty_id = 100
+    ///
+    /// Here relayer r1 has a_side connection 1 and b_side connection 101
```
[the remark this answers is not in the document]
ancazamfir: Ah right `a_id` and `b_id` were not shown in the first diagram which tries to explain the code. Just added them, let me know if it's more clear.

relayer/src/connection.rs
```diff
@@ -500,114 +410,231 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Connection<ChainA, ChainB> {
...
+    /// while on chain a the counterparty of connection 1 is 100. r1 needs to update
+    /// its b_side to 100
+    fn update_connection_and_query_states(&mut self) -> Result<(State, State), ConnectionError> {
+        let relayer_a_id = self.a_side.connection_id();
+        let relayer_b_id = self.b_side.connection_id().cloned();
+
+        let a_connection = self.a_connection(relayer_a_id)?;
+        let a_counterparty_id = a_connection.counterparty().connection_id();
+
+        if a_counterparty_id.is_some() && a_counterparty_id != relayer_b_id.as_ref() {
```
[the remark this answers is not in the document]
ancazamfir: It means they are not yet assigned/ known which is the case in `Init` state, i.e. before `OpenTry` or `OpenAck` steps complete. Added in the Case description.

relayer/src/connection.rs
seanchen1991: Ok, that makes sense :+1:

relayer/src/connection.rs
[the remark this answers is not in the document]
seanchen1991: Thanks! The diagram makes more sense now 🙂

relayer/src/connection.rs
```diff
@@ -500,114 +410,231 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Connection<ChainA, ChainB> {
...
+        .map_err(|err| {
+            error!(
+                "failed to open connection after {} retries",
+                retry_count(&err)
+            );
+            from_retry_error(
+                err,
+                format!("failed to finish connection handshake for {:?}", self),
+            )
+        })?;
```
seanchen1991, line 635: Does this `map_err` call get triggered upon every retry, or only once the maximum number of retries has been reached?

relayer/src/channel.rs
```diff
@@ -477,7 +477,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
 
             let dst_channel_id = channel
                 .dst_channel_id()
-                .ok_or_else(ChannelError::missing_counterparty_connection)?;
+                .ok_or_else(ChannelError::missing_counterparty_channel_id)?;
```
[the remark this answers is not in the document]
adizere: > Was returning a missing_counterparty_connection error a mistake?

Yes. It seems like this was a typo. We were trying to fetch a channel identifier, while the error type represents a "missing connection".

relayer/src/connection.rs
```diff
@@ -500,114 +410,231 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Connection<ChainA, ChainB> {
...
         }
+        Err(ConnectionError::handshake_finalize())
+    }
 
-        Err(ConnectionError::max_retry())
+    /// Executes the connection handshake protocol (ICS003)
+    fn handshake(&mut self) -> Result<(), ConnectionError> {
+        let max_block_times = self.max_block_times()?;
+        // retry every 1/10th of block time
+        let retry_delay = max_block_times / 10;
```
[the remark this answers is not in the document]
ancazamfir: tried in [b8fbfc6](https://github.com/informalsystems/ibc-rs/pull/2224/commits/b8fbfc695a2fd53c7c633ad713953228917d8df6)

relayer/src/connection.rs
```diff
@@ -33,16 +32,45 @@ pub use error::ConnectionError;
...
-pub fn from_retry_error(e: retry::Error<ConnectionError>, description: String) -> ConnectionError {
-    match e {
-        retry::Error::Operation {
-            error,
-            total_delay,
-            tries,
-        } => {
-            ConnectionError::max_retry(description, tries, total_delay, Box::new(error.0))
+// Includes utility methods and constants to configure the retry behavior
+// for the channel handshake algorithm.
```
seanchen1991, line 36: I think the way to properly document this sub-module would be to move these lines inside and at the top of the module, as well as change them to `//!` doc strings.

relayer/src/connection.rs
```diff
@@ -33,16 +32,45 @@ pub use error::ConnectionError;
...
+mod handshake_retry {
+    use crate::connection::ConnectionError;
+    use crate::util::retry::{clamp_total, ConstantGrowth};
+    use core::time::Duration;
+
+    // Approximate number of retries per block
```
seanchen1991, line 42: Let's change the comments on these constants to doc comments.

relayer/src/connection.rs
```diff
@@ -33,16 +32,45 @@ pub use error::ConnectionError;
...
+    const PER_BLOCK_RETRIES: u32 = 10;
+    // Keep the retry delay constant
+    const DELAY_INCREMENT: u64 = 0;
+    // Maximum retry delay expressed in number of blocks
+    const BLOCK_NUMBER_DELAY: u32 = 10;
+
+    // The default retry strategy.
+    // We retry with a constant backoff strategy. The strategy is parametrized by the
+    // maximum block time expressed as a `Duration`.
```
seanchen1991, line 51: Let's also change this to a doc comment.

relayer/src/connection.rs
```diff
@@ -33,16 +32,45 @@ pub use error::ConnectionError;
...
+        let retry_delay = max_block_times / PER_BLOCK_RETRIES;
+
+        clamp_total(
+            ConstantGrowth::new(retry_delay, Duration::from_secs(DELAY_INCREMENT)),
+            retry_delay,
+            max_block_times * BLOCK_NUMBER_DELAY,
+        )
+    }
+
+    pub fn from_retry_error(
```
seanchen1991, line 62: Could we also document this function? I'm not clear on what it's doing.

adizere (approved):
Tested with multiple scenarios, looks good. Great work Anca!

relayer/src/connection.rs
adizere: Got it! 9191e2ee

adizere: By the way, I hesitated originally to make doc comments because `mod handshake_retry` is supposed to be a private module. I guess even private module should do `//!` doc string then, right?

seanchen1991: I don't believe adding doc strings makes the module public. Documentation is still super useful for internal developers 😄

relayer/src/connection.rs
adizere: 7dd96ac4

relayer/src/connection.rs
adizere: 5aba4aa6

seanchen1991: Does the retry mechanism throw an error every time it retries? Or perhaps only when the maximum number of retries is reached?

adizere:
Blocked on input from Soares and/or Anca on the legitimacy of the refactoring I did in https://github.com/informalsystems/ibc-rs/pull/2224/commits/d34399163b88ca8c829c56ae60c64d68170d58ea. If we get the green there, I think we can merge this.

relayer/src/connection.rs
```diff
@@ -40,13 +40,7 @@ pub fn from_retry_error(e: retry::Error<ConnectionError>, description: String) -
...
             tries,
         } => {
-            let detail = error::ConnectionErrorDetail::MaxRetry(error::MaxRetrySubdetail {
-                description,
-                tries,
-                total_delay,
-                source: Box::new(error.0),
-            });
-            ConnectionError(detail, error.1)
+            ConnectionError::max_retry(description, tries, total_delay, Box::new(error.0))
```
mzabaluev: This is how it should be!

ancazamfir:
> Blocked on input from Soares and/or Anca on the legitimacy of the refactoring I did in [d343991](https://github.com/informalsystems/ibc-rs/commit/d34399163b88ca8c829c56ae60c64d68170d58ea). If we get the green there, I think we can merge this.

Looks good to me!

Issue #2223, opened 2022-05-18
Consolidate `ChainEndpoint::proven_*()` functions with their analogous `query_*()` version


plafer:
I just noticed that the `ChainEndpoint::proven_*()` methods each have an analogous `query_*()` method which performs the same query except that they don't return a proof.

I suggest we remove the `proven_*()` methods from `ChainEndpoint` and `ChainHandle`, and change the provable `query_*()` functions as follows:
```rust
// previous
fn query_channel(&self, request: QueryChannelRequest) -> Result<ChannelEnd, Error>;
fn proven_channel(&self, request: QueryChannelRequest) -> Result<(ChannelEnd, MerkleProof), Error>;

// new
fn query_channel(&self, request: QueryChannelRequest, include_proof: PerformQuery) -> Result<(ChannelEnd, Option<MerkleProof>), Error>;

enum PerformQuery {
    WithProof,
    WithoutProof,
};
```
I introduced `PerformQuery` to avoid [boolean blindness](http://sboosali.github.io/mirror/boolean-blindness.html).

Example calls:
```rust
chain.query_channel(request, PerformQuery::WithProof);
chain.query_channel(request, PerformQuery::WithoutProof);
```
An alternative name for `PerformQuery` I thought of:
```rust
enum IncludeProof {
    Yes,
    No,
};
chain.query_channel(request, IncludeProof::Yes);
chain.query_channel(request, IncludeProof::No);
```

I personally prefer merging the two functions into one, as we expose fewer methods while preserving the same functionality.

_Originally posted by @plafer in https://github.com/informalsystems/ibc-rs/pull/2200#issuecomment-1129177575_

PR #2226, opened 2022-05-19
Consolidate ChainEndpoint/Handle proven queries

plafer:
Closes: #2223 

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/relayer/src/chain/mock.rs
+++ b/relayer/src/chain/mock.rs
@@ -43,7 +43,7 @@ use crate::light_client::Verified;
 use crate::light_client::{mock::LightClient as MockLightClient, LightClient};
 
 use super::requests::{
-    QueryChannelsRequest, QueryClientConnectionsRequest, QueryClientStateRequest,
+    IncludeProof, QueryChannelsRequest, QueryClientConnectionsRequest, QueryClientStateRequest,
     QueryConnectionChannelsRequest, QueryConnectionRequest, QueryConnectionsRequest,
     QueryConsensusStateRequest, QueryConsensusStatesRequest, QueryHostConsensusStateRequest,
     QueryNextSequenceReceiveRequest, QueryPacketAcknowledgementsRequest,
@@ -190,14 +190,15 @@ impl ChainEndpoint for MockChain {
     fn query_client_state(
         &self,
         request: QueryClientStateRequest,
-    ) -> Result<AnyClientState, Error> {
+        _include_proof: IncludeProof,
+    ) -> Result<(AnyClientState, Option<MerkleProof>), Error> {
         // TODO: unclear what are the scenarios where we need to take height into account.
         let client_state = self
             .context
             .query_client_full_state(&request.client_id)
             .ok_or_else(Error::empty_response_value)?;
 
-        Ok(client_state)
+        Ok((client_state, None))
     }
 
     fn query_upgraded_client_state(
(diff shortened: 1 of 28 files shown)


relayer/src/chain/mock.rs
```diff
@@ -426,13 +417,20 @@ impl ChainEndpoint for MockChain {
...
+        assert!(matches!(include_proof, IncludeProof::No));
+
         let consensus_states = self.context.consensus_states(&request.client_id);
-        Ok(consensus_states
-            .into_iter()
-            .find(|s| s.height == request.consensus_height)
-            .unwrap()
-            .consensus_state)
+        Ok((
+            consensus_states
```
romac, line 427: Can we pull this expression out into a variable to improve readability?

relayer/src/chain/mock.rs
```diff
@@ -426,13 +417,20 @@ impl ChainEndpoint for MockChain {
...
+                .into_iter()
+                .find(|s| s.height == request.consensus_height)
+                .unwrap()
```
romac, line 430: While we are at it, can we either deal with the potential error here or at least document why this is safe?

romac (approved):
Preemptively approved, thanks a lot for going through this @plafer!

It's a bit unfortunate that we now cannot statically rely on the proof being present in the returned type when `IncludeProof` is `Yes` but we can perhaps find a way to recover this statically in a follow-up PR, perhaps with something like:

```rust
trait WithProof<Proof> {
    type Proof;

    fn include_proof(&self) -> bool;

    fn build_proof<E>(&self, build: impl FnOnce() -> Result<Proof, E>) -> Result<Self::Proof, E>;
}

struct IncludeProof;
struct NoProof;

impl<Proof> WithProof<Proof> for IncludeProof {
    type Proof = Proof;

    fn include_proof(&self) -> bool { true }

    fn build_proof<E>(&self, build: impl FnOnce() -> Result<Proof, E>) -> Result<Self::Proof, E> {
        build()
    }
}


impl<Proof> WithProof<Proof> for NoProof {
      type Proof = ();

      fn include_proof(&self) -> bool { false }

      fn build_proof<E>(&self, _build: impl FnOnce() -> Result<Proof, E>) -> Result<Self::Proof, E> {
          Ok(())
      }
}

impl ChainEndpoint for CosmosSdkChain {
   fn query_channel<P: WithProof<MerkleProof>>(
         &self,
         request: QueryChannelRequest,
         with_proof: P,
     ) -> Result<(ChannelEnd, P::Proof), Error> {
       // ...

         let res = self.query(
             ChannelEndsPath(request.port_id, request.channel_id),
             request.height,
             with_proof.include_proof(),
         )?;

         let channel_end = ChannelEnd::decode_vec(&res.value).map_err(Error::decode)?;

         let proof = with_proof.build_proof(|| res.proof.ok_or_else(Error::empty_response_proof))?;

        Ok(channel_end, proof)
     }
}
```


Not clear how that would interact with the chain runtime, but we might be able to solve this in one go once we move the queries out of the runtime and expose them directly from the `ChainHandle`.

@plafer @soareschen What do you think?

[Rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a40486e0b097347e5b5e1dc63e5a116f)

relayer/src/chain/mock.rs
plafer: [54c8127](https://github.com/informalsystems/ibc-rs/pull/2226/commits/54c8127586d33156b302d979a68984046999031a)

PR #2227, opened 2022-05-20
Working to support Gaia Rho

soareschen:
## Description

This is a work in progress to track support for [gaia/rho-ibc](https://github.com/cosmos/gaia/tree/rho-ibc), with the `gaiad` binary provided from https://github.com/informalsystems/cosmos.nix/pull/86.

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/relayer/src/chain/cosmos/wait.rs
+++ b/relayer/src/chain/cosmos/wait.rs
@@ -1,20 +1,20 @@
 use core::time::Duration;
 use ibc::core::ics24_host::identifier::ChainId;
 use ibc::events::IbcEvent;
-use ibc::query::{QueryTxHash, QueryTxRequest};
-use itertools::Itertools;
-use std::thread;
 use std::time::Instant;
+use tendermint::abci::transaction::Hash as TxHash;
+use tendermint_rpc::endpoint::tx_search::Response as TxSearchResponse;
 use tendermint_rpc::{HttpClient, Url};
-use tracing::{info, trace};
+use tokio::task;
+use tokio::time::sleep;
+use tracing::info;
 
-use crate::chain::cosmos::query::tx::query_txs;
-use crate::chain::cosmos::types::tx::TxSyncResult;
+use crate::chain::cosmos::query::tx::{all_ibc_events_from_tx_search_response, query_tx_response};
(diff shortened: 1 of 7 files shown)


relayer/src/chain/cosmos/wait.rs
```diff
@@ -1,114 +1,117 @@
...
+        });
+        join_handles.push(handle);
+    }
+
+    let mut responses = Vec::new();
+
+    for handle in join_handles.into_iter() {
+        let response = handle.await.map_err(Error::join)??;
+        responses.push(response);
+    }
```
mzabaluev, line 72: It's more efficient to use `FuturesOrdered`.

soareschen: Thanks for the suggestions! Loops are easier to write during quick hacks, though I have found the time to refactor them to use iterators and streams.

relayer/src/chain/cosmos/wait.rs
```diff
@@ -50,26 +50,16 @@ pub async fn wait_tx_hashes(
...
+    let tasks = tx_hashes
+        .iter()
+        .map(|tx_hash| async { wait_tx_hash(rpc_client, rpc_address, timeout, tx_hash).await })
+        .collect::<FuturesOrdered<_>>();
+
+    let responses = tasks
+        .collect::<Vec<_>>()
+        .await
+        .into_iter()
+        .collect::<Result<Vec<_>, _>>()?;
```
mzabaluev, line 62: A bit hard to follow the logic here. The first `.collect` makes sure all futures are executed, whereas `.try_collect` would bail out after one of them resolved an error. Then, the resolved vector of `Result` from all tasks is transformed to determine if the overall result is a success, anyway, and propagate the first error found. Would it be simpler to just use `.try_collect` here?

mzabaluev (commented):
Good use of future combinators!

relayer/src/chain/cosmos/wait.rs
soareschen: You are right, I managed to simplify it further!

PR #2238, opened 2022-05-24
Add test to reproduce clear packet failure

soareschen:
Closes: #XXX

## Description

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/tools/integration-test/src/tests/clear_packet.rs
+++ b/tools/integration-test/src/tests/clear_packet.rs
@@ -6,7 +6,20 @@ use ibc_test_framework::util::random::random_u64_range;
 fn test_clear_packet() -> Result<(), Error> {
     run_binary_channel_test(&ClearPacketTest)
 }
+
+#[test]
+fn test_clear_packet_recovery_dst() -> Result<(), Error> {
+    run_binary_channel_test(&ClearPacketRecoveryDstTest)
+}
+
+#[test]
+fn test_clear_packet_recovery_src() -> Result<(), Error> {
+    run_binary_channel_test(&ClearPacketRecoverySrcTest)
+}
+
 pub struct ClearPacketTest;
+pub struct ClearPacketRecoveryDstTest;
+pub struct ClearPacketRecoverySrcTest;
 
 impl TestOverrides for ClearPacketTest {
     fn modify_relayer_config(&self, config: &mut Config) {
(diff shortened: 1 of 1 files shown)


seanchen1991 (approved):
I'm a bit curious why you consider this a hacky fix. What would be an ideal fix for this issue?

soareschen:
> I'm a bit curious why you consider this a hacky fix. What would be an ideal fix for this issue?

Because the original code was already a hack, with mutation flags and complicated control flow. Adding yet another mutation flag further complicates the control flow and makes the code even harder to understand. A proper fix would require rewriting the functions and have simpler control flows.

relayer/src/worker/packet.rs
```diff
@@ -172,6 +175,7 @@ fn handle_packet_cmd<ChainA: ChainHandle, ChainB: ChainHandle>(
                 clear_interval,
                 height,
             ) {
+                scheduled_packet_clearing = true;
```
ancazamfir, line 178: Could we maybe instead remove the mutation of `is_first_run` from `should_clear_packets` (with this name it should never do that ;)) and instead only set it to `false` if `schedule_packet_clearing` is Ok below?

ancazamfir (dismissed):
@soareschen commented:
> Hmm unfortunately the original logic of should_clear_packets is pretty complicated, and it is possible for it to return false but also set is_first_run to false, or return true without mutating is_first_run. I extracted the original code out and made it into a function without changing the original logic, using the refactoring technique I have. We'd have to rewrite the logic to make it work without the mutation.


I looked a bit closer at this and I think we should take the opportunity and refactor, it is very messy indeed. Not sure why we split the `...schedule...` step away from the rest (`refresh_schedule` and `execute_schedule`).
For each `cmd` type we could do something like this (with appropriate schedule operation):
```
link.a_to_b.<schedule>
      .and_then(|_| link.a_to_b.refresh_schedule())
      .and_then(|_| link.a_to_b.execute_schedule()),
```
For the `WorkerCmd::NewBlock` we could just run this if `should_clear_packets()` is true and set `*is_fist_run = false` if all three steps are successful (i.e. remove it from `should_clear_packets()`). I made some quick [changes](https://github.com/informalsystems/ibc-rs/commit/7751a89608dcd1be1334ce031975a5e994574411) along these lines locally and tests are passing.
I am sure you will come with an even better solution.

soareschen:
@romac this is ready for review and merge.

relayer/src/worker/packet.rs
```diff
@@ -144,94 +108,98 @@ pub fn spawn_packet_cmd_worker<ChainA: ChainHandle, ChainB: ChainHandle>(
...
+
+    link.a_to_b
+        .execute_schedule()
+        .map_err(handle_link_error_in_task)
+        .map_err(|e| {
+            if let TaskError::Ignore(e) = &e {
+                error!("will retry: schedule execution encountered error: {}", e,);
+            }
+            e
+        })?;
```
seanchen1991, line 192: I'm not really a fan of the back-to-back `map_err` calls; they're a bit confusing to trace through. Could we do something like this instead?

```suggestion
link.a_to_b.execute_schedule().map_err(|e| {
  let task_error = handle_link_error_in_task(e);
  
  if let TaskError::Ignore(e) = &task_error {
    error!("will retry: schedule execution encountered error: {}", e,);
  }
  
  task_error
})?;
```

ancazamfir (approved):
Great refactor! Thanks @soareschen!

Issue #2154, opened 2022-04-27
Tutorial is outdated: `hermes create channel`


plafer:
## Summary of Bug
According to the [hermes tutorial](https://hermes.informal.systems/tutorials/local-chains/relay-paths/create-new-path.html), I should be able to run

```shell
hermes create channel ibc-0 ibc-1 --port-a transfer --port-b transfer
```

to create a client, connection and channel between chains `ibc-0` and `ibc-1`. This current fails with error

```shell
error: Invalid value "ibc-1" for '<CONNECTION_A>': identifier ibc-1 has invalid length 5 must be between 10-64 characters
```

## Version

v0.14.0

## Steps to Reproduce

1. Spawn two chains (I used basecoin chains)
2. Make sure your `hermes` config knows about both chains
3. Run `hermes create channel ibc-0 ibc-1 --port-a transfer --port-b transfer`
4. Notice the aforementioned error

## Acceptance Criteria

The `hermes create channel` command succeeds.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

romac:
The guide is outdated, the syntax for `create channel` is now

```
hermes create channel [OPTIONS] --port-a <PORT_A> --port-b <PORT_B> <CHAIN_A> [CONNECTION_A]
```

ie. the second positional argument is now the connection id.

I will update the guide later today, thanks for noticing!

PS: I am more and more convinced that we should eventually (for v1) move away from positional arguments and use flags everywhere.

adizere:
> PS: I am more and more convinced that we should eventually (for v1) move away from positional arguments and use flags everywhere.

++++1

romac:
Guide is now up to date: https://hermes.informal.systems/tutorials/local-chains/relay-paths/create-new-path.html

romac:
Output listed in the guide depends on https://github.com/informalsystems/ibc-rs/pull/2158

dylanschultzie:
@romac it appears the guide isn't updated still:
<img width="846" alt="Screen Shot 2022-05-26 at 1 54 42 PM" src="https://user-images.githubusercontent.com/9121234/170578573-9a995587-10ec-49bd-bd72-4b6b24dfc4ec.png">

adizere:
I was just about to approve #2245. Then I found this strange thing.

[...]

PR #2245, opened 2022-05-30
Fix wrong `create channel` command in "Create a new path" page in the guide

romac:
See https://github.com/informalsystems/ibc-rs/issues/2154#issuecomment-1139037920


adizere (approved):
Thanks Romain

Issue #2111, opened 2022-04-21
Add support for selecting specific wallet for packet clearing CLI


adizere:
## Summary

Add `-k <KEY>` option for `clear packets` CLI, just like Hermes provides one for `tx raw ft-transfer -k <KEY>`

## Problem Definition

Hermes operators typically run one `start` instance in the background and then sometimes require a second instance to do packet `clear`ing in parallel. This second instance should not be reusing the same wallet as the first, otherwise it would result in `account sequence mismatch` errors. Currently, it is not possible to start `clear packets` with a separate wallet.

## Acceptance Criteria

- [ ] Add support for using a specific <KEY> with option `-k` as part of `clear packets` CLI

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

seanchen1991:
I'm not sure what a `key` looks like. What's an example key I could use to test this flow?

ancazamfir:
I think we should have the `-k` for all CLIs. Is there a way to have a generic solution for all CLIs? i.e. overwrite the key in the chain config if -k flag is present.

PR #2252, opened 2022-05-31
Implement ability to select specific wallet for packet clearing CLI

seanchen1991:
Closes: #2111 

## Description

Add `--key <KEY>` and `--counterparty-key <KEY>` options for clear packets CLI, just like Hermes provides one for `tx ft-transfer -key <KEY>`.

______

### PR author checklist:
- [x] Linked to GitHub issue.
- [x] Updated code comments and documentation (e.g., `docs/`). Add a definition of what keys are in the "keys" chapter of the Hermes guide. 

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/features/ibc-telemetry/2175-pending-packet-tracking-metrics.md
+++ b/.changelog/unreleased/features/ibc-telemetry/2175-pending-packet-tracking-metrics.md
@@ -1,0 +1,4 @@
+- Added new metrics to track the number of relayed SendPacket and
+  WriteAcknowledgement, the sequence number and the timestamp of the
+  oldest pending SendPacket ([#2175](https://github.com/informalsystems/ibc-
+  rs/issues/2175))
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -739,6 +739,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "dashmap"
+version = "5.3.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3495912c9c1ccf2e18976439f4443f3fee0fd61f424ff99fde6a66b15ecb448f"
+dependencies = [
+ "cfg-if 1.0.0",
+ "hashbrown 0.12.1",
+ "lock_api",
+ "parking_lot_core 0.9.3",
+]
+
+[[package]]
 name = "deflate"
 version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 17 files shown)


seanchen1991:
It looks like as part of [ADR 10](https://github.com/informalsystems/ibc-rs/pull/2306) it was decided that CLI commands that require being able to specify parameters on both the source and destination chains should do so using `-host` and `-reference` suffixes. 

@adizere pointed out that the original design of this feature comes up a bit short because the `clear packet` flow sends messages to both a source and destination chain; specifying a single key via the flag doesn't make it clear which chain the key is associated with. 

The most straightforward way to address this is to split the proposed `key` flag into two flags that follow the convention laid out in the ADR. However, `--key-host` and `--key-reference` are confusing; `--host-key` and `--reference-key` are more appropriate. Though to be honest, "reference key" can also be confusing. 

Curious what folks thoughts are on how this question might be best resolved.

adizere:
> The most straightforward way to address this is to split the proposed `key` flag into two flags that follow the convention laid out in the ADR. However, `--key-host` and `--key-reference` are confusing; `--host-key` and `--reference-key` are more appropriate. Though to be honest, "reference key" can also be confusing.

I agree this is not ideal. The "host" + "reference" terminology seems to be useful for client-related queries only.

Just to understand the current situation: We're refactoring ([cf ADR 10](https://github.com/informalsystems/ibc-rs/blob/master/docs/architecture/adr-010-unified-cli-arguments-hermes.md#clear-packets)) the `clear packets` CLI as follows:

> clear packets --chain <CHAIN_ID> --port <PORT_ID> --chan <CHANNEL_ID>

This refactor is underway here https://github.com/informalsystems/ibc-rs/pull/2275. Note that whatever changes we introduce in the present PR will likely conflict with that refactor, so we should aim to minimize merge conflicts, but that's a separate topic.

I looked again through the ADR and found that the specifiers we typically use as of ADR 10 are as follows:

> Example of specifiers we currently use are `host`, `reference`, `a`, `b` and `counterparty`.

If we want to have a specifier to a `--key` option, the specifier should probably be one of the above, to avoid introducing new ones. So we are left with two options from what I can tell: 
1. `--a-key` + `--b-key` or 
2. `--key` + `--counterparty-key`.

I doubt option 1 would be ideal here since `clear packets` CLI has no "a" or "b" options.

Additionally, we should also be consistent with the [key name specifiers from ADR 010](https://github.com/informalsystems/ibc-rs/blob/master/docs/architecture/adr-010-unified-cli-arguments-hermes.md#commands-for-keys). So it's more like

1. `--a-key-name` + `--b-key-name` or 
2. `--key-name` + `--counterparty-key-name`.

PS. should we update ADR 010 with whichever decision we take? to clearly specify how `clear packets` looks like.

seanchen1991:
I agree that `a` and `b` specifiers in this case are not very helpful. It sounds like going with `--key-name` and `--counterparty-key-name` is the only viable option then. 

> should we update ADR 010 with whichever decision we take? to clearly specify how clear packets looks like.

Agree that the ADR should be updated.

relayer-cli/src/commands/clear.rs
```diff
@@ -47,6 +78,14 @@ impl Runnable for ClearPacketsCmd {
             Err(e) => Output::error(format!("{}", e)).exit(),
         };
 
+        // Now that we have access to the counterparty chain at this point, we can
+        // match on `counterparty_key_name` and if a String exists, then override
+        // the counterparty chain's `key_name` in its own Config
+        if let Some(ref counterparty_key_name) = self.counterparty_key_name {
```
adizere, line 84: I think you need to call `config` on an object of type `ChainHandle`, cannot call the method on the trait directly.

One way to do it is using the `chains` variable defined above (L71).

```rust
let dst_chain_cfg = chains.dst.config();
```

romac (approved):

Issue #2093, opened 2022-04-13
Duplicate packets on start


ancazamfir:
## Summary of Bug
When a packet worker is started (triggered by an IBC event) it will submit duplicate transaction triggered by clearing on start. I think we need to use `height.decrement()` here:
 https://github.com/informalsystems/ibc-rs/blob/184a811564d7d699ceb72ad1563f7746a8b98a39/relayer/src/worker/packet.rs#L171

## Version
master

## Steps to Reproduce
- `hermes create channel`
- `hermes start`
- `hermes tx raw ft-transfer ibc-1 ibc-0 transfer channel-0 10 -n 1 -o 30 -k user2`
- observe the errors for both `MsgRecvPacket` and `MsgAcknowledgement`:

```
2022-04-13T02:00:55.665191Z ERROR ThreadId(40) send_tx_check{id=aatwvhqFEt}:send_tx{id=ibc-0}: broadcast_tx_sync: Response { code: Err(22), data: Data([]), log: Log("packet messages are redundant"), hash: transaction::Hash(6EC112B981392A1EF527FC9FD88B34439D120B06D5E25CF97227E5B26AD3DF6D) }: diagnostic: unknown SDK error: 22
```

## Acceptance Criteria
Should not submit duplicate transactions

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
This is consistent with what we noticed yesterday in the [GH action logs](https://github.com/informalsystems/ibc-rs/runs/5988255616?check_suite_focus=true) --



>2022-04-12T10:41:46.595827Z  INFO ibc_test_framework::bootstrap::binary::channel: created new chain/client/connection/channel from ibc-0-f172a25a/07-tendermint-3/connection-20/channel-26 to ibc-0-f172a25a/07-tendermint-3/connection-21/channel-27
[9513](https://github.com/informalsystems/ibc-rs/runs/5988255616?check_suite_focus=true#step:8:9513)
2022-04-12T10:41:46.684534Z  INFO ibc_integration_test::tests::transfer: Sending IBC transfer from chain ibc-0-f172a25a to chain ibc-0-f172a25a with amount of 3625 coin24abc971
[...]

PR #2256, opened 2022-06-01
Duplicate packets on start

seanchen1991:
Closes: #2093 

## Description

This PR changes the `handle_packet_cmd` flow so that it initiates the clearing of packets upon receiving a batch of `IbcEvents`. 

The issue seems to have originated as a result of some non-determinism in how the relayer subscribes to chains. The usual expected flow is that upon subscribing to a chain, a `NewBlock` command is received at some height `h`, followed by an `IbcEvents` command also at height `h`. However, sometimes the `NewBlock` command at height `h` would be missed and the relayer would only receive the `IbcEvents` command at height `h`. This command would be processed at height `h`, and then later a subsequent `NewBlock` command at associated with height `h + 1` would arrive, which would trigger a `clear_pending_packets` flow at height `h`, which resulted in a "packet messages are redundant" error since the relayer would attempt to clear the `IbcEvents` at height `h`, but they had already been cleared. 

The key insight was realizing that the relayer needs to clear pending packets regardless of whether the `NewBlock` command arrives before the `IbcEvents` command or vice versa. Regardless of the ordering of these two commands, pending packets at height `h - 1` and before need to be cleared _once_ on startup if the relayer has been configured to do so. 
______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


relayer/src/worker/packet.rs
```diff
@@ -172,6 +169,9 @@ fn handle_packet_cmd<ChainA: ChainHandle, ChainB: ChainHandle>(
                 clear_interval,
                 height,
             ) {
+                // Decrement the height in order to avoid clearing packets at the same
+                // height at which the packet's events were queried
+                let height = height.decrement().unwrap();
```
seanchen1991, line 174: Is unwrapping here upon decrementing acceptable?

plafer:
Can we add a test here? Seems like the kind of bug that can resurface easily after a refactor

relayer/src/worker/packet.rs
ancazamfir: I think we can do something like:
```
    match height.decrement() {
        Err(e) => Err(LinkError::decrement_height(height, e)),
        Ok(height) => link.a_to_b.schedule_packet_clearing(Some(height)),
    }
```

ancazamfir:
Looked a bit through the logs and what happens is that depending on the event registration, hermes may miss the `NewBlock` for `h` but get an `IbcEvent` for `h`. So now it will process the events at `h` and on `NewBlock` for `h+1` it will clear for `h` (sending duplicate packets).

The fix is to clear packets at `h-1` either on `NewBlock` **or** `IbcEvent` command (with height `h`) from supervisor.

ancazamfir:
You may want to wait for @soareschen's PR https://github.com/informalsystems/ibc-rs/pull/2238 as the changes there will affect this PR and you can make use the refactoring there.

relayer/src/worker/packet.rs
```diff
@@ -115,34 +115,54 @@ fn handle_packet_cmd<ChainA: ChainHandle, ChainB: ChainHandle>(
...
+
+    if *should_clear_on_start {
+        *should_clear_on_start = false;
+        return true;
+    }
+
+    let height = height.unwrap();
+
     clear_interval != 0 && height.revision_height % clear_interval == 0
 }
```
seanchen1991, line 167: In order to reduce duplicate checks of the `should_clear_on_start` flag in the `handle_packet_cmd` function, the boolean was moved into this function, but I'm not terribly keen on this refactor, as it bloats the logical complexity of what was originally a very simple check. I'm also not entirely sure that this logic is entirely sound; it was merely the cleanest way I came up with to implement this.

relayer/src/worker/packet.rs
```diff
@@ -99,50 +99,87 @@ pub fn spawn_packet_cmd_worker<ChainA: ChainHandle, ChainB: ChainHandle>(
...
+    } else {
+        Ok(())
     }
 }
 
-/// Whether or not to clear pending packets at this `step` for the given height.
-/// Packets are cleared if `clear_interval` is not `0` and if we have reached the interval.
-fn should_clear_packets(clear_interval: u64, height: Height) -> bool {
+/// Whether or not to clear pending packets at this `step` for some height.
+/// If the relayer has been configured to clear packets on startup and that has not
```
ancazamfir, line 162:
```suggestion
/// If the relayer has been configured to clear packets on start and that has not
```

relayer/src/worker/packet.rs
```diff
@@ -99,50 +99,87 @@ pub fn spawn_packet_cmd_worker<ChainA: ChainHandle, ChainB: ChainHandle>(
...
+    //     return false;
+    // }
+
+    // if *should_clear_on_start {
+    //     *should_clear_on_start = false;
+    //     return true;
+    // }
+
+    // let height = height.unwrap();
+
```
ancazamfir, line 182: Can you remove the commented code? Also further down.

ancazamfir (approved):
The changes look good, thanks @seanchen1991! some commented out code needs to be cleaned. I will approve as I am off tomorrow.

Issue #2242, opened 2022-05-25
Use dashes instead of underscores in TOML config


romac:
## Summary

Using dashes to separate words in the TOML config looks more readable to me, which no semantic change.

It is also the preferred convention for the Tendermint `config.toml` and the Cosmos SDK `app.toml`.

Example config using dashes:

```toml
# The global section has parameters that apply globally to the relayer operation.
[global]
log-level = 'info'

[mode]

[mode.packets]
enabled = true
clear-interval = 100
clear-on-start = true
tx-confirmation = true

[rest]
enabled = true
host = '127.0.0.1'
port = 3000

[telemetry]
enabled = true
host = '127.0.0.1'
port = 3001

[[chains]]
id = 'ibc-1'
type = 'CosmosSdk'
rpc-addr = 'http://127.0.0.1:26557'
grpc-addr = 'http://127.0.0.1:9091'
websocket-addr = 'ws://127.0.0.1:26557/websocket'
rpc-timeout = '10s'
account-prefix = 'cosmos'
key-name = 'testkey'
store-prefix = 'ibc'
default-gas = 100000
max-gas = 400000
gas-price = { price = 0.001, denom = 'stake' }
gas-adjustment = 0.1
max-msg-num = 30
max-tx-size = 2097152
clock-drift = '5s'
max-block-time = '30s'
trusting-period = '14days'
trust-threshold = { numerator = '1', denominator = '3' }
address-type = { derivation = 'cosmos' }
```

### Acceptance criteria

- [ ] Update the default `config.toml` to use dashes
- [ ] Update the _Configuration_ page in the guide to use dashes

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2264, opened 2022-06-03
Refactored configuration file field names

ljoss17:
Closes: #2242

## Description

This PR refactored the TOML configuration file used by the ibc-relayer in order to have `-` (dash) instead of `_` (underscore) for the field names.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/breaking-changes/ibc-relayer/2242-replace-config-toml-underscores-with-dashes.md
+++ b/.changelog/unreleased/breaking-changes/ibc-relayer/2242-replace-config-toml-underscores-with-dashes.md
@@ -1,0 +1,3 @@
+- The ibc-relayer configuration file has been refactored to replace
+  all `_` with `-` in the field names. E.g. `log_level` is now `log-level`
+  ([#2242](https://github.com/informalsystems/ibc-rs/issues/2242))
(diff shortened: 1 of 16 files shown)


romac:
imho there is no need for a changelog entry here since this is a fully cosmetic change. On the other hand the guide needs to be updated as well.

ljoss17:
> imho there is no need for a changelog entry here since this is a fully cosmetic change. On the other hand the guide needs to be updated as well.

Regarding the changelog, since this PR will break any relayer using a custom config file which has the old format, is it not better to log it in breaking-changes ? The old field names will cause an unknown field name error.
As for the guide, I updated the Hermes guide, but there is one link, section 3.3 Configuration, which redirects to [https://github.com/informalsystems/ibc-rs/blob/v0.15.0/config.toml](https://github.com/informalsystems/ibc-rs/blob/v0.15.0/config.toml). As long as there isn't a new version with the updated config file format, I am not sure which link should be used.

romac:
Ah my bad, I thought the `toml` crate automatically mapped kebab-case fields to lower_dashes. Since this change is therefore not backward compatible, let's perhaps put it aside for now or figure out a way to make it work with both types of names.

PR #2266, opened 2022-06-03
Update error messages which returned from ibc-go

kubo39:
Closes: #XXX

## Description

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/relayer/src/sdk_error.rs
+++ b/relayer/src/sdk_error.rs
@@ -32,85 +32,85 @@ define_error! {
             |_| { "light client already exists" },
 
         InvalidLightClient
-            |_| { "light client already exists" },
+            |_| { "light client is invalid" },
 
         LightClientNotFound
-            |_| { "light client already exists" },
+            |_| { "light client not found" },
 
         FrozenLightClient
-            |_| { "light client already exists" },
+            |_| { "light client is frozen due to misbehaviour" },
 
         InvalidClientMetadata
-            |_| { "light client already exists" },
+            |_| { "invalid client metadata" },
 
         ConsensusStateNotFound
-            |_| { "light client already exists" },
+            |_| { "consensus state not found" },
 
         InvalidConsensusState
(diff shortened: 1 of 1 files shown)


romac:
Wow, thanks for noticing and fixing this!

adizere:
Good catch, thank you for the fix!

@ljoss17 and I bumped into this copy/past error just yesterday. We decided not to to anything with it until we understand what's up with the `sdk_error_from_tx_result` and the `ClientError` type.

My findings so far are:
- It seems like we introduced this new type as part of error type refactoring in #988 and never made use of it. 
- Originally, the code comes from a big lift of Soares to categorize all error types and improve error handling overall in the relayer pipeline: #950 ([file ref here](https://github.com/informalsystems/ibc-rs/pull/950/files#diff-9aebdde03ce94fdb3cbca7ed59e29fc27cb44f53569e776ecf0956e5ed1cdb73)). That PR was never merged, but I think the changes somehow slipped in master via #988.
- That being said, I am on the fence whether to delete the code altogether or keep it around. It's unused so not a big maintenance burden or liability + it may prove useful in the future.

Since you fixed it for us, it seems like a good idea to keep the code, unless Romain has a different opinion -- will defer to him.

seanchen1991:
Just for my own understanding, were these error message changes done so that they would match the errors returned by ibc-go?

adizere:
> Just for my own understanding, were these error message changes done so that they would match the errors returned by ibc-go?

The error string was wrong (copy/paste error). The changes were _not_ done so that they _match_ ibc-go, but so that the error string is consistent with the error type. The error codes have been copied from IBC-Go.

soareschen:
Thanks for the catch. Yes it was copy/paste errors when copying the error messages from the original definitions to the new error definitions using flex-error. Since there are so many error definitions, it was easy to slip up and copy wrongly.

Issue #2239, opened 2022-05-24
CLI: Use flags everywhere instead of positional arguments


romac:
## Summary

All CLI commands should use flags (`--flag`) options instead of positional arguments.

## Problem Definition

At the moment, a lot of the CLI commands use a mix of positional arguments and flags, for example
the `create channel` CLI:

```
hermes create client ibc-1 ibc-0 --trusting-period 14days
                     ^^^^^ ^^^^^ ^^^^^^^^^^^^^^^^^^^^^^^^
                      positional          flag
```

When writing the commands it is now always clear what the positional arguments denote (eg. which is the source chain and destination chain), and one often has to resort to the command documentation (`--help`) to figure it out.

Even worse, the number of positional arguments and their denotation for the `create channel` command change depending on which flags are passed, eg.

```
hermes create channel ibc-0         -c ibc-1 --port-a transfer --port-b transfer --new-client-connection
                      ^^^^^         ^^^^^^^^ ^^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^^
                      positional    flag     flag              flag              flag
```

vs

```
hermes create channel  ibc-0       connection-0 --port-a transfer --port-b transfer
                       ^^^^^       ^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^^
                    positional      positional        flag               flag
```

This is quite confusing, as expressed already by some operators: https://github.com/informalsystems/ibc-rs/issues/1935

## Proposal

I suggest using flags everywhere instead of positional arguments.

Positional arguments should be kept for the case where we want to pass a list of items to a command, which we don't have the need for currently.

## Acceptance Criteria

All CLI commands accept only flags and no positional arguments.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ljoss17:
Hey @adizere  and @romac .

I would like to use this issue in order to apply the following guidelines for the CLI command codes:

1. Explicitly give long and/or short flags. E.g:
```
    #[clap(short = 'a', long = "amount", required = true, help = "amount of stake")]
    amount: u64,
```
2. All flags which start with the character `c` would only have a long flag, to avoid conflicting with the global flag `-c/—config`
[...]

PR #2275, opened 2022-06-07
Replace positional arguments with flags in CLI

ljoss17:
Closes: #2239

## Description

This PR adds flags instead of positional arguments to all CLI commands. Long flags have been used in order to avoid conflicts or confusing flags.

### WIP

* At commit 51852ae51b91471e24f7b4b57116749b2d24d39c: Some short flags are still present for some commands, but a discussion in order to clarify if all the short flags will be removed or if some strategic short flags will be kept is ongoing.
* At commit 88474db7e7a2957f9ab0f3b5273f82ec3b2ee544 and a76599fa918e7fc6f55235f1b70a782ebc274b8c: Removed all short flags.

______

### PR author checklist:
- [x] Linked to GitHub issue.
- [x] Updated guide (e.g., `guide/`).

### Reviewer checklist:


plafer:
Just a note not to forget to update the `gm hermes cc` command. This is the current output:

```shell
% gm hermes cc
"./hermes" create channel ibc0 ibc1 --port-a transfer --port-b transfer
```

ljoss17:
> Just a note not to forget to update the `gm hermes cc` command. This is the current output:
> 
> ```shell
> % gm hermes cc
> "./hermes" create channel ibc0 ibc1 --port-a transfer --port-b transfer
> ```

Yes, I will check all the commands which use hermes and update them with the new flags.

relayer-cli/src/commands/keys/add.rs
```diff
@@ -179,3 +175,39 @@ pub fn restore_key(
...
+    #[test]
+    fn test_keys_add_no_file() {
+        assert!(KeysAddCmd::try_parse_from(&["test", "--chain", "chain_id"]).is_err());
+    }
+
+    #[test]
+    fn test_keys_add_key_and_mnemonic() {
+        assert!(KeysAddCmd::try_parse_from(&["test", "--chain", "chain_id", "--key-file", "key_file", "--mnemonic-file", "mnemonic_file"]).is_err());
+    }
+}
```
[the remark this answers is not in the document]
mzabaluev: I was hoping there is some framework to provide some handy test macros, but I can't find any at the moment.
This will do, we can add more cases to exercise.

If you have time, consider creating some helper macros like:

```rust
assert_valid_cmdline!("--chain", "chain_id", "--key-file", "key_file", "--mnemonic-file", "mnemonic_file");
assert_invalid_cmdline!("--chain", "chain_id", "--key-file", "key_file", "--mnemonic-file", "mnemonic_file");
```

mzabaluev:
I have found [trycmd](https://crates.io/crates/trycmd), which looks like a useful tool to organize CLI tests in a self-documented way.

This could also augment our manually maintained command examples in the guide mdbook. Not in scope for this PR, though.

relayer-cli/src/commands/keys/add.rs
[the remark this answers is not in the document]
ljoss17: Thank you for all the feedback! I will check this as soon as this PR is merged and I work on the unit-tests.

adizere:
I think I found a typo in the ADR.

```diff
--- a/docs/architecture/adr-010-unified-cli-arguments-hermes.md
+++ b/docs/architecture/adr-010-unified-cli-arguments-hermes.md
@@ -115,7 +115,7 @@ __Connection__
     * Optional: `[--height <HEIGHT>]`

 * `query connections --chain <CHAIN_ID>`
-    * Optional: `[--chain-counterparty <CHAIN_COUNTERPARTY_ID>] [--verbose]`
+    * Optional: `[--counterparty-chain <COUNTERPARTY_CHAIN_ID>] [--verbose]`
```

Shall we fix it in this PR?

ljoss17:
> I think I found a typo in the ADR.
> 
> ```diff
> --- a/docs/architecture/adr-010-unified-cli-arguments-hermes.md
> +++ b/docs/architecture/adr-010-unified-cli-arguments-hermes.md
> @@ -115,7 +115,7 @@ __Connection__
>      * Optional: `[--height <HEIGHT>]`
> 
>  * `query connections --chain <CHAIN_ID>`
> -    * Optional: `[--chain-counterparty <CHAIN_COUNTERPARTY_ID>] [--verbose]`
> +    * Optional: `[--counterparty-chain <COUNTERPARTY_CHAIN_ID>] [--verbose]`
> ```
> 
> Shall we fix it in this PR?

Good catch! I will update this branch in order to fix the ADR in this PR.

adizere:
@ljoss17 can you please confirm that (most) clap tests will be done in a separate PR?

For reviewing strategy, I suggest the following
- Based on [ADR 010](https://github.com/informalsystems/ibc-rs/blob/master/docs/architecture/adr-010-unified-cli-arguments-hermes.md), it should be clear what is the standard (or baseline) we have towards accepting a CLI as good and ready to merge. If that ADR is unclear we should update it in this PR.
- Reviewers should apply changes directly in the PR, no need for suggestions to Luca. We can do that and avoid git conflicts since we'll split the workload horizontally, just remeber to periodically do `git pull --rebase` or similar to update your clone.
- We need to check both the guide updates and the CLI output.
- The work is rather tedious, so we'll split it. As reviewers, I'd suggest the following sharded approach:
        - @adizere reviews the `query` namespace
        - @mzabaluev reviews the `clear`, `config`, `create`, `health-check`, `help` and `keys` namespaces
        -  @seanchen1991 reviews `listen`, `misbehavior`, `start`, `update`, `upgrade`, and `completions`.
- If I missed anything let me know!

ljoss17:
> @ljoss17 can you please confirm that (most) clap tests will be done in a separate PR?

Yes, all the unit-tests for clap parsing will be done in a separate PR. I removed the example tests (https://github.com/informalsystems/ibc-rs/pull/2275#discussion_r898098129) which were in this PR.

seanchen1991 (commented):
Preliminarily approving based on reviewing the `listen`, `misbehavior`, `start`, `update`, `upgrade`, and `completions` commands. I can approve the PR once @mzabaluev approves the other commands.

adizere (commented):
Preliminary approved on my side as well for all the CLIs in the `query` namespace.

romac:
Is there an easy way to allow both `--chan` and `--channel`? (respectively `conn`/`connection`, `seq`/`sequence`)

ljoss17:
> Is there an easy way to allow both `--chan` and `--channel`? (respectively `conn`/`connection`, `seq`/`sequence`)

Yes, with clap's `alias`. For example if we add the `alias` to the the `channel_id` in the `QueryChannelEndCmd`:
```
    #[clap(
        long = "chan",
        alias = "channel",
        required = true,
        value_name = "CHANNEL_ID",
        help = "identifier of the channel to query"
    )]
    channel_id: ChannelId,
```
Then both these tests will pass:
```
    #[test]
    fn test_query_channel_end_required_only() {
        assert_eq!(
            QueryChannelEndCmd{ chain_id: ChainId::from_string("chain_id"), port_id: PortId::from_str("port_id").unwrap(), channel_id: ChannelId::from_str("channel-07").unwrap(), height: None },
            QueryChannelEndCmd::parse_from(&["test", "--chain", "chain_id", "--port", "port_id", "--chan", "channel-07"])
        )
    }

    #[test]
    fn test_query_channel_end_required_only_alias() {
        assert_eq!(
            QueryChannelEndCmd{ chain_id: ChainId::from_string("chain_id"), port_id: PortId::from_str("port_id").unwrap(), channel_id: ChannelId::from_str("channel-07").unwrap(), height: None },
            QueryChannelEndCmd::parse_from(&["test", "--chain", "chain_id", "--port", "port_id", "--channel", "channel-07"])
        )
    }
```

romac:
Sweet, then let's do this. Remains to decide whether we want the long (`channel`) or shortened (`chan`) version to be the default one? What do you think? I would vote for the long one being the default and the shorter one as alias, but would like to hear what you and others think?

ljoss17:
I have a preference for the shortened version, but I think my opinion is biased due to using them when replacing and testing the commands.
Should the aliases be added in the PR adding the parsing unit-tests, which I will work on as soon as this PR is merged ? Or should they be added in a separate PR which would only add the aliases ?

seanchen1991:
I would also agree with sticking with the long versions, if only to be more explicit. I think this applies to `conn` and `connection` as well.

romac:
> Should the aliases be added in the PR adding the parsing unit-tests, which I will work on as soon as this PR is merged ? Or should they be added in a separate PR which would only add the aliases ?

I would rather add the aliases to this PR to keep PR churn down, if that's okay.

guide/src/commands/relaying/clear.md
```diff
@@ -9,28 +9,26 @@ and [packet-acks](../raw/packet.md#relay-acknowledgment-packets).
...
+        --port <PORT_ID>       identifier of the port
 ```
 
 ### Example
 
 1. Without Hermes running, send 3 packets over a channel, here `channel-13`:
 
 ```
-❯ hermes tx raw ft-transfer ibc1 ibc0 transfer channel-13 9999 -o 1000 -n 3
+❯ hermes tx raw ft-transfer --dst-chain ibc1 --src-chain ibc0 --src-port transfer --src-chan channel-13 --amount 9999 --timeout-height-offset 1000 --number-msgs 3
```
mzabaluev, line 31: We still have `--src-chain` and `--dst-chain` here, not `--a-chain` and `--b-chain` as ADR 010 stipulates?

guide/src/commands/path-setup/index.md
```diff
@@ -21,10 +21,12 @@ DESCRIPTION:
     Create objects (client, connection, or channel) on chains
 
 SUBCOMMANDS:
-    help       Get usage information
-    client     Create a new IBC client
-    connection Create a new connection between two chains
-    channel    Create a new channel between two chains
+    channel       Create a new channel between two chains using a pre-existing connection.
+                      Alternatively, create a new client and a new connection underlying the new
+                      channel if a pre-existing connection is not provided
```
mzabaluev, line 26: This looks too long for the subcommand summary.

Can we put it under the command's long help, leaving the summary as one line?

relayer-cli/src/commands/create/channel.rs
```diff
@@ -46,53 +46,61 @@ static HINT: &str = "Consider using the default invocation\n\nhermes create chan
...
         help = "The channel ordering, valid options 'unordered' (default) and 'ordered'",
         default_value_t
     )]
     order: Order,
 
     #[clap(
-        short,
-        long = "channel-version",
+        long = "chan-version",
         alias = "version",
```
mzabaluev, line 96: Might as well remove this alias now because it's ambiguous with clap's built-in `--version` flag.

relayer-cli/src/commands/create/channel.rs
```diff
@@ -46,53 +46,61 @@ static HINT: &str = "Consider using the default invocation\n\nhermes create chan
 #[clap(disable_version_flag = true)]
```
mzabaluev, line 46: If you remove the "version" flag alias below, remove this attribute as well.

relayer-cli/src/commands/create/connection.rs
```diff
@@ -17,37 +17,46 @@ use crate::prelude::*;
 #[derive(Clone, Command, Debug, Parser)]
 pub struct CreateConnectionCommand {
     #[clap(
+        long = "a-chain",
         required = true,
+        value_name = "A_CHAIN_ID",
         help = "identifier of the side `a` chain for the new connection"
     )]
     chain_a_id: ChainId,
```
mzabaluev, line 25: Maybe it's best to also rename the struct members to match the option names? Otherwise we'd be always confused.

relayer-cli/src/commands/listen.rs
```diff
@@ -61,11 +61,12 @@ impl FromStr for EventFilter {
 #[derive(Debug, Parser)]
 pub struct ListenCmd {
     /// Identifier of the chain to listen for events from
+    #[clap(long = "chain", required = true, value_name = "CHAIN_ID")]
     chain_id: ChainId,
 
     /// Add an event type to listen for, can be repeated.
     /// Listen for all events by default (available: Tx, NewBlock).
-    #[clap(short = 'e', long = "event", value_name = "EVENT")]
+    #[clap(long = "event", value_name = "EVENT")]
```
mzabaluev, line 69: This is where the zeal to remove short options may be doing more harm than good.
Note that this is a repeating option, so it will be annoying to type it over and over again if there are many events to filter.

mzabaluev (commented):
Good work!

I've reviewed "my turf"...

> `health-check`, `help`

These were easy: no parameters, nothing to change!
So I took it upon myself to look at some more commands.

mzabaluev:
Please also merge with recent master as there are some new conflicts.

guide/src/commands/relaying/clear.md
ljoss17: Yes, these are the `raw` commands, which are not in the ADR 010 since they are used only by developers. So they have not been changed.

relayer-cli/src/commands/listen.rs
ljoss17: Oh good catch, thank you!
I managed to find a solution, by adding the `multiple_values` it is not required to repeat the flag:

```
    /// Add an event type to listen for, can be repeated.
    /// Listen for all events by default (available: Tx, NewBlock).
    #[clap(long = "event", value_name = "EVENT", multiple_values = true)]
    events: Vec<EventFilter>,
```

This will have the following tests pass:

```
    #[test]
    fn test_listen_multiple_events() {
        assert_eq!(
            ListenCmd{ chain_id: ChainId::from_string("chain_id"), events: vec!(EventFilter::from_str("Tx").unwrap(), EventFilter::from_str("NewBlock").unwrap()) },
            ListenCmd::parse_from(&["test", "--chain", "chain_id", "--event", "Tx", "--event", "NewBlock"])
        )
    }

    #[test]
    fn test_listen_multiple_events_single_flag() {
        assert_eq!(
            ListenCmd{ chain_id: ChainId::from_string("chain_id"), events: vec!(EventFilter::from_str("Tx").unwrap(), EventFilter::from_str("NewBlock").unwrap()) },
            ListenCmd::parse_from(&["test", "--chain", "chain_id", "--event", "Tx", "NewBlock"])
        )
    }
```

I will update this with the other changes you suggested.

romac:
What's the reasoning behind having `--a-chain` instead of `--chain-a`? I am asking because the latter reads better to me.

ljoss17:
> What's the reasoning behind having `--a-chain` instead of `--chain-a`? I am asking because the latter reads better to me.

The choice of prefixing `--a-` and `--b-` is due to the following discussion in the ADR PR : https://github.com/informalsystems/ibc-rs/pull/2306#discussion_r899050167

guide/src/commands/path-setup/index.md
ljoss17: By having an `about` and `long_about` for the `create channel` subcommand will display only the short help (`about`). Meaning the commands `hemres create -h` and `hermes create --help` will both display:
```
Create objects (client, connection, or channel) on chains

USAGE:
    hermes create <SUBCOMMAND>

OPTIONS:
    -h, --help    Print help information

SUBCOMMANDS:
    channel       Create a new channel between two chains using a pre-existing connection
    client        Create a new IBC client
...
```
`--help` gives precedence to `about` over `long_about`, https://github.com/clap-rs/clap/issues/3193#issuecomment-996390269

But the command `hermes create channel -h` will display the short help:
```
Create a new channel between two chains using a pre-existing connection

USAGE:
    hermes create channel [OPTIONS] --a-chain <A_CHAIN_ID> --a-port <A_PORT_ID> --b-port <B_PORT_ID>
...
```
And the command `hermes create channel --help` will display the long help:
```
Create a new channel between two chains using a pre-existing connection.

Alternatively, create a new client and a new connection underlying the new channel if a pre-existing
connection is not provided.

USAGE:
    hermes create channel [OPTIONS] --a-chain <A_CHAIN_ID> --a-port <A_PORT_ID> --b-port <B_PORT_ID>
```
I think having the short version only for the `hermes create -h` and `hermes create --help`, and the short for `hermes create channel -h` and long for `hermes create channel --help` is a good option.
What do you think ?

mzabaluev: This is exactly how I think it should work, thanks.

mzabaluev: Dunno if help through doc comments automagically allows this, though, or whether it needs an explicit clap attribute.
What happens if the first line is set into its own paragraph?

relayer-cli/src/commands/listen.rs
mzabaluev: I think there was a reason why we wanted it to have the `multiple_occurrences` behavior, i.e. every occurrence needs its own flag repetition. Most likely because in `multiple_values` style as above, additional values were easily confused with positional parameters. But now we don't have any positional parameters! @ancazamfir @mircea-c any opinions here?

That said, I'd still prefer most frequently and likely used options to also have short flags, and this seems to be one such case.

guide/src/commands/path-setup/index.md
ljoss17: Having an empty line in the doc comments works. Like this:
```
    /// Create a new channel between two chains using a pre-existing connection.
    ///
    /// Alternatively, create a new client and a new connection underlying
    /// the new channel if a pre-existing connection is not provided.
    Channel(CreateChannelCommand),
```

relayer-cli/src/commands/create/connection.rs
romac: Let's keep the fields as-is, this PR has already gone through a lot of churn and I would rather merge it sooner than later. Let's perhaps do this if there's a follow-up PR (eg. in the one for unit tests)

relayer-cli/src/commands/listen.rs
romac: Let's use `multiple_values` since, as you say, we don't have any positional arguments anymore.

romac (approved):
Amazing work @ljoss17, thank you so much! Many thanks to all reviewers as well :)

Issue #2249, opened 2022-05-31
Account sequence mismatch causes unnecessary delay in relaying


ancazamfir:
## Summary of Bug
Sometimes Tx simulation failures due to account sequence mismatch causes hermes to unnecessary delay the relaying.
Background:
-  hermes simulates every transaction by calling the `simulateTx` (application gRPC). If successfully simulated, it uses the gas to determine the Tx fee and then 
- it `broadcast_tx_sync`-s the transaction (tendermint RPC), i.e. sent to tendermint mempool, which calls the application `checkTx`. 
- In both cases the verification happens on a copy of the application state, aka checkTx state. 
- the checkTx state is also updated by the application after a block is being committed (`deliverTx` is run for all Tx-es in a block)

Here are the details of the issue:
- hermes simulates and `broadcast_tx_sync`-s `N` transactions with following sequence numbers: `[1,.., K, K+1, ...,N]`
   - check Tx state **last account sequence is `N`.**
- tendermint creates a block with `[1,..,K]`.  This involves calling [`Commit()`](https://github.com/tendermint/tendermint/blob/85870def7b628effad73af942e638bbddf2ba8fd/state/execution.go#L211) which:
   - updates the application by calling `deliverTx` for `[1,..,K]` and then recreating the check Tx state with latest application state:
       - check Tx state **last account sequence is `K`.**
   - then it  [updates the mempool](https://github.com/tendermint/tendermint/blob/85870def7b628effad73af942e638bbddf2ba8fd/mempool/clist_mempool.go#L569). Part of the mempool update is [`recheckTxs()`](https://github.com/tendermint/tendermint/blob/85870def7b628effad73af942e638bbddf2ba8fd/mempool/clist_mempool.go#L631) which calls `checkTx` for `[K+1,..,N]`
      - check Tx state **last account sequence is `K+x`.**
- hermes tries to simulate a transaction with sequence number `N+1` sometime before `recheckTxs()` above finishes. It gets the error: 
    - **`"account sequence mismatch, expected K+x, got N+1: incorrect account sequence"`**, with x being the number of Tx-es that have been rechecked
- hermes recaches to `K+x` and tries to simulate again a bit later only to get: **`"account sequence mismatch, expected N+1, got K+x: incorrect account sequence"`**
- then recaches again to `N+1` and the simulation is usually successful as the `recheckTxs()` above is finished already.

[...]

PR #2298, opened 2022-06-14
Ignore acc seq mismath when expected < got

ancazamfir:
Closes: #2249

Changes:
- extracts the "expected" and "got" s.n. from error message
- uses default gas if mismatch s.n. occurs during simulation and expected < got

Testing:
- start two chains
- create a channel
- send 5000 transfers (this should already show the mismatch issue and CLI will finish early), repeat if error
- start hermes
  - before the fix: tx not sent, account is recached with wrong sequence, at next attempt the error is seen again, recaching again etc. (more details on what happens in #2249)
  - after the fix: tx is sent with default gas. For 5k sent packets, meaning 10k packet relays (recv and ack), I see the mismatch error with expected < got around 10 times
## Description

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/relayer/src/error.rs
+++ b/relayer/src/error.rs
@@ -6,6 +6,7 @@ use flex_error::{define_error, DisplayOnly, TraceClone, TraceError};
 use http::uri::InvalidUri;
 use humantime::format_duration;
 use prost::{DecodeError, EncodeError};
+use regex::Regex;
 use tendermint::Error as TendermintError;
 use tendermint_light_client::components::io::IoError as LightClientIoError;
 use tendermint_light_client::errors::{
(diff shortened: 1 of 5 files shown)


relayer/src/error.rs
```diff
@@ -554,7 +555,76 @@ impl GrpcStatusSubdetail {
             .trim_start()
             .starts_with("account sequence mismatch")
     }
+
+    pub fn is_account_sequence_mismatch_that_can_be_ignored(&self) -> bool {
+        match parse_sequences_in_mismatch_error_message(self.status.message()) {
+            None => false,
+            Some((expected, got)) => expected < got,
```
soareschen, line 562: Is this different from the usual case which the expected sequence is larger than the provided sequence? And is this safe to ignore because the actual SendTX allows provided account sequence to be higher than the next available sequence?

It may worth adding a comment on why the inverse can be ignored.

relayer/src/error.rs
```diff
@@ -554,7 +555,76 @@ impl GrpcStatusSubdetail {
...
+        }
+    }
+}
+
+fn parse_sequences_in_mismatch_error_message(message: &str) -> Option<(u64, u64)> {
+    let re = Regex::new(
+        r#"account sequence mismatch, expected (?P<expected>\d+), got (?P<got>\d+): incorrect account sequence"#
```
soareschen, line 569: We should have a comment to note that the regex may stop working if the error message is updated by the chain in the future.

relayer/src/error.rs
ancazamfir: > Is this different from the usual case which the expected sequence is larger than the provided sequence? And is this safe to ignore because the actual SendTX allows provided account sequence to be higher than the next available sequence?

Yes and yes. The details are in the issue.
expected > got = typically happens if some other agent uses the same wallet as the relayer, will happen in simulate and broadcast; we have to refresh in this case
expected > go = typically happens around the time when tm switches to a new block and we try to simulate a Tx after the new block and app state are created but while mempool Tx-es are "rechecked".  When this happens we can ignore the simulation failure and broadcast the Tx. The checkTx for this will happen after all other mempool Tx-es are rechecked so the mismatch error should not be seen anymore.

> It may worth adding a comment on why the inverse can be ignored.
tried in 132cd46276230574d3d0299b6192615cfe01cc78

relayer/src/error.rs
ancazamfir: tried in https://github.com/informalsystems/ibc-rs/commit/132cd46276230574d3d0299b6192615cfe01cc78

soareschen (approved):

Issue #2301, opened 2022-06-15
Hermes: Add support to fetch Tendermint version of a network


adizere:
## Summary

Given the upcoming breaking changes in #1560, Hermes will need to be able to fetch and parse the Tendermint semver that a network is running. There is already support in Hermes to fetch the version of the SDK and IBC-go modules, so we just need to extend that to additionally extract the Tendermint version spec.

## Problem Definition

__Why do we need this feature?__

To be able to detect in Hermes which version of Tendermint is network using. One important use-case is: this will allow Hermes to switch at runtime to a specific monitor (i.e., event subscriber) implementation:
- for Tendermint 0.34 and older we'll use the current Websocket-based monitor
- for Tendermint 0.35 and newer we'll switch to using the JSON-RCP event streaming

A note on priority of this issue: We're adding this feature in advance of breaking changes from Tendermint 0.35 to derisk that upgrade. We want to make sure Hermes is capable of finding & parsing Tendermint versions before this feature becomes urgent.

## Acceptance Criteria

- [ ] The chain endpoint for Cosmos chains should have support for detecting the Tendermint version of the network it is connected to

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

Issue #1337, opened 2021-09-08
Compatibility check doesn't seem to handle non semantic versions properyly, emits a warning


andynog:
## Crate

ibc-relayer

## Summary of Bug

Hermes might issue a warning when a theoretically valid version (but not semantic version).

## Version

v.0.6.2

## Steps to Reproduce

Running Hermes on a chain that has a version such as `0.42.0-rc0` might trigger this. For example the compatibility list requires `>=0.41.3 and <= 0.42.6`.

`Sep 08 10:30:46.443  WARN Hermes health check failed while verifying the application compatibility for chain regen-1:http://[ip_address]:9090/; caused by: SDK module at version '0.42.0-rc0' does not meet compatibility requirements >=0.41.3, <=0.42.6 for application regen:v1.0.0-1b7c80ef102d3ae7cc40bba3ceccd97a64dadbfd
Sep 08 10:30:46.443  WARN some Hermes features may not work in this mode!`

## Acceptance Criteria

I'm not sure if there's a easy way to check compatibility if chains don't follow a semantic versioning scheme. Since this is a warning it's not a high priority for now. Just reporting it.

____

## For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

romac:
This indeed follows the [semver spec](https://semver.org/spec/v2.0.0-rc.1.html#spec-item-10): (emphasis mine)

> A pre-release version indicates that the version is unstable and **might not** satisfy the intended compatibility requirements as denoted by its associated normal version

Which means that the spec leaves the handling of pre-release version to us. If we feel that's the right to do, I guess we could remove the pre-release version and match against that. What do you think @andynog @adizere?

PR #2302, opened 2022-06-15
Add support for fetching & parsing the Tendermint version of a chain

adizere:
Closes: #2301

## Description

Adds version detection support for the Tendermint dependency. In Hermes logs, this appears as:

> 2022-06-15T09:25:08.592954Z TRACE ThreadId(25) parsed version specification for gaiad v6.0.4@305668ab9d962431c79d718bb0ffdeec77a46439 -> SDK=0.44.6; Ibc-Go status=Some(Version { major: 2, minor: 0, patch: 3 }); Tendermint=Version { major: 0, minor: 34, patch: 15 }

______

### PR author checklist:
- [ ] ~~Added tests: integration (for Hermes) or unit/mock tests (for modules).~~
- [X] Linked to GitHub issue.
- [X] Updated code comments and documentation (e.g., `docs/`).

### Reviewer checklist:

### DIFF
--- a/relayer/src/chain/cosmos/version.rs
+++ b/relayer/src/chain/cosmos/version.rs
@@ -1,5 +1,5 @@
 //! Utilities for extracting and parsing versioning information
-//! of Cosmos-SDK chains. The extracted version specification
+//! of Cosmos-SDK networks. The extracted version specification
 //! is captured in a domain-type semver format in [`Specs`].
 
 use flex_error::define_error;
@@ -7,9 +7,9 @@ use tracing::trace;
 
 use ibc_proto::cosmos::base::tendermint::v1beta1::VersionInfo;
 
-/// Specifies the SDK & IBC-go modules path, as it is expected
+/// Specifies the SDK, IBC-go, and Tendermint modules path, as expected
 /// to appear in the application version information of a
-/// Cosmos chain.
+/// Cosmos-SDK network.
 ///
 /// The module identification is captured in a [`Module`]
 /// with the following structure as an example:
@@ -22,17 +22,19 @@ use ibc_proto::cosmos::base::tendermint::v1beta1::VersionInfo;
 /// ```
 const SDK_MODULE_NAME: &str = "cosmos/cosmos-sdk";
 const IBC_GO_MODULE_NAME: &str = "cosmos/ibc-go";
+const TENDERMINT_MODULE_NAME: &str = "tendermint/tendermint";
 
 /// Captures the version(s) specification of different
-/// modules of a chain.
+/// modules of a network.
 ///
-/// Assumes that the chain runs on Cosmos SDK.
+/// Assumes that the network runs on Cosmos SDK.
 /// Stores both the SDK version as well as
 /// the IBC-go module version (if existing).
 #[derive(Debug)]
 pub struct Specs {
     pub sdk_version: semver::Version,
     pub ibc_go_version: Option<semver::Version>,
+    pub tendermint_version: semver::Version,
 }
 
 define_error! {
@@ -42,7 +44,14 @@ define_error! {
                 address: String,
                 app: AppInfo,
             }
-            |e| { format!("node at {} running chain {} not caught up", e.address, e.app) },
+            |e| { format!("failed to find the SDK module dependency ('{}') for application {}", e.address, e.app) },
+
+        TendermintModuleNotFound
+            {
+                address: String,
+                app: AppInfo,
+            }
+            |e| { format!("failed to find the Tendermint dependency ('{}') for application {}", e.address, e.app) },
 
         VersionParsingFailed
             {
@@ -51,7 +60,7 @@ define_error! {
                 cause: String,
                 app: AppInfo,
             }
-            |e| { format!("failed parsing the SDK module ('{}') version number '{}' into a semver for application {}; cause: {}",
+            |e| { format!("failed parsing the module path ('{}') version number '{}' into a semver for application {}; cause: {}",
                     e.module_path, e.raw_version, e.app, e.cause) },
     }
 }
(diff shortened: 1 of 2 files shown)


relayer/src/chain/cosmos/version.rs
```diff
@@ -63,19 +72,22 @@ impl TryFrom<VersionInfo> for Specs {
         // Get the Cosmos SDK version
         let sdk_version = parse_sdk_version(&raw_version)?;
         let ibc_go_version = parse_ibc_go_version(&raw_version)?;
+        let tendermint_version = parse_tendermint_version(&raw_version)?;
 
         trace!(
-            "parsed version specification for {} {}@{} -> SDK={}; Ibc-Go status={:?}",
+            "parsed version specification for {} {}@{} -> SDK={}; Ibc-Go status={:?}; Tendermint={:?}",
```
mzabaluev, line 78: The log line could use KV pairs more, rather than format them into the message.

relayer/src/chain/cosmos/version.rs
```diff
@@ -142,6 +155,37 @@ fn parse_ibc_go_version(version_info: &VersionInfo) -> Result<Option<semver::Ver
     }
 }
 
+fn parse_tendermint_version(version_info: &VersionInfo) -> Result<semver::Version, Error> {
+    let module = version_info
+        .build_deps
+        .iter()
+        .find(|&m| m.path.contains(TENDERMINT_MODULE_NAME))
```
mzabaluev, line 162: `contains` should be practically OK, but for extra robustness it might be possible to use `starts_with`, or even match the whole string exactly?

relayer/src/chain/cosmos/version.rs
```diff
@@ -142,6 +155,37 @@ fn parse_ibc_go_version(version_info: &VersionInfo) -> Result<Option<semver::Ver
...
+        Error::version_parsing_failed(
+            module.path.clone(),
+            module.version.clone(),
+            e.to_string(),
+            AppInfo::from(version_info),
+        )
+    })?;
+
+    // Remove the pre-release version to ensure we don't give special treatment to pre-releases.
+    version.pre = semver::Prerelease::EMPTY;
```
mzabaluev, line 184: This could be problematic, since the -pre versions semantically precede the release without the pre suffix. So this is lossily bumping the version to a later one than what the module author intended.

relayer/src/chain/cosmos/version.rs
adizere: We use `contains` because some networks can have their own fork of tendermint, and they're not building on `tendermint/tendermint` but instead on `XXX/tendermint`. I know this is the case often for SDK, e.g., https://github.com/likecoin/cosmos-sdk or https://github.com/irisnet/cosmos-sdk or https://github.com/osmosis-labs/cosmos-sdk but seems for tendermint also: https://github.com/osmosis-labs/tendermint (unclear if they use it, just an example).

I would also like to be more strict here in what strings we accept, but not sure `starts_with` is the right one.

relayer/src/chain/cosmos/version.rs
adizere: fixed, thank you for catching that!

[1570530](https://github.com/informalsystems/ibc-rs/pull/2302/commits/1570530d368dffb907da85104e9e303a904bed7d)

relayer/src/chain/cosmos/version.rs
mzabaluev: > The issue was that networks were using an RC instead of the stable release, e.g.
> 
> SDK module at version '0.42.0-rc0' does not meet compatibility requirements >=0.41.3, <=0.42.6

This semver check behavior is stricter than what I assumed. I expected that 0.42.0-rc0 does succeed 0.41.3 and precede 0.42.6, but would not meet this requirement: >= 0.42.0. Are pre-release suffixes treated as completely incomparable by the comparison logic?

> Which was fixed by trimming the `Prerelease` element. Not sure if you're hinting at a different issue than the one we noticed in the production signaled in #1337.

Thanks for the exposition on this. A proper solution, I think, would be for the compatibility requirements to explicitly include any pre-releases they want to allow, but perhaps practicalities dictate otherwise.

relayer/src/chain/cosmos/version.rs
mzabaluev: If we want to exclude the forks, or any potential `tendermint/tendermint-foo` false matches, it makes even more sense to match with the exact string, doesn't it? A substring match seems too loose as a way to identify modules.

adizere: Ah, good point! I wasn't really clear in my comment above.

We actually want to _allow_ forks, since they are typically minor improvements, sometimes they include additional modules or exotic features that the vanilla versions don't have. Considering we want to allow forks, what would be a way for us to be more strict?

relayer/src/chain/cosmos/version.rs
adizere: > Do pre-release suffixes make versions completely incomparable by the comparison logic?

We could give this a try in a small experiment. Not sure what you want to try exactly, but I managed to reproduce the original issue here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c31232aaa272967f2979abf71e83f258

> A proper solution, I think, would be for the compatibility requirements to explicitly include any pre-releases they want to allow, but perhaps practicalities dictate otherwise.

Indeed, it's not feasible for us to track all pre-releases and test with each of them individually. We assume that pre-releases have the same API as their corresponding release.

relayer/src/chain/cosmos/version.rs
adizere: Ok, then let's leave it as it is for the moment. We'll gain more insight once it's out and people are experimenting with it, then we'll adjust the matching criteria to be more strict (or loose) if needed.

relayer/src/chain/cosmos/version.rs
mzabaluev: Thanks for the playground; it appears that the `semver` crate does treat non-empty pre-release segment as not matching any comparisons with regular versions:

> thread 'main' panicked at 'version 0.42.0-rc0 matching req >0.41.5', src/main.rs:14:5

The logic of `VersionReq::match` is poorly documented in [semver docs](https://docs.rs/semver/1.0.10/semver/index.html), even though the `Version` type as such implements a total ordering that _is_ documented to work in the way I described.
Then there is this:

> Where the various tools differ in their interpretation or implementation of the spec, this crate follows the implementation choices made by Cargo. If you are operating on version numbers from some other package ecosystem, you will want to use a different semver library which is appropriate to that ecosystem.

mzabaluev (approved):
Good to go, with reservations I have filed in #2309. These, however, apply to more than the code introduced in this PR, and can be fixed later.

PR #2284, opened 2022-06-09
Add Beefy & Near client implementations. ICS 02 refactor [CF work]

Wizdave97:
Closes: #XXX

## Description

______

### PR author checklist:
- [ ] Added changelog entry, using [`unclog`](https://github.com/informalsystems/unclog).
- [ ] Added tests: integration (for Hermes) or unit/mock tests (for modules). 
- [ ] Linked to GitHub issue.
- [ ] Updated code comments and documentation (e.g., `docs/`).

### Reviewer checklist:

- [ ] Reviewed `Files changed` in the GitHub PR explorer.
- [ ] Manually tested (in case integration/unit/mock tests are absent).

PR #2303, opened 2022-06-15
Wait for tx hash to succeed regardless of unrecognized IbcEvents

soareschen:
Closes: #2314

## Description

Extracted from #2227 to make the waiting of send_tx to succeed if all TxHash are committed successfully. Previously, send_tx would timeout and fail if unsupported transactions are submitted with events unrecognized by `IbcEvent`, even if the transaction were successful.

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/relayer/src/chain/cosmos/batch.rs
+++ b/relayer/src/chain/cosmos/batch.rs
@@ -1,12 +1,12 @@
 use ibc::events::IbcEvent;
 use ibc_proto::google::protobuf::Any;
 use prost::Message;
+use tendermint::abci::transaction::Hash as TxHash;
 use tendermint_rpc::endpoint::broadcast::tx_sync::Response;
 
 use crate::chain::cosmos::retry::send_tx_with_account_sequence_retry;
 use crate::chain::cosmos::types::account::Account;
 use crate::chain::cosmos::types::config::TxConfig;
-use crate::chain::cosmos::types::tx::TxSyncResult;
 use crate::chain::cosmos::wait::wait_for_block_commits;
 use crate::config::types::{MaxMsgNum, MaxTxSize, Memo};
 use crate::error::Error;
@@ -25,7 +25,7 @@ pub async fn send_batched_messages_and_wait_commit(
         return Ok(Vec::new());
     }
 
-    let mut tx_sync_results = send_messages_as_batches(
+    let tx_hashes = send_messages_as_batches(
         config,
         max_msg_num,
         max_tx_size,
@@ -36,19 +36,14 @@ pub async fn send_batched_messages_and_wait_commit(
     )
     .await?;
 
-    wait_for_block_commits(
+    let events = wait_for_block_commits(
         &config.chain_id,
         &config.rpc_client,
         &config.rpc_address,
         &config.rpc_timeout,
-        &mut tx_sync_results,
+        &tx_hashes,
     )
     .await?;
-
-    let events = tx_sync_results
-        .into_iter()
-        .flat_map(|el| el.events)
-        .collect();
 
     Ok(events)
 }
@@ -89,31 +84,24 @@ async fn send_messages_as_batches(
     account: &mut Account,
     tx_memo: &Memo,
     messages: Vec<Any>,
-) -> Result<Vec<TxSyncResult>, Error> {
+) -> Result<Vec<TxHash>, Error> {
     if messages.is_empty() {
         return Ok(Vec::new());
     }
 
     let batches = batch_messages(max_msg_num, max_tx_size, messages)?;
 
-    let mut tx_sync_results = Vec::new();
+    let mut tx_hashes = Vec::new();
 
     for batch in batches {
-        let events_per_tx = vec![IbcEvent::default(); batch.len()];
-
         let response =
             send_tx_with_account_sequence_retry(config, key_entry, account, tx_memo, batch, 0)
                 .await?;
 
-        let tx_sync_result = TxSyncResult {
-            response,
-            events: events_per_tx,
-        };
-
-        tx_sync_results.push(tx_sync_result);
+        tx_hashes.push(response.hash);
     }
 
-    Ok(tx_sync_results)
+    Ok(tx_hashes)
 }
 
 fn batch_messages(
--- a/relayer/src/chain/cosmos/wait.rs
+++ b/relayer/src/chain/cosmos/wait.rs
@@ -1,20 +1,20 @@
 use core::time::Duration;
+use futures::stream::{FuturesOrdered, TryStreamExt};
 use ibc::core::ics24_host::identifier::ChainId;
 use ibc::events::IbcEvent;
-use ibc::query::{QueryTxHash, QueryTxRequest};
-use itertools::Itertools;
-use std::thread;
 use std::time::Instant;
+use tendermint::abci::transaction::Hash as TxHash;
+use tendermint_rpc::endpoint::tx_search::Response as TxSearchResponse;
 use tendermint_rpc::{HttpClient, Url};
-use tracing::{info, trace};
+use tokio::time::sleep;
+use tracing::info;
 
-use crate::chain::cosmos::query::tx::query_txs;
-use crate::chain::cosmos::types::tx::TxSyncResult;
+use crate::chain::cosmos::query::tx::{all_ibc_events_from_tx_search_response, query_tx_response};
 use crate::error::Error;
 
 const WAIT_BACKOFF: Duration = Duration::from_millis(300);
 
-/// Given a vector of `TxSyncResult` elements,
+/// Given a vector of `TxHash` elements,
 /// each including a transaction response hash for one or more messages, periodically queries the chain
 /// with the transaction hashes to get the list of IbcEvents included in those transactions.
 pub async fn wait_for_block_commits(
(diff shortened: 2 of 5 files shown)


ancazamfir:
Is there an issue we can link here that describes the usecase of "unsuported" transactions? If not, can we open one and provide some context?

relayer/src/chain/cosmos/batch.rs
```diff
@@ -36,20 +36,15 @@ pub async fn send_batched_messages_and_wait_commit(
     )
     .await?;
 
-    wait_for_block_commits(
+    let events = wait_for_block_commits(
```
ancazamfir, line 39: The original intent for `send_batched_messages_and_wait_commit()` was:
- send `N` messages to the chain
- get back exactly `N` `IbcEvent`-s such that result `i` is for message `i`

Not sure this still holds. For non-IBC messages we could still get `IbcEvent::Empty`

soareschen:
> Is there an issue we can link here that describes the usecase of "unsuported" transactions? If not, can we open one and provide some context?

Filed #2314.

relayer/src/chain/cosmos/batch.rs
soareschen: I don't understand the rationale to expect the number of `IbcEvent`s to be exactly the same as the number of messages sent. In theory one message can result in any number of chain events, which can be translated into any number of `IbcEvent`s.

I think a more practical approach is to use `find` through the result list of `IbcEvent`s to find the event of interest, instead of looking at a specific position.

Also the change does not affect the semantics of existing transactions sent. So if the existing code produce messages that result in exact number of `IbcEvent`s, then they would still get that. It would only affect future changes, such as when the code is updated to send non-IBC messages mixed with IBC messages.

ancazamfir: > I don't understand the rationale to expect the number of `IbcEvent`s to be exactly the same as the number of messages sent. In theory one message can result in any number of chain events, which can be translated into any number of `IbcEvent`s.

It's always been the case in practice that an IBC message results in one and only one IBC event. While the events have not been documented in the IBC spec yet, some documentation exists [here](https://github.com/cosmos/ibc-go/blob/main/docs/ibc/events.md)

> 
> I think a more practical approach is to use `find` through the result list of `IbcEvent`s to find the event of interest, instead of looking at a specific position.

You still need to figure out the original message that caused a specific `IbcEvent`.  But sure there are other ways of coding this. Still trying to understand the usecase that breaks the current design.

> 
> Also the change does not affect the semantics of existing transactions sent. So if the existing code produce messages that result in exact number of `IbcEvent`s, then they would still get that.

Yes I was asking in general not necessarily pointing to your changes in this PR. 

> It would only affect future changes, such as when the code is updated to send non-IBC messages mixed with IBC messages.

Yeah, i still want to understand this. I think hermes should only deal with IBC messages and never send messages that have nothing to do with IBC. Probably we should start with what "IBC message" means? And what is a non-IBC message?

adizere:
> Probably we should start with what "IBC message" means? And what is a non-IBC message?

I think SDK upgrade proposals are one example of non-IBC transactions. I doubt there will be others.

@mzabaluev had experience with parsing events from non-IBC tx-es, specifically while working with upgrade proposals in #1909; we closed the PR because it required too many changes: https://github.com/informalsystems/ibc-rs/pull/1909#issuecomment-1096793079. We decide in that PR that Hermes will not wait on tx events after submitting the upgrade proposal, and instead it just prints the tx hash and exits (solution is https://github.com/informalsystems/ibc-rs/pull/1979).

relayer/src/chain/cosmos/wait.rs
```diff
@@ -1,114 +1,103 @@
...
-) -> Result<(), Error> {
-    let TxSyncResult { response, events } = tx_sync_result;
+    timeout: &Duration,
+    tx_hashes: &[TxHash],
+) -> Result<Vec<TxSearchResponse>, Error> {
+    let responses = tx_hashes
+        .iter()
+        .map(|tx_hash| wait_tx_hash(rpc_client, rpc_address, timeout, tx_hash))
+        .collect::<FuturesOrdered<_>>()
+        .try_collect()
```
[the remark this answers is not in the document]
mzabaluev: I believe the first `collect` populates the `FuturesOrdered` with tasks to complete, and `try_collect` is a `futures` method to drive the stream and collect the yielded values, failing fast on error.

mzabaluev:
> Probably we should start with what "IBC message" means? And what is a non-IBC message?

As far as I understand, the `IbcEvent` enum is our strictly typed way to represent various chain events.
It is defined in `ibc`. I don't know if it is used anywhere outside the relayer, but this has been put forward as a justification for keeping it "pure IBC", for example, in #1909.

A discussion on how to systematize chain events should go hand in hand with broader extensibility, which looks like a sane way to accommodate non-Cosmos chains: https://github.com/informalsystems/ibc-rs/pull/2284#issuecomment-1160132296

relayer/src/chain/cosmos/wait.rs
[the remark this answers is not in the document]
soareschen: You can see from Rust Analyzer the types for each step:

```rust
let responses = tx_hashes
  // Iterator of `Hash`
  .iter()
  // Iterator of future of `Result<TxSearchResponse, Error>`
  .map(|tx_hash| wait_tx_hash(rpc_client, rpc_address, timeout, tx_hash))
  // Future of iterator of `Result<TxSearchResponse, Error>`
  .collect::<FuturesOrdered<_>>()
  // Future of `Result<Vec<TxSearchResponse>, Error>`
  .try_collect()
  .await?;
```

Basically we have a collection that in simple form is something like `Vec<Future<Result<TxSearchResponse, Error>>>`, but we want to return `Result<Vec<TxSearchResponse>, Error>`. So we first turn it into `Future<Vec<Result<TxSearchResponse, Error>>>` using `.collect::<FuturesOrdered<_>>()`. Next we turn it into a `Future<Result<Vec<TxSearchResponse>, Error>>` using `.try_collect()`. Finaly we turn it into `Result<Vec<TxSearchResponse>, Error>` using `.await`.

ancazamfir:
> I think SDK upgrade proposals are one example of non-IBC transactions. I doubt there will be others.

Yes but part of that message is this [blob](https://github.com/cosmos/ibc-go/blob/9680cb91a61384593c4521484a071ad3c95c9b34/proto/ibc/core/client/v1/client.proto#L56-L74
) defined in IBC TAO (client). 
There is also an IBC [handler](https://github.com/cosmos/ibc-go/blob/9680cb91a61384593c4521484a071ad3c95c9b34/modules/core/02-client/keeper/proposal.go#L82) in the IBC core client impl in go. So it seems that this operation should be applicable, in theory at least, to any chain type. 

And imo if it is defined in the IBC TAO proto, implemented (even partially) by an IBC core (client in this case) handler then it is a transaction that should generate and IBC event.

relayer/src/chain/cosmos/batch.rs
ancazamfir: > The original intent for `send_batched_messages_and_wait_commit()` was:
> 
> * send `N` messages to the chain
> * get back exactly `N` `IbcEvent`-s such that result `i` is for message `i`
> 
> Not sure this still holds. For non-IBC messages we could get `IbcEvent::Empty`

i checked master and it is broken. When a Tx has N messages, if the Tx fails then all messages should be marked as failed and get back N error events. I have a quick fix in branch `anca/err_for_all_msgs_in_tx`. Also added and corrected some comments. 
I haven't checked/ fixed the packet worker (async sender), at first look it doesn't handle error events here: (https://github.com/informalsystems/ibc-rs/blob/194a1714cf9dc807dd7b4e33726319293848dc58/relayer/src/link/relay_path.rs#L609-L615). 
For the sync sender cases the code looks better.

adizere:
I will close this PR because it is a distraction from doing a stable v1 release.

### Summary

The discussion here & in the associated issue can be summarised as follows: It is unclear what benefits or problem we're solving by allowing non-IBC events to percolate through the Hermes pipeline. 

### Further details

There are 2 cases where non-IBC events could have been potentially interesting for Hermes:
- for chain upgrade proposals, which did not emit any IBC event, but fixed with https://github.com/cosmos/ibc-go/issues/1558
- for `MsgRegisterCounterpartyPayee` part of the ICS29 implementation flow, fixed with https://github.com/cosmos/ibc-go/issues/1566


For the above two cases, it makes sense to have **IBC events** associated with them, and the IBC-go team agreed to fix that. Other cases which will demand **non-IBC events** might occur in the future, and in that case we could reconsider the relayer architecture. Quoted the issue #2314:

> Generally, the same limitation probably also applies to all non-IBC messages, such as local token transfer, staking tokens, token swaps, calling smart contracts, etc. Although Hermes does not currently perform any of such transactions, it is an overly specialized limitation that should not be there.

Hermes is indeed specialized to IBC transactions. That could change in the future to become a more general off-chain orchestrator, but not clear at this point. Until concrete cases such as the above appear & inform our architectural decisions, I don't see a reason to generalize the relayer beyond IBC.

There is another implicit specialization (or assumption) that Hermes makes when submits transactions: it expects each message to result in one IBC event by filtering the `TxSyncResult` data it obtains from the chain. A suggestion by Soares is that Hermes should return a `Vec<Vec<IbcEvent>>` instead of `Vec<IbcEvent>`, allowing a vector of responses (likely with 1 element each) for each message.

Issue #2185, opened 2022-05-05
Client Ugrade Command Not Working in v0.14.1


rhuairahrighairidh:
## Summary of Bug

The command `hermes tx raw upgrade-client` is failing in v0.14.1, but working in v0.14.0.

I believe it's due to this issue: https://github.com/informalsystems/ibc-rs/pull/2021

The command tries to fetch the upgraded consensus state from the last block of the old chain but fails to find it, because it uses the wrong block height.

Previously it used the last block as reported by /status rpc, but now uses another method. When the chain halts, the two methods differ: /status returns the uncommited halted height (listed in the upgrade plan), the other returns the last commited height (one before the upgrade plan).

Note, although technically the last commited consensus state is for one block before the upgrade height, the cosmo-sdk upgrade querier stores it under the upgrade height.

## Version

v0.14.1

## Acceptance Criteria

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
Hmm, this is an interesting corner-case. Thanks for bringing it to our attention!

Issue #2300, opened 2022-06-14
Add mandatory flag `--upgrade-height` to `hermes upgrade client`


plafer:
## Summary
Add `--upgrade-height` to `hermes upgrade client` and have hermes wait for that height until it submits the transaction.

Also document that `hermes upgrade client` must be run when the chain is halted if no `--upgrade-height` is included for the command to succeed.

## Problem Definition

Currently `hermes upgrade client` needs to be run when the chain is halted for the command to succeed, and the error message is confusing when you don't.

We could add an optional flag `--upgrade-height` (or `--halt-height`?) and have hermes wait for that height until it submits the transaction. This would make it less confusing for users, as it is currently unclear that the command needs to be run when the chain is halted.

## Acceptance Criteria

- The `--upgrade-height` (or similar name) option is added to `hermes upgrade client`
- Document that the command must be run when the chain is halted if `--upgrade-height` is not specified

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2304, opened 2022-06-15
Fix client upgrade

plafer:
Closes: #2185

## Description
I did not add a `query_chain_height()` vs `query_latest_application_height()`, as the `query_chain_height()` would only be used in the special case of a client upgrade. Also, I now think it would be best if the cli for client upgrade forced the user to specify the chain upgrade height: it would be better UX IMO, and we wouldn't need to ever query the "chain height" (i.e. tendermint height). To be considered when dealing with #2300.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/relayer-cli/src/commands/tx/client.rs
+++ b/relayer-cli/src/commands/tx/client.rs
@@ -208,7 +208,18 @@ impl Runnable for TxUpgradeClientCmd {
         let client = ForeignClient::find(src_chain, dst_chain, &self.client_id)
             .unwrap_or_else(exit_with_unrecoverable_error);
 
-        let outcome = client.upgrade();
+        // Assumption: this query is run while the chain is halted as a result of an upgrade
+        let src_upgrade_height = {
+            let src_application_height = match client.src_chain().query_latest_height() {
+                Ok(height) => height,
+                Err(e) => Output::error(format!("{}", e)).exit(),
+            };
+
+            // When the chain is halted, the application height reports a height
+            // 1 less than the halted height
+            src_application_height.increment()
+        };
+        let outcome = client.upgrade(src_upgrade_height);
 
         match outcome {
             Ok(receipt) => Output::success(receipt).exit(),
@@ -280,7 +291,20 @@ impl TxUpgradeClientsCmd {
         src_chain: Chain,
     ) -> Result<Vec<IbcEvent>, Error> {
         let client = ForeignClient::restore(client_id, dst_chain, src_chain);
-        client.upgrade().map_err(Error::foreign_client)
+
+        let src_upgrade_height = {
+            let src_application_height = match client.src_chain().query_latest_height() {
+                Ok(height) => height,
+                Err(e) => Output::error(format!("{}", e)).exit(),
+            };
+
+            // When the chain is halted, the application height reports a height
+            // 1 less than the halted height
+            src_application_height.increment()
+        };
+        client
+            .upgrade(src_upgrade_height)
+            .map_err(Error::foreign_client)
     }
 }
 
--- a/relayer-cli/src/commands/tx/client.rs
+++ b/relayer-cli/src/commands/tx/client.rs
@@ -6,6 +6,7 @@ use abscissa_core::{Command, Runnable};
 use ibc::core::ics02_client::client_state::ClientState;
 use ibc::core::ics24_host::identifier::{ChainId, ClientId};
 use ibc::events::IbcEvent;
+use ibc::Height;
 use ibc_relayer::chain::handle::ChainHandle;
 use ibc_relayer::chain::requests::{
     IncludeProof, PageRequest, QueryClientStateRequest, QueryClientStatesRequest,
@@ -252,12 +253,29 @@ impl Runnable for TxUpgradeClientsCmd {
             Err(e) => Output::error(format!("{}", e)).exit(),
         };
 
+        let src_upgrade_height = {
+            let src_application_height = match src_chain.query_latest_height() {
+                Ok(height) => height,
+                Err(e) => Output::error(format!("{}", e)).exit(),
+            };
+
+            // When the chain is halted, the application height reports a height
+            // 1 less than the halted height
+            src_application_height.increment()
+        };
+
         let results = config
             .chains
             .iter()
             .filter_map(|chain| {
-                (self.src_chain_id != chain.id)
-                    .then(|| self.upgrade_clients_for_chain(&config, src_chain.clone(), &chain.id))
+                (self.src_chain_id != chain.id).then(|| {
+                    self.upgrade_clients_for_chain(
+                        &config,
+                        src_chain.clone(),
+                        &chain.id,
+                        &src_upgrade_height,
+                    )
+                })
             })
             .collect();
 
@@ -275,6 +293,7 @@ impl TxUpgradeClientsCmd {
         config: &Config,
         src_chain: Chain,
         dst_chain_id: &ChainId,
+        src_upgrade_height: &Height,
     ) -> UpgradeClientsForChainResult {
         let dst_chain = spawn_chain_runtime_generic::<Chain>(config, dst_chain_id)?;
 
@@ -286,7 +305,14 @@ impl TxUpgradeClientsCmd {
             .map_err(Error::relayer)?
             .into_iter()
             .filter_map(|c| (self.src_chain_id == c.client_state.chain_id()).then(|| c.client_id))
-            .map(|id| TxUpgradeClientsCmd::upgrade_client(id, dst_chain.clone(), src_chain.clone()))
+            .map(|id| {
+                TxUpgradeClientsCmd::upgrade_client(
+                    id,
+                    dst_chain.clone(),
+                    src_chain.clone(),
+                    src_upgrade_height,
+                )
+            })
             .collect();
 
         Ok(outputs)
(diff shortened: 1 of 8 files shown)


relayer/src/chain/cosmos.rs
```diff
@@ -779,11 +788,19 @@ impl ChainEndpoint for CosmosSdkChain {
         crate::telemetry!(query, self.id(), "query_upgraded_client_state");
 
         // Query for the value and the proof.
-        let tm_height =
+        let upgrade_height =
             Height::try_from(request.height.revision_height).map_err(Error::invalid_height)?;
+        let query_height = Height::try_from(
```
adizere, line 793: Wouldn't it work if we do `Height::decrement` ?

https://github.com/informalsystems/ibc-rs/blob/ad2d8c59083c4670c4ed0d8b6d446040fa7044f1/modules/src/core/ics02_client/height.rs#L65-L67

relayer/src/foreign_client.rs
```diff
@@ -430,25 +441,31 @@ impl<DstChain: ChainHandle, SrcChain: ChainHandle> ForeignClient<DstChain, SrcCh
...
-            ForeignClientError::client_upgrade(
-                self.id.clone(),
-                self.src_chain.id(),
-                "failed while querying src chain for latest height".to_string(),
-                e,
-            )
-        })?;
-
-        info!("[{}] upgrade Height: {}", self, src_height);
+    /// Create and send a transaction to perform a chain upgrade.
```
adizere, line 444: This method does not perform chain upgrade, only client state upgrade.


```suggestion
    /// Create and send a transaction to perform a client upgrade.
```

relayer/src/foreign_client.rs
```diff
@@ -942,7 +959,7 @@ impl<DstChain: ChainHandle, SrcChain: ChainHandle> ForeignClient<DstChain, SrcCh
 
     /// Returns a vector with a message for updating the client to height `target_height`.
     /// If the client already stores a consensus state for this height, returns an empty vector.
-    pub fn build_update_client_with_trusted(
+    pub fn wait_and_build_update_client_with_trusted(
```
adizere, line 962: Tip top!

relayer-cli/src/commands/tx/client.rs
```diff
@@ -208,7 +208,18 @@ impl Runnable for TxUpgradeClientCmd {
         let client = ForeignClient::find(src_chain, dst_chain, &self.client_id)
             .unwrap_or_else(exit_with_unrecoverable_error);
 
-        let outcome = client.upgrade();
+        // Assumption: this query is run while the chain is halted as a result of an upgrade
```
adizere, line 211: Consider outputting an `info` or `warn` here saying that Hermes is using `src_upgrade_height` and it assumes chain `src_chain` is halted for update at that height.

adizere (commented):
Left some improvement suggestions.

relayer-cli/src/commands/tx/client.rs
```diff
@@ -280,7 +291,20 @@ impl TxUpgradeClientsCmd {
...
+                Err(e) => Output::error(format!("{}", e)).exit(),
+            };
+
+            // When the chain is halted, the application height reports a height
+            // 1 less than the halted height
+            src_application_height.increment()
+        };
+        client
+            .upgrade(src_upgrade_height)
+            .map_err(Error::foreign_client)
```
ancazamfir, line 307: Should we do this only once in `upgrade_clients_for_chain()` or even earlier in `run()`?

relayer-cli/src/commands/tx/client.rs
```diff
@@ -208,7 +208,18 @@ impl Runnable for TxUpgradeClientCmd {
...
+        let src_upgrade_height = {
+            let src_application_height = match client.src_chain().query_latest_height() {
+                Ok(height) => height,
+                Err(e) => Output::error(format!("{}", e)).exit(),
+            };
+
+            // When the chain is halted, the application height reports a height
+            // 1 less than the halted height
+            src_application_height.increment()
+        };
```
ancazamfir, line 221: I guess this will go away once we have an `upgrade_height` flag? Will it be optional? In this case this code should probably go in the relayer foreign_client `upgrade()` and just pass an option there.

relayer/src/foreign_client.rs
ancazamfir, line 962: Maybe add a comment on what we are waiting for? Not your change but I think the use of `src_network_latest_...` here is misleading as it's `src_application_latest...`. 
And probably we will not be able to avoid to introduce `query_chain_latest..` vs `query_application_latest...`. 

Right now the code is confusing, just trying to figure out how much we can fix in this PR, or later in #2300 (where we will need to wait on chain) or yet another time.

relayer/src/chain/cosmos.rs
plafer: `Height` in this file is actually `tendermint::block::height::Height`, and it only has an `increment()`. I thought about renaming it to `TMHeight`, but decided against to reduce the number of merge conflicts with my other Height zero PR. I can address it there though.

The alternative way to write it is:
```rust
let query_height = Height::try_from(
            request
                .height
                .decrement()
                .expect("height overflow")
                .revision_height,
        )
        .map_err(Error::invalid_height)?;
```
I personally prefer the current one, since although not ideal, at least it's clearer that we're doing `upgrade_height - 1`.

relayer/src/foreign_client.rs
plafer: Good point!

relayer-cli/src/commands/tx/client.rs
plafer: Right, I am currently of the opinion that the cli should make the `upgrade_height` flag mandatory, as the operation really isn't a "give me the latest height", it's "give me the chain halt height" (otherwise it fails). We're only dealing with this "application height vs tendermint height in the case of an upgrade, and iff the user ran the command at the right time" mess because we allow the user to run the command "at just the right time", which is a bad design IMO. With the `upgrade_height` flag, all this complexity goes away.

relayer-cli/src/commands/tx/client.rs
plafer: This is only meant as a temporary hack until we introduce the `--upgrade_height` flag (which IMO should be before v1.0). See [this comment](https://github.com/informalsystems/ibc-rs/pull/2304/files#r899070005).

relayer-cli/src/commands/tx/client.rs
plafer: Do you mean the upgrade, or querying the `src_upgrade_height`? In both cases, I don't see how we can do it earlier. How I understand it:

+ `src_upgrade_height`: must be queried once per "source client", so we can't just query it once for all clients
+ `client.upgrade()`: we need to do this once per "source client" as well, so can't be done earlier.

Is this what you meant?

relayer/src/foreign_client.rs
plafer: Well really what is meant here is `src_application_latest_height`, right? As in we want the application to have reached that height before continuing, since we'll be querying for client state and what not? So simply changing the name here should be fine?

I can also add a comment to specify what we're waiting for; unfortunately the name is long enough as it is, I don't think we can pack more information in there 😅

plafer: [1807e7f](https://github.com/informalsystems/ibc-rs/pull/2304/commits/1807e7f96554bfbc212fe24e31d5f45bfcafa616)

relayer/src/chain/cosmos.rs
adizere: You're converting from `ibc::Height` to `tendermint::block::Height` in the line above, no?

```rust
Height::try_from(request.height.revision_height)
```

but this is a minor concern. It would improve readability if we could simplify the snipper, but feel free to ignore if the solution is not obvious.

relayer-cli/src/commands/tx/client.rs
ancazamfir: Just `src_upgrade_height`. This is the halt height for the chain that is upgraded, let it be `ibc-0`.  And let's assume we only have one other chain `ibc-1`.
When we issue `hermes upgrade clients ibc-0` we want all affected clients hosted on all of the other chains (on `ibc-1` in our example, clients that verify from `ibc-0`), to be upgraded so they can verify headers from the upgraded `ibc-0` chain.  

> + `src_upgrade_height`: must be queried once per "source client chain", so we can't just query it once for all clients

Not sure what "source client chain" is. Is it the chain for which a client verifies? Is it `ibc-0` in the example? In this case it can only be the one specified in the CLI (`ibc-0`). So we can do it in `TxUpgradeClientsCmd::run()`:
```
        let src_upgrade_height = {
            let src_application_height = match src_chain.query_latest_height() {
                Ok(height) => height,
                Err(e) => Output::error(format!("{}", e)).exit(),
            };

            // When the chain is halted, the application height reports a height
            // 1 less than the halted height
            src_application_height.increment()
        };
```
> + `client.upgrade()`: we need to do this once per "source client chain" as well, so can't be done earlier.

`client.upgrade()` should still be called for all clients on the other chains, `ibc-1` in this example.

relayer-cli/src/commands/tx/client.rs
ancazamfir: ok, let's leave it until #2300.

relayer-cli/src/commands/tx/client.rs
ancazamfir: btw just tried computing `src_upgrade_height` once for all,  [these](https://gist.github.com/ancazamfir/4586f5f8a535b30f2636552bbe969e9d) diffs seem ok

relayer-cli/src/commands/tx/client.rs
plafer: [83d6396](https://github.com/informalsystems/ibc-rs/pull/2304/commits/83d6396c5b34fe92d17088fce1ad4067e7dd907f)

relayer-cli/src/commands/tx/client.rs
plafer: Yep you're right, I got confused between source/dest again... Can't wait for `host` & `reference` naming 😅

plafer: [49df45f](https://github.com/informalsystems/ibc-rs/pull/2304/commits/49df45f7cac1934b116349e3006036df4f1aef51)

plafer:
I would vote for a different PR; I never looked into our integration framework, and I currently don't have the capacity to delve into it. But I agree that it is desirable; I'll create one and assign myself as an opportunity to learn it.

ancazamfir (approved):
Great work @plafer! Happy to see the original issue solved and looking forward to the #2300 improvement.

adizere (approved):

PR #2318, opened 2022-06-20
Deduplicate Docker images for gaiad

mzabaluev:
The `ibc-0` and `ibc-1` images were only really different in the content of the chain data directory, which is copied from the repository checkout.

To save building Gaia twice and occupying twice the space on the Docker repo, build and push the gaiad image once and reuse it for the two node container instances in CI, specifying volume mounts of the respective chain data directories from the repository checkout.


adizere:
What is the status and intention with this PR?

romac:
Superseded by #2471.

Issue #2333, opened 2022-06-24
Sending N messages to a chain does not always return N responses


ancazamfir:
## Summary of Bug
When hermes submits `N` messages to a chain it should get back `N` responses. This is currently broken when Tx errors occur.

When the messages are sent to cosmos-sdk, they may be split in `M` transactions. If one of the Tx-es fails all messages are considered failed (the index in the error indicates the position of the message that caused the failure). The Tx failure should recorded for all its messages. 

## Version
master

## Steps to Reproduce
- start two chains
- create channel
- send `N` transactions making sure `max_msg_num` is smaller than `N`
(tested with `max_msg_num = 5` and send `N=15`).
- there may be other ways to cause the error, i used the `ft-transfer` with a big amount such that the wallet is drained at some message `x, x >  max_msg_num`
(tested with wallet that has 100M samoleans and sent 15 transfers of 9M each)
- the output should show `N` results

## Acceptance Criteria

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2334, opened 2022-06-24
Propagate deliver-tx error for all messages in a failed Tx

ancazamfir:
Closes: #2333

## Description

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


adizere:
I followed the instructions in the issue but I'm likely doing something wrong, because I can't reproduce the error. Instead of getting `deliverTx` errors, I'm catching the error at the estimate gas step:


```
2022-06-28T11:08:53.131609Z ERROR ThreadId(11) send_tx_with_account_sequence_retry{id=ibc-1}:estimate_gas: failed to simulate tx. propagating error to caller: gRPC call failed with status: status: Unknown, message: "failed to execute message; message index: 4: 6998375samoleans is smaller than 9000000samoleans: insufficient funds", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "x-cosmos-block-height": "35000"} }
Error: transfer error: failed while submitting the Transfer message to chain ibc-1: gRPC call failed with status: status: Unknown, message: "failed to execute message; message index: 4: 6998375samoleans is smaller than 9000000samoleans: insufficient funds", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "x-cosmos-block-height": "35000"} }
```

This happens while I'm running eg.
> hermes tx raw ft-transfer ibc-1 ibc-0 transfer channel-0 9000000 -n 15 -o 200 -k wallet1

ancazamfir:
> I followed the instructions in the issue but I'm likely doing something wrong, because I can't reproduce the error. Instead of getting `deliverTx` errors, I'm catching the error at the estimate gas step:
> 
> ```
> 2022-06-28T11:08:53.131609Z ERROR ThreadId(11) send_tx_with_account_sequence_retry{id=ibc-1}:estimate_gas: failed to simulate tx. propagating error to caller: gRPC call failed with status: status: Unknown, message: "failed to execute message; message index: 4: 6998375samoleans is smaller than 9000000samoleans: insufficient funds", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "x-cosmos-block-height": "35000"} }
> Error: transfer error: failed while submitting the Transfer message to chain ibc-1: gRPC call failed with status: status: Unknown, message: "failed to execute message; message index: 4: 6998375samoleans is smaller than 9000000samoleans: insufficient funds", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "x-cosmos-block-height": "35000"} }
> ```
> 
> This happens while I'm running eg.
> 
> > hermes tx raw ft-transfer ibc-1 ibc-0 transfer channel-0 9000000 -n 15 -o 200 -k wallet1

Sorry forgot to clarify, you need to inspect the events, e.g. 
```
--- a/relayer/src/transfer.rs
+++ b/relayer/src/transfer.rs
@@ -182,6 +182,7 @@ pub fn build_and_send_transfer_messages<SrcChain: ChainHandle, DstChain: ChainHa
         .send_messages_and_wait_commit(TrackedMsgs::new_static(msgs, "ft-transfer"))
         .map_err(|e| TransferError::submit(packet_src_chain.id(), e))?;

+    dbg!(&events);
     // Check if the chain rejected the transaction
     let result = events
         .iter()
```

relayer/src/chain/cosmos/wait.rs
```diff
@@ -92,11 +94,39 @@ async fn update_tx_sync_result(
...
+                // This gives more information on the (first) message that caused the error.
+                // There is no evaluation of messages after the first failed message.
+                // If n messages are sent in a Tx, and the error index is i then:
+                //  - messages 1..i-1 passed deliverTx,
+                //  - message i failed deliverTx,
+                //  - deliverTx has not run for i+1..n
+                // Same is true for checkTx (above)
+                // TODO - maybe add at some point more error details to the `IbcEvent::ChainError()`
+                // below.
+                Some(err) => {
```
soareschen, line 120: I am not sure if I understand the error case here. I know that if one of the messages failed in the transaction, the entire transaction is reverted. Shouldn't the response code for `DeliverTx` be error if any of the message fails?

Based on my understanding, the result `Vec<IbcEvent>` given back from `query_txs` use [`all_ibc_events_from_tx_search_response`](relayer/src/chain/cosmos/query/tx.rs) to map the chain events to `IbcEvent`s, which in turns use [from_tx_response_event](https://github.com/informalsystems/ibc-rs/blob/05c8bc7df643202a6e74aa4b2abbb9396c7621bd/modules/src/events.rs#L342) for the conversion.

Since the `IbcEvent::ChainError` event is internal to ibc-rs and not correspond to any chain event, I don't see how `events_per_tx` can contain any `ChainError`.

soareschen (dismissed):

soareschen:
@romac @ancazamfir I added a new `TxStatus` field in `TxSyncResult` so that pending TXs have the status `TxStatus::Pending`, and that `wait_for_block_commits` returns when all status become `TxStatus::ReceivedResponse`. With that the `IbcEvent::Empty` event can be removed.

I also added a test `test_error_events` to test the number of error events when the transaction failed. If you copy the same test to `master` branch you should see the test failed, while on this branch it is fixed.

```
cargo test -p ibc-integration-test -- test_error_events
```

I also want to note that there is still [one call in `Link`](https://github.com/informalsystems/ibc-rs/blob/7f0bf8f52fde8b1b6ab5acbed8736cd54a23ae1b/relayer/src/link/pending.rs#L132-L134) that returns a single `ChainError` event. In this case the original message count is lost in translation, so it is not possible to fill in the error events with the original count. This in turn means that the [`RelaySummary` returned](https://github.com/informalsystems/ibc-rs/blob/7f0bf8f52fde8b1b6ab5acbed8736cd54a23ae1b/relayer/src/link/pending.rs#L233-L259) would still have only one error event.

If you think that needs to be fixed, we should do it in a separate PR and have this PR merged first.

romac:
> With that the IbcEvent::Empty event can be removed.

Sweet! That's a very nice solution, thanks a lot!

> I also want to note that there is still [one call in Link](https://github.com/informalsystems/ibc-rs/blob/7f0bf8f52fde8b1b6ab5acbed8736cd54a23ae1b/relayer/src/link/pending.rs#L132-L134) that returns a single ChainError event. In this case the original message count is lost in translation, so it is not possible to fill in the error events with the original count. This in turn means that the [RelaySummary returned](https://github.com/informalsystems/ibc-rs/blob/7f0bf8f52fde8b1b6ab5acbed8736cd54a23ae1b/relayer/src/link/pending.rs#L233-L259) would still have only one error event.

I am not fully sure I understand what's going on here, but let's get this merged already and follow-up on this at a later stage. Tracked in #2374, feel free to update the description.

romac:
Oh and thanks for the corresponding test 🙏

.changelog/unreleased/bug-fixes/ibc-relayer/2333-error-for-all-msgs.md
```diff
@@ -0,0 +1,2 @@
+- Replicate the deliver_tx error for all messages in the Tx
```
romac, line 1:
```suggestion
- When Hermes submits `N` messages to a chain, it will now always gets back `N` responses, even in the presence of errors.
```

romac (approved):

PR #2346, opened 2022-06-27
Remove the concept of a "zero Height"

plafer:
Closes: #1009

## Description

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


modules/src/core/ics03_connection/msgs/conn_open_ack.rs
```diff
@@ -29,11 +29,8 @@ pub struct MsgConnectionOpenAck {
 impl MsgConnectionOpenAck {
     /// Getter for accessing the `consensus_height` field from this message. Returns the special
     /// value `Height(0)` if this field is not set.
```
mzabaluev, line 31: The comments like this need to be updated too.

mzabaluev (commented):
Have you thought of making the `revision_height` member of `Height` a `NonZeroU64`?
And maybe making both members non-public.
This will enforce avoidance of 0, and ensure that `std::mem::size_of::<Option<Height>>()` is 16.

plafer:
> Have you thought of making the `revision_height` member of `Height` a `NonZeroU64`? And maybe making both members non-public while we are at it. This will enforce avoidance of 0, and ensure that `std::mem::size_of::<Option<Height>>()` is 16.

Yes, my original plan was to use `NonZeroU64`, but then creating a new `Height` looks like:
```rust
// e.g. when building from a tendermint header
Height::new(chain_id(), header_revision_number.try_into().map_err(|_| ...)?)
```
Basically since all the other types use `u64`, building a `NonZeroU64` from a `u64` requires a `try_into().map_err()`, and so I found it more elegant to have `new()` return a `Result` instead.

I am also planning on making the members private.

modules/src/applications/transfer/msgs/transfer.rs
```diff
@@ -63,8 +63,27 @@ impl TryFrom<RawMsgTransfer> for MsgTransfer {
...
+        // to this.
         let timeout_height: Option<Height> = raw_msg
             .timeout_height
+            .and_then(|raw_height| {
+                if raw_height.revision_number == 0 && raw_height.revision_height == 0 {
+                    None
+                } else {
+                    Some(raw_height)
+                }
+            })
```
romac, line 86: To clean this up a bit, can we perhaps add a `From<RawHeight>` impl for `Option<Height>`?

Or perhaps just use `Height::new(raw_height.revision_number, raw_height.revision_height).ok()`.

I think the former instance might be useful but perhaps the latter option is cleaner, what do you think?

plafer: `TimeoutHeight` def: [5660a99](https://github.com/informalsystems/ibc-rs/pull/2346/commits/5660a999ac192a0fc29d1de1797161ac212782c3)
Propagate changes: [69234fb](https://github.com/informalsystems/ibc-rs/pull/2346/commits/69234fbe7d3ee615f5dcb3dbca37203ba4aec30d)

modules/src/core/ics04_channel/packet.rs
```diff
@@ -188,19 +198,46 @@ impl core::fmt::Display for Packet {
...
+        // to this.
+        let packet_timeout_height: TimeoutHeight = raw_pkt
             .timeout_height
-            .ok_or_else(Error::missing_height)?
-            .into();
+            .try_into()
+            .map_err(|_| Error::invalid_timeout_height())?;
 
-        if packet_timeout_height.is_zero() && raw_pkt.timeout_timestamp == 0 {
+        if !packet_timeout_height.has_timeout() && raw_pkt.timeout_timestamp == 0 {
```
hu55a1n1, line 240: Unrelated to this PR, but I think the spec allows a packet without a timeout height and timestamp, so this is a bug and we shouldn't have this check here.

modules/src/core/ics04_channel/timeout.rs
```diff
@@ -0,0 +1,124 @@
...
+    pub fn has_expired(&self, height: &Height) -> bool {
+        match self.0 {
+            Some(timeout_height) => *height >= timeout_height,
+            // When there's no timeout, heights are never expired
+            None => false,
+        }
+    }
+
+    // FIXME: Find better name
+    pub fn has_timeout(&self) -> bool {
```
hu55a1n1, line 54: I find the `has_timeout()` name confusing, how about `fn never_expires() -> bool` because that is what we want to check elsewhere in the code. `TimeoutHeight` could also be modelled like so ->
```rust
pub enum TimeoutHeight {
  Never,
  At(Height)
}
```

modules/src/core/ics04_channel/packet.rs
```diff
@@ -106,7 +113,7 @@ pub struct Packet {
     pub destination_channel: ChannelId,
     #[serde(serialize_with = "crate::serializers::ser_hex_upper")]
     pub data: Vec<u8>,
-    pub timeout_height: Height,
+    pub timeout_height: TimeoutHeight,
     pub timeout_timestamp: Timestamp,
```
hu55a1n1, line 117: Should we add a similar `TimeoutTimestamp` newtype for `Timestamp` in the future (in a different PR)?

modules/src/core/ics04_channel/timeout.rs
hu55a1n1: Seems like we wouldn't need the `has_timeout()` function at all if we fix the bug I described earlier (https://github.com/informalsystems/ibc-rs/pull/2346/files#r910757508).

hu55a1n1:
Great work @plafer! :+1:  Just left some comments and questions.

modules/src/core/ics04_channel/packet.rs
plafer: As discussed, `Timestamp` already serves this purpose.

modules/src/core/ics04_channel/timeout.rs
plafer: Good idea!

[95eb992](https://github.com/informalsystems/ibc-rs/pull/2346/commits/95eb992deb7a32d06c44b41215601aff9311f3f8)

modules/src/core/ics04_channel/timeout.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    // Note: it is important for `revision_number` to also be `0`, otherwise
+    // packet commitment proofs will be incorrect (proof construction in
+    // `ChannelReader::packet_commitment()` uses both `revision_number` and
+    // `revision_height`). Note also that ibc-go conforms to this convention.
+    fn try_from(raw_height: RawHeight) -> Result<Self, Self::Error> {
+        if raw_height.revision_number == 0 && raw_height.revision_height == 0 {
+            Ok(TimeoutHeight(None))
+        } else {
+            let height: Height = raw_height.try_into()?;
```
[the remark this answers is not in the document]
plafer: As discussed, the current implementation is correct. I added a test and explanation here: [c0fafdb](https://github.com/informalsystems/ibc-rs/pull/2346/commits/c0fafdbbf9052699c8ed5747c5956fd7e3de46d4)

modules/src/core/ics03_connection/msgs/conn_open_ack.rs
plafer: I didn't find any other comments that referred to "Height zero"

[4a9dca7](https://github.com/informalsystems/ibc-rs/pull/2346/commits/4a9dca7bf1fba72aa4f25806a9442616a07c1c22)

romac (approved):
Awesome work @plafer, thanks so much!

Issue #2155, opened 2022-04-28
Hermes infinite retry clear packet


lyqingye:
## Summary of Bug

https://github.com/informalsystems/ibc-rs/blob/master/relayer/src/worker/packet.rs#L120
``` rust
 spawn_background_task(span, Some(Duration::from_millis(200)), move || {
        if let Ok(cmd) = cmd_rx.try_recv() {
            // infinity retry
            retry_with_index(retry_strategy::worker_stubborn_strategy(), |index| {
                handle_packet_cmd(
                    &mut is_first_run,
                    &link.lock().unwrap(),
                    clear_on_start,
                    clear_interval,
                    &path,
                    cmd.clone(),
                    index,
                )
            })
            .map_err(|e| TaskError::Fatal(RunError::retry(e)))?;
        }

        Ok(Next::Continue)
    })
```

Hermes will always clear the packet at a fixed height. When Hermes cannot build the packet proof at this height (the chain has been prune), Hermes will fall into an endless loop

## Version
hermes 0.14.0

## Acceptance Criteria

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
How does this problem manifest more precisely? It is it the case that Hermes stops operating on "live" packets and is stubbornly retrying "old" packets? Or can you explain the issue in more detail?

Note: The infinite retry approach was added as a requested feature from some operators (https://github.com/informalsystems/ibc-rs/issues/1290). I can see how this can be problematic when the chain data was pruned, so thank you for bringing this up!

lyqingye:
when Hermes executes clear packet, it will take too long in the following cases: 
1. If there is no packet to relay on the long-term chain, it will take a long time to build the update client next time(verify light client to target height)
2. an error was encountered during the clear packet process
3. rpc delay

when clear-height + pruning-keep-recent > latest height, the packet of this height will never be cleared. He cannot build the packet proof because the state  has been prune

conclusion：
when hermes cannot build proof, hermes should stop retrying, because retrying at this height will never succeed.

lyqingye:
I think Hermes should improve the performance of clear packet instead of relying on "real-time relay". You can check my suggestions under this [issue](https://github.com/informalsystems/ibc-rs/issues/2087)

adizere:
>conclusion：
[...]

PR #2348, opened 2022-06-28
Do not retry clearing with same height forever

ancazamfir:
Closes: #2155

## Description
`spawn_packet_cmd_worker` runs `handle_packet_cmd` passing a "command" that can be `NewBlock`.
The first thing it does is to try clearing packets. If this fails with ignorable error, the old code will retry immediately with same command and height. If the error persists (as in #2155) then the worker enters an infinite loop and new IBC events (coming via `IbcEvent` commands) are never processed.

I believe the infinite loop started with https://github.com/informalsystems/ibc-rs/pull/2238. Before that the worker would abort (due to `TaskError::Fatal(RunError::retry(e))`)

The propose change is to move to the next command even if previous has failed. The reasons are:
- it avoids the infinite loop
- the failed events will still be processed at next clear interval with a fresh height
- at different levels down in `handle_packet_cmd` there are retries mechanisms for MAX_RETRIES (current value hardcoded at 5). 

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/improvements/ibc-relayer/2155-remove-infinite-loop.md
+++ b/.changelog/unreleased/improvements/ibc-relayer/2155-remove-infinite-loop.md
@@ -1,2 +1,2 @@
-- Do not retry infinitely on cmd handling failure
+- Do not retry indefinitely on command handling failure in the packet worker
   ([#2155](https://github.com/informalsystems/ibc-rs/issues/2155))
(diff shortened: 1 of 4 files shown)


.changelog/unreleased/improvements/ibc-relayer/2155-remove-infinite-loop.md
```diff
@@ -0,0 +1,2 @@
+- Do not retry infinitely on cmd handling failure
```
romac, line 1:
```suggestion
- Do not retry infinitely on command handling failure in the packet worker
```

Issue #2349, opened 2022-06-28
Better message, usage, option help and completion for CLIs


ancazamfir:
## Summary
So not sure this is a feature or a bug or if there is anything we can do about it.
With the new CLIs in the USAGE and help are strange and not correct.
For example:
```
$ hermes create channel -h
hermes-create-channel
Create a new channel between two chains using a pre-existing connection

USAGE:
    hermes create channel [OPTIONS] --a-chain <A_CHAIN_ID> --a-port <A_PORT_ID> --b-port <B_PORT_ID>

OPTIONS:
        --a-chain <A_CHAIN_ID>
            Identifier of the side `a` chain for the new channel

        --a-connection <A_CONNECTION_ID>
            Identifier of the connection on chain `a` to use in creating the new channel.

        --a-port <A_PORT_ID>
            Identifier of the side `a` port for the new channel

        --b-chain <B_CHAIN_ID>
            Identifier of the side `b` chain for the new channel

        --b-port <B_PORT_ID>
            Identifier of the side `b` port for the new channel

        --channel-version <VERSION>
            The version for the new channel

    -h, --help
            Print help information

        --new-client-connection
            Indicates that a new client and connection will be created underlying the new channel

        --order <ORDER>
            The channel ordering, valid options 'unordered' (default) and 'ordered' [default:
            ORDER_UNORDERED]

        --yes
            Skip new_client_conn confirmation
```

There are a few issues for this one:
- `Create a new channel between two chains using a pre-existing connection` is wrong, that is not the only option. Seems that a blank line (`create.rs`) in the description drops everything after when displayed with -h
- the USAGE is wrong. not sure how much control we have with abscissa and clap but `--a-chain` appears both in `USAGE` and in OPTIONS (as if it could be specified twice). The required options should not a
-  the list of options in alphabetical order makes reading them hard, e.g. `-h, --help` appears in the middle of the option list.
- completion seems to work for a few options and then it stops working

Maybe most or all of these issues where there before but we should cleanup as much as possible before v1.0.

## Acceptance Criteria

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ljoss17:
> There are a few issues for this one:
> 
[...]

PR #2352, opened 2022-06-29
CLI unit-testing for all commands with at least one argument

ljoss17:
Closes: #2358 

## Description

This PR adds unit-testing parsing Hermes commands, ~as well as improving the commands help output.
The help output follows the request from issue https://github.com/informalsystems/ibc-rs/issues/2349~. Changes for issue https://github.com/informalsystems/ibc-rs/issues/2349 will be done in a separate PR.

### Changes

This PR has the following changes:

* Adding unit tests for all Hermes commands with at least one argument.
* Instead of manually validating the flags, the validation is done by `clap` for commands `create channel`, `create connection`, `keys add` and `keys delete`.
* Minor changes to the guide.
* Updated the ADR 010 with the new `--yes` optional flag for the command `create channel`.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


ljoss17:
> * Create a new channel between two chains using a pre-existing connection is wrong, that is not the only option. Seems that a blank line (create.rs) in the description drops everything after when displayed with -h
> * the USAGE is wrong. not sure how much control we have with abscissa and clap but --a-chain appears both in USAGE and in OPTIONS. The required flags should not be [OPTIONS]
>    * Not sure about this in fact as we wouldn't have a way to describe the required flags. Maybe have a [REQUIRED] section
> * the list of options in alphabetical order makes reading them hard, e.g. -h, --help appears in the middle of the option list. Is there a way to control the order in which they are listed?
> * completion seems to work for the first few options and then it stops working.

@ancazamfir regarding the issues you found, I was able to fix the following:
* Changing the short message for the short help, `-h`, of `create channel` and improved the long help, `--help`.
* Changed the USAGE for `create channel`, `create connection`, `keys add` and `keys delete` to include the two different usages.
* Added a FLAGS header for required flags, in addition to the OPTIONS header for optional flags, for all the commands.

What do you think about these fixes for the first three issues ?

Regarding the ordering, it is possible to arbitrarily reorder the list of options/flags, but I am not sure which order would be better than alphabetical order. Do you have a preference for this ?

I will try to reproduce the completion issue and if I manage to, try and fix it.

romac (approved):
❤️ these unit tests! Great job, @ljoss17 🎉

Issue #2174, opened 2022-05-03
Configuration and problems with gas_adjustment parameter


loredanacirstea:
## Summary

Allow configurable gas limits beyond the `gas_adjustment = 1.0`

`ibc-rs` tries to do a gas estimation before sending a transaction. If a gas estimation can be computed, then the gas limit value can only be tweaked with `gas_adjustment` (`0.0.` - `1.0`). If you need to increase this gas estimation more, you need to change the code and rebuild from source.

## Problem Definition

I personally needed to have custom gas because I was testing cutting-edge code and the gas estimation was off - `gas_adjustment = 1.0` was not enough.
I don't see drawbacks to introducing a gas limit overwrite variable that is properly commented.

## Proposal

1) increase `gas_adjustment` range (I am not aware of the drawbacks that made the team limit this value)
2) introduce a `gas_overwite` variable that can be used to provide a custom value that overrides all gas estimations

## Acceptance Criteria

- [ ] User is free to customize/overwrite the gas limit.
- [ ] The value in config.toml for `gas_adjustment` is currently set to `1.0`, and it should reflect the actual default (`0.1`)
    - https://github.com/informalsystems/ibc-rs/blob/3e7b2760ef214e43a16adc2220d53f95bc5e3772/config.toml#L163
- [ ] It seems that `gas_adjustment = 0.0` leads to a panic. We should disallow that value
    - set up a two network setup and configure the gas adjustment to `0.0` for `ibc-0`, leave it default for `ibc-1`
    - do `hermes create client ibc-0 ibc-1` -- this should work
    - do `hermes create client ibc-1 ibc-0` -- this should panic
    - >$ ./target/debug/hermes create client ibc-0 ibc-1
2022-05-30T08:38:26.450768Z  INFO ThreadId(01) using default configuration from '/Users/adi/.hermes/config.toml'
thread '<unnamed>' panicked at 'assertion failed: digits.len() == 1', relayer/src/chain/cosmos/gas.rs:54:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: foreign client error: error raised while creating client for chain ibc-0: failed sending message to dst chain : internal message-passing failure while receiving inter-thread request/response: receiving on an empty and disconnected channel

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
[...]

PR #2353, opened 2022-06-29
Deprecate `gas_adjustment` in favor of new `gas_multiplier` setting

romac:
Closes: #2174 

## Description

This PR does the following: (cf. https://github.com/informalsystems/ibc-rs/issues/2174#issuecomment-1168474595)

1. Let's deprecate `gas_adjustment`. The name of the setting does not properly convey what is used for imho, and its semantics (multiply the estimated gas by the value and add it to the estimation) are confusing.
2. Instead, replace it with a `gas_multiplier` setting, with the following semantics: take the estimated gas and multiply it by the value of the setting.

Meaning that when we previously had `gas_adjustment = 0.1`, which increases the gas by 10%, we would now have `gas_multiplier = 1.1`.

This PR enforces a lower bound of `1.0` for the new setting but no upper bound so that people are free to use as high gas as they want.

> @[loredanacirstea](https://github.com/loredanacirstea) That's perfect. Your proposal is also consistent with how cosmos sdk uses the gas-adjustment flag - as a multiplier for the gas value. https://github.com/informalsystems/ibc-rs/issues/2174#issuecomment-1168667944

## Note regarding backward incompatibility

This PR aims to hand-hold operators to use the new setting instead of the old one whenever they update to the Hermes version including these changes (1.0.0).

If an operator is still using the `gas_adjustment` setting in their config, they will be greeted with the following message when they try to run any command, as well as `config validate`:

> config file specifies deprecated setting `gas_adjustment` for the chain 'ibc-0', instead please use the `gas_multiplier` setting with value 1.1

Note: the message above arises when `gas_adjustment` is set to `0.1` in the config of chain `ibc-0`.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


relayer/src/chain/cosmos/gas.rs
```diff
@@ -42,21 +46,46 @@ pub fn mul_ceil(a: u64, f: f64) -> BigInt {
...
+    // Multiply the gas estimate by the gas_multiplier option
+    let (_, digits) = mul_floor(gas_amount, gas_multiplier).to_u64_digits();
+
+    let gas = if digits.len() == 1 {
+        // If the result fits in a u64, use that
+        digits[0]
+    } else {
+        // Otherwise, use u64::MAX
+        u64::MAX
+    };
```
ljoss17, line 85: The case where the `gas_amount` is simulated to 0 will cause to have the computed gas being u64::MAX, and thus the value returned by `adjust_estimated_gas` to be the `max_gas` configured.
Is this expected ? In the case this is expected, should there be a trace warning or error to inform that the simulated gas for the tx was 0 ?

romac: Good catch, thanks! Fixed in https://github.com/informalsystems/ibc-rs/pull/2353/commits/b0cbc4fd8ef8500f634448553867b01a265e07d8 and added short-circuiting for when `gas_amount = 0` or `gas_multiplier = 1.0` in [Short-circuit when gas_amount = 0 or gas_multiplier = 1](https://github.com/informalsystems/ibc-rs/pull/2353/commits/341774d1839f574ca193382e9e8903931f3fa966)

ljoss17 (approved):

Issue #2265, opened 2022-06-03
Release v1 for `ibc-relayer-cli` crate


adizere:
# Release Hermes v1.0.0

### Cleanup & dependencies

- [SDK 0.43 is discontinued for some time](https://github.com/cosmos/cosmos-sdk/releases/tag/v0.44.0), we should:
    - [ ] clearly signal that Hermes is no longer compatible with 0.43 in the changelog
    - [ ] deprecate it: I found two lingering references, but we should search the repo
        - ref 1: https://github.com/informalsystems/ibc-rs/blob/4e83aae8afdc9a88a4a126df9d217812672b3992/relayer/src/upgrade_chain.rs#L2
        - ref 2: https://github.com/informalsystems/ibc-rs/blob/4e83aae8afdc9a88a4a126df9d217812672b3992/relayer/src/chain/cosmos/simulate.rs#L17

⚡

### The usual todos

- [ ] Create a new release in the changelog, using [`unclog`](https://github.com/informalsystems/unclog).
  - If doing a release candidate (`rc`) version, then skip the `unclog release` step.
- [ ] Bump all crate versions to the new version.
- [ ] Reassign unfinished issues of previous milestone to the next milestone.
- [ ] Update Cargo.lock file (if re-publishing `ibc-relayer-cli`)

PR #2278, opened 2022-06-08
Generate gRPC server code

plafer:
Closes: #2277

## Description

Tested with basecoin.

Change done in pair programming with @hu55a1n1.

______

### PR author checklist:
- [x] Added changelog entry, using [`unclog`](https://github.com/informalsystems/unclog).
- [ ] Added tests: integration (for Hermes) or unit/mock tests (for modules). 
- [x] Linked to GitHub issue.
- [ ] Updated code comments and documentation (e.g., `docs/`).

### Reviewer checklist:

- [ ] Reviewed `Files changed` in the GitHub PR explorer.
- [ ] Manually tested (in case integration/unit/mock tests are absent).

PR #2355, opened 2022-06-29
Release Hermes v1.0.0

romac:
Closes: #2265

### Cleanup & dependencies

- [ ] The presentation for the `ibc-relayer-cli` crate is very minimal, we should improve that; we should signal clearly what each module is for and look for obvious boilerplate content to make it more specific or remove it from the documentation
    - https://docs.rs/ibc-relayer-cli/0.15.0/ibc_relayer_cli/

- [ ] We have an inconsistency between `query packet pending` and `query packet unreceived-acks`/`unreceived-packets`
    - We should stick to one adjective consistency: either pending or unreceived
    - Additionally, it's not accurate to use the word "acks" in `unreceived-acks` because "acks" are a type of packets; so when we say `unreceived-packets` that should actually include acks -> we should consider renaming `unreceived-packets` into `unreceived-send` or something similar to make the clear distinction between sends and acks.

⚡

### The usual todos

- [ ] Create a new release in the changelog, using [`unclog`](https://github.com/informalsystems/unclog).
  - If doing a release candidate (`rc`) version, then skip the `unclog release` step.
- [ ] Bump all crate versions to the new version.
- [ ] Reassign unfinished issues of previous milestone to the next milestone.
- [ ] Update Cargo.lock file (if re-publishing `ibc-relayer-cli`)


guide/src/commands/queries/packet.md
```diff
@@ -262,11 +262,11 @@ Success: [
...
-Use the `query packet unreceived-acks` command to query the sequence numbers of all packets that have not yet been acknowledged.
+Use the `query packet pending-acks` command to query the sequence numbers of all packets that have not yet been acknowledged.
 
 ```shell
 USAGE:
-    hermes query packet unreceived-acks --chain <CHAIN_ID> --port <PORT_ID> --channel <CHANNEL_ID>
+    hermes query packet pending-acks --chain <CHAIN_ID> --port <PORT_ID> --channel <CHANNEL_ID>
 
 DESCRIPTION:
     Query unreceived acknowledgments
```
hu55a1n1, line 272: Should we replace all occurrences of `unreceived` with `pending`? (in code, comments and the guide?)

guide/src/config.md
```diff
@@ -36,7 +36,7 @@ For example, if there are only two chains configured, then Hermes will only rela
 i.e. the two chains will serve as a source for each other, and likewise as a destination for each other's relevant events.
 Hermes will ignore all events that pertain to chains which are unknown (ie. not present in config.toml).
 
-To restrict relaying on specific channels, or uni-directionally, you can use [packet filtering policies](https://github.com/informalsystems/ibc-rs/blob/v0.15.0/config.toml#L207-L224).
+To restrict relaying on specific channels, or uni-directionally, you can use [packet filtering policies](https://github.com/informalsystems/ibc-rs/blob/v1.0.0-rc.0/config.toml#L207-L224).
```
hu55a1n1, line 39: Nit: lines in the link are misplaced. Might be better to use a permalink here or just link to the config file so we don't have to check it on every release.

proto/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name         = "ibc-proto"
-version      = "0.18.0"
+version      = "0.19.0"
```
hu55a1n1, line 3: Doesn't seem like there have been breaking changes in crates other than `ibc` and `ibc-relayer-cli`, so should we continue to bump versions for all crates?

hu55a1n1 (commented):
Nicely done @romac! Left some questions and minor nits.

relayer-cli/src/error.rs
```diff
@@ -1,3 +1,5 @@
+//! All errors which can be reaised from a command.
```
soareschen, line 1:
```suggestion
//! All errors which can be raised from a command.
```

soareschen (approved):

proto/Cargo.toml
romac: The `ibc-proto` crate contains [these changes](https://github.com/informalsystems/ibc-rs/pull/2278) which are indeed not breaking but are also a sizable feature so to me it makes sense to bump its major version as well. What do you think?

guide/src/commands/queries/packet.md
romac: Fixed the guide, let's do code changes after rc0.

proto/Cargo.toml
hu55a1n1: Right, I agree it makes sense to bump the version. :+1: What do you suggest regarding the other worksapce crates?

guide/src/config.md
romac: Fixed thanks!

proto/Cargo.toml
romac: For now let's bump them all, we can figure out a proper process for this after the rc0 release, in the form of an ADR.

hu55a1n1 (approved):

PR #2356, opened 2022-06-29
ADR 11: Light client extraction

hu55a1n1:
## Description
An ADR for light client extraction by removal of `Any*` enums and replacing them with trait objects. 

[Rendered](https://github.com/informalsystems/ibc-rs/blob/hu55a1n1/adr011-light-client-extraction/docs/architecture/adr-011-light-client-crates-extraction.md)
______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


docs/architecture/adr-011-light-client-crates-extraction.md
```diff
@@ -0,0 +1,292 @@
...
+```rust
+pub enum AnyClientState {
+    Tendermint(client_state::ClientState),
+
+    #[cfg(any(test, feature = "mocks"))]
+    Mock(MockClientState),
+}
+```
+
+See [ADR003 - Dealing with chain-specific datatypes](https://github.com/informalsystems/ibc-rs/blob/master/docs/architecture/adr-003-handler-implementation.md#dealing-with-chain-specific-datatypes)
```
seanchen1991, line 26: Adding a summary of the rationale behind the initial decision for the `Any*` enum here so that folks reading this ADR don't have to click away to a different doc (unless they want the additional context) would be helpful here.

docs/architecture/adr-011-light-client-crates-extraction.md
```diff
@@ -0,0 +1,292 @@
...
+for the rationale behind the `Any*` enum design choice.
+
+Additionally, there are places where the core modules code (indirectly) depends on light client specific types. e.g.:
```
seanchen1991, line 29: Pointing out the exact dependencies in the code snippet would be helpful, something like:
```suggestion
Additionally, there are places where the core modules code (indirectly) depends on light client specific types, such as the `tendermint::abci::Event` and `IbcEvent` types in the following code snippet:
```

docs/architecture/adr-011-light-client-crates-extraction.md
```diff
@@ -0,0 +1,292 @@
...
+
+* `AnyClient` implements `ClientDef`.
+* `AnyClientState` implements `ClientState`.
+* `AnyConsensusState` implements `ConsensusState`.
+* `AnyHeader` implements `Header`.
+* `AnyMisbehaviour` implements `Misbehaviour`.
+
+In order to replace the `Any*` enums with trait objects, these light client traits must be object safe. This
+essentially means that these traits cannot have a `Self: Sized` requirement and their methods cannot have type
+parameters & cannot use `Self`. These restrictions, their implications and possible workarounds are documented below.
```
seanchen1991, line 67: Oxford commas just make more sense 😄
```suggestion
parameters & cannot use `Self`. These restrictions, their implications, and possible workarounds are documented below.
```

docs/architecture/adr-011-light-client-crates-extraction.md
```diff
@@ -0,0 +1,292 @@
...
+```
+
+Thankfully, the core modules code doesn't use the `serde` derivations except in logs and errors. Host and light-client
+implementations can optionally choose to downcast to the concrete type and use it's `serde` derivations directly (if
+available).
+
+#### Light client traits cannot have constructors
+
+This restriction comes from the fact that trait methods of object safe traits cannot return `Self`.  
+However, we would need a ctor to be able to create a `ClientState` and `ConsensusState` in the `create_client` handler.
```
seanchen1991, line 143: Not sure what a "ctor" is. Or is that a typo?

docs/architecture/adr-011-light-client-crates-extraction.md
```diff
@@ -0,0 +1,292 @@
...
+## Status
+
+Proposed
+
+## Consequences
+
+### Positive
+
+* Light client implementations can be hosted in separate crates outside the ibc-rs repo.
+* Light client implementations can be maintained & audited independently. clearer ownership
```
seanchen1991, line 266:
```suggestion
* Light client implementations can be maintained & audited independently, resulting in clearer ownership.
```

seanchen1991 (commented):
Left some nits.

docs/architecture/adr-011-light-client-crates-extraction.md
hu55a1n1: Good point, done! https://github.com/informalsystems/ibc-rs/pull/2356/commits/f4a7747255c310b92b2ca4b878cd88c7875efd11

docs/architecture/adr-011-light-client-crates-extraction.md
hu55a1n1: `ctor` is shorthand for constructor but I decided to not use it. :) Fixed in https://github.com/informalsystems/ibc-rs/pull/2356/commits/5d2495c3e645ca215f9c76f34eff2080c2d0283d

docs/architecture/adr-011-light-client-crates-extraction.md
hu55a1n1: Fixed in https://github.com/informalsystems/ibc-rs/pull/2356/commits/5d2495c3e645ca215f9c76f34eff2080c2d0283d

docs/architecture/adr-011-light-client-crates-extraction.md
```diff
@@ -0,0 +1,375 @@
...
+
+### Light client registry
+
+With the proposals in this ADR, the `ibc` crate would be light client agnostic, however, the host implementation must
+still be aware of all light clients that it wishes to support. For e.g., after an upgrade a blockchain must be able to
+deserialize types from the persistent store.
+Furthermore, protobuf has emerged as the canonical serialization scheme for IBC, and IBC's message definitions usually
+serialize light client types using the `google::protobuf::Any` type where the `type_url` is accepted to uniquely
+represent specific light client types, although this has not been standardized yet.
+It is proposed to standardize this and provide a `ibc-client-registry` crate which standardizes this and collects all
```
hu55a1n1, line 299: A cool crate that might be helpful in implementing such a registry -> https://github.com/dtolnay/inventory. (Thanks to @romac for suggesting it).

docs/architecture/adr-011-light-client-crates-extraction.md
```diff
@@ -0,0 +1,375 @@
...
+```rust
+#[derive(Deserialize, Serialize)]
+pub struct UpdateClient {
+    pub common: Attributes,
+    pub header: Option<AnyHeader>,
+}
+```
+
+This can be solved using the `ibc_proto::google::protobuf::Any` type instead and having the light client traits provide
+an encoding to `Any` ->
```
plafer, line 132: Instead of using `Option<Any>` for `header`, why not implement `Deserialize` ourselves, and apply the transformation `encode_any()` in `deserialize()`? 

`serde` might also even provide an attribute that we could tag `header` with that specifies to apply this transformation before deserializing. Not sure though.

docs/architecture/adr-011-light-client-crates-extraction.md
```diff
@@ -0,0 +1,375 @@
...
+}
+```
+
+This is problematic because it means the module code must be aware of a `ClientType` -> `ClientDef` mapping. This can be
+solved by requiring the `ClientState` trait to provide us with its `ClientDef` implementation.
+
+```rust
+pub trait ClientState {
+    /// Return the associated `ClientDef` implementation
+    fn client_def(&self) -> Box<dyn ClientDef>;
```
plafer, line 232: Instead of having `ClientState` provide the `client_def()`, why not have `ClientReader` have a `client_def(client_id)` method instead?

As discussed, this is a different way of thinking about it. I see the `ClientState` as simply encapsulating the parameters of a light client. To me, it's more natural for the chain/context to know the type of all its clients (indexed by `clientId`), rather than the parameters knowing about their corresponding client implementation.

docs/architecture/adr-011-light-client-crates-extraction.md
```diff
@@ -0,0 +1,375 @@
...
+
+    /* ... */
+}
+```
+
+Now, we can use the `ClientState` instance to get the `ClientDef` implementation ->
```
plafer, line 238: Can we also use this ADR to rename `ClientDef` to, say, `Client` or `LightClient`?

docs/architecture/adr-011-light-client-crates-extraction.md
```diff
@@ -0,0 +1,375 @@
...
+    match any_header.type_url {
+        TENDERMINT_HEADER_TYPE_URL => TendermintHeader::try_from(any_header).map_err(|e| Error::decode_header(e)),
+        /* ... */
+        _ => Err(Error::unsupported_client())
+    }
+}
+```
+
+This way we have a standardized list of supported light clients so that host implementations would not have to directly
+import every single light client crate that they wish to support. The crate could feature gate light client support to
```
plafer, line 318: How would light client versioning work in this system?

Given a fixed type url, say `"/ibc.lightclients.tendermint.v1.Header"`, the `tendermint-light-client` crate could change its domain type from version `1.0` to `1.1`. Which version would the registry use? What if the registry uses `1.0`, and my chain uses `1.1`?

I think we need to carefully consider this issue, and then decide whether or not the registry is a good idea, and if it would end up causing more headaches than it solves.

docs/architecture/adr-011-light-client-crates-extraction.md
```diff
@@ -0,0 +1,375 @@
...
+Common traits that the light client traits depend on that have the `Self: Sized` requirement
+are `core::convert::{Into<T>, From<T>, ...}`, `tendermint_proto::Protobuf`, `Clone`, etc.
+
+An easy workaround is to add these methods directly to the trait, for e.g. we add an `encode_vec()` method to
+the `Header` trait directly.
+
+```rust
+pub trait ClientState {
+    /// Encode to canonical binary representation
+    fn encode_vec(&self) -> Vec<u8>;
```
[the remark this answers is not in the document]
hu55a1n1: This is an interesting idea and I thought this should be theoretically possible so I tried to experiment with it here -> 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0017151c945b0eebb412dbda9a06db20

The `Protobuf` trait requires the raw type to implement `Message + From<Self> + Default` and the domain type to implement `Sized + Clone + TryFrom<T>`. The raw type part is easy as prost already derives `Message` and `default` for us. For the domain types, we need `Clone` and `TryFrom` without the `Self: Sized` requirement, so we could write our own. And it all seems to work until we decide to call a `Protobuf` method on our trait object. 

This isn't allowed because methods such as `Protobuf::encode_vec()` must be able to convert the domain type to a raw type and at that point, the domain type's size must be known. This is the problem that the code in the playground runs into. I can't think of a good solution to this problem now but happy to revisit it if we find a decent workaround.

docs/architecture/adr-011-light-client-crates-extraction.md
hu55a1n1: I think the problem with that approach is that the deserialize impl would depend on `Header::decode()` which must have a `Self: Sized` requirement. This is because we must be aware of the concrete type of the header at the time of deserialization but that's not possible.

hu55a1n1: Seems like there are some ready-made solutions like [erased_serde](https://github.com/dtolnay/erased-serde) but I am not sure if they can work for our specific use case.

hu55a1n1: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=de9aa123fcec49b410fec8bb23d5f53e

docs/architecture/adr-011-light-client-crates-extraction.md
hu55a1n1: I see your point, I have been thinking about this myself as well. To me, the `ClientState` represents an instance of a light client and it's the only 1:1 mapping of a client id v/s light client that is stored on-chain. ~~And the `ClientState` trait is something that the light client will implement. So it makes more sense to me to ask the light client to give us access to its verification functions instead of asking the host. Besides, if `client_def()` was a method on `ClientReader`, that would require the host to maintain a `ClientId` <> `ClientType` mapping.~~
Just checked the spec and it already requires hosts to store a `ClientId` <> `ClientType` mapping and your interpretation seems to be more in line with the spec.

docs/architecture/adr-011-light-client-crates-extraction.md
hu55a1n1: Great idea! Interestingly, the spec calls the verification methods on the `ClientType` itself -> https://github.com/cosmos/ibc/tree/master/spec/core/ics-002-client-semantics#create
I wonder if it'd be better to rename `ClientDef` to `ClientType`.

docs/architecture/adr-011-light-client-crates-extraction.md
hu55a1n1: d66271656205207c92e497a5c69f0708bbba469e

docs/architecture/adr-011-light-client-crates-extraction.md
```diff
@@ -0,0 +1,375 @@
...
+to/from `tendermint::abci::Event` are required for `deliverTx` implementation.
+
+It is recommended that most client specific code is moved either to the `ibc-relayer` crate or the light client crates
+and issues of this kind be dealt with on a case-by-case basis.
+
+### Light client registry
+
+With the proposals in this ADR, the `ibc` crate would be light client agnostic, however, the host implementation must
+still be aware of all light clients that it wishes to support. For e.g., after an upgrade a blockchain must be able to
+deserialize types from the persistent store.
```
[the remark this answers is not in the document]
hu55a1n1: I believe this is possible with the in-place migration feature -> https://docs.cosmos.network/master/core/upgrade.html. But decided to use the more common example of chain restart.

docs/architecture/adr-011-light-client-crates-extraction.md
hu55a1n1: I agree this needs careful consideration and we must ask ourselves questions such as these. :+1:

> Which version would the registry use? What if the registry uses 1.0, and my chain uses 1.1?

For the specific problem you described above, I think one of the benefits of having a client registry is that host implementations will not have to import or deal with light client crates/types directly, they will only depend on the light client crates via the registry. Furthermore, both the host and the module code will only ever access light client types from behind a `&dyn Trait` (if they use the registry), so changes to the concrete type would not have any adverse effects (AFAICT). 

The registry can be thought of as hosting the `Any*` enums that we currently have in our codebase. Interestingly, this is indeed one of the ways in which such a registry can be implemented. So usability wise not much should change for the host implementations. Note that the versioning of light client crates does not have an immediate effect on the versioning scheme of the registry crate because the registry never exposes the concrete light client types. We might need to decide on a versioning scheme for the registry crate, but I don't think that will be too difficult. 

To me, the fact that there isn't a canonical mapping of a `type_url` <> `ClientType` is a clear sign that we need such a registry. 
Would be nice to get more input from the team here. cc: @romac @mzabaluev @soareschen

docs/architecture/adr-011-light-client-crates-extraction.md
plafer: As discussed, ibc-go uses `ClientState` as a combination of our `ClientState` and `ClientDef`: it has the state, and the verify functions. I believe we should follow this approach as well.

So basically, remove `ClientDef`, and put all those methods in our new `ClientState` trait.

docs/architecture/adr-011-light-client-crates-extraction.md
plafer: As discussed, I don't think this is an issue anymore. Even if the chain implementing the keepers and readers uses `tendermint-light-client` crate on a different version than whatever the registry uses, Rust sees the types coming from them as separate. For example,
```rust
// The tendermint-light-client crate as declared in the chain's Cargo.toml
use tendermint_light_client::Header;

// even if `light_client_registry` crate re-exports the tendermint_light_client
// crate, then any type is exports (here, `TmHeader`) is a different type from
// `Header`
use light_client_registry::tendermint_light_client::Header as TmHeader;
```
Ultimately, the registry can choose to update its dependencies freely without fear of breaking code that depends on it.

docs/architecture/adr-011-light-client-crates-extraction.md
plafer: > This is because we must be aware of the concrete type of the header at the time of deserialization but that's not possible.

This is a good point. To address this issue, we could modify the `IntoRef` trait [suggested here](https://github.com/informalsystems/ibc-rs/pull/2356/files/619e34127c43c7372af0756a846bea2fef230dc7#r916169288) to inject some context into the conversion:
```rust
trait IntoRef<T: ?Sized> { 
    // Same applies to `try_into_ref()`
    fn into_ref(&self, ctx: &ConversionContext) -> Box<T>;
}
```
`ConversionContext` (or any context more appropriate) would be implemented by the chain (along with all other `Reader`s and `Keeper`s). It would look something like:
```rust
trait ConversionContext {
    fn decode_header(&self, raw_header: &Any) -> Result<Box<dyn Header>, Error>;
    fn decode_client_state(&self, raw_cs: &Any) -> Result<Box<dyn ClientState>, Error>;
    fn decode_consensus_state(&self, raw_cs: &Any) -> Result<Box<dyn ConsensusState>, Error>;
    ...
}
```
Note that we could facilitate the implementation of this trait with the light client registry crate proposed in this ADR (for all known light clients known to the registry).

However, the downside of this approach then is that the `ibc` library doesn't know how to convert `Raw` -> `Domain` for types who use, for example, `dyn Header`. This is because by design, we want the chains to tell us how to decode the `Header` (and `ClientState`, etc) for the light clients that it knows about. 

 We then need to introduce an additional trait `TryFromContext`:
```rust
trait TryFromContext<T> {
    type Error;
    
    fn try_from_ctx(value: T, ctx: ConversionContext) -> Result<Self, Self::Error>;
}
```


which would be used with, say, `MsgUpdateClient`:

```rust
struct MsgUpdateClient {
    pub client_id: ClientId,
    pub header: Box<dyn Header>,
    pub signer: Signer,
}

impl TryFromContext<RawMsgUpdateClient> for MsgUpdateClient {
    type Error = Error;

    // We need to use `TryFromContext` instead of `TryFrom` so that 
    // the `ConversionContext` can be used to decode the raw `Header`
    fn try_from_ctx(raw: RawMsgUpdateClient, ctx: &ConversionContext) -> Result<Self, Self::Error> {
        Ok(MsgUpdateClient {
            client_id: ...,
            header: ctx.decode_header(&raw)?,
            signer: ...,
        })
    }
```

Ultimately, I still prefer this solution to using the raw `Any` for `Header` et al, as it allows us to have proper domain types (where all fields are also domain types), at the cost of more complexity in the conversion code.

WDYT?

hu55a1n1: `erased_serde` uses a cool trick explained here -> https://github.com/dtolnay/erased-serde/blob/master/explanation/main.rs. It allows one to have a separate object safe version of a trait that can operate with the original (non object safe version) seamlessly. 
In our case, I don't think we need to support interoperability with the `tendermint_proto::Protobuf` trait so we can define our own trait that is object safe and just use that instead. We can still use `erased_serde`'s blanket impl trick to keep the original `tendermint_proto::Protobuf` API on the implementation side. 
See experimental PR where `Protobuf` is object safe and requires no other changes on the implementation side -> https://github.com/informalsystems/ibc-rs/pull/2412/files#diff-2b95cbfaf034ca5aa2b9a3f02436e8bd7a7a937df7c5ffb6dba81a0237141941R70 =>
```rust
mod erased {
  use core::convert::{Into as CoreInto, TryFrom as CoreTryFrom};
  
  mod sealed {
      use super::*;
  
      pub trait SealedInto<T: ?Sized> {}
      impl<T, U: Clone + CoreInto<T>> SealedInto<T> for U {}
      pub trait SealedTryFrom<T> {}
      impl<T, U: CoreTryFrom<T>> SealedTryFrom<T> for U {}
  }
  
  pub trait Into<T: ?Sized>: sealed::SealedInto<T> {
      fn into(&self) -> Box<T>;
  }
  
  impl<T, U: Clone + CoreInto<T>> Into<T> for U {
      fn into(&self) -> Box<T> {
          Box::new(self.clone().into())
      }
  }
  
  pub trait TryFrom<T>: sealed::SealedTryFrom<T> {
      type Error;
  
      fn try_from(t: T) -> Result<Self, Self::Error>
      where
          Self: Sized;
  }
  
  impl<T, U: CoreTryFrom<T>> TryFrom<T> for U {
      type Error = <Self as CoreTryFrom<T>>::Error;
  
      fn try_from(t: T) -> Result<Self, Self::Error>
      where
          Self: Sized,
      {
          <Self as CoreTryFrom<T>>::try_from(t)
      }
  }
}

pub trait Protobuf<T: Message + Default>
where
    Self: erased::TryFrom<T> + erased::Into<T>,
    <Self as erased::TryFrom<T>>::Error: Display,
{ /* ... */ }
```

hu55a1n1: `TryFromContext` isn't object safe because the trait function is static (and doesn't have a `self` parameter) and it returns `Self`. This is the same problem we faced with the `From` trait which we solved using the `IntoRef` solution, but that doesn't work here with `TryFromContext`, that's because an equivalent `TryIntoContext` trait bound would look like `T: TryIntoContext<Self>` and that isn't allowed due object safety restrictions. 

Furthermore, to be able to decode using `TryFromContext` we would need to specify the additional context at the point of decoding which may or may not be accessible there, for e.g. Rust's serde deserialization is stateless and so it is not straightforward to access host context at the time of deserialization. 

For these reasons, I believe it is better to not use the `TryFromContext` trait and instead stick to the `TryFrom` trait with a `where Self: Sized` bound.

docs/architecture/adr-011-light-client-crates-extraction.md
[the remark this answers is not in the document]
hu55a1n1: Added this to the ADR -> https://github.com/informalsystems/ibc-rs/pull/2356/commits/0b085570e24119a7f921f178a156ce4dfaa97507

romac (approved):
Amazing work, @hu55a1n1 and @plafer! 💐 

The restriction introduced under *Domain types containing light client specific types* is unfortunate but does seems indeed like the best way forward for now.

docs/architecture/adr-011-light-client-crates-extraction.md
```diff
@@ -0,0 +1,484 @@
...
+use ibc_proto::google::protobuf::Any;
+
+pub trait ClientState {
+    fn decode(any: Any) -> Box<dyn ClientState> where Self: Sized;
+
+    /* ... */
+}
+```
+
+### Downcasting support
```
plafer, line 198: Why would we ever need to downcast to a specific concrete type? Isn't the interface provided by, say, `Header` enough?

hu55a1n1: I believe this will be required (for example) in cases where the light client wants to access its state stored on the host - the host is oblivious to the light client state's concrete type so it would return a trait object (such as `&dyn ConsensusState`) and the light client would need to downcast it to the concrete light client type (e.g. `TmConsensusState`) before it can use light client specific fields from it (such as `next_validators_hash`).

hu55a1n1:
Already merged via https://github.com/informalsystems/ibc-rs/pull/2423 and impld in https://github.com/informalsystems/ibc-rs/pull/2483.

PR #2357, opened 2022-06-29
Add `--upgrade-height` parameter to `upgrade client` command

seanchen1991:
Closes: #2300 

## Description

Adds a flag `--upgrade-height` parameter that receives a Height parameter from the user that specifies the application height at which the host chain should halt before performing the client upgrade. 

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


relayer-cli/src/commands/tx/client.rs
```diff
@@ -240,24 +248,32 @@ impl Runnable for TxUpgradeClientCmd {
...
+        };
+
+        let mut src_application_latest_height = match client.src_chain().query_latest_height() {
+            Ok(height) => height,
+            Err(e) => Output::error(format!("{}", e)).exit(),
+        };
+
+        // Wait until the client's application height reaches the target application upgrade height
+        while src_application_latest_height < target_application_upgrade_height {
+            thread::sleep(Duration::from_millis(200));
```
romac, line 268: This seems a bit low to me given that blocks typically take >1 seconds to be emitted. Let's use a 1 second interval perhaps? What do you think @seanchen1991 @adizere?

relayer/src/foreign_client.rs
```diff
@@ -977,7 +977,7 @@ impl<DstChain: ChainHandle, SrcChain: ChainHandle> ForeignClient<DstChain, SrcCh
 
         // Wait for the source network to produce block(s) & reach `target_height`.
         while src_application_latest_height()? < target_height {
-            thread::sleep(Duration::from_millis(100))
+            thread::sleep(Duration::from_millis(100));
```
romac, line 980: Same here, but since this is on the relay path it makes sense to have a smaller interval than in the CLI, perhaps 500ms?

adizere: I suspect the caching layer will take the brunt of the call to `src_application_latest_height` (cache expires every 200ms or so). So from a relay path perspective, not sure if we should make this value 500ms, we can probably leave it as is. I recall having this discussion with Anca: whether we shouldn't increase the value here, and the argument was that the `latest_height` query is the cheapest and we can afford to retry often. Before the caching layer we were retrying every 100ms; after we introduced the caching layer that means we're retrying every ~250ms or so (backoff 100ms from the sleep here + 200ms backoff from cache TTL).

https://github.com/informalsystems/ibc-rs/blob/e3d0b108831629e5f7fb541c06396e345af0228c/relayer/src/cache.rs#L20

romac: Makes sense, let's leave it as-is then!

relayer-cli/src/commands/tx/client.rs
romac: Right, probably does not matter too much indeed, but let's go with 500ms still?

seanchen1991:
I'm attempting to test the new flag by following the [Testing Client Upgrade](https://hermes.informal.systems/commands/upgrade/test.html) section of the Hermes guide. Running the `upgrade client` command specified in step 6, it appears the `while src_application_latest_height < target_application_upgrade_height` loop doesn't terminate; the `src_application_latest_height` seems to always stay at height `0-1` and never increases.

relayer-cli/src/commands/tx/client.rs
```diff
@@ -214,6 +215,14 @@ pub struct TxUpgradeClientCmd {
...
     )]
     client_id: ClientId,
+
+    #[clap(
+        long = "upgrade-height",
+        required = true,
+        value_name = "SRC_UPGRADE_HEIGHT",
+        help = "The height at which the source chain should halt in order to perform the client upgrade"
+    )]
+    target_upgrade_height: Height,
```
plafer, line 225: Let's use `u64` here instead of `Height`; it's easier for users and more consistent with the rest of the cli

relayer-cli/src/commands/tx/client.rs
```diff
@@ -250,24 +259,36 @@ impl Runnable for TxUpgradeClientCmd {
...
+            "Source application latest height: {}",
+            src_application_latest_height
+        );
+
+        // In order to perform the client upgrade, the chain is paused at the height specified by
+        // the user. When the chain is paused, the application height reports a height of 1 less
+        // than the height according to Tendermint. As a result, the target height at which the
+        // upgrade occurs at (the application height) is 1 less than the height specified by
+        // the user, hence the strictly less-than check in the while loop.
+        while src_application_latest_height < self.target_upgrade_height {
```
plafer, line 277: How about we change this to 
```rust
while src_application_latest_height != self.target_upgrade_height - 1
```
It makes it more explicit what the desired state is for when we come out of this loop. Also, this currently infinite loops because `src_application_height` will always be smaller than `target_upgrade_height` (by at least 1)

relayer-cli/src/commands/tx/client.rs
```diff
@@ -214,18 +215,26 @@ pub struct TxUpgradeClientCmd {
         help = "Identifier of the client to be upgraded"
     )]
     client_id: ClientId,
+
+    #[clap(
+        long = "upgrade-height",
+        required = true,
+        value_name = "HOST_TARGET_UPGRADE_HEIGHT",
```
plafer, line 222: This should be `"REFERENCE_UPGRADE_HEIGHT"`, as the reference chain is the one performing the upgrade at height `upgrade-height`.

relayer-cli/src/commands/tx/client.rs
```diff
@@ -214,18 +215,26 @@ pub struct TxUpgradeClientCmd {
...
+        help = "The height at which the host chain should halt in order to perform the client upgrade"
```
plafer, line 223: This should be `"The height at which the reference chain halts for the upgrade"`

guide/src/commands/upgrade/test.md
```diff
@@ -86,7 +86,7 @@ gaiad version --log_level error --long | head -n4
      > `gm ports` in order to see a list of the ports being used.
 
     ```shell
-    gaiad --node tcp://localhost:<RPC PORT> query gov proposal 1 --home $HOME/.gm/ibc-0/
+    gaiad --node tcp://localhost:27000 query gov proposal 1 --home $HOME/.gm/ibc-0/
```
seanchen1991, line 89: Should the RPC port here be specified? Is it always the case that this is always the correct port?

guide/src/commands/upgrade/test.md
```diff
@@ -180,7 +181,7 @@ gaiad version --log_level error --long | head -n4
     This command must be issued while the proposal status is `PROPOSAL_STATUS_VOTING_PERIOD`. Confirm transaction when prompted.
 
     ```shell
-    gaiad --node tcp://localhost:<RPC PORT> tx gov vote 1 yes --home $HOME/.gm/ibc-0/data/ --keyring-backend test --keyring-dir $HOME/.gm/ibc-0/ --chain-id ibc-0 --from validator
+    gaiad --node tcp://localhost:27000 tx gov vote 1 yes --home $HOME/.gm/ibc-0/data/ --keyring-backend test --keyring-dir $HOME/.gm/ibc-0/ --chain-id ibc-0 --from validator
```
seanchen1991, line 184: Same point here about specifying the RPC port.

guide/src/commands/upgrade/test.md
```diff
@@ -193,7 +194,7 @@ gaiad version --log_level error --long | head -n4
      Note the `final tally_result` that includes the vote submitted in the previous step.
 
       ```shell
-      gaiad --node tcp://localhost:<RPC PORT> query gov proposal 1 --home $HOME/.gm/ibc-0/
+      gaiad --node tcp://localhost:27000 query gov proposal 1 --home $HOME/.gm/ibc-0/
```
seanchen1991, line 197: Same point here about specifying the RPC port.

guide/src/commands/upgrade/test.md
plafer: Yes, it's always `27000` since we specify the `gm` config at the beginning of the tutorial

plafer (approved):
🎉

seanchen1991:
Huge thanks to @plafer and @ljoss17 for helping to get this over the finish line! ❤️

PR #2361, opened 2022-07-01
Change ChannelId representation to String

mzabaluev:
Closes: #2330

## Description

ICS 024 does not restrict channel IDs to the "channel-{N}" format.

Change the internal representation of `ChannelId` to `String`, and modify the API accordingly to the similar ID types.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


modules/src/core/ics24_host/validate.rs
```diff
@@ -67,11 +67,19 @@ pub fn validate_port_identifier(id: &str) -> Result<(), Error> {
     validate_identifier(id, 2, 128)
 }
 
+/// Default validator function for Channel identifiers.
+///
+/// A valid Identifier must be between 10-64 characters and only contain lowercase
+/// alphabetic characters,
+pub fn validate_channel_identifier(id: &str) -> Result<(), Error> {
+    validate_identifier(id, 8, 64)
```
romac, line 75: There is discrepancy between the [spec](https://github.com/cosmos/ibc/blob/master/spec/core/ics-024-host-requirements/README.md#paths-identifiers-separators) and the code here. The spec says channel ids must be 10 characters long or more but that would mean that `channel-0` is not a valid channel id. We should clarify that.

romac: Clarified offline. Let's update the comment above then

modules/src/core/ics24_host/validate.rs
```diff
@@ -67,11 +67,19 @@ pub fn validate_port_identifier(id: &str) -> Result<(), Error> {
     validate_identifier(id, 2, 128)
 }
 
+/// Default validator function for Channel identifiers.
+///
+/// A valid Identifier must be between 10-64 characters and only contain lowercase
```
romac, line 72:
```suggestion
/// A valid Identifier must be between 8-64 characters and only contain lowercase
```
```suggestion
/// A valid Identifier must be between 10-64 characters and only contain lowercase
```

romac (changes requested):

modules/src/core/ics24_host/validate.rs
```diff
@@ -67,11 +67,19 @@ pub fn validate_port_identifier(id: &str) -> Result<(), Error> {
     validate_identifier(id, 2, 128)
 }
 
+/// Default validator function for Channel identifiers.
+///
+/// A valid Identifier must be between 8-64 characters and only contain lowercase
```
romac, line 72:
```suggestion
/// A valid identifier must be between 8-64 characters and only contain
```

romac (approved):
Thanks for grinding through all this 🙏

Issue #2376, opened 2022-07-05
`hermes tx raw upgrade-chain` still uses `src`/`dst` convention


plafer:
## Summary of Bug
The `hermes tx raw upgrade-chain` still uses the old convention of `src`/`dst` instead of `reference`/`host`.

```shell
hermes tx raw upgrade-chain [OPTIONS] --dst-chain <DST_CHAIN_ID> --src-chain <SRC_CHAIN_ID> --src-client <SRC_CLIENT_ID> --amount <AMOUNT> --height-offset <HEIGHT_OFFSET>
```

## Version
0.15

## Acceptance Criteria

`host`/`reference`/`a`/`b` convention is used instead of `src`/`dst`

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #2315, opened 2022-06-17
Update `hermes tx raw` commands to drop the `raw` subcommand


ljoss17:
## Summary

Update the commands `hermes tx raw` in order to drop the `raw` subcommand.

## Problem Definition

Following discussion https://github.com/informalsystems/ibc-rs/pull/2306#discussion_r898326303 the `hermes tx raw` commands would be clear even after dropping the `raw`, so having `hermes tx <subcommand>`.

## Acceptance Criteria

* Remove `raw` from Hermes commands:
  - `hermes tx raw chan-close-confirm`
  - `hermes tx raw chan-close-init`
  - `hermes tx raw chan-open-ack`
  - `hermes tx raw chan-open-confirm`
  - `hermes tx raw chan-open-init`
  - `hermes tx raw chan-open-try`
  - `hermes tx raw conn-ack`
  - `hermes tx raw conn-confirm`
  - `hermes tx raw conn-init`
  - `hermes tx raw conn-try`
  - `hermes tx raw create-client`
  - `hermes tx raw ft-transfer`
  - `hermes tx raw packet-ack`
  - `hermes tx raw packet-recv`
  - `hermes tx raw update-client`
  - `hermes tx raw upgrade-chain`
  - `hermes tx raw upgrade-client`
  - `hermes tx raw upgrade-clients`

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #2383, opened 2022-07-08
Update ADR 010 to include `tx raw` commands


ljoss17:
## Summary

Updating the ADR 010 to include the `tx raw` commands.

## Problem Definition

The ADR 010 doesn't have the `tx raw` commands, which require changes in order to have a similar syntax as the other commands.

## Proposal

* Adding the `tx raw` commands
* Removing the `raw` subcommand
* Updating the flags in order to replace outdated prefixes `--src-` and `--dst-`

## Acceptance Criteria

ADR 010 complete with the new `tx raw` commands.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2404, opened 2022-07-13
Refactor `tx raw` commands

ljoss17:
Closes: #2315, closes: #2376

## Description

This PR refactors the `tx raw` commands in order to be synchronised with the ADR 010. The following modifications are applied:

* Remove the `raw` subcommand from all the `tx raw` commands
* Remove the four duplicate commands:
  * `tx raw update-client`, which is the same as `update client`
  * `tx raw upgrade-client`, which is the same as `upgrade client`
  * `tx raw upgrade-clients`, which is the same as `upgrade clients`
  * `tx raw create-client`, which is the same as `create client`
* Replace all flag prefixes `--src-` and `--dst-` with more appropriate prefixes
* Add unit tests for all `tx raw` commands

______

### PR author checklist:
- [x] Linked to GitHub issue.
- [x] Updated code comments and documentation (e.g., `guide/`).

### Reviewer checklist:


guide/src/commands/raw/index.md
```diff
@@ -1,13 +1,11 @@
-# Raw Transactions
+# Transactions
```
romac, line 1: Let's rename this file to `guide/src/commands/tx/index.md`

romac (commented):
Awesome, thanks so much @ljoss17! 💐 

Let's rename the `guide/src/commands/raw` folder to `guide/src/commands/tx` and we're good to go!

ancazamfir:
I know this is merged but I am not sure about some of these changes.
 
I find it confusing, especially for handshake Tx-es, where we now use `a-chain` and `b-chain` but then in the help strings to make things clear we still use source and destination. For packets seems a bit better but again we use source and destination in the help to explain things.
The general idea for tx (raw) commands is that, regardless of the type of action (relay packet, client or handshake message), you instruct the relayer to send one or more messages to a (destination) chain by retrieving state values and proofs from another (source) chain. So I find using src/ dst more clear here. Or chain/ counterparty.
`a-chain` and `b-chain` work fine in the `create connection/ channel` where you just want to say "create a connection/channel" between chains `a` and `b`. And it results in sending Tx-es to both `a` and `b`. 

I also find it confusing the sender/ receiver in `tx upgrade-chain`. We use host/ reference in the high level upgrade. We should use the same here.

Nit: there is stil `tx raw packet-recv` in the ADR.

adizere:
Is it a reasonable compromise if we make the explicit relation between `a-chain` being a destination always (wherever applicable) and `b-chain` being the source? I think this might be the case already.

ancazamfir:
> Is it a reasonable compromise if we make the explicit relation between `a-chain` being a destination always (wherever applicable) and `b-chain` being the source? I think this might be the case already, we would just document it as such throughout help and guide.

Not sure what you mean by "make an explicit relation". And btw, `a-chain` is the source/ counterparty chain (i just looked at all tx CLIs, at least we are consistent).
IMO we should fix this and not use `a-chain` and `b-chain` in the tx(raw) and fix the upgrade params names also.

ancazamfir:
I propose we use chain/ counterparty instead of b-chain/ a-chain. Except the chain upgrade where we can use host-chain/ reference-chain (with the note that the reference-chain is the chain to be upgraded)

seanchen1991:
> I propose we use chain/ counterparty instead of b-chain/ a-chain
Just to be clear, are you proposing we remove all `a-chain`/`b-chain` language entirely from the CLI?

ancazamfir:
> Just to be clear, are you proposing we remove all `a-chain`/`b-chain` language entirely from the CLI?

Only from the tx CLIs (old `tx raw`-s)

ljoss17:
As I see the process for the commands syntaxe, it usually has two steps:

1. One issue/PR to create/update an ADR which generates a discussion up to some mutual agreement
2. One issue/PR to update the commands in order to match the ADR

So I think this should be discussed in the issue regarding updating ADR 010 https://github.com/informalsystems/ibc-rs/issues/2383 or in a new issue. This would avoid having a mismatch between ADR 010 and the commands, as well as making it easier to track the discussion which lead to the chosen syntaxe.
What do you think ?

seanchen1991:
I opened #2410 to capture @ancazamfir's request 🙂

Issue #2408, opened 2022-07-14
Some issues and wrinkles with telemetry metrics


ancazamfir:
## Summary of Bug
I was trying to make sense of the telemetry metrics and while I find some of them very useful (like the cache hits, number of queries, workers, etc) I am struggling with others. Here are some notes I took:

missing:
 - light client queries and latencies for queries, verification, etc.
 - there is `Number of WriteAcknowledgement relayed` but not `Number of SendPacket relayed`. Should probably fix wording as those are events, hermes acts on events but relays messages, e.g. on `SendPacket` event it relays a `MsgRecvPacket` message, on `WriteAcknowledgement` event it relays `MsgAcknowledgement`.
 - there is a `Number of SendPacket relayed through ClearPendingPackets` but not `Number of WriteAcknowledgement relayed through ClearPendingPackets`. Or is this supposed to count only once per packet lifetime? (i.e. potentially two message). Could this be clarified?
 - number of update client messages

confusing:
 - relayed per channel only shows with `tx_confirmation = true`, help should mention this. But since almost no one runs with this set, I am wondering if we should record the times we send a msg successfully (even if we don't have a confirmation that it was not a duplicate or deliverTx has not failed). Not sure how it will work with retries.
 - `tx_latency_submitted_bucket` always show `le="+Inf"}`. I see the comments below but not clear why we don't have some buckets. We should be able to configure something like `10, 50, 100, 200, 500, 1000, 2000, Inf` no? At least explain there are no buckets or remove it altogether.
      ```
      // TODO: Once quantile sketches are supported, replace histograms with that.
      // For the moment, disable histogram buckets since no values make sense for all use-cases.
      ```

incorrect (?):
the oldest sequence/ timestamp. I send some packets and start hermes with `clear_on_start = false`. Hermes will not relay those packets and can be confirmed with `query packet pending` CLI. However the metrics show:
```
# HELP oldest_sequence The sequence number of the oldest pending SendPacket. If this value is 0, it means there are no pending SendPacket
# TYPE oldest_sequence gauge
oldest_sequence{chain="ibc-1",channel="channel-0",counterparty="ibc-0",port="transfer"} 0
# HELP oldest_timestamp The timestamp of the oldest sequence number in seconds
# TYPE oldest_timestamp gauge
oldest_timestamp{chain="ibc-1",channel="channel-0",counterparty="ibc-0",port="transfer"} 0
``` 
[...]

Issue #2175, opened 2022-05-03
Report on slow/stalled channel traffic


greg-szabo:
## Summary

User story: I need to better understand when a channel is being relayed properly. For monitoring and alerting purposes, I need to be able to query the oldest sequence number that is still in the queue for a specific channel and find out how old (date) the packet is.

## Problem Definition
We need to better monitor if a channel is being relayed properly or not. Out-of-band monitoring has the benefit of not relying on the technology actually doing the relaying but it has the disadvantage that it has to describe the infrastructure and application setup yet again from scratch. For example the hermes config details the relationship among networks so well, that if I say "channel-0" on the Osmosis network, everyone (including a program) understands exactly what that means (which endpoint represents it, what wallet can I use to manage it, etc).

Implementing this (and subsequent monitoring related) feature in Hermes takes advantage of the already existing configuration and library knowledge of endpoints. (Writing `curl` scripts to poll endpoint health is not fun. Especially, on gRPC.)

Including this and similar requests makes Hermes "ready with batteries" for production use, including monitoring assets. (Well, prometheus endpoints or HTTP API calls or somesuch. The operator still need to gather the data somewhere and present it, say, using Grafana.)

Disadvantage of this kind of feature is that it opens up topic that is not strictly IBC as a protocol but more on the side of "IBC as a product used in servers". Personally, I think it shows the maturity of a project, but others might have differing opinions. This request is fairly specific which might be good (when everyone needs it) or not so good (when it only serves one specific use-case of an operator).

## Proposal
There is a monitoring bot on Discord that essentially does something similar. The goal is to find out if a channel has "stuck" packets: we define "stuck" packets as packets that haven't been relayed for 5 minutes.

One implementation idea:
One or more prometheus metric(s) per-channel configured in Hermes, that displays the oldest sequence number on the channel still in the queue as well as the submission date associated with the sequence number. (Extra query to the channel.)

This could be picked up by any monitoring tool and alert on it every 5 minutes (or whatever the operator configures).

Alternatively, if this doesn't fit the prometeus metrics specs, it could be a HTTP web API call that responds with the data in a JSON object. Somehow, I feel prometheus should fit here, but we're open to other implementations (even a CLI command, if necessary). The one implementation that doesn't work for us is plugging this data into the log file. The data has to be independently queryable, mostly separated from the current operational state of Hermes. (As mentioned in the disadvantages.)

## Acceptance Criteria
There is a way to query the "stuck state" of a channel.

____

[...]

PR #2409, opened 2022-07-15
Improve metrics descriptions and naming

ljoss17:
Closes: #2408

## Description

This PR applies the following improvements to some telemetry metrics:

* Renamed the following metric: `cleared_count` to `cleared_send_packet_count`
* Added new metric: `cleared_acknowledgment_count`
* Improve help message for:
  * `ibc_receive_packets`
  * `ibc_acknowledgment_packets`
  * `ibc_timeout_packets`
  * `send_packet_count`
  * `acknowledgement_count`
  * `cleared_send_packet_count`
  * `oldest_sequence`
* Added histogram buckets for:
  * `tx_latency_submitted`
  * `tx_latency_confirmed`
* Changed default value of `tx_confirmation` to `false` for `config.toml`

Improvements suggested by operators:
- for oldest sequence number & timestamp, once that packet is relayed, the metric should be updated with a default (`0`) value so that it no longer signals any oldest sn/timestamp

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


telemetry/src/state.rs
```diff
@@ -495,8 +523,14 @@ impl AggregatorSelector for CustomAggregatorSelector {
             // https://docs.rs/opentelemetry-prometheus/0.10.0/src/opentelemetry_prometheus/lib.rs.html#411-418
             // TODO: Once quantile sketches are supported, replace histograms with that.
             // For the moment, disable histogram buckets since no values make sense for all use-cases.
-            "tx_latency_submitted" => Some(Arc::new(histogram(descriptor, &[]))),
-            "tx_latency_confirmed" => Some(Arc::new(histogram(descriptor, &[]))),
+            "tx_latency_submitted" => Some(Arc::new(histogram(
+                descriptor,
+                &[10.0, 50.0, 100.0, 200.0, 500.0, 1000.0, 2000.0],
+            ))),
```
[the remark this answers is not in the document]
ljoss17: Ok, I will set it up so it can be tested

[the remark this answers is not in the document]
ljoss17: I am getting higher values for `tx_latency_confirmed` than `tx_latency_submitted`. Is it expected or is it due to testing locally on my machine ?

`tx_latency_submitted`:

```
tx_latency_submitted_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="200"} 0
tx_latency_submitted_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="500"} 3
tx_latency_submitted_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="1000"} 4
tx_latency_submitted_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="2000"} 4
tx_latency_submitted_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="5000"} 4
tx_latency_submitted_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="10000"} 4
tx_latency_submitted_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="+Inf"} 4
tx_latency_submitted_sum{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer"} 1316
```

`tx_latency_confirmed`:

```
tx_latency_confirmed_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="200"} 0
tx_latency_confirmed_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="500"} 0
tx_latency_confirmed_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="1000"} 0
tx_latency_confirmed_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="2000"} 0
tx_latency_confirmed_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="5000"} 4
tx_latency_confirmed_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="10000"} 4
tx_latency_confirmed_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="+Inf"} 4
```

If this difference is expected, should we keep the same buckets anyway ?

guide/src/telemetry.md
```diff
@@ -46,10 +46,11 @@ The following table describes the metrics currently tracked by the telemetry ser
...
 | `query_cache_hits`           | Number of cache hits for queries emitted by the relayer, per chain and query type | `u64` Counter |
-| `send_packet_count`          | Number of SendPacket relayed | `u64` Counter |
-| `acknowledgement_count`      | Number of WriteAcknowledgement relayed               | `u64` Counter       |
-| `cleared_count`              | Number of SendPacket relayed through ClearPendingPackets | `u64` Counter   |
-| `oldest_sequence`            | The sequence number of the oldest pending SendPacket. If this value is 0, it means there are no pending SendPacket | `u64` ValueRecorder |
+| `send_packet_count`          | Number of SendPacket events processed | `u64` Counter |
+| `acknowledgement_count`      | Number of WriteAcknowledgement events processed      | `u64` Counter       |
+| `cleared_send_packet_count`    | Number of SendPacket events processed during the initial and periodic clearing | `u64` Counter   |
+| `cleared_acknowledgment_count` | Number of WriteAcknowledgement events processed during the initial and periodic clearing | `u64` Counter   |
```
ancazamfir, line 52: Could you try to get this counter in the example below? I tried and was not able to get it. Here is what I tried:
- send 20 packets to ibc-0
`hermes tx ft-transfer --sender-chain ibc-0 --receiver-chain ibc-1 --sender-channel channel-0 --sender-port transfer --amount 90 --timeout-height-offset 3000 --number-msgs 20 --key-name user2`
- check pending (shows 20 unreceived_packets)
`hermes  query packet pending --chain ibc-0 --channel channel-0 --port transfer`
- receive the 20 packets on ibc-1
`hermes tx packet-recv --receiver-chain ibc-1 --sender-chain ibc-0 --sender-port transfer --sender-channel channel-0`
- check pending (shows 20 unreceived_acks)
`hermes  query packet pending --chain ibc-0 --channel channel-0 --port transfer`
- send 10 more packets to ibc-0
`hermes tx ft-transfer --sender-chain ibc-0 --receiver-chain ibc-1 --sender-channel channel-0 --sender-port transfer --amount 90 --timeout-height-offset 3000 --number-msgs 10 --key-name user2`
- check pending (shows 10 unreceived_packets and 20 unreceived_acks)
`hermes  query packet pending --chain ibc-0 --channel channel-0 --port transfer`
- start hermes
`hermes start`
- wait a bit for hermes to finish clearing and then check metrics. I get these (cleared_acknowledgment_count):
```
# HELP acknowledgement_count Number of WriteAcknowledgement events processed
# TYPE acknowledgement_count counter
acknowledgement_count{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer"} 10

# HELP cleared_send_packet_count Number of SendPacket events processed during the initial and periodic clearing
# TYPE cleared_send_packet_count counter
cleared_send_packet_count{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer"} 10

# HELP msg_num How many messages Hermes submitted to the chain, per chain
# TYPE msg_num counter
msg_num{chain="ibc-0"} 32
msg_num{chain="ibc-1"} 11

# HELP send_packet_count Number of SendPacket events processed
# TYPE send_packet_count counter
send_packet_count{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer"} 10 
```
For me also the latency metrics are missing for one chain (`ibc-1`)

telemetry/src/state.rs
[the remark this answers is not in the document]
ancazamfir: ah good point, i have not yet run with confirmed but will do a bit later.  Yes they should be different. The confirmed ones in the order of the block time which typically is between 1 and 10 seconds. So if we cover 1..20 seconds it should be fine.
The submitted is just the `broadcast_tx_sync` rpc time.

guide/src/telemetry.md
ljoss17: Indeed I am not getting the `cleared_acknowledgment_count` as expected. I will investigate this issue.

ljoss17: There was an issue in where the `cleared_acknowledgment_count` was counted. I pushed a fix and running the example you gave now correctly has the following metrics:

```
acknowledgement_count{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer"} 10
cleared_acknowledgment_count{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer"} 20
send_packet_count{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer"} 10
cleared_send_packet_count{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer"} 10
```

There is indeed an issue recording latencies with the clearing:
* clearing unreceived_acks doesn't record any latency
* clearing unreceived_packets only record latencies for the sender chain
* sending packets when Hermes is running correctly record the latencies

I will check and see how this can be corrected

ancazamfir: excellent, thanks for looking into this!

telemetry/src/state.rs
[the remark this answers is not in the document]
adizere: There is a bit of background on the buckets work https://github.com/informalsystems/ibc-rs/pull/2182#issuecomment-1133007587. Originally we had 3 buckets but they were confusing (to me) and not useful in production (they were always 0). I decided to remove them, as we had three options:
1. populate bucket options with values that work for most cases, didn't do this as it's not clear which values make sense, and they vary between production/dev and between `confirmed`/`submitted`
2. add telemetry configuration options in `config.toml` and let operators define the bucket values
    - didn't do this as it didn't seem important enough (operators didn't say it's a limitation if there are no buckets)
3. remove the buckets

I opted for the simplest (last) option. But if we find values for option 1 that would be probably best. I suspect it's not worth the effort to permit configuration à la option 2.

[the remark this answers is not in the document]
ancazamfir: thanks for the ref, good points. i think 1. is fine and better to what we have now (all zeros and everything in Inf). We can start with some values (I think latest in this PR should do) and adjust as we get more metrics from operators. 2. is nice but we still need defaults which are basically case 1.

telemetry/src/state.rs
```diff
@@ -495,8 +523,14 @@ impl AggregatorSelector for CustomAggregatorSelector {
             // https://docs.rs/opentelemetry-prometheus/0.10.0/src/opentelemetry_prometheus/lib.rs.html#411-418
             // TODO: Once quantile sketches are supported, replace histograms with that.
             // For the moment, disable histogram buckets since no values make sense for all use-cases.
```
romac, line 525: Let's remove this line since it's outdated.

config.toml
```diff
@@ -59,9 +59,10 @@ clear_on_start = true
 # (to any chain in this config file) have been successfully delivered.
 # If they have not been, and `clear_interval = 0`, then those packets are
 # queued up for re-submission.
-# Experimental feature. Affects telemetry if set to false.
-# [Default: true]
-tx_confirmation = true
+# Experimental feature. If set to false this will disable telemetry
```
romac, line 62:
```suggestion
# Experimental feature. If set to false this will disable the following telemetry metrics:
```

guide/src/telemetry.md
```diff
@@ -35,9 +35,9 @@ The following table describes the metrics currently tracked by the telemetry ser
...
 | `ibc_client_misbehaviours`   | Number of misbehaviours detected per client          | `u64` Counter       |
-| `ibc_receive_packets`        | Number of receive packets relayed per channel        | `u64` Counter       |
-| `ibc_acknowledgment_packets` | Number of acknowledgment packets relayed per channel | `u64` Counter       |
-| `ibc_timeout_packets`        | Number of timeout packets relayed per channel        | `u64` Counter       |
+| `ibc_receive_packets`        | Number of confirmed receive packets relayed per channel. Available if relayer runs with Tx confirmation enabled        | `u64` Counter       |
+| `ibc_acknowledgment_packets` | Number of confirmed acknowledgment packets relayed per channel. Available if relayer runs with Tx confirmation enabled | `u64` Counter       |
+| `ibc_timeout_packets`        | Number of confirmed timeout packets relayed per channel. Available if relayer runs with Tx confirmation enabled        | `u64` Counter       |
 | `wallet_balance`             | The balance of each wallet Hermes uses per chain. Please note that when converting the balance to f64 a loss in
```
ancazamfir, line 41: I have this broken output for this (see the `\n`, looks weird in the browser)
\# HELP wallet_balance The balance of each wallet Hermes uses per chain. Please note that when converting the balance to f64 a loss in\n                precision might be introduced in the displayed value

guide/src/telemetry.md
ancazamfir: back to the counters, for the example I gave I get now (*):
```
acknowledgement_count{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer"} 10
cleared_acknowledgment_count{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer"} 20

send_packet_count{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer"} 10
cleared_send_packet_count{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer"} 10

msg_num{chain="ibc-0"} 32
msg_num{chain="ibc-1"} 11
```
should `acknowledgement_count` be 30 if it is the total? Meaning there were 20 pending acks plus the 10 that came in live from the processing of the 10 send_packet.

guide/src/telemetry.md
```diff
@@ -65,127 +66,133 @@ see Prometheus-encoded metrics.
...
-send_packet_count{chain="ibc-0",channel="channel-0",counterparty="ibc-2",port="transfer"} 5
-send_packet_count{chain="ibc-1",channel="channel-0",counterparty="ibc-0",port="transfer"} 6
-send_packet_count{chain="ibc-2",channel="channel-1",counterparty="ibc-1",port="transfer"} 7
+send_packet_count{chain="ibc-1",channel="channel-0",counterparty="ibc-0",port="transfer"} 35
+# HELP tx_latency_confirmed The latency for all transactions submitted & confirmed to a specific chain, i.e. the difference between the moment when Hermes received a batch of events until the corresponding transaction(s) were confirmed. Milliseconds.
+# TYPE tx_latency_confirmed histogram
+tx_latency_confirmed_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="10"} 0
+tx_latency_confirmed_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="50"} 0
+tx_latency_confirmed_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="100"} 0
+tx_latency_confirmed_bucket{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer",le="200"} 0
```
ancazamfir, line 145: Can we update the metrics here with some nicer numbers?

relayer/src/config.rs
```diff
@@ -51,7 +51,7 @@ pub mod default {
     }
 
     pub fn tx_confirmation() -> bool {
-        true
+        false
```
ancazamfir, line 54: This should be documented explicitly in the release notes as it changes a default.

guide/src/telemetry.md
ljoss17: I ran some commands in order to have better telemetry output example for the Hermes guide

relayer/src/config.rs
ljoss17: I updated the changelog entry to explicitly state the change of default value

guide/src/telemetry.md
ancazamfir: For me this is confusing.  Also `send_packet_count` will not mirror `acknowledgment_count` if any packets have timed out. IMO we should be consistent for send and ack metrics. But if we keep things the way they are we need to explain very clearly what the metrics mean in the help strings. The name of the counters would also need to reflect the difference (`send_packet_count` is a total while the `acknowledgment_count` counts the live events only). Something like:

```
live_acknowledgement_count{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer"} 10
cleared_acknowledgment_count{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer"} 20

total_send_packet_count{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer"} 10
cleared_send_packet_count{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer"} 10
```

We could also have `live_send_packet_count` instead of `total`.

adizere:
> Changed default value of tx_confirmation to false for config.toml


I'm curious, what is the rationale for making this default false? Some operators are running with this enabled. Some are disabling it for optimization reasons.

ancazamfir:
> > Changed default value of tx_confirmation to false for config.toml
> 
> I'm curious, what is the rationale for making this default false? Some operators are running with this enabled. Some are disabling it for optimization reasons.

It involves more complex processing and logic. The config also says "experimental". And for some reason I thought most operators run with it disabled. Is there any way to gauge this?

guide/src/telemetry.md
ljoss17: Oh I hadn't taken into account the timeouts. The points raised regarding these metrics make me realise that these metrics seem more confusing than helpful.

The addition of these metrics was originally to spot slow/stalled traffic https://github.com/informalsystems/ibc-rs/issues/2175, but in my opinion the `oldest_sequence` and `oldest_timestamp` should be enough to track this.

Furthermore these metrics, `send_packet_count`, `acknowledgment_count`, `cleared_send_packet_count` and `cleared_acknowledgement_count` seem to aim at corner cases rather than general cases, they work well if there are no timeouts, and/or pending packets.

The more points were raised the more I had trouble convincing myself of the usefulness of these metrics. I am wondering if it is a good idea to keep them. What do you think ?

ancazamfir: I think we should keep something, they are helpful as there is no other hint on possible packets relayed when `tx_confirmation` is false. It's the best we can do to say we processed x, y type events. 
What is not completely clear is how to reduce the confusion. As I said, if we keep what we have, we can clarify by having good help and metric names. Since these are already in the code base I would say let's do it the best we can in this PR. Once we experiment more we can change later along the other lines I mentioned or something else or even remove them.

adizere:
>  Is there any way to gauge this?
I'm not sure what you mean. Do you mean measure/analytics what % of operators have `tx_confirmation = true`?

We discussed yesterday with operators and here are 3 takeaways, also based on my observations so far:

1. Even after a channel is cleared, `oldest_*` metrics can remain to the same value (i.e., not reset to `0`).
    - I'm not sure why that would be the case, because in v1-rc.0 we _do_ reset the field to `0`
    - https://github.com/informalsystems/ibc-rs/blob/b80bceab591a291203af8c41732b016cf612a296/telemetry/src/state.rs#L468-L470

2. Some operators do a _discovery_ phase to understand the Prometheus schema, but Hermes upon starting up does not populate most of the metric fields
    - the most important fields that we should initially populate with some zero value would be the `oldest_*` ones and the ones related to activity (`msg_num` and `send_packet_count` for example).


3. I think we should clarify what the `oldest_timestamp` is, it seems this field is a _local_ timestamp to the Hermes process, not an on-chain packet timestamp (when the packet was created), which is not clear from the telemetry help message, specifically:
    > \# HELP oldest_timestamp The timestamp of the oldest sequence number in seconds
    > \# TYPE oldest_timestamp gauge
    oldest_timestamp{chain="ibc-0",channel="channel-3",counterparty="ibc-1",port="transfer"} 0
    - https://github.com/informalsystems/ibc-rs/blob/7e422494746273001ea223a75ebcb3a1c14eb9af/telemetry/src/state.rs#L424-L429

ancazamfir:
> I'm not sure what you mean. Do you mean measure/analytics what % of operators have `tx_confirmation = true`?

Yes.

I agree with all of your other 3 points @adizere. I was also confused about the lack of "discovery".

adizere (approved):
Great improvements in here! Did some tests and looks like a clear improvement over what is currently in `master`. Will pick up the work in a follow-up issue + PR.

Issue #2335, opened 2022-06-24
Extract light client implementations into separate crates


hu55a1n1:
## Summary
The module and relayer code relies heavily on enums such as `AnyClientState`, `AnyClient`, etc. which define one variant per type of light-client supported. As we look to add support for more light-clients, it would be great for the module code to be agnostic to the light-client implementation. This will also light-client implementations to be hosted outside of the ibc-rs repo and maintained independently. 

## Problem Definition
Tight coupling with `Any*` types and circular dependency between light-client code and module code.

## Proposal
Remove all occurrences of Any* enums and solve circular dependencies between crates if any. 

## Acceptance Criteria
Light clients can be hosted outside of the ibc-rs repo. 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2412, opened 2022-07-16
Object safe Protobuf

hu55a1n1:
Partially addresses: #2335 

Replaces `tendermint_proto::Protobuf` with an object safe variant. 

Note: The target branch is a long-lived branch [hu55a1n1/light-client-extraction](https://github.com/informalsystems/ibc-rs/tree/hu55a1n1/light-client-extraction). The plan is to merge PRs (addressing #2335) such as this into [hu55a1n1/light-client-extraction](https://github.com/informalsystems/ibc-rs/tree/hu55a1n1/light-client-extraction) and ultimately merge that into master. 
______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


proto/src/protobuf/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+/// }
+///
+/// impl From<MyDomainType> for MyRawType {
+///     fn from(value: MyDomainType) -> Self {
+///         Self { a: value.a, b: value.b }
+///     }
+/// }
+///
+/// impl Protobuf<MyRawType> for MyDomainType {}
+///
```
seanchen1991, line 70: These comments seem to document an example of what this module enables, and don't necessarily only pertain to the `Protobuf` trait. As a result, I wonder if these comments would be more appropriate as module-level docs using `//!` at the top of the file. What do you think?

seanchen1991 (commented):
I love the thoroughness of the documentation! Just have a small question on the docs for the `Protobuf` trait.

proto/src/protobuf/mod.rs
hu55a1n1: Good point! Done -> https://github.com/informalsystems/ibc-rs/pull/2412/commits/8de194b12486bcdbdba981cca6b93afca4c4dfab
PS: the doc comments were copied from `tendermint-rs` so credits to them for the thorough documentation. :tada: :sweat_smile:

proto/src/protobuf/erased.rs
```diff
@@ -0,0 +1,40 @@
+use alloc::boxed::Box;
+use core::convert::{Into as CoreInto, TryFrom as CoreTryFrom};
+
+mod sealed {
```
plafer, line 4: We can remove the `sealed` module entirely. It basically defines two "marker traits", and "marks" all types that satisfy the constraints:
+ `SealedInto<T>` for `U: Clone + CoreInto<T>`
+ `SealedTryFrom<T>` for `U: CoreTryFrom<T>`
These markers are then used as supertraits to our `Into` and `TryFrom`.

We see the same constraints being applied to our blanket implementations. So, the supertraits to our `Into` and `TryFrom` can simply be removed; the blanket implementation takes care of the constraints.

TLDR: if we remove the `sealed` module, the `sealed::SealedInto<T>` supertrait on `Into`, and the `sealed::SealedTryFrom<T>` supertrait on `TryFrom`, everything compiled and we've reduced the complexity of this module.

proto/src/protobuf/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+use core::fmt::Display;
+
+use bytes::Buf;
+use prost::encoding::encoded_len_varint;
+use prost::Message;
+
+pub use self::error::Error;
+
+/// Object safe equivalent of `tendermint_proto::Protobuf`.
+pub trait Protobuf<T: Message + Default>
```
plafer, line 71: How about we rename `T` to `Raw` or `RawType`? I always need an extra mental step to remember whether we do `Protobuf<Raw> for Domain`, or the opposite.

proto/src/protobuf/erased.rs
```diff
@@ -0,0 +1,40 @@
...
+    use super::*;
+
+    pub trait SealedInto<T: ?Sized> {}
+    impl<T, U: Clone + CoreInto<T>> SealedInto<T> for U {}
+    pub trait SealedTryFrom<T> {}
+    impl<T, U: CoreTryFrom<T>> SealedTryFrom<T> for U {}
+}
+
+pub trait Into<T: ?Sized>: sealed::SealedInto<T> {
+    fn into(&self) -> Box<T>;
```
plafer, line 14: I think we should give a different name to `into()`, such as `into_ref()`. With the current version, I noticed weird behaviors in my editor. When hovering over a call to `erased::Into::into()` in one of the functions of our `Protobuf` implementation, rust-analyzer claims that the `into()` being referenced is the one in `core::convert::Into`. I even tried to explicitly write
```rust
let raw: Box<T> = self.into();
```
and rust-analyzer still thinks that I'm using `core::convert::into`. I haven't tested other language servers.

I then played around a bit in a rust playground [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ea46d1568133bf5089af0a1665b04114). I'm not sure why it's not working; perhaps it has something to do with `core::convert::Into` being in the prelude?

In any case, using another name for `into()` avoids these headaches.

proto/src/protobuf/mod.rs
adizere: Please do!

proto/src/protobuf/erased.rs
hu55a1n1: This is an interesting problem. I used `erased::Into` because that is the convention that erased_serde uses and I thought that it was a common convention, but it seems we're seeing problems because we're using it with std lib traits (in the prelude). IMHO, `into_ref()` would be a method that takes `self` by value, so a better name would perhaps be `to_boxed()`. WDYT?

proto/src/protobuf/erased.rs
hu55a1n1: The sealed traits signal that they are not supposed to be implemented by the user. I think it's nice to have them because we provide blanket impls for these traits.

proto/src/protobuf/mod.rs
hu55a1n1: Done! https://github.com/informalsystems/ibc-rs/pull/2412/commits/eca98a1f4826d40d111f1ba86c29396d6e5522e8

proto/src/protobuf/erased.rs
plafer: But what does having them provide that my suggestion doesn't? I understand that they're not supposed to be implemented by the user; I just don't see why they're useful at all.

>TLDR: if we remove the sealed module, the sealed::SealedInto<T> supertrait on Into, and the sealed::SealedTryFrom<T> supertrait on TryFrom, everything compiles and we've reduced the complexity of this module.

proto/src/protobuf/erased.rs
plafer: I agree that `into_ref()` suggests taking ownership. On the flip slide, for me, `to_boxed()` suggests boxing `self` (i.e. no conversion to another type), such as

```rust
fn to_boxed(&self) -> Box<Self> {
    Box::new(self.clone())
}
```

Brainstorming alternatives:
+ `into_boxed()`: still `into` suggests taking ownership of `self`
+ `clone_boxed()`: I don't like it, doesn't suggest a conversion either
+ `clone_into_boxed()`: quite wordy, but perhaps captures best the conversion we're doing?
+ `as_boxed()`: `as_*` in `std` take a `&self`, and suggest a conversion?
+ ... ?

plafer (approved):

Issue #2381, opened 2022-07-06
Prometheus metric `wallet_balance` incorrect


mturkia:
## Summary of Bug

The wallet_balance prometheus metrics converts the account balance by dividing value by 10^6, which works for most Cosmos chains, but does not work e.g. for Crypto.org chain where it should be 10^8, or evmos with 10^18. There are also other chains with base unit something else than micro.

## Version

0.15.0+e3d0b108

## Steps to Reproduce

Configure Hermes with crypto.org (or fetch, evmos, provenance) and add e.g. 1 cro (100000000basecro) to the account. Observe that wallet_balance gives you 100cro as balance.

## Acceptance Criteria

Correct chain specific denom exponent is used to calculate account balance.

____

## For Admin Use

- [x] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ljoss17:
Hi @mturkia we can suggest the following solution. Instead of parametrising the value used to divide the balance, we will correct the description of the `wallet_balance` metric to be accurate. If I take your example of a wallet balance being 100000000 basecro, the new output would look like this:
```
wallet_balance{account=<ACCOUNT_ADDRESS>,chain=<CHAIN_ID>,denom="basecro 10E-6"} 100
```
Would this solution be acceptable ?

romac:
Some operators are reporting that they would like to keep the raw balance as was previously reported. In fact, they use wallets with small amount of coins which fit into a u64 and are therefore not interested in the divided value which is throwing off their metrics. I would suggest we still add a setting to specify by how much to divide or more simply by how many decimals to remove from the balance (eg. a setting of 6 would divide by 10^6).

mturkia:
I think that from usage point of view, the raw balance would be better than always dividing by 6. So either raw value, or correct base unit value.

romac:
We cannot do the raw value only because some operators use wallets with with a high balance which overflows the u64 counter (the Rust OpenTelemetry crate currently only support 64-bits counters) but with this change you should be able to get the raw value by eg. setting the exponent to 0 if we take it to mean "divide by 10^{exponent}". Would that work for you?

mturkia:
Should be fine for me

PR #2425, opened 2022-07-19
Configure wallet balance unit in telemetry

ljoss17:
Closes: #2381 

## Description

This PR adds a configuration, `wallet_metric_scale`, per chain which allows to specify the factor by which the `wallet_balance` from the telemetry metrics is divided by.

The default value is `wallet_metric_scale = 0`, which means that if the raw balance is displayed. If we set `wallet_metric_scale = 6` and the raw balance is `10'000'000`, then the telemetry metric will display `wallet_balance = 10`.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


.changelog/unreleased/features/ibc-telemetry/2381-configure-wallet-balance-divider.md
```diff
@@ -0,0 +1,3 @@
+- Added a new configuration, `wallet_metric_scale`, to `config.toml` to specify the
+  value used to divide the `wallet_balance` displayed in the telemetry metrics.
```
romac, line 2:
```suggestion
  value used to scale down the `wallet_balance` displayed in the telemetry metrics
  in case the raw value does not fit into a unsigned 64-bits integer,
  eg. because the relayer wallet contains a high amount of coins.
```

config.toml
```diff
@@ -188,6 +188,14 @@ clock_drift = '5s'
 # Note: This MUST be the same as the `max_expected_time_per_block` genesis parameter for Tendermint chains.
 max_block_time = '30s'
 
+# Specify the order of magnitude used to convert a balance to a readable balance.
```
romac, line 191:
```suggestion
# Specify the order of magnitude used to scaled down the wallet balance reported via telemetry to a readable balance.
```

config.toml
```diff
@@ -188,6 +188,14 @@ clock_drift = '5s'
...
+# This configuration will only affect the value displayed by the telemetry metric `wallet_balance`.
+# E.g. with wallet_metric_scale = 6 the raw balance will be divided by 10^6, so if the
```
romac, line 193:
```suggestion
# E.g. with `wallet_metric_scale = 6`, the raw balance will be divided by 10^6, so if the
```

config.toml
```diff
@@ -188,6 +188,14 @@ clock_drift = '5s'
...
+# raw balance is 10'000'000, the telemetry will display
+# `wallet_balance{account=<ACCOUNT_ID>, chain=<CHAIN_ID>, denom=<DENOM>} 10`
+# Default wallet_metric_scale 0
```
romac, line 196:
```suggestion
# Default: 0 (no scaling is applied and the raw value is reported directly)
```

relayer/src/worker/wallet.rs
```diff
@@ -52,18 +53,18 @@ pub fn spawn_wallet_worker<Chain: ChainHandle>(chain: Chain) -> TaskHandle {
         } else {
             trace!(
                 %amount, denom = %balance.denom, account = %key.account,
-                "amount cannot be scaled down to fit into u64 and therefore won't be reported to telemetry"
+                "amount cannot be scaled down to fit into u64 and therefore won't be reported to telemetry.\nPlease increase the value of `wallet_metric_scale` in Hermes `config.toml`.\nTo convert from uatom to atom set `wallet_metric_scale = 6`."
```
romac, line 56:
```suggestion
                "amount cannot be scaled down to fit into u64 and therefore won't be reported to telemetry. \
                Please increase the value of `wallet_metric_scale` in Hermes `config.toml`.\
                For example, to report the balance as atoms instead of uatoms, set `wallet_metric_scale = 6`."
```

romac (changes requested):

relayer/src/config.rs
```diff
@@ -73,6 +73,10 @@ pub mod default {
     pub fn connection_delay() -> Duration {
         ZERO_DURATION
     }
+
+    pub fn wallet_metric_scale() -> u32 {
+        0
```
adizere, line 78: Default = 6?

ljoss17: We ended up deciding to use 0 as a default. I will update the PR description

PR #2430, opened 2022-07-19
Test and fix integration tests on Osmosis and Wasmd

soareschen:
## Description

Add CI integration test for Osmosis and Wasmd via Cosmos.nix.

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/tools/test-framework/src/types/single/node.rs
+++ b/tools/test-framework/src/types/single/node.rs
@@ -138,7 +138,7 @@ impl FullNode {
             default_gas: None,
             max_gas: Some(3000000),
             gas_adjustment: None,
-            gas_multiplier: None,
+            gas_multiplier: Some(1.1),
             fee_granter: None,
             max_msg_num: Default::default(),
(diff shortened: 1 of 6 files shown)


tools/test-framework/src/types/single/node.rs
```diff
@@ -138,7 +138,7 @@ impl FullNode {
             default_gas: None,
             max_gas: Some(3000000),
             gas_adjustment: None,
-            gas_multiplier: None,
+            gas_multiplier: Some(1.1),
```
soareschen, line 141: This needs to be set to 1.1 for the integration tests on Osmosis and Wasmd to pass. Should there be adjustment on the default values so that there is no need for users to configure such magic value depending on the chains?

romac: Yes, let's change it to 1.1 in a different issue/PR since that seems to work in most if not all cases.

romac (approved):

Issue #2451, opened 2022-07-21
Telemetry: introduce `backlog_*` metrics, plus minor fixes


adizere:
This is a follow-up to the work for #2408 and PR #2409.

### Suggestions

We discussed yesterday with operators and here are some takeaways, also based on my observations so far:

1. Even after a channel is cleared, `oldest_*` metrics can remain to the same value (i.e., not reset to `0`).
    - I'm not sure why that would be the case, because in v1-rc.0 we _do_ reset the field to `0`
    - https://github.com/informalsystems/ibc-rs/blob/b80bceab591a291203af8c41732b016cf612a296/telemetry/src/state.rs#L468-L470

3. I think we should clarify what the `oldest_timestamp` is, it seems this field is a _local_ timestamp to the Hermes process, not an on-chain packet timestamp (when the packet was created), which is not clear from the telemetry help message, specifically:
    > \# HELP oldest_timestamp The timestamp of the oldest sequence number in seconds
    > \# TYPE oldest_timestamp gauge
    oldest_timestamp{chain="ibc-0",channel="channel-3",counterparty="ibc-1",port="transfer"} 0
    - https://github.com/informalsystems/ibc-rs/blob/7e422494746273001ea223a75ebcb3a1c14eb9af/telemetry/src/state.rs#L424-L429

3. Let's rename `oldest_*` metrics to `backlog_*` and additionally:
   - make it clear these metrics are per-channel
    - add a `backlog_size` metric to capture the number of pending packets in a channel.

AlianBenabdallah:
About 1., if a packet times out then it stays recorded.

AlianBenabdallah:
After more investigations, I think that `sequences_histories` is correctly handled only when a packet is successfully transferred and an ack is received. 

Any other case will lead to inconsistencies in the metrics.

Goldmindsx:
> This is a follow-up to the work for #2408 and PR #2409.
> 
> ### Suggestions
> We discussed yesterday with operators and here are 3 takeaways, also based on my observations so far:
> 
> 1. Even after a channel is cleared, `oldest_*` metrics can remain to the same value (i.e., not reset to `0`).
>    
>    * I'm not sure why that would be the case, because in v1-rc.0 we _do_ reset the field to `0`
>    * https://github.com/informalsystems/ibc-rs/blob/b80bceab591a291203af8c41732b016cf612a296/telemetry/src/state.rs#L468-L470
[...]

PR #2457, opened 2022-07-25
Metric updates when timeout occurs

AlianBenabdallah:
This PR fixes the following:

- [X] #2451
- [x] #2467 
- [X] #2469 
- [ ] "There is an inconsistency between send packets count and acknowledgements count in the displayed metrics."
    - This is Anca's comment from #2451, need to investigate further
- [ ] Discover phase as described in #2451 

## Description

> 1. Even after a channel is cleared, oldest_* metrics can remain to the same value (i.e., not reset to 0).

> 3. I think we should clarify what the oldest_timestamp is, it seems this field is a local timestamp to the Hermes process, not an on-chain packet timestamp (when the packet was created), which is not clear from the telemetry help message, specifically:

> 4. There is an inconsistency between send packets count and acknowledgements count in the displayed metrics.

Previously, `oldest_*` metrics were not updated when a timeout would occur. The `oldest_*` metrics take their values from the `sequences_histories` map which was updated only when an ACK was received. 

This PR handles the case of a timeout by calling `record_ack_history` when a timeout occurs. 

The description of the `oldest_*` metric is also updated to incorporate the fact that the `oldest_*` is not only updated when an ACK occurs.
______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


.changelog/unreleased/bug-fixes/ibc-relayer-cli/2429-fix-oldest-sequence-timeout.md
```diff
@@ -0,0 +1,2 @@
+- Updates sequence_histories when a timeout occurs
```
romac:
```suggestion
- Ensure `oldest_sequence` and `oldest_timestamp` metrics are updated when a timeout occurs
```

relayer/src/worker/packet.rs
```diff
@@ -267,17 +267,32 @@ fn acknowledgment_metrics(path: &Packet, summary: &RelaySummary) {
...
-    let count = summary
+    let mut count = 0;
+    summary
         .events
         .iter()
-        .filter(|e| matches!(e, TimeoutPacket(_)))
-        .count();
+        .filter(|event| matches!(event, TimeoutPacket(_)))
+        .for_each(|e| {
+            let height = e.height().revision_height(); // TODO : remove height argument
```
romac, line 276: Let's remove the height arguments everywhere in this PR?

.changelog/unreleased/bug-fixes/ibc-relayer-cli/2429-fix-oldest-sequence-timeout.md
romac: I am suggesting the change above because changelog entries should be written with users of the relayer in mind, to the extent that it is possible of course.

relayer/src/worker/packet.rs
AlianBenabdallah: I thought maybe opening a new PR would be good as well. If removing height introduces new bugs, breaks some tests or if it becomes useful to revert it, having two separate PRs could be helpful. What do you think about it ?

.changelog/unreleased/bug-fixes/ibc-relayer-cli/2429-fix-oldest-sequence-timeout.md
AlianBenabdallah: Done !

seanchen1991 (approved):
Other than Romain's comments, this looks good to me! 👍

adizere (approved):
Amazing work Ali!

In additional to the bugs attached to the PR, we did some work refining the metrics and try to make them more useful, specifically:

- renamed `oldest_sequence` -> `backlog_oldest_sequence`
- renamed `oldest_timestamp` -> `backlog_oldest_timestamp`
- introduced `backlog_size` to complement the other `backlog_*` data, as a metric reporting how many packets are pending on a channel


### What's left to do:

- [ ] "There is an inconsistency between send packets count and acknowledgements count in the displayed metrics."
    - This is Anca's comment from https://github.com/informalsystems/ibc-rs/issues/2451, need to investigate further
- [ ]  Discovery phase as described in https://github.com/informalsystems/ibc-rs/issues/2451

I think both of these should be done in a follow-up PR. The present PR is linked to 3 issues and will be difficult to track what change is fixing what issue.

#### Discovery phase

Not sure how we can implement the discovery phase, because Hermes does not start packet workers for any channel until there's packets to relay on that channel, so not sure how we can populate the channel-specific metrics. For the other (non-channel-specific) metrics, we can probably initialize them with zero.

#### Inconsistency
For the "inconsistency between send packets count and acknowledgements count" we need some more digging. We actually have a lot of complexity around these metrics, there's four of them:

1. send_packet_count
2. acknowledgement_count
3. cleared_send_packet_count
4. cleared_acknowledgment_count

and its not clear how operators can use them (aside from inconsistencies). At the very least, we should document what is the purpose and insights they offer. Do they indicate activity? Lack of activity? Problems relaying? I'm not sure I know what they do.

AlianBenabdallah:
I tested again and the bug is correctly fixed.

PR #2473, opened 2022-07-28
Fix some broken links

romac:


adizere (approved):
Thanks Romain!

This is not caught by markdown lint checker btw?
I'm confused why we have periodic PRs like this.

romac:
We run the link checker nightly rather than on every PR these days because of too many false positives, ie. when checking links, those often time out or are unreachable temporarily.

Issue #2477, opened 2022-07-29
Hermes may break the `max_tx_size` configuration param


ancazamfir:
## Summary of Bug
When we batch messages in a Tx we may exceed the `max_tx_size` if `max_msg_num` is not hit. The problem code is here:
https://github.com/informalsystems/ibc-rs/blob/1d188187721051a34d20841c0146b8c16b5b413b/relayer/src/chain/cosmos/batch.rs#L152-L157
We keep in the batch the msg that caused the crossing of `max_tx_size`.

It becomes an issue with chains like osmosis where Tx-es with size bigger than some threshold may be priced differently resulting in Tx simulation failure and no recovery (see https://github.com/informalsystems/ibc-rs/issues/2422, the threshold is exceeded with 3 msgs, one update client and two packet messages)

In addition, the actual Tx will be slightly bigger due to extra auth encoding and we should do something about it. 

## Version
all recent

## Steps to Reproduce
The code is obviously wrong but:
- add `assert!(current_size < max_tx_size);` after L154
- configure hermes with:
```
max_msg_num = 50
max_tx_size = 15000
```
- send 100 packets

## Acceptance Criteria

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

soareschen:
Since `batch_messages` is now a pure function, we should write some unit tests that test the behavior of `batch_messages` directly without having to run the full integration test.

mzabaluev:
> Since `batch_messages` is now a pure function, we should write some unit tests that test the behavior of `batch_messages` directly without having to run the full integration test.

Thanks for reminding, it's good unit testable function indeed.

seanchen1991:
@mzabaluev I've started working on adding some unit tests

PR #2480, opened 2022-07-29
relayer: Fix one-past-the-limit flaw in message batching

mzabaluev:
Closes: #2477

## Description

Prevent the batch data size from exceeding the given tx size limit.
Previous code pushed the offending last message to the end of the tx,
resulting in an oversized transaction.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/unreleased/bug-fixes/ibc-relayer/2477-prevent-oversized-tx-batch.md
+++ b/.changelog/unreleased/bug-fixes/ibc-relayer/2477-prevent-oversized-tx-batch.md
@@ -1,0 +1,3 @@
+- Fix code that could result in message batch size growing above
+  the transaction size limit
+  ([#2477](https://github.com/informalsystems/ibc-rs/issues/2477)).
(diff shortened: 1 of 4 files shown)


soareschen:
Since `batch_messages` is now a pure function, let's have some unit tests that test the behavior of `batch_messages` directly without having to run the full integration test.

ancazamfir:
thanks for fixing this @mzabaluev !
Adding a note here, I checked the difference between the total length of the msgs in a batch and the final len of the serialized Tx and it's 278bytes. This is on a local gaia setup. We could probably recommend then for osmosis to allow for ~2k bytes for overhead until we figure out how to get that more exactly.

ancazamfir (approved):
Looks good! Thanks for the quick fix @mzabaluev !

romac (changes requested):
Nice! Let's just add some unit tests before merging, please :)

seanchen1991:
Tangentially related, but I noticed that there's no need for the `batch_messages` function to return a `Result`, should we change the signature of the function to just return `Vec<Any>`?

seanchen1991 (approved):

romac:
Can we expand the unit tests a bit, eg. to test for corner cases etc?

seanchen1991:
@romac I added a few more unit tests. Let me know if there are additional corner cases that you think should be covered 🙂

romac (approved):

Issue #2479, opened 2022-07-29
Telemetry: last fixes for v1


adizere:
## Proposal

There's a few things left:

- There is an inconsistency between send packets count and acknowledgements count in the displayed metrics.
    - Ensure the documentation on `send_packet_count`, `acknowledgement_count`, `cleared_send_packet_count` and `cleared_acknowledgment_count` is clear, as these four metrics seem to be difficult to use

- Discovery phase: Some operators do a _discovery_ phase to understand the Prometheus schema, but Hermes upon starting up does not populate most of the metric fields
    - the most important fields that we should initially populate with some zero value would be the `backlog_*` ones and the ones related to activity (`msg_num` and `send_packet_count` for example).

## Acceptance Criteria

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

AlianBenabdallah:
> Discovery phase: Some operators do a _discovery_ phase to understand the Prometheus schema, but Hermes upon starting up does not populate most of the metric fields
> 
> * the most important fields that we should initially populate with some zero value would be the `backlog_*` ones and the ones related to activity (`msg_num` and `send_packet_count` for example).

The branch `ali/metrics_discovery` now implements a discovery phase when exact filters are given in the config file. Every metric should be initialized except for `tx_latency_*`. I'm still unsure about how to init these metrics as they use histograms.

If no exact filter is given, the relayer needs to query every channel. I'll see if it's possible to do it with one query but I guess that it might hurt performances if there are too many channels.

AlianBenabdallah:
Regarding the metrics, we have `acknowledgment_packets` and `acknowledgement_count`. I think the latest should be renamed as `acknowledgement_events`. 

Also, would it be useful to add a metric `timeout_events`  just like for acknowledgements ?

PR #2505, opened 2022-08-02
Metrics discovery phase

AlianBenabdallah:
Partially closes: #2479

## Description

Introduces a discovery phase to initialize the metrics.

Changes :
- `state.rs` now contains functions to initialize metrics specific to a chain / client / channel / path.
- Hermes now scans connections and channels in every case. Previously, the relayer would scan channels only when exact filters were specified in the config, otherwise it would just scan every client. 
- Minor code improvements

Note : 
- tx_latency* metrics are still not initialized because a histogram can only be initialized by adding a value and values can't be removed.
- The impact of the discovery phase on performances is still unknown. On a small test net, there is no degradation.
______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


relayer/src/supervisor/scan.rs
```diff
@@ -339,6 +347,15 @@ impl<'a, Chain: ChainHandle> ChainScanner<'a, Chain> {
...
                 }) => {
+                    let counterparty_chain_id = client.client_state.chain_id();
+                    init_telemetry(
+                        &chain.id(),
+                        &client.client_id,
+                        &counterparty_chain_id,
+                        channel_id,
+                        port_id,
+                    );
+
```
ljoss17, line 358: As the `client.client_state.chain_id()` clones the `ChainId` value I was wondering if:

1. It is more optimal to directly pass the reference `&client.client_state.chain_id()` to the `init_telemetry` call:
```
 init_telemetry(
    ...
    &client.client_state.chain_id()
    ...
    )
```
2. It is more optimal to declare a variable as it is done here
3. It does not matter as the compiler will handle both cases the same way

@romac @soareschen Do you have an idea ?

ljoss17 (commented):
Great job @AlianBenabdallah !
It might be better to only discover the metrics which depend on the configuration `tx_confirmation = true` if the configuration is set. The metrics are `ibc_receive_packets`, `ibc_acknowledgment_packets`, `ibc_timeout_packets` and `tx_latency_confirmed `

AlianBenabdallah:
> Great job @AlianBenabdallah ! It might be better to only discover the metrics which depend on the configuration `tx_confirmation = true` if the configuration is set. The metrics are `ibc_receive_packets`, `ibc_acknowledgment_packets`, `ibc_timeout_packets` and `tx_latency_confirmed `

Regarding `tx_latency*`, do you know how to initialize these metrics without adding a new value ?

relayer/src/supervisor/scan.rs
AlianBenabdallah: I think using a variable improves readability. 

Regarding reference, I used a reference so `init_telemetry()` would not consume `counterparty_chain_id` and eventually reuse it (line [684](https://github.com/informalsystems/ibc-rs/blob/25bf5d7f0836c05da22d2bcfdb0a40f85f71cc0a/relayer/src/supervisor/scan.rs#L684) and [689](https://github.com/informalsystems/ibc-rs/blob/25bf5d7f0836c05da22d2bcfdb0a40f85f71cc0a/relayer/src/supervisor/scan.rs#L689)).

ljoss17:
> > Great job @AlianBenabdallah ! It might be better to only discover the metrics which depend on the configuration `tx_confirmation = true` if the configuration is set. The metrics are `ibc_receive_packets`, `ibc_acknowledgment_packets`, `ibc_timeout_packets` and `tx_latency_confirmed `
> 
> Regarding `tx_latency*`, do you know how to initialize these metrics without adding a new value ?

I haven't found a way to initialise the latencies without adding a new value thus distorting the histogram. But in my opinion it is not critical to discover the two metrics `tx_latency_submitted` and `tx_latency_confirmed`.

AlianBenabdallah:
> > > Great job @AlianBenabdallah ! It might be better to only discover the metrics which depend on the configuration `tx_confirmation = true` if the configuration is set. The metrics are `ibc_receive_packets`, `ibc_acknowledgment_packets`, `ibc_timeout_packets` and `tx_latency_confirmed `

It should be solved in the last commit. I also took into account `ibc_client_updates`, `ibc_client_misbehaviours` and `clear_*` metrics.

relayer/src/supervisor/scan.rs
```diff
@@ -815,3 +854,67 @@ fn query_connection_channels<Chain: ChainHandle>(
         })
         .map_err(Error::query)
 }
+
+/// Telemetry discovery is only done for metrics which will be recorded.
+/// For example if the client workers or client misbehaviour detection have
+/// been disabled in the configuration, the `client_misbehaviours_submitted`
+/// will never record any value as no misbehaviour will be submitted.
+fn init_telemetry(
```
seanchen1991, line 862: I think we can clean this function up a bit from a logic perspective by rearranging how the checks are done. Something like this:
```rust
fn init_telemetry(
    chain_id: &ChainId,
    client: &ClientId,
    counterparty_chain_id: &ChainId,
    channel_id: &ChannelId,
    port_id: &PortId,
    config: &Config,
) {
    // Boolean flag that is toggled if any of the tx workers are enabled
    let mut tx_worker_enabled = false;

    let clear_packets = config.mode.packets.enabled
        && (config.mode.packets.clear_on_start || config.mode.packets.clear_interval > 0);

    if config.mode.packets.enabled {
        tx_worker_enabled = true;

        telemetry!(
            init_per_path,
            chain_id,
            counterparty_chain_id,
            channel_id,
            port_id,
            clear_packets
        );

        telemetry!(init_worker_by_type, WorkerType::Packet);

        if config.mode.packets.tx_confirmation {
            telemetry!(init_per_channel, chain_id, channel_id, port_id);
        }
    }

    if config.mode.clients.enabled {
        tx_worker_enabled = true;
        telemetry!(init_worker_by_type, WorkerType::Client);
    }

    if config.mode.connections.enabled {
        tx_worker_enabled = true;
        telemetry!(init_worker_by_type, WorkerType::Connection);
    }

    if config.mode.channels.enabled {
        tx_worker_enabled = true;
        telemetry!(init_worker_by_type, WorkerType::Channel);
    }

    // Now we can just check this boolean
    if tx_worker_enabled {
        telemetry!(
            init_per_client,
            chain_id,
            counterparty_chain_id,
            client,
            config.mode.clients.misbehaviour && config.mode.clients.enabled
        );
    }

    telemetry!(init_worker_by_type, WorkerType::Wallet);
}
```
This removes the `if config.mode.clients.enabled || config.mode.channels.enabled || config.mode.connections.enabled || config.mode.packets.enabled` check so there's a bit less redundancy. 

Of course, if it turns out the metrics need to be initialized in a particular such that this refactor doesn't work, then disregard 🙂

ljoss17: This is great!! From what I checked it works, I will test it and if I don't encounter any issue I will push the changes

seanchen1991 (approved):
Looks good to me 👍

Issue #2187, opened 2022-05-06
Add initial support for Ocular to automate part of Hermes configuration


adizere:
## Summary

Add a `hermes config add <CHAIN_ID>` CLI to automatically populate a new section in the Hermes configuration file with data for chain `<CHAIN_ID>` based on data from [the chain-registry](https://github.com/cosmos/chain-registry).

## Problem Definition

Developers and beginner relayer users do not have access to their own full nodes for relaying or reading on-chain state from production. Even with full nodes available, it can difficult to configure correctly the `config.toml` of Hermes.

## Proposal

We can use the chain-registry data to populate correctly a config.toml section for Hermes, for any chain present in the registry. [Ocular](https://github.com/PeggyJV/ocular) is a Rust library to access the information stored in the chain-registry. In [here](https://github.com/PeggyJV/ocular/issues/8) we describe some ways in which Hermes can make use of Ocular.

In particular for the present issue, we'll likely depend on the ocular API method `fn chain_info(chain_id: String) -> ChainInfo` to be able to fetch chain information from the chain registry. The description of `ChainInfo` is not yet available, but the section "Ocular requirement B" from the [issue linked above](https://github.com/PeggyJV/ocular/issues/8) provides details about it. From the resulting `ChainInfo` we can generate a Hermes chain configuration section in `config.toml`.

Update: On the ocular side, initial support is now available for our use-case: https://github.com/PeggyJV/ocular/pull/63

## Acceptance Criteria

- [ ] add `hermes config add <CHAIN_ID>` with the details provided above
- [ ] we're not interested path information (i.e., channel filter), so we can skip populating the `[chains.packet_filter]` section
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2533, opened 2022-08-05
Introduce crate `ibc-chain-registry` and `hermes config auto` to automatically generate a config file

AlianBenabdallah:
Closes: #2187

## Status
Ready for review 

<details>
<summary>config.toml</summary>

```toml
[global]
log_level = 'info'

[mode.clients]
enabled = true
refresh = true
misbehaviour = true

[mode.connections]
enabled = false

[mode.channels]
enabled = false

[mode.packets]
enabled = true
clear_interval = 100
clear_on_start = true
tx_confirmation = false

[rest]
enabled = false
host = '127.0.0.1'
port = 3000

[telemetry]
enabled = false
host = '127.0.0.1'
port = 3001

[[chains]]
id = 'cosmoshub-4'
type = 'CosmosSdk'
rpc_addr = 'https://rpc-cosmoshub.ecostake.com/'
websocket_addr = 'ws://rpc-cosmoshub.ecostake.com/websocket'
grpc_addr = 'https://grpc-cosmoshub-ia.notional.ventures/'
rpc_timeout = '10s'
account_prefix = 'cosmos'
key_name = 'testkey1'
key_store_type = 'Test'
store_prefix = 'ibc'
default_gas = 100000
max_gas = 400000
gas_multiplier = 1.1
max_msg_num = 30
max_tx_size = 2097152
clock_drift = '5s'
max_block_time = '30s'
memo_prefix = ''


[chains.trust_threshold]
numerator = '1'
denominator = '3'

[chains.gas_price]
price = 0.1
denom = 'uatom'

[chains.packet_filter]
policy = 'allowall'

[chains.address_type]
derivation = 'cosmos'

[[chains]]
id = 'osmosis-1'
type = 'CosmosSdk'
rpc_addr = 'https://rpc-osmosis.ecostake.com/'
websocket_addr = 'ws://rpc-osmosis.ecostake.com/websocket'
grpc_addr = 'https://grpc-osmosis-ia.notional.ventures/'
rpc_timeout = '10s'
account_prefix = 'osmo'
key_name = 'testkey2'
key_store_type = 'Test'
store_prefix = 'ibc'
default_gas = 100000
max_gas = 400000
gas_multiplier = 1.1
max_msg_num = 30
max_tx_size = 2097152
clock_drift = '5s'
max_block_time = '30s'
memo_prefix = ''


[chains.trust_threshold]
numerator = '1'
denominator = '3'

[chains.gas_price]
price = 0.1
denom = 'uosmo'

[chains.packet_filter]
policy = 'allowall'

[chains.address_type]
derivation = 'cosmos'
```

</details>

## Description

- Introduces a new crate able to fetch data from the [chain-registry](https://github.com/cosmos/chain-registry). This crate is an alternative to [Ocular](https://github.com/PeggyJV/ocular) which is not stable. Ocular can introduce dependency issues and does not verify that the websocket interface is enabled.
- Introduces command `hermes config auto [OPTIONS] --output <PATH> -chains <CHAIN_NAME_1[:<KEY1>] CHAIN_NAME_2[:<KEY2>]> [--commit <COMMIT_HASH>]` which fetches data from the chain-registry and automatically generates a config file.
- Gas settings are set to default value.

## Design choices 
- If a ChainConfig can not be retrieved. The command fails and outputs and error. 
- Data is fetched from the latest commit of the chain-registry by default. 
______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


chain-registry/src/relayer_config.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+}
+
+/// Selects a healthy RPC from a list of RPCs
+/// An RPC is healthy if it can provide the RpcMandatoryData
+async fn select_healthy_rpc(rpcs: &[String]) -> Result<RpcMandatoryData, RegistryError> {
+    let mut futures = FuturesUnordered::new();
+    rpcs.iter().for_each(|rpc| {
+        futures.push(query_rpc(rpc));
+    });
```
seanchen1991, line 62: The use of iterators here could be a bit cleaner:
```suggestion
    let mut futures = rpcs
        .iter()
        .map(|rpc| query_rpc(rpc))
        .collect::<FuturesUnordered<_>>();
```

chain-registry/src/relayer_config.rs
```diff
@@ -0,0 +1,239 @@
...
+        .map_err(|_| RegistryError::unable_to_connect_with_grpc())?;
+    Ok(tendermint_url)
+}
+
+/// Select a healthy grpc address from a list of grpc addresses
+async fn select_healthy_grpc(grpcs: &[String]) -> Result<tendermint_rpc::Url, RegistryError> {
+    let mut futures = FuturesUnordered::new();
+    grpcs.iter().for_each(|grpc| {
+        futures.push(health_check_grpc(grpc));
+    });
```
seanchen1991, line 113:
```suggestion
    let mut futures: FuturesUnordered<_> = 
        grpcs.iter().map(|grpc| health_check_grpc(grpc)).collect();
```

chain-registry/src/relayer_config.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    while let Some(result) = futures.next().await {
+        if result.is_ok() {
+            return result;
+        }
+    }
+    Err(RegistryError::no_healthy_grpc())
+}
+
+/// Parses or build a valid uri from a grpc address
```
seanchen1991, line 123:
```suggestion
/// Parses or builds a valid uri from a grpc address
```

chain-registry/src/relayer_config.rs
```diff
@@ -0,0 +1,239 @@
...
+    grpcs.iter().for_each(|grpc| {
+        futures.push(health_check_grpc(grpc));
+    });
+
+    while let Some(result) = futures.next().await {
+        if result.is_ok() {
+            return result;
+        }
+    }
+    Err(RegistryError::no_healthy_grpc())
```
seanchen1991, line 120:
```suggestion
    }
    
    Err(RegistryError::no_healthy_grpc())
```

relayer-cli/src/commands/config/auto.rs
```diff
@@ -0,0 +1,95 @@
...
+    )]
+    path: PathBuf,
+
+    #[clap(
+        long = "chains",
+        required = true,
+        multiple = true,
+        value_name = "CHAIN_NAME1 CHAIN_NAME2...",
+        help_heading = "REQUIRED",
+        help = "Identifier of the chain to include in the config"
```
seanchen1991, line 33:
```suggestion
        help = "Identifier of the chains to include in the config"
```

relayer-cli/src/commands/config/auto.rs
```diff
@@ -0,0 +1,95 @@
...
+    chain_names: &[String],
+    key_names: &[String],
+) -> Vec<Result<ChainConfig, RegistryError>> {
+    let futures = FuturesUnordered::new();
+    chain_names
+        .iter()
+        .zip(key_names.iter())
+        .for_each(|(chain_name, key_name)| {
+            futures.push(hermes_config(chain_name.as_str(), key_name.as_str()));
+        });
```
seanchen1991, line 58:
```suggestion
    let futures: FuturesUnordered<_> = chain_names
        .iter()
        .zip(key_names.iter())
        .map(|(chain_name, key_name)| hermes_config(chain_name.as_str(), key_name.as_str()))
        .collect();
```

relayer-cli/src/commands/config/auto.rs
```diff
@@ -0,0 +1,95 @@
...
+        required = true,
+        multiple = true,
+        value_name = "KEY_NAME_CHAIN1 KEY_NAME_CHAIN2...",
+        help_heading = "REQUIRED",
+        help = "key names to include in the config"
+    )]
+    key_names: Vec<String>,
+}
+
+async fn get_chain_configs(
```
seanchen1991, line 48: Might be a good idea to validate that `chain_ids.len() == key_names.len()` at the start of this function.

seanchen1991 (commented):
It doesn't seem like the CLI parameter names for the `hermes config auto` command adhere to [ADR 010](https://github.com/informalsystems/ibc-rs/blob/master/docs/architecture/adr-010-unified-cli-arguments-hermes.md). Are users supposed to be able to pass in as many chain IDs and key names as they would like? Or are users only able to supply two chain IDs and two key names? 

There's also a fair bit of duplicated code. For example, `AssetList::fetch` and `ChainData::fetch` are essentially the same functions. `select_healthy_rpc` and `select_healthy_grpc` as well. While you'd typically abstract these into traits, it looks like doing that with these functions isn't currently possible due to them being `async`, unless we want to pull in the [async-trait](https://github.com/dtolnay/async-trait) crate, but that doesn't really seem worth it to me.

AlianBenabdallah:
> It doesn't seem like the CLI parameter names for the `hermes config auto` command adhere to [ADR 010](https://github.com/informalsystems/ibc-rs/blob/master/docs/architecture/adr-010-unified-cli-arguments-hermes.md). 
I will rename every parameter. I also think `key-names` could be optionnal. 

> Are users supposed to be able to pass in as many chain IDs and key names as they would like? Or are users only able to supply two chain IDs and two key names?
They are able to pass in as many chain_id and key_names as they'd like. 

> There's also a fair bit of duplicated code. For example, `AssetList::fetch` and `ChainData::fetch` are essentially the same functions. `select_healthy_rpc` and `select_healthy_grpc` as well. While you'd typically abstract these into traits, it looks like doing that with these functions isn't currently possible due to them being `async`, unless we want to pull in the [async-trait](https://github.com/dtolnay/async-trait) crate, but that doesn't really seem worth it to me.
Indeed, my first approach was to implement a trait `Fetch` for `AssetList` and `ChainData` but it isn't currently possible. I did not know about `async-trait`. I think it would be much cleaner to use async traits here.

Regarding `select_healthy*` functions, I think it is possible to use generic data types however, I would like to confirm their correctness before trying to improve them. We might find that some things are missing. 

Also, I think `parse_or_build_grpc_endpoint` should be inside `select_healthy_grpc` instead of `health_check_grpc`.

relayer-cli/src/commands/config/auto.rs
AlianBenabdallah: Should I remove the check at line 68 to avoid checking twice the same condition ?

seanchen1991: Ah, missed that. Nevermind then 🙂

chain-registry/src/relayer_config.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    #[test]
+    fn test_fetch_chain_config() {
+        use tokio::runtime::Runtime;
+        let rt = Runtime::new().unwrap();
+
+        let mut handles = Vec::with_capacity(TEST_CHAINS.len());
+        for chain in TEST_CHAINS {
+            handles.push(fetch_chain(chain));
```
seanchen1991, line 249: This test confuses me. `hermes_config` returns a `Result<ChainConfig>`, but the `fetch_chain` function, according to its signature, doesn't return anything. But then here the output of `fetch_chain` is pushed to a vector. Upon checking the length of the vector when running it the test, it does contain elements inside, which is the confusing part. 

Changing the the signature of `fetch_chain` to specify that it returns a `ChainConfig` still works, and is less confusing, though it doesn't answer the question of why this works as-is in the first place.

seanchen1991 (commented):
There's something that's confusing me about the unit tests. My preference would be to get rid of the `async` helper functions in order to test their inner functionality directly in the tests. Tests can be made `async` by changing `#[test]` to `#[tokio::test]`. I tried that and moved the async functionality from the helper functions directly into the tests, but that didn't seem to work. I'm not sure why, and I'm also not sure how much work it would require to get it working. 

We don't necessarily have to do it that way. I suggested some changes that stay in line with the approach you've taken with `async` helper functions in the tests.

chain-registry/src/relayer_config.rs
AlianBenabdallah: I have rewritten this test without using a helper function. I forgot to use `tokio::spawn()`. The vector should contain a `JoinHandle`.

chain-registry/src/relayer_config.rs
```diff
@@ -0,0 +1,243 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::utils::TEST_CHAINS;
+
+    async fn fetch_chain(chain: &str) {
+        hermes_config(chain, "testkey").await.unwrap();
+    }
```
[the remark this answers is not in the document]
AlianBenabdallah: This function was removed

chain-registry/src/utils.rs
```diff
@@ -10,48 +10,49 @@ pub const REF: &str = "ffdfbff3a21c7d2dbaac6f1e4730f47063f4bd72";
...
-}
 
-#[async_trait]
-pub trait Fetch<T: FileName + DeserializeOwned> {
-    async fn fetch(chain_name: String) -> Result<T, RegistryError> {
-        match fetch_data::<T>(chain_name).await {
+    async fn fetch(chain_name: String) -> Result<Self, RegistryError> {
+        match Self::fetch_data(chain_name).await {
             Ok(body) => match serde_json::from_str(&body) {
                 Ok(config) => Ok(config),
```
seanchen1991, line 57: Nit: I think we could name the wrapped value `parsed` or something like that instead of `config`, as this trait could certainly be used to fetch data other than config data.
```suggestion
                Ok(parsed) => Ok(parsed),
```

chain-registry/src/utils.rs
```diff
@@ -10,48 +10,49 @@ pub const REF: &str = "ffdfbff3a21c7d2dbaac6f1e4730f47063f4bd72";
 
 pub const TEST_CHAINS: &[&str] = &["cosmoshub", "juno", "evmos", "osmosis", "regen"];
 
-pub trait FileName {
+#[async_trait]
```
seanchen1991, line 13: We should document this trait and its methods. A suggestion:
```suggestion
/// `Fetchable` represents the basic expectations for external data or resources that
/// can be fetched.
#[async_trait]
```

chain-registry/src/utils.rs
```diff
@@ -10,48 +10,49 @@ pub const REF: &str = "ffdfbff3a21c7d2dbaac6f1e4730f47063f4bd72";
...
+pub trait Fetchable
+where
+    Self: DeserializeOwned,
+{
     fn file_name() -> String;
```
seanchen1991, line 18:
```suggestion
    /// The file name of the fetchable resource.
    fn file_name() -> String;
```

chain-registry/src/utils.rs
```diff
@@ -10,48 +10,49 @@ pub const REF: &str = "ffdfbff3a21c7d2dbaac6f1e4730f47063f4bd72";
...
-            response.status().as_u16(),
-        ))
     }
-}
 
-#[async_trait]
-pub trait Fetch<T: FileName + DeserializeOwned> {
-    async fn fetch(chain_name: String) -> Result<T, RegistryError> {
-        match fetch_data::<T>(chain_name).await {
+    async fn fetch(chain_name: String) -> Result<Self, RegistryError> {
```
seanchen1991, line 54:
```suggestion
    /// Fetches the fetchable resource.
    // The default implementation fetches config data from a chain registry. This
    // should be overridden if you're looking to fetch any other type of resource.
    async fn fetch(chain_name: String) -> Result<Self, RegistryError> {
```

chain-registry/src/utils.rs
```diff
@@ -10,48 +10,49 @@ pub const REF: &str = "ffdfbff3a21c7d2dbaac6f1e4730f47063f4bd72";
...
-        .scheme(PROTOCOL)
-        .authority(HOST)
-        .path_and_query(
-            format!(
-                "{}/{}/{}/{}",
-                REGISTRY_PATH,
-                REF,
-                chain_name,
-                T::file_name()
+    async fn fetch_data(chain_name: String) -> Result<String, RegistryError> {
```
seanchen1991, line 20: I don't see much reason to split the fetching logic between two different trait methods; it might not make sense for some future use-case of this trait. It is a bit clunky, but I think the logic of `fetch_data` should be folded into `fetch`.

seanchen1991 (commented):
Using `#[tokio::test]` really cleans things up nicely! 🙂

chain-registry/src/relayer_config.rs
```diff
@@ -0,0 +1,249 @@
...
+    let uri = parse_or_build_grpc_endpoint(grpc_address)?;
+    let tendermint_url = uri.to_string().parse().unwrap();
+    QueryClient::connect(uri)
+        .await
+        .map_err(|_| RegistryError::unable_to_connect_with_grpc())?;
+    Ok(tendermint_url)
+}
+
+/// Select a healthy rpc/grpc address from a list of urls
+async fn select_healthy<'a, RES, FUNC, FUTURE>(
```
romac, line 104: In Rust, multi-characters type parameters are typically written as CamelCase rather than uppercase.

```suggestion
async fn select_healthy<'a, Res, Func, Fut>(
```

chain-registry/src/utils.rs
```diff
@@ -0,0 +1,62 @@
...
+
+/// `Fetchable` represents the basic expectations for external data or resources that
+/// can be fetched.
+#[async_trait]
+pub trait Fetchable
+where
+    Self: DeserializeOwned,
+{
+    /// The file name of the fetchable resource.
+    fn file_name() -> String;
```
romac, line 21: File names and paths are typically represented as a `PathBuf`:

```suggestion
    fn file_name() -> PathBuf;
```

chain-registry/src/utils.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    /// Fetches the fetchable resource.
+    // The default implementation fetches config data from a chain registry. This
+    // should be overridden if you're looking to fetch any other type of resource.
+    async fn fetch(chain_name: String) -> Result<Self, RegistryError> {
```
romac, line 26: No need for the `chain_name` to be owned, let's just use a `&str`:

```suggestion
    async fn fetch(chain_name: &str) -> Result<Self, RegistryError> {
```

relayer-cli/src/commands/config/auto.rs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+impl Runnable for AutoCmd {
+    fn run(&self) {
+        // Assert that for every chain, a key name is provided
+        if self.chain_ids.len() != self.keys.len() {
+            Output::error("Must provide a key name for every chain").exit();
+        }
+        let runtime = Builder::new_multi_thread()
+            .worker_threads(1)
```
romac, line 70: I think we can get away by using the single threaded Tokio runtime here:

```suggestion
        let runtime = Builder::new_current_thread()
```

chain-registry/src/relayer_config.rs
```diff
@@ -0,0 +1,249 @@
...
+        .authority(uri.authority().unwrap().clone())
+        .path_and_query("/websocket")
+        .build()
+    {
+        Ok(uri) => Ok(tendermint_rpc::Url::from_str(uri.to_string().as_str()).unwrap()),
+        Err(e) => Err(RegistryError::unable_to_build_websocket_endpoint(
+            rpc_endpoint.to_string(),
+            e,
+        )),
+    }
```
romac, line 88: This can be expressed in a slightly more readable way as:

```suggestion
    let uri = Uri::builder()
        .scheme("wss")
        .authority(uri.authority().unwrap().clone())
        .path_and_query("/websocket")
        .build();

    match uri {
        Ok(uri) => Ok(tendermint_rpc::Url::from_str(uri.to_string().as_str()).unwrap()),
        Err(e) => Err(RegistryError::unable_to_build_websocket_endpoint(
            rpc_endpoint.to_string(),
            e,
        )),
    }
```

romac (commented):
Great stuff! Left some minor suggestions inline.

relayer-cli/src/commands/config/auto.rs
```diff
@@ -0,0 +1,93 @@
...
+
+use ibc_relayer::config::{store, ChainConfig, Config};
+
+use tokio::runtime::Builder;
+
+#[derive(Clone, Command, Debug, Parser, PartialEq)]
+pub struct AutoCmd {
+    #[clap(
+        long = "path",
+        short = 'p',
```
romac, line 19: As per ADR 010, we do not use short flags anymore in CLI commands.

```suggestion
```

relayer-cli/src/commands/config/auto.rs
```diff
@@ -0,0 +1,93 @@
...
+        help = "Key names to include in the config"
+    )]
+    keys: Vec<String>,
+}
+
+async fn get_chain_configs(
+    chain_names: &[String],
+    keys: &[String],
+) -> Vec<Result<ChainConfig, RegistryError>> {
+    let futures: FuturesUnordered<_> = chain_names
```
romac, line 52: Let's add an `assert!` to enforce the invariant used in `zip` below:

```suggestion
) -> Vec<Result<ChainConfig, RegistryError>> {
    assert_eq!(chain_names.len(), keys.len());

    let futures: FuturesUnordered<_> = chain_names
```

chain-registry/src/relayer_config.rs
```diff
@@ -0,0 +1,249 @@
...
+/// Generates a ChainConfig for a given chain by fetching data from
+/// https://github.com/cosmos/chain-registry.
+/// Gas settings are set to default values.
+pub async fn hermes_config(chain_name: &str, key_name: &str) -> Result<ChainConfig, RegistryError> {
+    let name_string = chain_name.to_string();
+    let chain_data_handle = tokio::spawn(async move { ChainData::fetch(name_string).await });
+    let name_string = chain_name.to_string();
+    let assets_handle = tokio::spawn(async move { AssetList::fetch(name_string).await });
+
+    let chain_data = chain_data_handle.await.unwrap()?;
```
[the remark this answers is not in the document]
AlianBenabdallah: The only `unwrap` are in the tests. Thanks for catching that !

ljoss17:
Great job! Some unit-tests for the the new command `hermes config auto` would be great.

AlianBenabdallah:
I added `path.rs` which contains structs and functions to fetch [IBC data](https://github.com/cosmos/chain-registry/tree/master/_IBC) from the chain repo.

Now I need to modify the API in order to incorporate this change. I'll most likely add a function which takes `chain_names : &[&str]` as argument and returns a `Vec<ChainConfig>`.

adizere:
> keys is a required argument. It populates key_name in ChainConfig.

Discussing with Luca today and we're wondering why would the keys option be required? Hermes could automatically detect in the [default keystore location](https://github.com/informalsystems/ibc-rs/blob/master/relayer/src/keyring.rs#L30) whether each chain has some wallets already (see `keys list --chain <CHAIN_ID>`), and select the first one among them. It could also output warnings if there are no wallets for the given chain, and also log to warn the user that it picked a certain wallet. Then the user could manually alter `config.toml` to choose a different wallet if they wish.

AlianBenabdallah:
> > keys is a required argument. It populates key_name in ChainConfig.
> 
> Discussing with Luca today and we're wondering why would the keys option be required? Hermes could automatically detect in the [default keystore location](https://github.com/informalsystems/ibc-rs/blob/master/relayer/src/keyring.rs#L30) whether each chain has some wallets already (see `keys list --chain <CHAIN_ID>`), and select the first one among them. It could also output warnings if there are no wallets for the given chain, and also log to warn the user that it picked a certain wallet. Then the user could manually alter `config.toml` to choose a different wallet if they wish.

It makes sense, I'll put it in the TODO list

chain-registry/src/relayer_config.rs
```diff
@@ -0,0 +1,286 @@
...
+mod tests {
+    use super::*;
+    use crate::utils::TEST_CHAINS;
+
+    #[tokio::test]
+    async fn fetch_chain_config() -> Result<(), RegistryError> {
+        let mut handles = Vec::with_capacity(TEST_CHAINS.len());
+
+        for chain in TEST_CHAINS {
+            handles.push(tokio::spawn(hermes_config(chain, "testkey")));
```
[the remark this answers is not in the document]
AlianBenabdallah: I think this piece of code is outdated.

chain-registry/src/relayer_config.rs
```diff
@@ -0,0 +1,286 @@
...
+use tendermint_light_client_verifier::types::TrustThreshold;
+use tendermint_rpc::{Client, SubscriptionClient, WebSocketClient};
+use tokio;
+use tokio::time::timeout;
+
+// ----------------- RPC ------------------
+
+/// Data retrievable from RPC endpoints.
+#[derive(Clone, Debug)]
+pub struct RpcMandatoryData {
```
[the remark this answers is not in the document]
AlianBenabdallah: I wonder if I should rename this struct as `RpcData`. 

This struct should contain all the data that we must retrieve from RPC endpoints in order to build a `ChainConfig`.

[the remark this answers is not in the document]
seanchen1991: > This struct should contain all the data that we must retrieve from RPC endpoints in order to build a ChainConfig
This is a good description that should probably be included in the docs for the struct.

Changing the name to `RpcData` sounds good to me :+1:

AlianBenabdallah:
Maybe I should try to remove `ChainConfig` dependencies from the `chain-registry` crate in order to separate concerns.

chain-registry/src/formatter.rs
```diff
@@ -0,0 +1,233 @@
...
+    fn parse_or_build_address(input: &str) -> Result<Self::OutputFormat, RegistryError> {
+        // Remove the last character if it is a '/'
+        let input = match input.ends_with('/') {
+            false => input,
+            true => {
+                let mut chars = input.chars();
+                chars.next_back();
+                chars.as_str()
+            }
+        };
```
seanchen1991: It's more idiomatic to use an `if` statement on boolean checks:
```suggestion
        let input = if input.ends_with('/') {
            let mut chars = input.chars()
            chars.next_back();
            chars.as_str()
        } else {
            input
        };
```

chain-registry/src/utils.rs
```diff
@@ -0,0 +1,268 @@
...
+    "secretnetwork-terra.json",
+    "secretnetwork-terra2.json",
+    "sentinel-sifchain.json",
+    "sentinel-starname.json",
+];
+
+/// `Fetchable` represents the basic expectations for external data or resources that
+/// can be fetched.
+#[async_trait]
+pub trait Fetchable
```
seanchen1991, line 220: I think it would make more sense to move the `Fetchable` trait to its own file `chain-registry/src/fetchable.rs`.

AlianBenabdallah:
> > keys is a required argument. It populates key_name in ChainConfig.
> 
> Discussing with Luca today and we're wondering why would the keys option be required? Hermes could automatically detect in the [default keystore location](https://github.com/informalsystems/ibc-rs/blob/master/relayer/src/keyring.rs#L30) whether each chain has some wallets already (see `keys list --chain <CHAIN_ID>`), and select the first one among them. It could also output warnings if there are no wallets for the given chain, and also log to warn the user that it picked a certain wallet. Then the user could manually alter `config.toml` to choose a different wallet if they wish.

@adizere @seanchen1991 This feature is now implemented !

relayer-cli/src/commands/config/auto.rs
```diff
@@ -0,0 +1,200 @@
...
+    } else {
+        None
+    }
+}
+
+/// The data structure that represents the arguments when invoking the `config auto` CLI command.
+///
+/// The command has two required arguments and an optional one which is used to manually specify the keys to use in the config file:
+///
+/// `config auto [OPTIONS] --path <PATH> --chain <CHAIN_NAME_1 CHAIN_NAME_2...>`
```
romac, line 38: Do we really need the chain name here since there are embedded in the path JSON file?

relayer-cli/src/commands/config/auto.rs
```diff
@@ -0,0 +1,200 @@
...
+///
+/// OR
+///
+/// `config auto [OPTIONS] --path <PATH> --chain <CHAIN_NAME_1 CHAIN_NAME_2...> --keys <KEY_NAME_1 KEY_NAME_2...>`
```
romac, line 42: I see how they are useful here for providing the corresponding keys in the same order, but perhaps we could improve the ux a little bit in this case by asking for `CHAIN:KEY` pairs? eg.

```
hermes config auto [OPTIONS] --path <PATH> --keys <CHAIN_1:KEY_1 CHAIN_2:KEY_2...>
```

What do you think?

relayer-cli/src/commands/config/auto.rs
AlianBenabdallah: `--path` is just the path where the config will be stored. 

I chose not to overwrite `~/.hermes/config.toml` by default.

relayer-cli/src/commands/config/auto.rs
AlianBenabdallah: Nice idea! However, maybe we should only keep `--chains`. The command would be :
```
hermes config auto [OPTIONS] --path <PATH> --chains <CHAIN_1[:KEY_1] CHAIN_2[:KEY_2]...>
```
If a key is provided then it will be used, otherwise the command would check if a key already exists otherwise it would leave a blank `key_name` argument.

AlianBenabdallah:
> I see how they are useful here for providing the corresponding keys in the same order, but perhaps we could improve the ux a little bit in this case by asking for `CHAIN:KEY` pairs? eg.
> 
> ```
> hermes config auto [OPTIONS] --path <PATH> --keys <CHAIN_1:KEY_1 CHAIN_2:KEY_2...>
The final commit incorporates this change. The command now fetches data from the last commit of the chain-registry. I think the PR is now ready to merge.

romac:
Awesome, great work Ali! Before we merge, can we also change `--path` to `--output` to avoid confusion wit/h between the path of the config file that's generated and a relay path between two blockchains

AlianBenabdallah:
> Awesome, great work Ali! Before we merge, can we also change `--path` to `--output` to avoid confusion between the path of the config file that's generated and a relay path between two blockchains?
@romac Done !

romac:
Yep, let's re-open all the PRs in v1.1 milestone

seanchen1991:
@AlianBenabdallah Looks like there are some merge conflicts that need to be resolved, but once that's done, we can merge this 🙂

AlianBenabdallah:
It is done and ready to merge !

seanchen1991 (approved):

Issue #2487, opened 2022-08-01
Hermes corner-cases related to gas


adizere:
Three errors found while looking throught the logs from an operator.
---

## Error 1: "Too much gas wanted"

The error looks like this:
>  send_tx_with_account_sequence_retry{id=osmosis-1}: broadcast_tx_sync: Response { code: Err(11), data: Data([]), log: Log("Too much gas wanted: 35000000, maximum is 25000000: out of gas"), hash: transaction::Hash(2E74921779770D2497B5720F6376BC1B75EB231AC66B60E834F84820218C3FBD) }: diagnostic: the gas requirement is higher than the configured maximum gas! please check the Hermes config.toml

### Consequence

In this case, the tx never makes it into the mempoool. Relayer's wallet is not being drained.
The relayer drop the transaction and the IBC packet is never relayed.

### Explanation

The reason why this error occurs is because Hermes is submitting the tx with 35M `gas_limit`, but the network `osmosis-1` disallows transactions
that have [over a certain amount of gas `MaxGasWantedPerTx` to be rejected during `CheckTx`](https://github.com/osmosis-labs/osmosis/blob/4800c2be72bf643ad0bb68386aaa5a27824f2da3/x/txfees/keeper/feedecorator.go#L47-L48).
This seems an osmosis-specific requirements. The [default limit is 25M gas](https://github.com/osmosis-labs/osmosis/blob/5879583fa0c82857a1cb18b4bab8f18173b43de1/x/txfees/types/options.go#L21).

It seems that Hermes uses 35M gas because it's using `default_gas` parameter, which is due to account sequence mismatch error. Concretely:

>   1842 Jul 26 00:03:38 Beastie hermes[1950306]: 2022-07-26T00:03:38.980489Z  WARN ThreadId(79) send_tx_with_account_sequence_retry{id=osmosis-1}:estimate_gas: failed to simulate tx, falling back on default gas because the error is potentially recoverable: gRPC call failed with status: status: Unknown, message: "account sequence mismatch, expected 234849, got 234850: incorrect account sequence", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "x-cosmos-block-height": "5311844"} }

This error [is ignorable](https://github.com/informalsystems/ibc-rs/blob/54ae389e8595e0360005ba99a9f5446490069cd0/relayer/src/error.rs#L587),
it can occur when Hermes is submitting multiple transactions in a row.
[...]

PR #2537, opened 2022-08-08
Handle some gas corner-cases

ljoss17:
Closes: #2487

## Description

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/relayer/src/error.rs
+++ b/relayer/src/error.rs
@@ -506,11 +506,17 @@ define_error! {
             },
 
         EmptyDenomTrace
-        { hash: String }
-        |e| {
-            format_args!(
-                "Query/DenomTrace RPC returned an empty denom trace for trace hash: {}", e.hash)
-        },
+            { hash: String }
+            |e| {
+                format_args!(
+                    "Query/DenomTrace RPC returned an empty denom trace for trace hash: {}", e.hash)
+            },
+
+        MessageExceedsMaxTxSize
+            { len: usize }
+            |e| {
+                format_args!("message length {} exceeds maximum transaction size", e.len)
+            }
     }
 }
 
--- a/relayer/src/error.rs
+++ b/relayer/src/error.rs
@@ -459,6 +459,15 @@ define_error! {
                     e.chain_id, e.default_gas, e.max_gas)
             },
 
+        ConfigValidationGasMultiplierLow
+            {
+                chain_id: ChainId,
+                gas_multiplier: f64,
+            }
+            |e| {
+                format!("semantic config validation failed for option `gas_mutliplier` on chain '{}', reason: gas multiplier ({}) is smaller than 1.1, this could trigger gas fee errors", e.chain_id, e.gas_multiplier)
+            },
+
         SdkModuleVersion
             {
                 chain_id: ChainId,
--- a/relayer/src/foreign_client.rs
+++ b/relayer/src/foreign_client.rs
@@ -29,7 +29,7 @@ use ibc::core::ics02_client::msgs::upgrade_client::MsgUpgradeAnyClient;
 use ibc::core::ics02_client::trust_threshold::TrustThreshold;
 use ibc::core::ics24_host::identifier::{ChainId, ClientId};
 use ibc::downcast;
-use ibc::events::{IbcEvent, WithBlockDataType};
+use ibc::events::{IbcEvent, IbcEventType, WithBlockDataType};
 use ibc::timestamp::{Timestamp, TimestampOverflowError};
 use ibc::tx_msg::Msg;
 use ibc::Height;
@@ -778,6 +778,25 @@ impl<DstChain: ChainHandle, SrcChain: ChainHandle> ForeignClient<DstChain, SrcCh
     }
 
     pub fn refresh(&mut self) -> Result<Option<Vec<IbcEvent>>, ForeignClientError> {
+        match self.try_refresh() {
+            Ok(res) => {
+                if let Some(ibc_events) = res.clone() {
+                    match ibc_events.clone()
+                        .into_iter()
+                        .find(|e| e.event_type() == IbcEventType::ChainError) {
+                            Some(ev) => {
+                                return Err(ForeignClientError::chain_error_event(self.dst_chain().id(), ev));
+                            },
+                            None => {}
+                    }
+                }
+                Ok(res)
+            }
+            Err(e) => Err(e),
+        }
+    }
+
+    fn try_refresh(&mut self) -> Result<Option<Vec<IbcEvent>>, ForeignClientError> {
         let (client_state, elapsed) = self.validated_client_state()?;
 
         // The refresh_window is the maximum duration
--- a/relayer/src/worker/client.rs
+++ b/relayer/src/worker/client.rs
@@ -32,7 +32,7 @@ pub fn spawn_refresh_client<ChainA: ChainHandle, ChainB: ChainHandle>(
                 src_chain = %client.src_chain.id(),
                 dst_chain = %client.dst_chain.id(),
             ),
-            Some(Duration::from_secs(1)),
+            Some(Duration::from_secs(5)),
             move || {
                 let res = client.refresh().map_err(|e| {
                     if e.is_expired_or_frozen_error() {
--- a/.changelog/unreleased/bug-fixes/ibc-relayer/2433-wait-consensus-height.md
+++ b/.changelog/unreleased/bug-fixes/ibc-relayer/2433-wait-consensus-height.md
@@ -1,0 +1,4 @@
+- For the `ConnOpenTry` and `ConnOpenAck` steps, wait for the destination
+  app height to be higher than the consensus height, otherwise we fail to
+  complete the handshake when the block times of the two chains involved differ
+  significantly ([#2433](https://github.com/informalsystems/ibc-rs/issues/2433))
--- a/.changelog/unreleased/bug-fixes/ibc-relayer/2477-prevent-oversized-tx-batch.md
+++ b/.changelog/unreleased/bug-fixes/ibc-relayer/2477-prevent-oversized-tx-batch.md
@@ -1,0 +1,3 @@
+- Fix code that could result in message batch size growing above
+  the transaction size limit
+  ([#2477](https://github.com/informalsystems/ibc-rs/issues/2477)).
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -40,3 +40,8 @@ updates:
     directory: "relayer-cli"
     schedule:
       interval: "weekly"
+
+  - package-ecosystem: "cargo"
+    directory: "ci/no-std-check"
+    schedule:
+      interval: "weekly"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -97,9 +97,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.58"
+version = "1.0.60"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bb07d2053ccdbe10e2af2995a2f116c1330396493dc1269f6a91d0ae82e19704"
+checksum = "c794e162a5eff65c72ef524dfe393eb923c354e350bb78b9c7383df13f3bc142"
 
 [[package]]
 name = "arc-swap"
@@ -136,9 +136,9 @@ dependencies = [
 
 [[package]]
 name = "async-trait"
-version = "0.1.56"
+version = "0.1.57"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "96cf8829f67d2eab0b2dfa42c5d0ef737e0724e4a82b01b3e292456202b19716"
+checksum = "76464446b8bc32758d7e88ee1a804d9914cd9b1cb264c029899680b0be29826f"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -369,9 +369,9 @@ checksum = "14c189c53d098945499cdfa7ecc63567cf3886b3332b312a5b4585d8d3a6a610"
 
 [[package]]
 name = "bytes"
-version = "1.2.0"
+version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f0b3de4a0c5e67e16066a0715723abd91edc2f9001d09c46e1dca929351e130e"
+checksum = "ec8a7b6a70fde80372154c65702f00a0f56f3e1c36abbc6c440484be248856db"
 
 [[package]]
 name = "camino"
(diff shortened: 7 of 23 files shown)


relayer/src/error.rs
```diff
@@ -459,6 +459,15 @@ define_error! {
                     e.chain_id, e.default_gas, e.max_gas)
             },
 
+        ConfigValidationGasMultiplierLow
+            {
+                chain_id: ChainId,
+                gas_multiplier: f64,
+            }
+            |e| {
+                format!("semantic config validation failed for option `gas_mutliplier` on chain '{}', reason: gas multiplier ({}) is smaller than 1.1, this could trigger gas fee errors", e.chain_id, e.gas_multiplier)
```
adizere, line 468:
```suggestion
                format!("semantic config validation failed for option `gas_multiplier` of chain '{}', reason: gas multiplier ({}) is smaller than `1.1`, which could trigger gas fee errors in production", e.chain_id, e.gas_multiplier)
```

relayer/src/foreign_client.rs
```diff
@@ -778,6 +778,29 @@ impl<DstChain: ChainHandle, SrcChain: ChainHandle> ForeignClient<DstChain, SrcCh
     }
 
     pub fn refresh(&mut self) -> Result<Option<Vec<IbcEvent>>, ForeignClientError> {
+        match self.try_refresh() {
+            Ok(res) => {
+                if let Some(ibc_events) = res.clone() {
+                    match ibc_events
+                        .into_iter()
+                        .find(|e| e.event_type() == IbcEventType::ChainError)
```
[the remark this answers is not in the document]
adizere: Or, alternatively, we could have strong type guarantees on this. CC @soareschen @romac

relayer/src/foreign_client.rs
```diff
@@ -778,6 +778,29 @@ impl<DstChain: ChainHandle, SrcChain: ChainHandle> ForeignClient<DstChain, SrcCh
...
+                    {
+                        Some(ev) => {
+                            return Err(ForeignClientError::chain_error_event(
+                                self.dst_chain().id(),
+                                ev,
+                            ));
+                        }
```
adizere, line 793: There are many different execution branches and it's not clear what are the post-conditions of the `refresh` method. Can we briefly describe that?

Let's maybe also put a comment to describe what is this match arm for. There is also an `Ok(res)` and a `Err(e` branches.

relayer/src/worker/client.rs
```diff
@@ -24,6 +27,18 @@ pub fn spawn_refresh_client<ChainA: ChainHandle, ChainB: ChainHandle>(
         );
         None
     } else {
+        // Compute the refresh interval as a fraction of the client's trusting period
+        // If the trusting period or the client state is not retrieved, fallback to a default value.
+        let refresh_interval = match client.validated_client_state() {
```
adizere, line 32:
```suggestion
        let interval_pause = match client.validated_client_state() {
```

relayer/src/worker/client.rs
```diff
@@ -14,6 +14,9 @@ use crate::{
 
 use super::WorkerCmd;
 
+const TRUSTING_PERIOD_DIVIDER: u32 = 100;
```
adizere, line 17: It seems like for Tendermint clients we would refresh every ~3 hours if the trusting period is 18 days. This seems a bit too infrequest. We should retry more frequently.

adizere:
> The interval between client refresh calls has been changed to be the client's trusting_period/100. In addition, the client refresh method has been updated in order to catch ChainError resulting events.

Notes from discussing with Luca:

It seems that there are 3 cases:
1. no update required at the moment -> the client worker can backoff with a large window eg several hours, a fraction of the `refresh_period`
2. successful update -> similar to case 1, but the refresh window could be even larger, still a fraction of `refresh_period`
3. failed to update -> the client worker should retry aggressively; if the errors keep coming up (such as it is signaled in #2487 error (3)), then the backoff between retries can be increased (eg Fibonacci or linear increase, but not [constant as it currently](https://github.com/informalsystems/ibc-rs/blob/7f5106b7426eda8c8e73ba331ac57def31d01023/relayer/src/worker/client.rs#L34) is).

relayer/src/worker/client.rs
```diff
@@ -14,6 +14,9 @@ use crate::{
...
+const DEFAULT_REFRESH_INTERVAL: u64 = 5;
```
adizere, line 18:
```suggestion
const DEFAULT_REFRESH_INTERVAL_SECONDS: u64 = 5;
```

soareschen:
@ljoss17 your changes are causing the integration tests to hang indefinitely. Can you merge  #2567 to your branch and also fix the tests?

relayer/src/foreign_client.rs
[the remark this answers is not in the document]
romac: Like @soareschen, I think we should get rid of these `ChainError` events altogether and figure out a better way to bubble up such errors.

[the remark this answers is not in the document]
romac: Tracked in https://github.com/informalsystems/ibc-rs/issues/2565

mzabaluev (changes requested):

relayer-cli/src/config.rs
```diff
@@ -170,16 +170,6 @@ fn validate_trust_threshold(
 }
 
 fn validate_gas_settings(id: &ChainId, config: &ChainConfig) -> Result<(), Diagnostic<Error>> {
-    // Check that the gas_multiplier is greater than or equal to 1.0
-    if let Some(gas_multiplier) = config.gas_multiplier {
-        if gas_multiplier < 1.0 {
-            return Err(Diagnostic::Error(Error::invalid_gas_multiplier(
```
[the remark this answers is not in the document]
ljoss17: Oh thank you, I forgot to remove it.

relayer/src/chain/cosmos/gas.rs
```diff
@@ -71,8 +71,6 @@ fn adjust_estimated_gas(
         gas_amount,
     }: AdjustGas,
 ) -> u64 {
-    assert!(gas_multiplier >= 1.0);
-
```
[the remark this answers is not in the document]
ljoss17: The `assert!` was replaced so that the test, in `client_refresh.rs`, using a `gas_multiplier = 0.8` to trigger a failed `refresh()` runs correctly. The checks are now in the new `GasMultiplier` struct:
https://github.com/informalsystems/ibc-rs/blob/5c522c4d72659ceec7b81ee31ecc3dddf9460700/relayer/src/config/gas_multiplier.rs#L18-L27
I am not sure if there is a way to easily reproduce the failure while keeping a `gas_multiplier >= 1.0`.

relayer/src/error.rs
```diff
@@ -459,6 +459,15 @@ define_error! {
                     e.chain_id, e.default_gas, e.max_gas)
             },
 
+        ConfigValidationGasMultiplierLow
```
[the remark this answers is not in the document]
ljoss17: I will open an issue for this

relayer/src/foreign_client.rs
```diff
@@ -783,6 +783,36 @@ impl<DstChain: ChainHandle, SrcChain: ChainHandle> ForeignClient<DstChain, SrcCh
     }
 
     pub fn refresh(&mut self) -> Result<Option<Vec<IbcEvent>>, ForeignClientError> {
+        match self.try_refresh() {
+            Ok(res) => {
+                // If elapsed < refresh_window for the client, `try_refresh()` will
+                // be successful with an empty vector.
+                if let Some(ibc_events) = res.clone() {
```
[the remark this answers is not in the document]
ljoss17: I see, so I removed the `clone()` in commit 47a49b95c32b5cbf481d67d897b8ddba9fc51e6a, but I am not sure I understand how `Option::map` could be used here. Would the idea be to apply the `.map()` to `res` in order to avoid the `Ok(None)` when `res` is `None`?

relayer/src/config/gas_multiplier.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+impl From<GasMultiplier> for f64 {
+    fn from(m: GasMultiplier) -> Self {
+        m.0
+    }
+}
+
+#[cfg(test)]
+#[allow(dead_code)] // the fields of the structs defined below are never accessed
```
mzabaluev, line 77: Does this still apply? There are no structs defined in this module.

relayer/src/foreign_client.rs
```diff
@@ -787,17 +787,17 @@ impl<DstChain: ChainHandle, SrcChain: ChainHandle> ForeignClient<DstChain, SrcCh
             Ok(res) => {
                 // If elapsed < refresh_window for the client, `try_refresh()` will
                 // be successful with an empty vector.
-                if let Some(ibc_events) = res.clone() {
+                if let Some(ibc_events) = &res {
```
mzabaluev, line 790: This is better, but I think you can just destructure `res` since this scope owns it and it's not used anywhere else.
```suggestion
                if let Some(ibc_events) = res {
```

relayer/src/foreign_client.rs
```diff
@@ -787,17 +787,17 @@ impl<DstChain: ChainHandle, SrcChain: ChainHandle> ForeignClient<DstChain, SrcCh
...
                     // The assumption is that only one IbcEventType::ChainError will be
                     // in the resulting Vec<IbcEvent> if an error occurred.
                     match ibc_events
-                        .into_iter()
-                        .find(|e| e.event_type() == IbcEventType::ChainError)
+                        .iter()
+                        .find(|&e| e.event_type() == IbcEventType::ChainError)
```
mzabaluev, line 795: This iterator could again take ownership:

```suggestion
                        .into_iter()
                        .find(|e| e.event_type() == IbcEventType::ChainError)
```

relayer/src/foreign_client.rs
```diff
@@ -787,17 +787,17 @@ impl<DstChain: ChainHandle, SrcChain: ChainHandle> ForeignClient<DstChain, SrcCh
...
                     {
                         Some(ev) => {
                             return Err(ForeignClientError::chain_error_event(
                                 self.dst_chain().id(),
-                                ev,
+                                ev.to_owned(),
```
mzabaluev, line 800: And this could be just moved without a copy.

```suggestion
                                ev,
```

But I may be missing some remaining borrow checker issue here, try it out and see if something pops out. Weirdly I get some error on not being able to use `==` with `IbcEventType` values.

relayer/src/foreign_client.rs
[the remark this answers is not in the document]
mzabaluev: Yes, see if `res.map(|ibc_events| { ... })` could be more elegant here.

relayer/src/config/gas_multiplier.rs
ljoss17: The struct is in the first test: https://github.com/informalsystems/ibc-rs/blob/47a49b95c32b5cbf481d67d897b8ddba9fc51e6a/relayer/src/config/gas_multiplier.rs#L87-L89

But I will update the comment to:
```suggestion
#[allow(dead_code)] // the field of the struct `DummyConfig` defined below is never accessed
```

relayer/src/foreign_client.rs
ljoss17: So the issue I am having is that `res` is used at this line.

https://github.com/informalsystems/ibc-rs/blob/47a49b95c32b5cbf481d67d897b8ddba9fc51e6a/relayer/src/foreign_client.rs#L804

In order to have a more elegant `refresh()` method, a part of it was extracted in a function.

romac: I think you can rewrite `refresh` as follows to avoid such nesting:

```rust
    pub fn refresh(&mut self) -> Result<Option<Vec<IbcEvent>>, ForeignClientError> {
        match self.try_refresh() {
            // If elapsed < refresh_window for the client, `try_refresh()` will
            // be successful with an empty vector.
            Ok(Some(ibc_events)) => Self::verify_chain_error(ibc_events, self.dst_chain().id()),

            // Return the successful empty vector.
            Ok(None) => Ok(None),

            // Return the error
            Err(e) => Err(e),
        }
    }
```


or by using the `?` operator on the result of `try_refresh`:

```rust
    pub fn refresh(&mut self) -> Result<Option<Vec<IbcEvent>>, ForeignClientError> {
        // If elapsed < refresh_window for the client, `try_refresh()` will
        // be successful with an empty vector.
        match self.try_refresh()? {
            Some(events) => Self::verify_chain_error(events, self.dst_chain().id()),
            None => Ok(None),
        }
    }
```

or even like this, but I find the latter harder to read:

```rust
    pub fn refresh(&mut self) -> Result<Option<Vec<IbcEvent>>, ForeignClientError> {
        // If elapsed < refresh_window for the client, `try_refresh()` will
        // be successful with an empty vector.
        self.try_refresh().and_then(|opt_events| {
            opt_events
                .map(|events| Self::verify_chain_error(events, self.dst_chain().id()))
                .transpose()
                .map(Option::flatten)
        })
    }
```

romac: Another alternative is to have `verify_chain_error` (renamed below to `check_no_errors`) return a `Result<(), _>`:

```rust
    pub fn refresh(&mut self) -> Result<Option<Vec<IbcEvent>>, ForeignClientError> {
        fn check_no_errors(
            ibc_events: &[IbcEvent],
            dst_chain_id: ChainId,
        ) -> Result<(), ForeignClientError> {
            // The assumption is that only one IbcEventType::ChainError will be
            // in the resulting Vec<IbcEvent> if an error occurred.
            let chain_error = ibc_events
                .iter()
                .find(|&e| e.event_type() == IbcEventType::ChainError);

            match chain_error {
                None => Ok(()),
                Some(ev) => Err(ForeignClientError::chain_error_event(
                    dst_chain_id,
                    ev.to_owned(),
                )),
            }
        }

        // If elapsed < refresh_window for the client, `try_refresh()` will
        // be successful with an empty vector.
        if let Some(events) = self.try_refresh()? {
            check_no_errors(&events, self.dst_chain().id())?;
            Ok(Some(events))
        } else {
            Ok(None)
        }
    }
```

romac: I think I like the latter solution the best, as it seems a bit weird to me to have `veriy_chain_errors` return an `Option<Vec<_>>` where the `Option` will always be `Some`, ie. the signature does not really tell us what is going on, and could also have us think that the events that are returned are not the very same than the one which are passed in

relayer/src/foreign_client.rs
romac: I don't think we want to take ownership here since we want to be able to keep using the events.

relayer/src/foreign_client.rs
ljoss17: The proposed solutions are great, and I agree that the last one is better, as the `check_no_errors` makes it more understandable.

mzabaluev (approved):
Looks good to me.

Issue #2539, opened 2022-08-08
Remove `height()` from `IbcEvent`


plafer:
`IbcEvent::height()` doesn't belong in the `IbcEvent`. This field was only tacked on `IbcEvent` for relayer purposes. We should instead create a `IbcEventWithHeight` struct in the relayer that tracks extra height data.

```rust
struct IbcEventWithHeight {
    event: IbcEvent,
    height: Height,
}
```

Contributes towards #2538.

PR #2542, opened 2022-08-09
Remove height information from `IbcEvent`

plafer:
Closes: #2539

## Description

- Removes `height()` and `set_height()` from `IbcEvent`
- In the relayer, replaces uses of `IbcEvent` that needed height information with new `IbcEventWithHeight`
- We still use the `extract_attributes_from_tx()` in the `TryFrom<&AbciEvent> for IbcEventStruct` temporarily. All client events don't have the same attributes, so we need a different solution. Similarly for connection and channel events.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


modules/src/core/ics02_client/events.rs
```diff
@@ -297,3 +294,86 @@ impl From<UpgradeClient> for AbciEvent {
...
+        let key = tag.key.as_ref();
+        let value = tag.value.as_ref();
+        if key == HEADER_ATTRIBUTE_KEY {
+            return AnyHeader::decode_from_string(value);
+        }
+    }
+    Err(Error::missing_raw_header())
+}
+
+#[cfg(test)]
```
romac, line 336: Sweet!

modules/src/core/ics03_connection/events.rs
```diff
@@ -4,39 +4,26 @@ use serde_derive::{Deserialize, Serialize};
...
+
 /// The content of the `key` field for the attribute containing the connection identifier.
-pub const HEIGHT_ATTRIBUTE_KEY: &str = "height";
 pub const CONN_ID_ATTRIBUTE_KEY: &str = "connection_id";
 pub const CLIENT_ID_ATTRIBUTE_KEY: &str = "client_id";
 pub const COUNTERPARTY_CONN_ID_ATTRIBUTE_KEY: &str = "counterparty_connection_id";
 pub const COUNTERPARTY_CLIENT_ID_ATTRIBUTE_KEY: &str = "counterparty_client_id";
 
-#[derive(Debug, Deserialize, Serialize, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Debug, Default, Deserialize, Serialize, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
romac, line 19: Nice!

relayer/src/event.rs
```diff
@@ -1,3 +1,33 @@
...
+    pub event: IbcEvent,
+    pub height: Height,
+}
+
+impl IbcEventWithHeight {
+    pub fn new(event: IbcEvent, height: Height) -> Self {
+        Self { event, height }
+    }
+
+    pub fn event(&self) -> &IbcEvent {
```
romac, line 20: We don't need those if the fields are public imho

relayer/src/event.rs
```diff
@@ -1,3 +1,33 @@
...
+        &self.event
+    }
+
+    pub fn height(&self) -> &Height {
```
romac, line 24: Same here

relayer/src/link/relay_path.rs
```diff
@@ -941,12 +957,21 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
             .send_messages_and_wait_commit(tm)
             .map_err(LinkError::relayer)?;
 
-        info!("result: {}", PrettyEvents(&dst_tx_events));
+        info!(
+            "result: {}",
+            PrettyEvents(
+                dst_tx_events
+                    .iter()
+                    .map(|ev| ev.event.clone())
```
romac, line 965: Is there no way around cloning all the events here? Feel free to disregard

relayer/src/event.rs
romac: Alternatively, we can make the the fields private.

romac (commented):
Great work @plafer, left some minor comments inline.

relayer/src/chain/cosmos/types/events/client.rs
```diff
@@ -1,125 +0,0 @@
...
-            .ok(),
-        Ok(IbcEventType::UpgradeClient) => extract_attributes_from_tx(event)
-            .map(UpgradeClient)
-            .map(IbcEvent::UpgradeClient)
-            .ok(),
-        _ => None,
-    }
-}
-
-fn extract_attributes_from_tx(event: &AbciEvent) -> Result<Attributes, Error> {
```
hu55a1n1, line 36: IMHO, this code should continue to live in the relayer crate because module users wouldn't be interested in conversions from `AbciEvent` to `IbcEvent`.

hu55a1n1, line 36: IMHO, this code should continue to live in the relayer crate because module users wouldn't be interested in conversions from `AbciEvent` to `IbcEvent`.

modules/src/events.rs
```diff
@@ -328,6 +336,73 @@ impl TryFrom<IbcEvent> for AbciEvent {
     }
 }
 
+impl TryFrom<&AbciEvent> for IbcEvent {
```
hu55a1n1, line 339: Same here, I think it would be better to have this code in the relayer crate. If the orphan rule precludes us from moving it there, we could have an `impl TryFrom<AbciEvent> for IbcEventWithHeight` along with a `impl From<IbcEventWithHeight> for IbcEvent` in the relayer crate. WDYT?

modules/src/events.rs
```diff
@@ -336,59 +411,6 @@ impl IbcEvent {
         }
     }
 
-    pub fn height(&self) -> Height {
```
hu55a1n1, line 339: :ok_hand:

relayer/src/event/rpc.rs
```diff
@@ -142,48 +147,36 @@ pub fn get_all_events(
...
                             {
                                 tracing::trace!(event = "SendPacket", "tx hash: {}", hash);
                             }
                         }
-                        vals.push((height, chan_event));
+                    } else {
+                        tracing::trace!("extracted ibc event {}", ibc_event);
                     }
+
+                    events_with_height.push(IbcEventWithHeight::new(ibc_event, height));
```
hu55a1n1, line 166: We previously only returned events based on the query and now we're returning all `IbcEvent`s in the tx result. Could this have unintended effects elsewhere?

hu55a1n1 (commented):
Thanks, @plafer! Nicely done! :pray: Just added a question and a suggestion.

modules/src/events.rs
plafer: I indeed put it there because of the orphan rule. I don't like `impl TryFrom<AbciEvent> for IbcEventWithHeight` because there is never any `height` information in the `AbciEvent`, so you can never build an `IbcEventWithHeight` from it. While I understand that what you're suggesting is that we should always do `AbciEvent` -> `IbcEventWithHeight` -> `IbcEvent`, I find it hacky for the reason explained above.

I thought about putting it behind a `"relayer"` feature flag. WDYT?

relayer/src/event/rpc.rs
plafer: Good catch! [a60e524](https://github.com/informalsystems/ibc-rs/pull/2542/commits/a60e52483df7d3e11f7238464ed81a425bac5d2d)

modules/src/events.rs
hu55a1n1: I like the relayer feature flag idea. :+1: But I am concerned about the light client refactoring and `UpdateClientEvent`'s header field; keeping it in the relayer allows us to use `AnyHeader` there (in `extract_header_from_tx`). Would it be possible to use a `impl TryFrom<(AbciEvent, Height)> for IbcEventWithHeight` instead and still keep this code in the relayer crate?

relayer/src/event.rs
plafer: [ab9c8d2](https://github.com/informalsystems/ibc-rs/pull/2542/commits/ab9c8d2e3d4651b41342c22672a43ca2d60ae5f0)

relayer/src/event.rs
plafer: [083c2ad](https://github.com/informalsystems/ibc-rs/pull/2542/commits/083c2ad9ea6368183ea8a756d7b0489b2a34161a)

relayer/src/link/relay_path.rs
plafer: [714b03c](https://github.com/informalsystems/ibc-rs/pull/2542/commits/714b03c7ecd737684e64da95dfe9a76837db72e0)

relayer/src/event/rpc.rs
plafer: Try 2: [377b60b](https://github.com/informalsystems/ibc-rs/pull/2542/commits/377b60bbf963d670eb969bb72cdc6392339f3d0f)

relayer/src/event/rpc.rs
```diff
@@ -143,22 +148,19 @@ pub fn get_all_events(
 
             for abci_event in &tx_result.result.events {
                 if query == queries::ibc_client().to_string() {
```
hu55a1n1, line 150: I think we need something like -> 
```rust
if let Ok(ibc_event) = IbcEvent::try_from(abci_event) {
  fn is_type_client(ev: &IbcEvent) -> bool {
      matches!(
          ev,
          IbcEvent::CreateClient(_)
              | IbcEvent::UpdateClient(_)
              | IbcEvent::UpgradeClient(_)
              | IbcEvent::ClientMisbehaviour(_)
      )
  }

  if query == queries::ibc_client().to_string() && is_type_client(&ibc_event) {
      tracing::trace!("extracted ibc_client event {}", client_event);
      events_with_height.push(IbcEventWithHeight::new(ibc_event, height));
  } /* else if ... */
}
```

relayer/src/event/rpc.rs
hu55a1n1: https://github.com/informalsystems/ibc-rs/pull/2542/files#r943587248

modules/src/events.rs
hu55a1n1: I agree, `ibc_event_try_from_abci_event()` seems like the best solution. :+1:

plafer: [785cbad](https://github.com/informalsystems/ibc-rs/pull/2542/commits/785cbada83692347edd63dc1ee9dfa54a50505b6) & [1bdf27f](https://github.com/informalsystems/ibc-rs/pull/2542/commits/1bdf27f4710f8e65c0bb37deaff679e4f519715e)

relayer/src/event/rpc.rs
plafer: [12484d9](https://github.com/informalsystems/ibc-rs/pull/2542/commits/12484d9dd8eca10d554f8c52e5132e2285961f99)

hu55a1n1 (approved):
Awesome work! Pre-approved (pending Ci fix).

Issue #2350, opened 2022-06-29
Support for priority mempool chains


ancazamfir:
## Summary
Starting with tm v0.34.20 we may see cosmos-sdk chains using priority enabled mempool. We need to make sure we support these chains.

## Problem Definition
Here is my understanding about how prio mempool will be used from hermes pov.

Let’s say hermes submits 2 Tx-es, `tx1` and `tx2`, with incrementing nonces, using `broadcast_tx_sync`. The application runs `checkTx` for each of them and returns an `abci.ResponseCheckTx` that includes a `priority`. The prio mempool then stores this and when it’s time to create a block it picks Tx-es in priority order.
In sdk chains, during `checkTx`, the `priority` is set by a `TxFeeChecker` and the default is defined in the auth module: https://github.com/cosmos/cosmos-sdk/blob/main/x/auth/ante/validator_tx_fee.go.

In this case the priority is computed based on the Tx fee. Different apps/chains will have the ability to overwrite this handler. 
But staying with the default behavior for now, it is possible  that tx2 will be included in a block before tx1, so `deliverTx(tx2)` will be executed first and will fail with account sequence mismatch.  `tx1` will be successful.

If we try to use multiple wallets to go around the account sequence mismatch we will still see issues, e.g.:
- tx2 may be rejected if tx1 includes a client update needed for msgs in tx2, or
- the msgs in the tx-es are for ordered channels.

While we push for a solution in tendermint or SDK  (e.g. to not reorder Tx-es from same account or allow out of order execution during `deliverTx`), we should remove our assumption that Tx-es are executed in order,  figure out all the cases where we are exposed to this and how to handle them.  
For the examples above, include update client in every tx (see also #2191), don’t send multiple Tx-es for ordered channels in same block (i.e. wait for tx1 hash to show up in a block before sending tx2), etc. 

## Acceptance Criteria
hermes works with prio mempool chains

- [x] depends on #2455 
- [x] update compatibility.rs to mark compat w/ sdk 0.46
    - https://github.com/informalsystems/ibc-rs/blob/5df35478607452baa8db004efb1d7aa38dfd8d89/relayer/src/chain/cosmos/compatibility.rs#L8-L23

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
An update on this:

[...]

PR #2543, opened 2022-08-09
Add hidden configuration for sending batched messages sequentially

soareschen:
Partial fix for #2350

## Description

This PR adds a hidden chain configuration `sequential_batch_tx`, which will cause the relayer to submit batched messages in serial after the previous transaction is committed to the chain.

The config has default value `false`, which disables sequential sending by default. However we can instruct relayer operators to set this to `true`, in the rare case that the batched transactions are failing due to priority mempool being turned on by validators of a chain.

Note that enabling sequential sending may cause degradation in performance. Hence we do not expose this config in the public documentation, to prevent relayer operators from shooting themselves in the foot.

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


relayer/src/chain/cosmos/batch.rs
```diff
@@ -105,33 +148,88 @@ async fn send_messages_as_batches(
...
-        }
+        let tx_sync_result = response_to_tx_sync_result(&config.chain_id, message_count, response);
+
+        tx_sync_results.push(tx_sync_result);
     }
 
     Ok(tx_sync_results)
 }
 
+async fn sequential_send_messages_as_batches(
```
mzabaluev, line 159: I'm concerned about more code duplication here.
When contemplating refactoring for batching fixes respecting actual tx size, already there are two functions to care about, this adds a third one.

Could this be incorporated in the regular `send_messages_as_batches` under a dynamic flag?

soareschen: The core logic has already been abstracted into helper functions like `batch_messages`, `send_tx_with_account_sequence_retry`, and `wait_for_block_commits`. So functions like `sequential_send_messages_as_batches` only act as higher level functions that wire up the lower level functions. Such low level functions are already there to serve the code deduplication purpose.

Using a dynamic flag can obscure the logic that differentiates `send_messages_as_batches` from `sequential_send_messages_as_batches`. With them being separate functions, it is much more clear of how the behavior are different from one another.

romac (approved):

Issue #1538, opened 2021-11-03
Log Improvements [meta issue]


adizere:
- [x] #1800 
- [x] Put more structured data in logs, instrument more functions
- [x] #2846
- [x] #2142
- [x] #1013
- [x] #1529 
- [x] #1488
- [x] #1537
    - [ ] Document the relayer pipeline model (e.g., in an ADR or in the Hermes guide) as captured in this issue; the goal is that the relayer operators can use this documentation as a reference for understanding the log

- [x] #2031  

- [x] Refactoring suggestion by @mzabaluev:
    - https://github.com/informalsystems/ibc-rs/pull/2089#discussion_r854062751

____

#### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate milestone (priority) applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

adizere:
All done, closing. Go team!

PR #2544, opened 2022-08-10
Clean up and improve logs

romac:
See: #1538

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


romac:
> I do not see a case where this log is very useful, so I tend to be of the opinion to remove it from the CLIs.

Agreed!

relayer/src/chain/cosmos.rs
```diff
@@ -124,6 +124,23 @@ impl CosmosSdkChain {
...
+        self.config.max_tx_size.into()
+    }
+
+    fn key(&self) -> Result<KeyEntry, Error> {
+        self.keybase()
+            .get_key(&self.config.key_name)
+            .map_err(Error::key_base)
+    }
+
+    fn trusting_period(&self, unbonding_period: Duration) -> Duration {
```
seanchen1991, line 138:
```suggestion
    /// Fetches the trusting period as a `Duration` from the chain config.
    /// If no trusting period exists in the config, the trusting period is calculated
    /// as two-thirds of the `unbonding_period`. 
    fn trusting_period(&self, unbonding_period: Duration) -> Duration {
```

relayer/src/supervisor.rs
```diff
@@ -700,8 +681,54 @@ fn process_batch<Chain: ChainHandle>(
     Ok(())
 }
 
+fn send_telemetry<Src, Dst>(src: &Src, dst: &Dst, events: &[IbcEvent], path: &Packet)
```
seanchen1991, line 684: Can we document this function? Some questions that I think would be helpful for the docs to answer:
1. Do the metrics get sent to both the source and destination? Only one of them? In which situations? 
2. A brief description of each of the metrics that are sent based on the `IbcEvent` variant in the `match`.

seanchen1991 (commented):
Let's add some more docs, but other than that, looks good to me 🙂

relayer/src/supervisor.rs
ljoss17: The idea for these metrics is the following:

* These metrics count the number of `SendPacket`, `WriteAcknowledgement` and `TimeoutPacket` events observed by the supervisor.
* In order to make it easier to see the duality `SendPacket`/`WriteAcknowledgement` or `SendPacket`/`TimeoutPacket`, the `WriteAcknowledgement` and `TimeoutPacket` will have the destination chain labelled as the source as these events are being "sent back" to the chain which submitted the corresponding `SendPacket` event.

So a `SendPacket` event from `ibc-0` will increment the `send_packet_events{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer"}` and the returning `WriteAcknowledgment` event will increment the `acknowledgement_events{chain="ibc-0",channel="channel-0",counterparty="ibc-1",port="transfer”}`.

But I realise that this might cause confusion. I can update it to label the `chain` and `counterparty` to be the submitting and receiving chains for the isolated events. Meaning the new send/acknowledgement/timeout events metrics label would have:
* `chain` is always the chain that submits the event
* `counterparty` is always the chain that will receive the packet generated from the event

Instead of being the submitting and receiving chain for the whole transfer. Current send/acknowledgement/timeout events metrics label is:
* `chain` being the chain that sends the `SendPacket` event and receives the corresponding `WriteAcknowledgment`, `TimeoutPacket` events
* `counterparty` being the chain that receives the `SendPacket` event and sends the `WriteAcknowledgment`, `TimeoutPacket` events

What do you think?

relayer/src/channel.rs
```diff
@@ -659,7 +659,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
             }
 
             (State::Open, State::Open) => {
-                info!("channel handshake already finished for {:#?}\n", self);
```
romac, line 659: We should also remove the trailing newline here and in the other cases below

romac: Let's also use the Display implementation for logs levels error, warn and info.

relayer/src/supervisor.rs
seanchen1991: > chain is always the chain that submits the event
counterparty is always the chain that will receive the packet generated from the event

I agree that labeling these in absolute terms, instead of relative terms, is probably a bit easier to parse 🙂

PR #2559, opened 2022-08-15
Remove `AnyClientState`

hu55a1n1:
Closes: #XXX

## Description

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/modules/src/core/ics02_client/client_state.rs
+++ b/modules/src/core/ics02_client/client_state.rs
@@ -1,11 +1,16 @@
 use core::marker::{Send, Sync};
 use core::time::Duration;
 
+use dyn_clone::DynClone;
+use erased_serde::Serialize as ErasedSerialize;
 use ibc_proto::google::protobuf::Any;
-use ibc_proto::protobuf::Protobuf;
-use serde::{Deserialize, Serialize};
-
 use ibc_proto::ibc::core::client::v1::IdentifiedClientState;
+use ibc_proto::ibc::lightclients::tendermint::v1::ClientState as RawClientState;
+#[cfg(any(test, feature = "mocks"))]
+use ibc_proto::ibc::mock::ClientState as RawMockClientState;
+use ibc_proto::protobuf::Protobuf;
+use ibc_proto::protobuf::Protobuf as ErasedProtobuf;
+use serde::{Deserialize, Serialize};
 
 use crate::clients::ics07_tendermint::client_state;
 use crate::clients::ics07_tendermint::client_state::ClientState as TmClientState;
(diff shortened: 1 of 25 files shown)


modules/src/core/ics02_client/client_state.rs
```diff
@@ -1,39 +1,36 @@
...
+{
     /// Return the chain identifier which this client is serving (i.e., the client is verifying
     /// consensus states from this chain).
     fn chain_id(&self) -> ChainId;
 
     /// Type of client associated with this state (eg. Tendermint)
     fn client_type(&self) -> ClientType;
 
-    /// Latest height of consensus state
+    /// Latest height of client state
```
plafer, line 33: I think the previous comment was actually correct. The client state doesn't get updated on every header; the consensus state does. I think this comment comes from the ibc-go implementation (e.g. [here](https://github.com/cosmos/ibc-go/blob/0a887e47318c21dc4668c1056ceae7a7e795a28c/modules/core/02-client/keeper/events.go#L24) the consensus height is being set with `ClientState.latest_height()`.

We should probably find a better name when we refactor the context traits.

modules/src/core/ics02_client/msgs/create_client.rs
```diff
@@ -16,17 +15,13 @@ pub const TYPE_URL: &str = "/ibc.core.client.v1.MsgCreateClient";
 /// A type of message that triggers the creation of a new on-chain (IBC) client.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct MsgCreateAnyClient {
-    pub client_state: AnyClientState,
+    pub client_state: Any,
```
plafer, line 18: Rust is giving me the following error here:
>the trait bound `ibc_proto::google::protobuf::Any: core::cmp::Eq` is not satisfied
the trait `core::cmp::Eq` is not implemented for `ibc_proto::google::protobuf::Any`

I'm not sure why?

plafer (approved):
LGTM!

modules/src/core/ics02_client/client_state.rs
hu55a1n1: I found `Latest height of consensus state` ambiguous because there would be multiple consensus states stored for a single light client (or client state). I think it should be something like `height of most recently installed header` or `Latest height the client was updated to` (this is what the [proto def has](https://github.com/informalsystems/ibc-rs/blob/master/proto/src/prost/ibc.lightclients.tendermint.v1.rs#L22)). WDYT?

plafer: Good point, "Latest height the client was updated to" sounds good!

hu55a1n1: https://github.com/informalsystems/ibc-rs/pull/2559/commits/8d6bfcd42970ed86daf11838ae8ff771f4785841

Issue #2560, opened 2022-08-15
Maximum tx size enforcement does not account for full transaction encoding


mzabaluev:
## Summary of Bug

When splitting messages into batches to form transactions, the `max_tx_size` limit only accounts for message content. It's still possible that the full `TxRaw` encoded message exceeds this limit.

This raises some questions:

* Does this configuration parameter mean the full transaction size including the signatures and AuthInfo data, as encoded in protobuf? Or is there any other way to calculate the tx size that could be more useful to relayer operators, e.g. for purposes of limiting fees?
* How likely does this discrepancy matter in practical use, to decide on the priority of fixing this?
* How important it is to pack the messages exhaustively up to the tx size limit? For example, if the relayer bumps one message off a transaction to the next batch due to using some size estimation heuristics, while in perfect implementation the message could still fit within the tx, how costly this miscalculation could be?

## Version

1.0.0-rc.2

## Steps to Reproduce

Not observed specifically for this cause, but see #2422 for possible symptoms.

## Acceptance Criteria

* There is a documented meaning for the `max_tx_size` chain config parameter, and it is correctly enforced.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2575, opened 2022-08-18
Account for full transaction encoding when batching messages

mzabaluev:
Simple, Cosmos-specific fix for #2560

Closes: #2560 

## Description

Use the estimated encoded transaction size to ensure that the transaction created from each message batch does not exceed `max_tx_size`.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


relayer/src/chain/cosmos/batch.rs
```diff
@@ -187,11 +219,19 @@ fn batch_messages(
...
-    use crate::config::types::{MaxMsgNum, MaxTxSize};
+    use crate::chain::cosmos::types::account::Account;
+    use crate::chain::cosmos::types::config::TxConfig;
+    use crate::config::types::{MaxMsgNum, MaxTxSize, Memo};
+    use crate::keyring::KeyEntry;
     use ibc_proto::google::protobuf::Any;
 
+    fn test_fixture() -> (TxConfig, KeyEntry, Account) {
+        todo!()
+    }
```
mzabaluev, line 230: @soareschen Is there anything in the test framework to provide usable values here? Unfortunately, the unit tests become more complicated.

relayer/src/chain/cosmos/batch.rs
```diff
@@ -139,27 +157,41 @@ async fn send_messages_as_batches(
 }
 
 fn batch_messages(
+    config: &TxConfig,
     max_msg_num: MaxMsgNum,
     max_tx_size: MaxTxSize,
+    key_entry: &KeyEntry,
+    account: &Account,
+    tx_memo: &Memo,
```
mzabaluev, line 165: This parameter lists is used repetitively, maybe there should be a struct to arrange some of them.

soareschen: It was originally not repetitive, but you made them repetitive now. The point of explicit parameters is to make it obvious of cascading dependencies and help us avoid that.

Instead, why don't just pass `tx_envelope_len` as a parameter and call `encoded_tx_len` from the call site? It also looks like `encoded_tx_len` can be estimated once and applicable for all transactions, rather than recomputing it every time.

relayer/src/chain/cosmos/batch.rs
soareschen: There could be a mock instance for `TxConfig`, if you don't call the API URLs. For `KeyEntry` you would have to construct a private key pair. For `Account` the main essential information is the account sequence, which could be set to something like `0` or `999999`.

mzabaluev: Good, I'm creating mock values (not pushed yet), but had doubts about all fields of `TxConfig`.

relayer/src/chain/cosmos/batch.rs
mzabaluev: @soareschen The same parameters are used in other functions in this module, I did not want to break the pattern at this point. As you point out, this flattening was done in an earlier refactoring to reduce unnecessary data dependencies and useless mutability, but I think we should follow it by some creative parameter grouping where it makes sense, to avoid excessive repetition.

> It also looks like `encoded_tx_len` can be estimated once and applicable for all transactions

I was not so sure, there may be some variation in how e.g. addresses are encoded, and there's also the memo field. There needs to be some context above the `send_batched_messages_and_wait_*` API to safely make this assumption.

soareschen: The key challenge for using a common struct in this case is that the references are borrowed from different places. So if we define a struct like `TxContext` we would have to move or clone the values into it. And if we define a struct containing references like `TxContext<'a>`, then it does not really contribute much other than being a syntactic sugar. It would also risk overgeneralizing functions, especially when catering for those few places that requires a `&mut Account`.

That said, if we want to refactor the code, it would better to define a proper `TxHandler` struct that specializes in sending transactions, and decouple it from the `ChainHandle` implementation, which is the real source of this mess.

The address should all have the same length, and is the same within the same `ChainHandle`. Same for the memo field, which has the same value configured. The only thing that could potentially change length is the account sequence, unless it has a fixed length encoding like `u64`.

mzabaluev: > t would better to define a proper `TxHandler` struct that specializes in sending transactions, and decouple it from the `ChainHandle` implementation, which is the real source of this mess.

I agree, splitting the "god trait" into smaller traits is a good way to refactor.

> Same for the memo field, which has the same value configured.

So there might be a case for a struct that contains these less-variable parameters.

> The only thing that could potentially change length is the account sequence, unless it has a fixed length encoding like u64.

Yes, there is a number of these varint fields and length delimiters that can grow between one tx to another, so you either have to compute the length on the fully composed transaction as it would be sent, or put in some upper bound for the overhead, neither of which is an ideal approach.

PR #2585, opened 2022-08-22
Fix guide commands

AlianBenabdallah:
## Status 

Work in progress 

## Description

- Some guide commands are fixed.

- In section `Identifiers`, `hermes tx conn-init --dst-chain ibc-1 --src-chain ibc-0 --dst-client 07-tendermint-0 --src-client 07-tendermint-0` can be confusing because we are using a real identifier (07-tendermint-0 on src) for a dummy client which does not even exist. I propose to change this identifier to `dummyclientname`.

- `gm key` does not exist and should be replaced by `gm keys`.
 
- I added a troubleshooting paragraph after the first commands using `gm` to make sure everything works.
______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


AlianBenabdallah:
@romac I think there is an issue with the instructions given at [6.1.3](https://hermes.informal.systems/tutorials/local-chains/start.html). Chains need to be started before doing `gm hermes config`, otherwise, the command will output an error because some files don't exist :
`stat $HOME/.gm/ibc-0/wallet_seed.json: no such file or directory`.

I will fix it in the guide [new design](https://github.com/informalsystems/ibc-rs/pull/2579). Should I leave it as it is for this PR ?

romac:
> @romac I think there is an issue with the instructions given at [6.1.3](https://hermes.informal.systems/tutorials/local-chains/start.html). Chains need to be started before doing `gm hermes config`, otherwise, the command will output an error because some files don't exist : `stat $HOME/.gm/ibc-0/wallet_seed.json: no such file or directory`.
> 
> I will fix it in the guide [new design](https://github.com/informalsystems/ibc-rs/pull/2579). Should I leave it as it is for this PR ?

Good catch! If it's just a matter of re-ordering the commands, can we please fix it in this PR?

AlianBenabdallah:
It's solved. I did not go through every command in section 7 but at least the tutorial should be fixed. I think this PR can be merged now.

romac (approved):

Issue #2598, opened 2022-08-29
Implement Telemetry Wrapper Components


soareschen:
## Summary

Add telemetry wrapper to various relayer components so that we can log telemetry without affecting the core logic.

To keep the telemetry component modular, we will make the logging dynamic typed. i.e. metrics are recorded as opaque objects. The methods are meant to work directly with types like `opentelemetry::Counter`.

To keep the task scope small, we will focus on only logging a few `Counter` type telemetry.

Some sub traits `HasTelemetry` and `Telemetry` has been defined at `relayer-framework/src/core/traits/contexts/telemetry.rs` and `relayer-framework/src/core/traits/runtime/telemetry.rs`. We can modify the methods as necessary.

## Step 1: Implement Telemetry traits

1. In `ibc_relayer_cosmos`, define a new type `struct CosmosTelemetry(Arc<TelemetryState>)`, and have it implement `Telemetry`.
2. Add an associated `type Telemetry: Telemetry`, and a method `fn telemetry(&self) -> &Self::Telemetry` for `OfaChain` at `relayer-framework/src/one_for_all/traits/chain.rs` and `OfaRelay`.
3. Add a new field `telemetry: CosmosTelemetry` in `CosmosChainContext`. Modify the `new()` constructors to fill the field with value from `ibc_telemetry::global()`. This is meant as a stub for initial implementation.
4. Implement the missing `OfaChain` types and methods for `CosmosChainContext`.
5. Implement `HasTelemetry` for `OfaChainContext` at `relayer-framework/src/one_for_all/impls/chain.rs`.
6. Add `HasTelemetry` trait bound to `AfoChainContext` at `relayer-framework/src/all_for_one/traits/chain.rs`
7. Verifies that the code compiles and commit.

## Step 2: Implement telemetry for chain status query

1. Create a new component `ChainStatusTelemetryQuerier<InQuerier>` at `relayer-framework/src/core/impls/queries`.
2. Implement `ChainStatusQuerier` for `ChainStatusTelemetryQuerier`. In the trait bounds, require the chain context to implement `HasTelemetry` and `InQuerier` to also implement `ChainStatusQuerier`.
3. Call the telemetry method `chain.telemetry().add_counter("query", 1, ...)` to add the query metric for chain status query.
4. Verify the code compiles and commit.

## Step 3: Add `ChainStatusTelemetryQuerier` to the present component graphs

[...]

PR #2604, opened 2022-08-31
Telemetry wrapper components

AlianBenabdallah:
Closes: #2598

## Description

Implements the Telemetry Wrapper component.

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


soareschen (changes requested):

relayer-cosmos/src/cosmos/core/types/telemetry.rs
```diff
@@ -0,0 +1,126 @@
...
+}
+
+impl HasLabel for CosmosTelemetry {
+    type Label = KeyValue;
+    fn new_label(key: &str, value: &str) -> Self::Label {
+        KeyValue::new(key.to_string(), value.to_string())
+    }
+}
+
+impl HasMetric<TelemetryCounter> for CosmosTelemetry {
```
soareschen, line 35: We can define a `OfaTelemetryCounter` that implements the preset metrics in `BasicTelemetryContext`. Something like:

```
pub trait OfaTelemetry {
  fn update_counter_metric(&self, ...);
  fn update_value_recorder_metric(&self, ...);
  fn update_up_down_counter_metric(&self, ...);
}
```

That way, `CosmosTelemetry` can implement only `OfaTelemetry` instead of implementing multiple `HasMetric` traits.

soareschen (changes requested):

relayer-cosmos/src/cosmos/all_for_one/chain.rs
```diff
@@ -32,6 +34,7 @@ pub trait AfoCosmosChainContext<Counterparty>:
     WriteAcknowledgementEvent = WriteAcknowledgement,
     ConsensusState = ConsensusState,
     ChainStatus = ChainStatus,
+    Telemetry = OfaTelemetryWrapper<CosmosTelemetry>,
```
soareschen, line 37: There is no need for a concrete type for `Telemetry`, if there is no compile error and the external user has no need to know about the concrete type. In general the AllForOne traits are used for finding the smallest closure dependencies required for specific uses.

soareschen (approved):

relayer-cosmos/src/cosmos/all_for_one/chain.rs
AlianBenabdallah: Thank you ! I removed it in the latest commit.

Issue #2442, opened 2022-07-20
Support Evmos in integration tests


soareschen:
## Summary

A number of changes are needed for running `ibc-integration-test` using `evmosd`.

## Problem Definition

There are a number of incompatibilities when running the current integration test suite with `evmosd`:

- `evmosd` requires a very specific format for the chain ID during chain initialization, in the form `evmos_9000-{digit}`. This is incompatible with the current chain ID used in integration test, such as `ibc-0-f5a2a988`.
- The `address_type` field in the relayer config has to be the value:
  ```
  config::AddressType::Ethermint {
     pk_type: "/ethermint.crypto.v1.ethsecp256k1.PubKey".to_string(),
  }
  ```
- The HD Path used for importing private key is `m/44'/60'/0'/0/0`.

It is also unclear how we can make the above parameters to be customizable when running the integration tests. There are several options available:

1. Introduce new environment variable for each parameter. e.g. `CHAIN_IDS`, `$HD_PATH`, `$ADDRESS_TYPE`.
2. Introduce a test config file to capture all test parameters without environment variables.
3. Special case during test when the chain binary is `evmosd`. This can be done by turning `FullNode` and `ChainDriver` into generic traits, with multiple implementations for different chains.

A further complication is that the parameters need to work with multiple chain binaries that will be introduced in #2003, such as running tests with an Evmos chain connect to a Gaia chain. With that, option 3 seems to be the simplest to implement.

Support for running integration tests with Evmos is required to triage issues such as #2433, and use CI to ensure that IBC relaying from Evmos remain smooth in the future.

## Proposal

- Modify `FullNode` and `ChainDriver` in `ibc-test-framework` to work generically and behave differently when the chain binary `evmosd` is used.

## Acceptance Criteria

Tests pass when running:

```
CHAIN_COMMAND_PATH=evmos cargo test -p ibc-integration-test
```
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

romac:
I agree that your proposal is probably the best way forward!

PR #2606, opened 2022-08-31
Added support for Evmos in integration tests

ljoss17:
Closes: #2442

## Description

This PR adds a new enum `ChainBinary` in order to easily adapt the configuration required to run Evmos in the integration tests.

The changes required to run Evmos are:
* The `chain_id` must have the following format `evmos_9000-{digits}` with `{digits}` being at least `1` and not containing any letters
* The address type needs to be: `/ethermint.crypto.v1.ethsecp256k1.PubKey`
* The command `evmosd start` needs an additional flag `--json-rpc.address` when starting multiple instances, since omitting the flag will default to the port `8545`. This will result in only the first `evmosd start` to run successfully.
* The HD Path needs to be: `m/44'/60'/0'/0/0`

This solution fits well as long as the differences between the configuration of tested chains is small, but if chains require more changes, turning `ChainDriver` and `FullNode` to generic traits, as suggested in the issue, might be better.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


tools/test-framework/src/chain/chain_binary.rs
```diff
@@ -0,0 +1,68 @@
...
+use crate::util::random::{random_u32, random_unused_tcp_port};
+
+const COSMOS_HD_PATH: &str = "m/44'/118'/0'/0/0";
+const EVMOS_HD_PATH: &str = "m/44'/60'/0'/0/0";
+
+#[derive(Clone, Debug)]
+pub enum ChainBinary {
+    CosmosChain,
+    EvmosChain,
+}
```
romac, line 13:
```suggestion
pub enum ChainType {
    Cosmos,
    Evmos,
}
```

tools/test-framework/src/chain/chain_binary.rs
```diff
@@ -0,0 +1,68 @@
...
+
+impl ChainBinary {
+    pub fn get_hd_path(&self) -> &str {
```
romac, line 16:
```suggestion
    pub fn hd_path(&self) -> &str {
```

tools/test-framework/src/chain/chain_binary.rs
```diff
@@ -0,0 +1,68 @@
...
+        match self {
+            Self::CosmosChain => COSMOS_HD_PATH,
+            Self::EvmosChain => EVMOS_HD_PATH,
+        }
+    }
+
+    pub fn get_chain_id(&self, prefix: &str, use_random_id: bool) -> ChainId {
```
romac, line 23:
```suggestion
    pub fn chain_id(&self, prefix: &str, use_random_id: bool) -> ChainId {
```

tools/test-framework/src/chain/chain_binary.rs
```diff
@@ -0,0 +1,68 @@
...
+                    ChainId::from_string(&format!("ibc-{}-{:x}", prefix, random_u32()))
+                } else {
+                    ChainId::from_string(&format!("ibc-{}", prefix))
+                }
+            }
+            Self::EvmosChain => ChainId::from_string(&format!("evmos_9000-{}", prefix)),
+        }
+    }
+
+    pub fn get_json_rpc_address(&self) -> (String, String) {
```
romac, line 36: Let's have this return an option instead, and perhaps call it `extra_start_args` or something similar?

```suggestion
    pub fn extra_start_args(&self) -> Option<String> {
```

tools/test-framework/src/chain/chain_binary.rs
```diff
@@ -0,0 +1,68 @@
...
+        match self {
+            Self::CosmosChain => ("".to_owned(), "".to_owned()),
+            Self::EvmosChain => (
+                "--json-rpc.address".to_owned(),
+                format!("localhost:{}", json_rpc_port),
+            ),
+        }
+    }
+
+    pub fn get_default_address_type(&self) -> AddressType {
```
romac, line 47:
```suggestion
    pub fn address_type(&self) -> AddressType {
```

tools/test-framework/src/chain/chain_binary.rs
```diff
@@ -0,0 +1,68 @@
...
+        match self {
+            Self::CosmosChain => AddressType::default(),
+            Self::EvmosChain => AddressType::Ethermint {
+                pk_type: "/ethermint.crypto.v1.ethsecp256k1.PubKey".to_string(),
+            },
+        }
+    }
+}
+
+impl From<String> for ChainBinary {
```
romac, line 57: This should probably be a `TryFrom` instance, or even better a `FromStr` instance.

tools/test-framework/src/chain/chain_binary.rs
ljoss17: Having a single `String` for the flag and argument trigger the following error `Error: unknown flag: --json-rpc.address localhost:50613`.
I will go with the solution by renaming to `extra_start_args` as this would still be clear if there are other flags which are required by other chains.

soareschen (changes requested):

tools/test-framework/src/framework/binary/node.rs
```diff
@@ -144,7 +144,7 @@ where
     fn run(&self, config: &TestConfig, builder: &ChainBuilder) -> Result<(), Error> {
         let node = bootstrap_single_node(
             builder,
-            "alpha",
+            "11",
```
soareschen, line 147: I think we can keep `alpha` here.

soareschen (approved):

Issue #2605, opened 2022-08-31
Integrate mdbook-template with the guide


seanchen1991:
## Summary

Make it easier to keep the guide in sync with the code base by taking advantage of mdbook [anchors](https://rust-lang.github.io/mdBook/format/mdbook.html#including-portions-of-a-file) and [mdbook-template](https://github.com/sgoudham/mdbook-template) in order to integrate code snippets and files into the guide. 

## Problem Definition and Proposal

Keeping the guide in sync with the code base is currently an extremely manual process. This can be somewhat alleviated by taking advantage of mdbook [anchors](https://rust-lang.github.io/mdBook/format/mdbook.html#including-portions-of-a-file) and [mdbook-template](https://github.com/sgoudham/mdbook-template) in order to integrate code snippets and files into the guide. These files can then be checked and validated via a CI job on every PR to ensure that the snippets and files are kept up-to-date. 

## Acceptance Criteria

- [ ] Use mdbook templates in the Commands Reference section of the guide
- [ ] Validate these files and code snippets on every PR with a CI job. 
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2625, opened 2022-09-07
Script to automatically generate template files for the guide

AlianBenabdallah:
Closes: #2605 

## Description

This PR integrates `mdbook-template` with the guide.

## How does mdbook-template work ?
A template file contains constant strings and arguments which can take default values. i.e.:
__my_template.md__
```
[[#arg1 "default"]] hello world [[#arg2]]
```
In another file, this template file can be inserted with the following macro : 
```
{{#my_template.md arg2="!"}}
```
The macro will be replaced by:
```
default hello world !
```
## Issues of mdbook-template
- `mdbook-template` will not log any error if you call the macro with an argument which does not exist. For example:
`{{#my_template.md arg3="hello"}}` will not generate any log or crash. It's a silent issue.
- It will also not log anything if an argument is missing.

## Content of this PR 
- In the guide : 
  - Replaces every command description by `{{#include templates/help_templates/file.md}}` which inserts the content of `help_templates/file.md` in the markdown file. This file is automatically updated.
  - Replaces every hermes command by `{{#template templates/commands/hermes/command_n.md flag1=value1}}`. As explained above, it inserts the content of `command_n.md` in the markdown file and can take arguments. The template file is automatically generated from the USAGE of the command.

- New script `auto_gen_templates.sh` with two different modes : 
  - "check" Which verifies that the templates in `templates/help_templates` and `templates/commands/hermes` are up to date.
  - "update" Which generates or updates template files. 

- New package `check-guide`. It contains a binary which checks that every command in the guide can be correctly parsed.

- New CI job which runs `auto_gen_templates.sh --mode "check"` and runs `check-guide`'s main binary. 

## Possible improvements
- `mdbook-templates` can be removed. Instead, we can use a script to extract every Hermes command from the guide and verify that they can be parsed. 

Tasks :
- [x] Function to automatically find every subcommand of Hermes
- [x] Function to automatically generate a help file for every command
- [x] Function to automatically generate a template file for every command used in the guide
- [x] CI job to use the script
- [x] Guide update

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


seanchen1991 (commented):
Thanks for taking this work on Ali! Just so I understand how this will work: is the idea that the `auto_gen_templates` script will be run on every PR (likely as part of a CI job) to re-generate these template files?

AlianBenabdallah:
> Thanks for taking this work on Ali! Just so I understand how this will work: is the idea that the `auto_gen_templates` script will be run on every PR (likely as part of a CI job) to re-generate these template files?

It is exactly what I am trying to do =D It's not very costly to generate these template files.

guide/src/tutorials/local-chains/start-relaying.md
```diff
@@ -87,7 +87,7 @@ Now, let's exchange `samoleans` between two chains.
 
 - Transfer back these tokens to ibc-0:
     ```shell
-    {{#template ../../templates/commands/hermes/transfer_with_denom dst-chain=ibc-1 src-chain=ibc-0 src-port=transfer src-channel=channel-0 amount=100000 timeout-seconds=10000 --denom ibc/C1840BD16FCFA8F421DAA0DAAB08B9C323FC7685D0D7951DC37B3F9ECB08A199}}
+    {{#template ../../templates/commands/hermes/tx/ft-transfer.md DST_CHAIN_ID=ibc-1 SRC_CHAIN_ID=ibc-0 SRC_PORT_ID=transfer SRC_CHANNEL_ID=channel-0 AMOUNT=100000 OPTIONS=--timeout-seconds=10000 --denom ibc/C1840BD16FCFA8F421DAA0DAAB08B9C323FC7685D0D7951DC37B3F9ECB08A199}}
```
AlianBenabdallah, line 90: @seanchen1991 There are many cases in the guide where some commands need optional flags and I have to rely on "OPTIONS=". 

If we rename these flags then the command will not work. I am not sure how to deal with this.

AlianBenabdallah:
The current script generates templates from the USAGE section of the `help` command however some command have multiple usages. For example : 
```
USAGE:
    hermes create channel [OPTIONS] --a-chain <A_CHAIN_ID> --a-connection <A_CONNECTION_ID> --a-port <A_PORT_ID> --b-port <B_PORT_ID>

    hermes create channel [OPTIONS] --a-chain <A_CHAIN_ID> --b-chain <B_CHAIN_ID> --a-port <A_PORT_ID> --b-port <B_PORT_ID> --new-client-connection
```
I will have to deal with this before updating the guide.

guide/src/tutorials/local-chains/start-relaying.md
AlianBenabdallah: Exactly. I can't think of any solution that is 100% automatic. A possible solution would be to detect if an optional flag has changed for a given command and to tell the user to update the guide accordingly.

seanchen1991: Yea, I was thinking that's probably the best way to go about this as well

AlianBenabdallah: The guide can also break if we modify the name of a command. For example, if instead of `hermes tx subcommand` we had `hermes tx raw subcommand`.

seanchen1991: Yea, I think the best thing to do in either of those cases would be for the CI job that generates the template files to fail, indicating to the PR author that the commands need to be updated in the guide.

AlianBenabdallah: I looked at the source code of `mdbook-template`.

I think it's possible to retrieve all the pieces of code that use a command template in the guide, inject the dynamic arguments ([example](https://github.com/sgoudham/mdbook-template/blob/5da795954563a401b29bf0ac9dc4e5fbf9d5678e/src/lib.rs#L173)) and verify that the command can be parsed with [`try_parse_from()`](https://docs.rs/clap/3.0.14/clap/trait.Parser.html#method.parse_from). 

It won't work with new commands but it will work with current ones.

AlianBenabdallah: Nevermind, the function `replace_args` is not public. I can still make my own.

seanchen1991: > I think it's possible to retrieve all the pieces of code that use a command template in the guide, inject the dynamic arguments ([example](https://github.com/sgoudham/mdbook-template/blob/5da795954563a401b29bf0ac9dc4e5fbf9d5678e/src/lib.rs#L173)) and verify that the command can be parsed with [try_parse_from()](https://docs.rs/clap/3.0.14/clap/trait.Parser.html#method.parse_from).

I'm not clear this will change the usage flow from what we previously mentioned.

AlianBenabdallah: > > I think it's possible to retrieve all the pieces of code that use a command template in the guide, inject the dynamic arguments ([example](https://github.com/sgoudham/mdbook-template/blob/5da795954563a401b29bf0ac9dc4e5fbf9d5678e/src/lib.rs#L173)) and verify that the command can be parsed with [try_parse_from()](https://docs.rs/clap/3.0.14/clap/trait.Parser.html#method.parse_from).
> 
> I'm not clear this will change the usage flow from what we previously mentioned.

It's true however I think it could be an improvement. With the current usage flow, the script will just detect a change 
without giving details about what changed and what needs to be updated. It will be the responsability of the user to update the guide and make sure every command is syntaxically correct.

Another solution is to verify that every command can be parsed. It's a bit more work but the script will throw errors until every command can be parsed.

AlianBenabdallah: Maybe we can leave it as a suggestion for now and eventually come back to it later.

AlianBenabdallah:
The PR is now ready for review. I also updated the `chain upgrade` command to close #2630.

.github/workflows/guide-templates.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - telemetry/**
+      - scripts/**
+
+env:
+  CARGO_INCREMENTAL: 0
+  CARGO_PROFILE_DEV_DEBUG: 1
+  CARGO_PROFILE_RELEASE_DEBUG: 1
+  RUST_BACKTRACE: short
+  CARGO_NET_RETRY: 10
+  RUSTUP_MAX_RETRIES: 10
```
adizere, line 30: I think we don't need any of these env variables.

guide/src/advanced/troubleshooting/help-command.md
```diff
@@ -14,17 +14,7 @@ hermes help create
 will provide details about all the valid invocations of the `create` CLI command.
 
 ```
-USAGE:
```
adizere, line 17: I'm running `mdbook serve` locally from this branch, and the output seems incomplete. Any ideas?

> http://localhost:3000/advanced/troubleshooting/help-command.html

Screenshot 2022-09-14 at 15 21 27

adizere (commented):
Still digesting this PR, left a couple of comments.

guide/src/advanced/troubleshooting/help-command.md
AlianBenabdallah: Hey, I think you need to install `mdbook-template` with `cargo install mdbook-template`.

AlianBenabdallah:
I added a new package `check-guide` to verify that every command can be correctly parsed. It works as follows : 
- It extracts every lines containing `{{#template .*templates/command/hermes}}`
- Generates the command by retrieving the template file's content and replacing arguments by their values.
- Verifies that an `EntryPoint` can be parsed from the String.
  - If it can not be parsed, it logs a message in `stderr`
- Returns the number of error encountered.

We do not need `mdbook-template` as long as we can execute a script with a similar logic which extracts every command in the guide and verify that they all can be parsed.

AlianBenabdallah:
The maintainer of `mdbook-template` released version 1.1.0 which makes the function `replace_template()` public. The script is now improved and much easier to read.

tools/check-guide/src/main.rs
```diff
@@ -0,0 +1,113 @@
+//! This crate verifies the correctness of every Hermes command in the guide by:
+//! 1. Extracting every line in the guide with '{{#template *templates/commands/hermes*}}', a macro call for mdbook template.
+//! 2. Replace every template call with the content of the template. It will replace macro call by what should be an Hermes command.
```
seanchen1991, line 3:
```suggestion
//! 2. Replace every template call with the content of the template. It will replace the macro call with what should be a Hermes command.
```

tools/check-guide/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+            eprintln!("{}:{}: {:?}", path.display(), line_number, e);
+            error_founds += 1;
+        }
+    });
+    error_founds
+}
+
+fn verify_file(path: &Path) -> i32 {
+    // Verifies that every template macro call in the file can be replaced by a valid Hermes command.
+    // Returns the number of invalid commands found.
```
seanchen1991, line 78: Let's turn these into doc comments for the function.
```suggestion
/// Verifies that every template macro call in the file can be replaced by a valid Hermes command.
/// Returns the number of invalid commands found.
fn verify_file(path: &Path) -> i32 {
```

tools/check-guide/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+    T: IntoIterator<Item = &'a str>,
+    // Returns an error if the command cannot be parsed.
+{
+    EntryPoint::try_parse_from(command)?;
+    Ok(())
+}
+
+fn verify_line(line: &str, path: &Path, line_number: i32) -> i32 {
+    // Extracts macro calls from a command, replace them with the content of the template and check that the command is correct.
+    // Returns the number of errors encountered.
```
seanchen1991, line 61: Let's turn these into doc comments for the function.
```suggestion
/// Extracts macro calls from a command, replace them with the content of the template and check that the command is correct.
/// Returns the number of errors encountered.
fn verify_line(line: &str, path: &Path, line_number: i32) -> i32 {
```

tools/check-guide/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+    let parent = path.parent().unwrap_or_else(|| Path::new(&*GUIDE_PATH));
+    let mut error_founds = 0;
```
seanchen1991, line 63:
```suggestion
    let mut errors_found = 0;
```

tools/check-guide/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    TEMPLATE_RE.find_iter(line).for_each(|m| {
+        let template = m.as_str();
+        let template_replaced = replace_template(template, &*FILEREADER, &parent, "", 0);
+        if let Err(e) = check_correctness(template_replaced.split_whitespace()) {
+            eprintln!("{}:{}: {:?}", path.display(), line_number, e);
+            error_founds += 1;
```
seanchen1991, line 70:
```suggestion
            errors_found += 1;
```

tools/check-guide/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+        }
+    });
+    error_founds
```
seanchen1991, line 73:
```suggestion
    errors_found
```

tools/check-guide/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+fn verify_file(path: &Path) -> i32 {
+    // Verifies that every template macro call in the file can be replaced by a valid Hermes command.
+    // Returns the number of invalid commands found.
+    let mut error_founds = 0;
```
seanchen1991, line 79:
```suggestion
    let mut errors_found = 0;
```

tools/check-guide/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+    let file = File::open(path);
+    let reader =
+        BufReader::new(file.unwrap_or_else(|_| panic!("File not found: {}", path.display())));
+    let mut line_number = 1;
+
+    for line in reader.lines() {
+        let line = line
+            .unwrap_or_else(|_| panic!("{} : Failed to read line {}", path.display(), line_number));
+        error_founds += verify_line(&line, path, line_number);
```
seanchen1991, line 88:
```suggestion
        errors_found += verify_line(&line, path, line_number);
```

tools/check-guide/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+        line_number += 1;
+    }
+    error_founds
```
seanchen1991, line 91:
```suggestion
    errors_found
```

tools/check-guide/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+        .filter_entry(|e| {
+            !EXCLUSIONS.contains(
+                e.file_name()
+                    .to_str()
+                    .expect("Unwrapping a file_name to str failed."),
+            )
+        }) // Filter out the excluded directories
+        .map(|e| e.expect("Failed to get an entry."))
+        .filter(|e| e.file_type().is_file()) // Keep only files
+        .filter(|e| e.path().extension() == Some(OsStr::new("md"))) // Keep only markdown files
```
seanchen1991, line 107: Can these two `filter` calls be consolidated into one?
```suggestion
        .filter(|e| e.file_type().is_file() && e.path().extension() == Some(OsStr::new("md"))) // Keep only markdown files
```

seanchen1991 (commented):
This is super cool work Ali 🙂
Left a couple of nits; otherwise it looks great to me!

AlianBenabdallah:
Thanks for reviewing this PR @seanchen1991 ! I think we should merge this PR now.

seanchen1991 (commented):
One last question: is it clear which files need to updated when the CI job fails, and how those files should be updated?

AlianBenabdallah:
> One last question: is it clear which files need to updated when the CI job fails, and how those files should be updated?

We can consider two CI jobs : 
- The first one will run `auto_gen_templates.sh --mode "check"`. If it fails because of a file, it will output "$filename is not up to date" ([source](https://github.com/informalsystems/ibc-rs/blob/913948e603b952efef573135335d67ad06ef8200/scripts/auto_gen_templates.sh#L150)). It does not print any instruction however running `auto_gen_templates.sh --mode "update"` will fix that.
- The second one will run `check-guide`'s binary. If it finds a template that is remplaced by an invalid Hermes command, it outputs `file-line-error`. The guide will need to be updated manually.

seanchen1991:
That sounds pretty good to me 🙂

seanchen1991 (approved):

PR #2627, opened 2022-09-08
Refactor coin and amount constructs into separate modules

soareschen:
## Description

Extracting the work done from #1981.

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


modules/src/applications/transfer/coin.rs
```diff
@@ -0,0 +1,157 @@
...
+impl<D: FromStr> Coin<D>
+where
+    D::Err: Into<Error>,
+{
+    pub fn from_string_list(coin_str: &str) -> Result<Vec<Self>, Error> {
+        coin_str.split(',').map(FromStr::from_str).collect()
+    }
+}
+
+impl<D: FromStr> FromStr for Coin<D>
```
hu55a1n1, line 40: I believe this is Cosmos SDK-specific code and is not being used in the modules. Can we move it (and any related tests) out of the modules crate?

soareschen: This is needed for parsing the `IncentivizedPacket` event from `AbciEvent` in #1981. We can remove this after splitting the IBC module away from the relayer code and remove the parsing for ABCI events.

hu55a1n1: Sounds good. :+1:

modules/src/applications/transfer/coin.rs
```diff
@@ -0,0 +1,157 @@
...
+        PrefixedCoin {
+            denom: coin.denom.into(),
+            amount: coin.amount,
+        }
+    }
+}
+
+impl<D: Display> Display for Coin<D> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), FmtError> {
+        write!(f, "{}{}", self.amount, self.denom)
```
soareschen, line 102: I change the format of coins to follow ibc-go, which has the format `"{amount}{denom}"` instead of `"{amount}-{denom}"`.

hu55a1n1: Previously we only provided a `Display` impl for `PrefixedCoin` because that is being used by the library for logging (in `send_transfer()`). Providing a `Display` impl for all coin types would mean that `PrefixedCoin` would be displayed as `10transfer/channel-0/transfer/channel-1/uatom`. I think I like the old version better. 
An alternative would be to not provide a `Display` impl at all and print the amount and denom separately in logs. Note that a `Coin's` fields are public so users can implement `Display` as they like.

soareschen: Hmm but I try to refactor the code to use the `Coin` type to format output rather than putting `format!("{}{}", amount, denom)` everywhere. Otherwise I guess I'd just define a separate `Token` type inside integration test rather than reusing the `Coin` type.

hu55a1n1: Could you use a `Coin` with a concrete `Denom` and impl `Display` on it in the integration test?

soareschen: Other crates cannot implement traits if they don't own either the type or the trait. In any case, the display format for `PrefixedCoin` also does not match the standard, which requires the path to be hashed.

I wonder if a better way is to use `Debug` for internal use of displaying coins including what type of coin it is, and use `Display` for standard compliance display.

hu55a1n1: I previously meant to suggest you use a newtype to overcome the orphan rule. 
Regarding the display format for `PrefixedCoin`, the ICS doesn't require the trace path to be hashed, it is a Cosmos SDK thing. (I also had this confusion initially. :sweat:)

hu55a1n1: If you like, we can also keep this for now and remove it from the modules after the split as we decided to do with the `IncentivizedPacket` related code earlier.

soareschen: If possible let's keep it for now, since I don't have time to rework #1981 before vacation, but want to keep it in a merge-able state. I can define a separate `Token` type for testing later, which should be easier on one hand, albeit less consistent on the other hand.

hu55a1n1 (approved):

PR #2695, opened 2022-10-04
Pin nightly version for no-std check until it works again on latest nightly

romac:
The `no_std` check for Substrate is broken on current nightly because of an internal compiler error (ICE).


hu55a1n1 (approved):
:ok_hand:

PR #2687, opened 2022-10-03
Fix clearing packets on ordered channels

romac:
Closes: #2670

## Description

______

### PR author checklist:
- [x] Added changelog entry, using [`unclog`](https://github.com/informalsystems/unclog).
- [x] Added tests: integration (for Hermes) or unit/mock tests (for modules).
- [x] Linked to GitHub issue.
- [x] Updated code comments and documentation (e.g., `docs/`).
- [x] Tagged *one* reviewer who will be the one responsible for shepherding this PR.

### Reviewer checklist:

- [ ] Reviewed `Files changed` in the GitHub PR explorer.
- [ ] Manually tested (in case integration/unit/mock tests are absent).

PR #2704, opened 2022-10-05
Add test for clearing packets on an ordered channel

romac:
Temporary PR to exercise the packet clearing on an ordered channel described in #2670 and tentatively fixed in #2687.

Will eventually be merged into #2687.


romac:
Merged into #2687

Issue #2707, opened 2022-10-06
Cross-Chain Queries (ICS 031)


SangyunOck:
## Summary
Implementation of ics-031 for the relayer side.

## Problem Definition

Requirement of ics-31 is to light-weight the process of interchain-queries without using full IBC stack so that
deals with query requests as many as possible while keeping other ibc transactions safe. 

The only method to handle cross chain query message via current version of IBC-rs is using packet, but there would be bottleneck when sharing worker for txs. Thus, separated runtime is needed for crosschain-queries.

## Proposal

This implementation includes:

1. Subscribing crosschain-query event which is different from ibc-packet(ics-04)
2. Separated worker for cross-chain-query-packet(struct of crosschain-query event)
3. Separated runtime to request query to rest gateway enabled node, runtime for txs used replying back to querying chain.

Also, there is a demo chain implemented ics-31 spec in the module side (in progress), and e2e test repository.

ibc-rs with ics-031: https://github.com/validance/ibc-rs
demo chain: https://github.com/decipherhub/interchain-queries
e2e test: https://github.com/validance/cross-chain-query-e2e

PR: https://github.com/informalsystems/hermes/pull/2709

adizere:
Update: See this discussion for more insights into use-cases for ICS 031 (and ICS 032): https://github.com/cosmos/ibc-go/discussions/2866

ljoss17:
Hermes extracts the IBC events from RPC events regarding channels (`OpenInit`, `OpenTry`, `CloseInit`, etc...) as well as `SendPacket` in the `RpcEventData::NewBlock`.
From `RpcEventData::Tx` it extracts all the events, with `SendPacket`, `ReceivePacket`, `WriteAcknowledgement`, `AcknowledgePacket`, `TimeoutPacket` and `TimeoutOnClosePacket` being part of `message.module: ibc_channel`.

It seems that Stride and Persistence send the Cross-Chain Queries in the EndBlock as part of `message.module: interchainquery`:

* https://github.com/cosmos/cosmos-sdk/blob/v0.46.7/types/events.go#L241-L244
* https://github.com/Stride-Labs/stride/blob/v4.0.3/x/interchainquery/keeper/abci.go#L26 and https://github.com/Stride-Labs/stride/blob/v4.0.3/x/interchainquery/types/events.go
[...]

PR #2709, opened 2022-10-06
Cross-Chain Queries (ICS 031)

SangyunOck:
Closes: #2707

## Description

Workflow

1. Module emits event for cross-chain-query.
2. Relayer subscribes query events, parses into object(CrossChainQueryPacket).
3. Just like ics-04 packet, collect query events into batch type according to block height.
4. Spawn a new worker for CrossChainQueryPacket.
5. Worker checks the query path, get data using query-runtime from via gateway enabled node not to block process for other queries. 
6. Send result to querying chain sending async transaction message.
7. On the module side, user can query the result which is stored in KVstore.

To-Be
1. Implement fee model for relayer
2. Query proofs

demo chain: https://github.com/decipherhub/interchain-queries
e2e test: https://github.com/validance/cross-chain-query-e2e
______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


crates/proto/src/prost/ibc.applications.ibc_query.v1.rs
```diff
@@ -0,0 +1,158 @@
+/// CrossChainQuery
+#[derive(Clone, PartialEq, ::prost::Message)]
```
adizere, line 2: Where did you get the .proto file to generate this? I don't see it in https://github.com/cosmos/ibc-go/tree/main/proto/ibc/applications.

scripts/gm/bin/gm.toml
```diff
@@ -0,0 +1,6 @@
+[global]
```
adizere, line 1: Probably should not be merged with this hardcoded values. Please rm this file from the PR.

adizere (changes requested):
Amazing, thanks for opening this contribution!

Also appreciate the testing recipe from https://github.com/cosmos/ibc/tree/main/spec/app/ics-031-crosschain-queries.

We will need to be able to test this feature in our CI so we can maintain it in the future. For that, we would need your help. Might be good to open a separate PR so that either:

-   a. you get familiar with our relayer [test framework](https://github.com/informalsystems/hermes/tree/master/tools/test-framework) and implement your own cross-chain query tests in that framework. Here is a [basic example of fungible token transfers test](https://github.com/informalsystems/hermes/blob/master/tools/integration-test/src/tests/transfer.rs).

-    b. or you add new CI/CD infra support to capture your own tests as you have them in https://github.com/validance/cross-chain-query-e2e.

scripts/gm/bin/gm.toml
SangyunOck: Done!

crates/proto/src/prost/ibc.applications.ibc_query.v1.rs
SangyunOck: Proto files are created by my cooperators who are working on cross-chain-query module. 

https://github.com/decipherhub/interchain-queries/tree/main/proto/ibc_query/v1

crates/relayer/src/chain/mock.rs
```diff
@@ -471,6 +490,34 @@ impl ChainEndpoint for MockChain {
     ) -> Result<(AnyConsensusState, MerkleProof), Error> {
         unimplemented!()
     }
+
+    fn cross_chain_query(
```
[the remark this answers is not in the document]
boosik: didn't use gRPC right?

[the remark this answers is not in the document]
scalalang2: sry, it mislead, let me correct.
the result of query defined in the module is not stored in blockchain.
It's evaluated at runtime and it wasn't proved by any validators, so that module receiving result from relayer cannot validate the value is valid or not.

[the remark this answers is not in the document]
SangyunOck: Thank you the idea of potential failure of system.
I'll try to find query path that can be proofed.

crates/relayer/src/chain/cosmos.rs
```diff
@@ -618,54 +631,12 @@ impl ChainEndpoint for CosmosSdkChain {
         Ok(HealthCheck::Healthy)
     }
 
-    /// Fetch a header from the chain at the given height and verify it.
```
romac, line 621: While I agree that the current order of definitions is subpar, I would rather not move code around in this PR to help with reviewing it and keeping the diff as small as possible.

Can you please undo these changes and keep the original ordering?

romac (changes requested):

crates/relayer/src/chain/cosmos.rs
SangyunOck: Replaced all the methods based on cosmos.rs in the main branch.

crates/proto/src/prost/ibc.applications.ibc_query.v1.rs
SangyunOck: Working with test framework(option a). Will open a new PR when it's ready!

adizere:
Hi @SangyunOck
I've been trying to reach your team through email for the last week. Can you contact me `adi@informal.systems` ? Before we accept your PR we need to gain a better understanding of your organization and the use-case here. Many thanks !

ljoss17:
I will close this PR since the PR https://github.com/informalsystems/hermes/pull/2723 contains the work done in this PR, with the addition of tests and some other additions.

Issue #2710, opened 2022-10-06
Rename leftover modules crate to `ibc-relayer-types` and trim it down


romac:
## Summary

Rename leftover modules crate to `ibc-relayer-types` and clean it up to remove everything not needed by the relayer, eg. the handlers.

## Problem Definition

Now that the `ibc` and `ibc-proto` crates have been split out into their own repositories, we are left with a copy of the `ibc` crate in the `crates/modules` folder.

## Proposal

- [x] Rename the `ibc` crate in the `crates/modules` folder to `ibc-relayer-types` 
- [x] Rename `crates/modules` to `crates/relayer-types`
- [x] Update the workspace `Cargo.toml` and the `[dependencies]` section of all crates which depend on `ibc`
- [x] Update all the import paths in all the crates which now depend on `ibc-relayer-types`
- [x] Remove all the handlers from `ibc-relayer-types` and all the code which depend on it
   - [x] Handlers themselves
   - [x] Contexts, readers and keepers
   - [x] Mock contexts, mock chain, etc.
   - [x] MBT tests, tests against the mock chain
- [x] Remove references to the mock chain and mock client from the `ibc-relayer `crate

### Follow-up:

- [ ] #2715

## Acceptance Criteria

All steps above have been carried out.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #2711, opened 2022-10-06
Rename leftover modules crate to ibc-relayer-types and trim it down

romac:
Closes: #2710

## Description

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


crates/relayer-cli/src/commands/query/packet/commitment.rs
```diff
@@ -82,7 +82,7 @@ impl QueryPacketCommitmentCmd {
                     sequence: self.sequence,
                     height: self.height.map_or(QueryHeight::Latest, |revision_height| {
                         QueryHeight::Specific(
-                            ibc::Height::new(chain.id().version(), revision_height)
+                            ibc_relayer_types::Height::new(chain.id().version(), revision_height)
```
ljoss17, line 85: Is the `ibc_relayer_types::` required since it is imported at line 10 ? A similar case in file `crates/relayer-cli/src/commands/query/packet/ack.rs` should we have `use ibc_relayer_types::Height` and call directly `Height` or should we remove `use ibc_relayer_types::Height` and use directly `ibc_relayer_types::Height` instead ? Not sure which is preferred, but I think being consistent everywhere would be nice.

romac: Agreed let's use the non-qualified version, ie. just Height

seanchen1991 (approved):
Seems good to me 👍

Issue #2484, opened 2022-07-29
Useless NewBlock events sent to connection and channel workers


ancazamfir:
## Summary of Bug
I believe that connection and channel handshake workers do not need the NewBlock events. The supervisor scans for unfinished handshakes and spawns a worker for each one. The worker itself should just act on connection or channel IBC events depending on the worker type.
Currently during handshake, on each new block event the worker performs some useless queries just to find out that there is no work to be done.

## Version
all

## Steps to Reproduce
- start two chains
- ```hermes --config psql_no.toml create channel --a-chain ibc-0 --b-chain ibc-1 --a-port transfer --b-port transfer --new-client-connection --yes```
- `hermes start`
- check the logs and metrics (`starts processing block event` in the logs with no constructive action, queries in metrics, e.g. `query_client_connections`, `query_connection`, etc)

## Acceptance Criteria

____

## For Admin Use

- [x] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2742, opened 2022-10-18
Fix handshake workers to process pending events

ancazamfir:
Closes: #2743
At first look I thought it might help with #2608 but the hermes releases don't match. Also it is not clear what exactly doesn't work there. Will investigate in parallel.

## Description
#2484 removed all new block event processing such that now pending handshakes are not picked up. Instead we should process the pending events on first NewBlock event. 
It didn't help that the e2e script has been broken for a long time therefore not catching this issue.

- [x] Fix e2e scripts 
- [x] Process first NewBlock event
- [x] Stop handshake on OpenAck and OpenConfirm events

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/relayer/src/channel.rs
+++ b/crates/relayer/src/channel.rs
@@ -762,6 +762,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
             _ => None,
         };
 
+        // Abort if the channel is at OpenAck or OpenConfirm stage, as there is nothing more for the worker to do
         match event {
             Some(IbcEvent::OpenConfirmChannel(_)) | Some(IbcEvent::OpenAckChannel(_)) => {
                 Ok((event, Next::Abort))
--- a/crates/relayer/src/channel.rs
+++ b/crates/relayer/src/channel.rs
@@ -747,7 +747,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
         &mut self,
         state: State,
     ) -> Result<(Option<IbcEvent>, Next), ChannelError> {
-        let res = match (state, self.counterparty_state()?) {
+        let event = match (state, self.counterparty_state()?) {
             (State::Init, State::Uninitialized) => Some(self.build_chan_open_try_and_send()?),
             (State::Init, State::Init) => Some(self.build_chan_open_try_and_send()?),
             (State::TryOpen, State::Init) => Some(self.build_chan_open_ack_and_send()?),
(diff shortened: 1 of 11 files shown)


crates/relayer/src/connection.rs
```diff
@@ -726,7 +726,12 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Connection<ChainA, ChainB> {
             _ => None,
         };
 
-        Ok((event, Next::Continue))
+        match event {
```
romac, line 729:
```suggestion
        // Abort if the connection is at OpenAck or OpenConfirm stage, as there is nothing more for the worker to do
        match event {
```

crates/relayer/src/channel.rs
```diff
@@ -762,7 +762,12 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
             _ => None,
         };
 
-        Ok((res, Next::Continue))
+        match event {
```
romac, line 765:
```suggestion
        // Abort if the channel is at OpenAck or OpenConfirm stage, as there is nothing more for the worker to do
        match event {
```

crates/relayer/src/worker/channel.rs
```diff
@@ -19,6 +19,7 @@ pub fn spawn_channel_worker<ChainA: ChainHandle, ChainB: ChainHandle>(
     chains: ChainHandlePair<ChainA, ChainB>,
     cmd_rx: Receiver<WorkerCmd>,
 ) -> TaskHandle {
+    let mut clear_pending = true;
```
romac, line 22: Would `complete_handshake_on_new_block` be a better name for this, since we are not actually clearing packets in the worker?

crates/relayer/src/worker/connection.rs
```diff
@@ -19,6 +19,7 @@ pub fn spawn_connection_worker<ChainA: ChainHandle, ChainB: ChainHandle>(
     chains: ChainHandlePair<ChainA, ChainB>,
     cmd_rx: Receiver<WorkerCmd>,
 ) -> TaskHandle {
+    let mut clear_pending = true;
```
romac, line 22: Same here.

crates/relayer/src/worker/channel.rs
ancazamfir: we are clearing a pending (vs live) handshake event. but i also like your proposal, will change.

romac:
Looks like there is already a test for completing connection and channel handshakes: https://github.com/informalsystems/hermes/blob/93e68aca490dfc78436aff28d23471de15a7d8fe/tools/integration-test/src/tests/supervisor.rs#L60-L76

Not sure why it didn't pick up this failure? Maybe because of `clear_on_start = true`?

ancazamfir:
> Looks like there is already a test for completing connection and channel handshakes:
> 
> https://github.com/informalsystems/hermes/blob/93e68aca490dfc78436aff28d23471de15a7d8fe/tools/integration-test/src/tests/supervisor.rs#L60-L76
> 
> Not sure why it didn't pick up this failure? Maybe because of `clear_on_start = true`?

I think it only tests handshake completion with the relayer already started. Also `clear_on_start = true` it a packet mode configuration.

romac (approved):

Issue #2733, opened 2022-10-13
Implement first iteration of `MockChainContext`


soareschen:
## Summary

The new relayer is fully generic and can work with any concrete chain implementation. With that we can create a simplified mock chain to test the relayer logic.

The mock chain can be implemented by implementing the OneForAll traits and instantiated with naive types such as:

```rust
struct PacketKey {
  client_id: u128,
  channel_id: u128,
  port_id: u128,
  sequence: u128,
  ...
}

struct MockChain {
  packet_commitments: HashMap<PacketKey, String>,
}

impl OfaChainTypes for MockChain {
  type Height = u128;
  type ClientId = u128;
  ...
}
```

Using this, we can test the correctness of the relayer by inspecting the end state of the mock chain, and see whether it records the correct state.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

soareschen:
@ljoss17 there are some follow up fixes that we still need to work on after #2785 is merged:

- The semantics of `process_messages` should be atomic: the messages need to be either all being processed successfully at the same height, or all failed. So this means that methods like `self.receive_packet` should not directly update the chain state. Instead they should take the current pending chain state, and return a new chain state to be committed. The next message handler would take the chain state returned from the previous message handler. And if any message handling failed, the whole pending chain state is discarded.

- The `query_write_ack_event` method introduced by #2907 is currently stubbed with `todo!()`, and needs to be implemented.

- The `Message::RecvPacket`, `Message::AckPacket`, and `Message::TimeoutPacket` should not contain the client ID. The client ID should be retrieved by having a mapping from channel ID to client ID in the local chain state.

- The `PacketKey` used in `Message` should contain the following fields: `source_channel_id`, `source_port_id`, `destination_channel_id`, `destination_port_id`, `sequence`, `timeout_height`, `timeout_timestamp`. Whether to read from the source or channel ID depends on the specific message. For example, processing of `RecvPacket` would use the destination channel ID as the local channel ID.

seanchen1991:
To be able to track this more cleanly, I opened #2957 to keep track of the follow-up work.

PR #2785, opened 2022-10-28
Mock chain context

ljoss17:
Closes: #2733

## Description

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


tools/test-framework/src/relayer_mock/base/types/state.rs
```diff
@@ -0,0 +1,47 @@
+use std::collections::HashSet;
+use std::fmt::Display;
+
+#[derive(Clone, Default)]
+pub struct State {
+    recv_events: HashSet<String>,
```
soareschen, line 6: We can use `HashSet<(PortId, ChannelId, Sequence)>` as key.

It would be good to define type aliases such as `type PortId = String`, so that we can change it to other types like `u8` easily in the future.

soareschen (changes requested):

crates/relayer-framework/src/tests/relayer_mock/base/types/runtime.rs
```diff
@@ -0,0 +1,51 @@
...
+            LogLevel::Trace => tracing::trace!(message),
+        }
+    }
+
+    async fn sleep(&self, duration: Duration) {
+        sleep(duration).await;
+    }
+
+    fn now(&self) -> Instant {
+        Instant::now()
```
soareschen, line 37: The time itself should be mocked. You can have an internal `MockTimestamp` type that is a newtype over `u128` as seconds or milliseconds.

When `now()` is called, the mock runtime can return the current timestamp, and then increment the internal timestamp by one so that the next call will return a slightly later time.

crates/relayer-framework/src/tests/relayer_mock/base/types/runtime.rs
```diff
@@ -0,0 +1,51 @@
...
+            LogLevel::Error => tracing::error!(message),
+            LogLevel::Warn => tracing::warn!(message),
+            LogLevel::Info => tracing::info!(message),
+            LogLevel::Debug => tracing::debug!(message),
+            LogLevel::Trace => tracing::trace!(message),
+        }
+    }
+
+    async fn sleep(&self, duration: Duration) {
+        sleep(duration).await;
```
soareschen, line 33: Sleeping is a little tricky to handle, as they makes the test non-deterministic. We might need a sleep queue so that we can manually wake the tasks that are sleeping. Something like

```rust
struct MockRuntime {
  current_time: Timestamp,
  sleep_tasks: Arc<Mutex<Vec<(Timestamp, SenderOnce<()>)>>>,
}

async fn sleep(&self, duration: Duration) {
  let wake_time = self.current_time.add(duration);
  let (sender, receiver) = channel_once();
  self.sleep_tasks.lock().push((wake_time, sender));
  receiver.await;
}
```

crates/relayer-framework/src/tests/relayer_mock/base/types/runtime.rs
```diff
@@ -0,0 +1,51 @@
...
+    fn duration_since(time: &Instant, other: &Instant) -> Duration {
+        time.duration_since(*other)
+    }
+
+    fn spawn<F>(&self, task: F)
+    where
+        F: Future + Send + 'static,
+        F::Output: Send + 'static,
+    {
+        self.runtime.spawn(task);
```
soareschen, line 49: Concurrency is something that is very difficult to test. We can leave this for now, but I am rethinking if we should use approaches that are more similar to TLA+ so that we can mock and test the relayer in a more deterministic way.

In the future, we might also want to consider splitting the one-for-all runtime also into minimal and full variants, so that we would only implement a minimal runtime for tests that can be tested using the mock context we define here.

crates/relayer-framework/src/tests/relayer_mock/base/impls/chain.rs
```diff
@@ -0,0 +1,171 @@
...
+    async fn send_messages(
+        &self,
+        messages: Vec<Self::Message>,
+    ) -> Result<Vec<Vec<Self::Event>>, Error> {
+        let mut res = vec![];
+        for m in messages {
+            match m {
+                MockMessage::SendPacket(_, _, h, p) => {
+                    self.receive_packet(p)?;
+                    res.push(Event::RecvPacket(h));
```
soareschen, line 103: I think here should return a write acknowledgement event?

crates/relayer-framework/src/tests/relayer_mock/base/impls/chain.rs
```diff
@@ -0,0 +1,171 @@
...
+                MockMessage::AckPacket(_, _, p) => {
+                    self.acknowledge_packet(p)?;
+                }
+                MockMessage::UpdateClient(from, _, s) => {
+                    self.insert_consensus_state(from, s);
+                }
+                _ => {}
+            }
+        }
+        Ok(vec![res])
```
soareschen, line 114: The result events should be a nested list of events where the outer list have the same length as the number of messages. So if a message do not have any event, we should push an empty inner list to the result event.

crates/relayer-framework/src/tests/relayer_mock/contexts/chain.rs
```diff
@@ -0,0 +1,118 @@
...
+    <tests::relayer_mock::contexts::chain::MockChainContext as OfaChainTypes>::ClientId;
+
+/// The 'state' represents the received messages and received acks at a given height.
+/// The 'consensus_states' represent the known whole states of clients.
+/// Usually the 'consensus_states' would be the Merkle Tree root hashes, but for
+/// the MockChain the whole state is used to avoid needing to implement Merkle
+/// Trees and Proofs.
+pub struct MockChainContext {
+    pub name: String,
+    pub states: Arc<Mutex<HashMap<Height, State>>>,
```
soareschen, line 25: Let's also have a `current_state` field here so that the access to the latest state always succeed.

crates/relayer-framework/src/tests/relayer_mock/contexts/chain.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub consensus_states: Arc<Mutex<HashMap<ClientId, ChainState>>>,
```
soareschen, line 26: The `consensus_states` should be a `HashMap<ClientId, HashMap<Height, ChainState>>`, since a client can have many consensus states at different heights.

Let's also try and explain why we want what the `states` and `consensus_states` are for, and why they have heights and client IDs as the hash map keys.

crates/relayer-framework/src/tests/relayer_mock/base/types/state.rs
```diff
@@ -0,0 +1,48 @@
...
+use std::{collections::HashSet, fmt::Display};
+
+type PortId = String;
+type ChannelId = String;
+type Sequence = u128;
+
+#[derive(Clone, Default, Debug)]
+pub struct State {
+    recv_events: HashSet<(PortId, ChannelId, Sequence)>,
+    ack_events: HashSet<(PortId, ChannelId, Sequence)>,
```
soareschen, line 11: These are more like received packets and acked packets.

crates/relayer-framework/src/tests/relayer_mock/base/types/state.rs
```diff
@@ -0,0 +1,48 @@
+use alloc::string::{String, ToString};
+use std::{collections::HashSet, fmt::Display};
+
+type PortId = String;
+type ChannelId = String;
+type Sequence = u128;
+
+#[derive(Clone, Default, Debug)]
+pub struct State {
```
soareschen, line 9: We should also have a `sent_packets` field that records packets that are being sent. We would then modify this state directly during tests, to insert the packets we want to send.

crates/relayer-framework/src/tests/relayer_mock/base/impls/relay.rs
```diff
@@ -0,0 +1,157 @@
...
+            height.clone(),
+            state,
+        )])
+    }
+
+    async fn build_dst_update_client_messages(
+        &self,
+        height: &<Self::SrcChain as OfaChainTypes>::Height,
+    ) -> Result<Vec<<Self::DstChain as OfaChainTypes>::Message>, Self::Error> {
+        let state = self.dst_chain().chain.state();
```
soareschen, line 111: We should give back the state at the specific height. Although since our mock chain do not have a state at every height, it should give back the state that is at the earliest after the specified height. If all heights are before the given height, it should return error.

crates/relayer-framework/src/tests/relayer_mock/base/impls/relay.rs
```diff
@@ -0,0 +1,157 @@
...
+            state,
+        )])
+    }
+
+    async fn build_receive_packet_message(
+        &self,
+        height: &<Self::SrcChain as OfaChainTypes>::Height,
+        packet: &Self::Packet,
+    ) -> Result<<Self::DstChain as OfaChainTypes>::Message, Self::Error> {
+        let h = self.dst_chain().chain.get_latest_height().unwrap();
```
soareschen, line 124: We should get the latest state and check if the packet is in the list of sent packets. If not, it should return error.

crates/relayer-framework/src/tests/relayer_mock/base/impls/relay.rs
```diff
@@ -0,0 +1,157 @@
...
+        ))
+    }
+
+    async fn build_ack_packet_message(
+        &self,
+        destination_height: &<Self::DstChain as OfaChainTypes>::Height,
+        packet: &Self::Packet,
+        _ack: &<Self::DstChain as OfaChainTypes>::WriteAcknowledgementEvent,
+    ) -> Result<<Self::SrcChain as OfaChainTypes>::Message, Self::Error> {
+        Ok(MockMessage::AckPacket(
```
soareschen, line 139: We should get the latest state and check if the packet is in the list of received packets. If not, it should return error.

crates/relayer-framework/src/tests/relayer_mock/base/impls/relay.rs
```diff
@@ -0,0 +1,157 @@
...
+            packet.clone(),
+        ))
+    }
+
+    async fn build_timeout_unordered_packet_message(
+        &self,
+        destination_height: &<Self::DstChain as OfaChainTypes>::Height,
+        packet: &Self::Packet,
+    ) -> Result<<Self::SrcChain as OfaChainTypes>::Message, Self::Error> {
+        Ok(MockMessage::TimeoutPacket(
```
soareschen, line 151: We should check if the packet timeout height is less than the current state's height, and the timeout timestamp is less than the current state's timestamp. If not, it should return error.

crates/relayer-framework/src/tests/relayer_mock/base/impls/chain.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// If the message is an `UpdateClient` update the consensus state.
+    async fn send_messages(
+        &self,
+        messages: Vec<Self::Message>,
+    ) -> Result<Vec<Vec<Self::Event>>, Error> {
+        let mut res = vec![];
+        for m in messages {
+            match m {
+                MockMessage::SendPacket(_, _, h, p) => {
+                    self.receive_packet(p)?;
```
soareschen, line 102: We should check on the consensus state on whether there is a consensus state at the _exact_ height given. If so, we should also check if the consensus state contains the packet in the list of sent packets. If not, it should return an error.

The same applies to `AckPacket`.

crates/relayer-framework/src/tests/relayer_mock/base/types/message.rs
```diff
@@ -0,0 +1,25 @@
+use std::fmt::{Display, Formatter, Result};
+
+use crate::tests::relayer_mock::base::types::{height::Height, packet::PacketKey};
+use crate::tests::relayer_mock::contexts::chain::{ChainState, ClientId};
+
+#[derive(Debug)]
+pub enum Message {
+    SendPacket(ClientId, Height, Height, PacketKey),
```
soareschen, line 8: This should be a `RecvPacket` message. SendPacket events are initiated by external transactions like IBC transfer.

We should also document what each of these messages correspond to in the actual IBC implementation.

crates/relayer-framework/src/tests/relayer_mock/mock.rs
```diff
@@ -0,0 +1,98 @@
...
+                &packet.port_id,
+                &packet.channel_id,
+                &packet.sequence
+            ));
+        }
+
+        // Source chain must be higher than destination chain
+        src_chain.chain.new_block()?;
+
+        let events = runtime.block_on(async { relay_context.relay_packet(&packet).await });
```
soareschen, line 56: There is no need to block on async if our test is an async function.

soareschen (changes requested):
Great work at the initial mock chain implementation!

Although the mock chain is a simplified implementation of the actual chain, let's try to get the semantics of IBC as close to the actual chain behavior as possible. I have added some comments on the checks the mock chain should do in order to ensure that invalid packets cannot be sent accidentally.

The mock chain should also serve as a good learning material for people to learn about how IBC works conceptually. We should add some documentation on why the mock chain is designed in certain ways, and what behavior we are simulating or simplifying from the real IBC spec. For example, we want to explain that we use the actual chain state in place of merkle roots, since that is easier to implement and debug in a mock environment.

Since we already some basic tests, we should be able to improve the mock chain implementation incrementally.

crates/relayer-framework/src/lib.rs
```diff
@@ -5,6 +5,12 @@
 
 extern crate alloc;
 
+#[cfg(test)]
+extern crate std;
+
+#[cfg(test)]
+pub mod tests;
```
soareschen, line 12: Looking at it now, I think it might be better if we create a dedicated `relayer-framework-test` crate for this. It would be better if we can avoid using feature flags and dev dependencies to keep the `relayer-framework` crate itself clean of all possible depedencies.

crates/relayer-framework-test/src/relayer_mock/base/types/aliases.rs
```diff
@@ -0,0 +1,17 @@
...
+use ibc_relayer_framework::base::one_for_all::traits::chain::OfaChainTypes;
+
+use crate::relayer_mock;
+use crate::relayer_mock::base::types::height::Height as MockHeight;
+
+pub type PacketUID = (PortId, ChannelId, Sequence);
+pub type ConsensusState =
+    <relayer_mock::contexts::chain::MockChainContext as OfaChainTypes>::ConsensusState;
+pub type ChainState = HashMap<MockHeight, ConsensusState>;
+pub type ClientId = <relayer_mock::contexts::chain::MockChainContext as OfaChainTypes>::ClientId;
```
soareschen, line 12: Here we should define the types directly such as:

```rust
type ConsensusState = State;
type ClientId = String;
```

And then inside the implementation of `OfaChainTypes`, we will refer to these type aliases.

crates/relayer-framework-test/src/relayer_mock/base/impls/chain.rs
```diff
@@ -0,0 +1,190 @@
...
+    /// If the message is a `SendPacket`, update the received packets,
+    /// and add a `RecvPacket` event to the returned array of events.
+    /// If the message is an `AckPacket`, update the received acknowledgment
+    /// packets.
+    /// If the message is an `UpdateClient` update the consensus state.
+    async fn send_messages(
+        &self,
+        messages: Vec<Self::Message>,
+    ) -> Result<Vec<Vec<Self::Event>>, Error> {
+        let mut res = vec![];
```
soareschen, line 112: Let's move the implementation body to a `process_messages` method in `MockChainContext`, so that we keep the trait implementation body simple.

crates/relayer-framework-test/src/relayer_mock/contexts/chain.rs
```diff
@@ -0,0 +1,236 @@
...
+use crate::relayer_mock::base::types::runtime::{MockChainRuntimeContext, MockRuntimeContext};
+use crate::relayer_mock::base::types::{
+    height::Height as MockHeight, packet::PacketKey, state::State,
+};
+use ibc_relayer_framework::base::one_for_all::traits::runtime::OfaRuntimeContext;
+
+pub struct MockChainContext {
+    pub name: String,
+    pub consensus_states: Arc<Mutex<HashMap<MockHeight, ChainState>>>,
+    pub consensus_state: Arc<Mutex<ChainState>>,
```
soareschen, line 33: The consensus state refers to the state of other chains recorded on the current chain. For the current chain's state, we can just call it `current_state`. It looks like the `consensus_states` field here should actually be called `past_chain_states`.

crates/relayer-framework-test/src/relayer_mock/contexts/chain.rs
```diff
@@ -0,0 +1,236 @@
...
+    pub client_states: Arc<Mutex<HashMap<ClientId, HashMap<MockHeight, ChainState>>>>,
```
soareschen, line 34: This is actually the `consensus_states` we talked about. The `client_states` field should be something like:

```rust
client_states: Arc<Mutex<HashMap<ClientId, ClientState>>>
```

Where `ClientState` is a different type that contains information such as the client type and whether the client has been frozen. However since we do not model that part at the moment, we can skip defining the actual `client_states` field.

soareschen (changes requested):

crates/relayer-framework-test/src/relayer_mock/util/mutex.rs
```diff
@@ -0,0 +1,19 @@
+//! MutexUtil is a helper trait to simplify the error handling
+//! when locking a Mutex.
+
+use std::sync::{Arc, Mutex, MutexGuard};
+
+use crate::relayer_mock::base::error::Error;
+
+pub trait MutexUtil<T> {
+    fn acquire_mutex(&self) -> Result<MutexGuard<T>, Error>;
```
soareschen, line 9: Sorry for not being clear, I think we can have a non-fallible version of `acquire_mutex` here. For this specific case, we can treat poisoned mutex errors as fatal errors during tests that they can just panic. We can add an `#[allow()]` pragma here for an exception for allowing panic.

With that, the functions that only use the mutex can return the result straight without needing to return `Result`.

crates/relayer-framework-test/src/relayer_mock/base/types/runtime.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    // Increment the shared MockClock by the duration is milliseconds.
+    async fn sleep(&self, duration: Duration) {
+        if self.clock.increment_millis(duration.as_millis()).is_err() {
+            tracing::warn!("MockClock failed to sleep for {}ms", duration.as_millis());
+        }
+    }
+
+    fn now(&self) -> Instant {
+        Instant::now()
```
soareschen, line 60: `now()` should use the mock clock.

crates/relayer-framework-test/src/relayer_mock/base/types/aliases.rs
```diff
@@ -0,0 +1,15 @@
+use std::collections::HashMap;
+
+use crate::relayer_mock::base::types::chain::MockChainStatus;
+use crate::relayer_mock::base::types::height::Height as MockHeight;
+use crate::relayer_mock::base::types::state::State;
+
+pub type PacketUID = (PortId, ChannelId, Sequence);
+pub type ConsensusState = State;
+pub type ChainState = HashMap<MockHeight, ConsensusState>;
```
soareschen, line 9: Seems like there is some confusion here. The `ChainState` and `ConsensusState` should be just `State`.

crates/relayer-framework-test/src/relayer_mock/contexts/chain.rs
```diff
@@ -0,0 +1,302 @@
...
+};
+use crate::relayer_mock::util::clock::MockClock;
+use crate::relayer_mock::util::mutex::MutexUtil;
+
+use ibc_relayer_framework::base::one_for_all::traits::runtime::OfaRuntimeContext;
+
+pub struct MockChainContext {
+    pub name: String,
+    pub past_chain_states: Arc<Mutex<HashMap<MockHeight, ChainState>>>,
+    pub current_state: Arc<Mutex<ChainState>>,
```
soareschen, line 37: I think what you need here is a `current_height` field and a `current_state` field, so that both operations are infallible inside the implementation.

crates/relayer-framework-test/src/relayer_mock/contexts/chain.rs
```diff
@@ -0,0 +1,302 @@
...
+                MockMessage::RecvPacket(receiver, h, p) => {
+                    let client_consensus =
+                        self.query_consensus_state_at_height(receiver.clone(), h.clone())?;
+                    let state = client_consensus.get(&h).ok_or_else(|| {
+                        Error::no_consensus_state_at_height(self.name().to_string(), h.0)
+                    })?;
+                    if !state.check_sent(&p.port_id, &p.channel_id, &p.sequence) {
+                        return Err(Error::generic(eyre!("chain `{}` got a RecvPacket, but client `{}` state doesn't have the packet as sent", self.name(), receiver)));
+                    }
+                    // Check that the packet is not timed out. Current height < packet timeout height.
```
soareschen, line 275: It seems like the timeout packet logic has not yet been implemented.

crates/relayer-framework-test/src/relayer_mock/contexts/chain.rs
```diff
@@ -0,0 +1,302 @@
...
+                    self.receive_packet(p)?;
```
soareschen, line 276: Should the validation logic above be implemented inside this function, so that the `process_messages` logic can be kept simple?

soareschen (changes requested):
This looks mostly ready now. Just some minor revisions needed.

crates/relayer-framework-test/src/relayer_mock/contexts/chain.rs
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    /// If the message is a `SendPacket`, update the received packets,
+    /// and add a `RecvPacket` event to the returned array of events.
+    /// If the message is an `AckPacket`, update the received acknowledgment
+    /// packets.
+    /// If the message is an `UpdateClient` update the consensus state.
+    /// When a RecvPacket and AckPacket are received, verify that the client
+    /// state has respectively sent the message and received the message.
+    pub fn process_messages(&self, messages: Vec<MockMessage>) -> Result<Vec<Vec<Event>>, Error> {
```
soareschen, line 315: Actually there is one last fix needed for the semantics of `process_messages`: the messages need to be either all being processed successfully at the same height, or all failed.

So this means that methods like `self.receive_packet` should not directly update the chain state. Instead they should take the current pending chain state, and return a new chain state to be committed. The next message handler would take the chain state returned from the previous message handler. And if any message handling failed, the whole pending chain state is discarded.

soareschen (changes requested):

soareschen (approved):

crates/relayer-framework-test/src/relayer_mock/contexts/chain.rs
soareschen: In any case, I will merge this first so that we don't get too many merge conflicts from too many parallel branches.

PR #2793, opened 2022-10-31
Updated the gm.toml file

harveenSingh:
## Description

- Update "PATH-TO-HERMES-BINARY" in gm.toml for clarity
- In reference to the issue in gm repo (https://github.com/informalsystems/gm/issues/2)

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.changelog/v1.1.0/improvements/guide/update-gm.toml-template.md
+++ b/.changelog/v1.1.0/improvements/guide/update-gm.toml-template.md
@@ -1,0 +1,2 @@
+- Update gm.toml file for better clarity in reference to the issue in gm repo
+https://github.com/informalsystems/gm/issues/2
--- a/guide/src/templates/files/gm/default_gm.toml
+++ b/guide/src/templates/files/gm/default_gm.toml
@@ -10,7 +10,7 @@
   wallet_mnemonic = ""
 
   [global.hermes]
-    binary = "./hermes" #change this path according to your setup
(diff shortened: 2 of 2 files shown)


seanchen1991 (approved):

adizere:
Thanks Harveen, Sean for the quick turnaround!

There is a way to make `gm` more resilient, but we can leave that for later. What I had in mind is to change the logic in `lib-gm` so that command `gm hermes keys` does not fail silently upon misconfigured `binary` path, but finding the right place to fix in lib-gm seems tricky, will keep it in mind for a later time.

greg-szabo:
Ah, I had this long-winded description on we can do this properly and then the GitHub page timed out on me...

In short:
* If we [change](https://github.com/informalsystems/hermes/pull/2793/files) the example config.toml, we should also change the [configuration default](https://github.com/informalsystems/gm/blob/b3bc17d88e50076f8c9a954a437454605cc42e84/bin/lib-gm#L48) in the code.
* Hermes binary is not checked "early" in the code because it could fail the code when something completely unrelated (like a `gm init`) is happening. Having a valid hermes binary should not be a requirement to initialize, start, stop, configure networks.
* Checking hermes execution each time for errors seems overkill (and not readable). Maybe we should check each "first execution" of a hermes command in the `hermes_*` functions. That would gives a decent way to get out if there's no hermes binary without killing execution for unrelated commands.

Issue #2781, opened 2022-10-28
Error output when upgrading client with `hermes upgrade client`


ljoss17:
## Summary of Bug

There is an issue with the `hermes upgrade client` which seems to be related to the `Any` types.

Error output:

```
ERROR failed while trying to upgrade client id 07-tendermint-0 for chain ibc-0: failed while fetching from
chain the upgraded client state: conversion from a protobuf `Any` into a domain type failed: conversion
from a protobuf `Any` into a domain type failed:
error converting message type into domain type: the client state was not found
```

## Version

`master` branch 28.10.22, `hermes 1.0.0+3aba0ff27`

## Steps to Reproduce

Follow the guide to upgrade the chain, https://hermes.informal.systems/commands/upgrade/test.html.
Note that the command at step `2.` should actually be: `hermes tx upgrade-chain --reference-chain ibc-0 --host-chain ibc-1 --host-client 07-tendermint-0 --amount 10000000 --height-offset 60`

## Acceptance Criteria

- [ ] `upgrade client` works as expected
- [ ] https://github.com/informalsystems/hermes/issues/2312

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

plafer:
just tried it with the latest master and it worked

```
SUCCESS [
    UpdateClient(
        UpdateClient {
            common: Attributes {
                client_id: ClientId(
                    "07-tendermint-0",
                ),
                client_type: Tendermint,
                consensus_height: Height {
                    revision: 0,
                    height: 80,
                },
            },
            header: Some(
                 Header {...},
            ),
        },
    ),
    UpgradeClient(
        UpgradeClient(
            Attributes {
                client_id: ClientId(
                    "07-tendermint-0",
                ),
                client_type: Tendermint,
                consensus_height: Height {
                    revision: 0,
                    height: 81,
                },
            },
        ),
    ),
]
```

ljoss17:
It seems that this error is triggered when calling `hermes upgrade client --host-chain ibc-1 --client 07-tendermint-0 --upgrade-height <UPGRADE_HEIGHT>` with the incorrect `UPGRADE_HEIGHT`.
Otherwise, if the correct `UPGRADE_HEIGHT` is passed, everything works fine.

Issue #2312, opened 2022-06-16
Add an integration test for client upgrade


plafer:
Can we add an integration test for client upgrade? Maybe in a different PR?

_Originally posted by @ancazamfir in https://github.com/informalsystems/ibc-rs/issues/2304#issuecomment-1157811442_

plafer:
Let's wait for #2300 to be addressed too

PR #2795, opened 2022-11-01
Add upgrade client integration tests

ljoss17:
Closes: #2781, closes: #2312

## Description

~This PR fixes the error when upgrading a client caused by failing the conversion from a protobuf `Any` into a domain type.~
The conversion from protobuf `Any` is triggered when the incorrect upgrade height is passed to the client upgrade call.
The error message has been updated in order to suggest checking the height used when upgrading a client.

In addition this PR adds integration test for client upgrade.

### Testing

Follow the testing instructions https://hermes.informal.systems/documentation/commands/upgrade/test.html

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


tools/integration-test/src/tests/client_upgrade.rs
```diff
@@ -0,0 +1,577 @@
...
+}
+
+struct ClientUpgradeTest;
+
+impl TestOverrides for ClientUpgradeTest {
+    /// Update the genesis file in order to reduce the time required to upgrade the chain.
+    fn modify_genesis_file(&self, genesis: &mut serde_json::Value) -> Result<(), Error> {
+        use serde_json::Value;
+
+        let max_deposit_periods = genesis
```
soareschen, line 58: We can define a helper function in `tools/test-framework/src/chain/config.rs` or similar location to modify the config parameters.

tools/integration-test/src/tests/client_upgrade.rs
```diff
@@ -0,0 +1,577 @@
...
+            .expect("Upgrade height cannot be 1");
+
+        let mut reference_application_latest_height = foreign_clients
+            .client_a_to_b
+            .src_chain()
+            .query_latest_height()
+            .ok()
+            .unwrap();
+
+        while reference_application_latest_height != target_reference_application_height {
```
soareschen, line 174: We may want to poll for a limited time so that the test don't run indefinitely when failed.

tools/integration-test/src/tests/client_upgrade.rs
```diff
@@ -0,0 +1,577 @@
...
+
+        assert!(outcome.is_err(), "{:?}", outcome);
+
+        Ok(())
+    }
+}
+
+struct HeightTooHighClientUpgradeTest;
+
+impl TestOverrides for HeightTooHighClientUpgradeTest {
```
soareschen, line 237: If there are multiple identical test overrides, we can group them into a common override and implement `HasOverrides` instead. See `ExpirationTestOverrides` in `tools/integration-test/src/tests/client_expiration.rs` for example.

tools/integration-test/src/tests/client_upgrade.rs
```diff
@@ -0,0 +1,577 @@
...
+
+        // Wait for the proposal to be processed
+        std::thread::sleep(Duration::from_secs(2));
+
+        let driver = chains.node_a.chain_driver().0;
+
+        // Retrieve the height which should be used to upgrade the client
+        let upgrade_height = driver
+            .runtime
+            .block_on(query_upgrade_proposal_height(
```
soareschen, line 326: We can define extension methods in `tools/test-framework/src/chain/ext` so that here we can call `driver.query_upgrade_proposal_height()` while preserving the type tags.

tools/integration-test/src/tests/client_upgrade.rs
```diff
@@ -0,0 +1,577 @@
...
+        let dst_chain_config = chains.handle_a().config().unwrap();
+
+        let upgraded_chain_id = chains.chain_id_a().0.clone();
+
+        let src_client_id = foreign_clients.client_id_b().0.clone();
+
+        // Create and send an chain upgrade proposal
+        let opts = UpgradePlanOptions {
+            dst_chain_config,
+            src_chain_config,
```
soareschen, line 117: What is the purpose to pass in the chain configs inside `UpgradePlanOptions`? I tried removing the fields, and it does not indicate that the configs are used anywhere.

tools/integration-test/src/tests/client_upgrade.rs
```diff
@@ -0,0 +1,577 @@
...
+            .unwrap();
+
+        let client_upgrade_height = Height::new(
+            foreign_clients.client_a_to_b.src_chain().id().version(),
+            upgrade_height,
+        )
+        .unwrap();
+
+        // Vote on the proposal so the chain will upgrade
+        vote_proposal(
```
soareschen, line 155: We can add an extension method in `tools/test-framework/src/chain/ext` for calling `driver.vote_proposal()`.

soareschen (changes requested):

tools/integration-test/src/tests/client_upgrade.rs
ljoss17: This has been addressed in this commit https://github.com/informalsystems/hermes/pull/2795/commits/88e15de2a355dce396d45ed1688db9b2d12f6a6e

tools/integration-test/src/tests/client_upgrade.rs
ljoss17: This has been addressed in this commit https://github.com/informalsystems/hermes/pull/2795/commits/fa7ad9cc3e09eb4e8d1e8852d0976045f408ae78

tools/integration-test/src/tests/client_upgrade.rs
ljoss17: This has been addressed in this commit https://github.com/informalsystems/hermes/pull/2795/commits/3e6b163b86f05455777d928039875cdcd28260ba

tools/integration-test/src/tests/client_upgrade.rs
ljoss17: This has been addressed in this commit https://github.com/informalsystems/hermes/pull/2795/commits/b5b046afc91fc53c3d10a40115cdd0140d7b0010

tools/integration-test/src/tests/client_upgrade.rs
ljoss17: Indeed it seems these two fields are never used. I removed them in this commit https://github.com/informalsystems/hermes/pull/2795/commits/fde11207a6f03fb3475f1225973493d0c2860a38

tools/integration-test/src/tests/client_upgrade.rs
ljoss17: This was addressed in this commit https://github.com/informalsystems/hermes/pull/2795/commits/b5b046afc91fc53c3d10a40115cdd0140d7b0010

tools/test-framework/src/chain/ext/wait_chain.rs
```diff
@@ -0,0 +1,51 @@
...
+    // Query the latest height of the chain
+    let mut reference_application_latest_height = foreign_clients
+        .client_a_to_b
+        .src_chain()
+        .query_latest_height()
+        .ok()
+        .unwrap();
+
+    let start = std::time::SystemTime::now();
+    while reference_application_latest_height != target_reference_application_height {
```
romac, line 22: Let's perhaps to this to avoid having the chain skip over the target height? It's unlikely to occur but just to be safe.

```suggestion
    while reference_application_latest_height < target_reference_application_height {
```

tools/test-framework/src/chain/ext/wait_chain.rs
```diff
@@ -0,0 +1,51 @@
...
+) -> Result<(), Error> {
+    // Query the latest height of the chain
+    let mut reference_application_latest_height = foreign_clients
+        .client_a_to_b
+        .src_chain()
+        .query_latest_height()
+        .ok()
+        .unwrap();
+
+    let start = std::time::SystemTime::now();
```
romac, line 21: Let's add a `use` statement for `SystemTime` since we are using it multiple times?

tools/test-framework/src/chain/ext/proposal.rs
```diff
@@ -0,0 +1,91 @@
...
+use ibc_proto::cosmos::gov::v1beta1::{query_client::QueryClient, QueryProposalRequest};
+use ibc_proto::ibc::core::client::v1::UpgradeProposal;
+use ibc_relayer::error::Error as RelayerError;
+
+use crate::chain::cli::upgrade::vote_proposal;
+use crate::chain::driver::ChainDriver;
+use crate::error::Error;
+use crate::types::tagged::*;
+
+pub trait ChainProposalMethodsExt<Chain> {
```
soareschen, line 13: I think there is no need for a generic `Chain` parameter here.

tools/integration-test/src/tests/client_upgrade.rs
soareschen: Did you missed updating the use of the update method here?

tools/test-framework/src/chain/config.rs
```diff
@@ -141,3 +141,43 @@ pub fn set_mode(config: &mut Value, mode: &str) -> Result<(), Error> {
...
+        Err(eyre!("failed to update max_deposit_period in genesis file"))
+    }
+}
+
+pub fn set_voting_period(genesis: &mut serde_json::Value, period: &str) -> Result<(), Error> {
+    let voting_period = genesis
+        .get_mut("app_state")
+        .and_then(|app_state| app_state.get_mut("gov"))
+        .and_then(|gov| gov.get_mut("voting_params"))
+        .and_then(|voting_params| voting_params.as_object_mut());
```
soareschen, line 170: We can use `.ok_or_else()?` to extract the `Option`.

tools/test-framework/src/chain/ext/wait_chain.rs
```diff
@@ -0,0 +1,53 @@
...
+        .client_a_to_b
+        .src_chain()
+        .query_latest_height()
+        .ok()
+        .unwrap();
+
+    let start = SystemTime::now();
+    while reference_application_latest_height < target_height_of_a {
+        // Check if the wait time has timed out.
+        match SystemTime::now().duration_since(start) {
```
soareschen, line 26: We can do `SystemTime::now().duration_since(start).map_err(handle_generic_error)?`

tools/test-framework/src/chain/ext/wait_chain.rs
```diff
@@ -0,0 +1,53 @@
...
+                )));
+            }
+        }
+        std::thread::sleep(Duration::from_millis(500));
+
+        // Query the latest height of the chain
+        reference_application_latest_height = foreign_clients
+            .client_a_to_b
+            .src_chain()
+            .query_latest_height()
```
soareschen, line 48: I believe we can just do `.query_latest_height()?` here. If not, we can add a `From` implementation for the test `Error` type.

tools/test-framework/src/chain/ext/proposal.rs
```diff
@@ -0,0 +1,91 @@
...
+                    Some(plan) => {
+                        let h = plan.height;
+                        Ok(h as u64)
+                    }
+                    None => Err(Error::empty_plan()),
+                },
+                Err(_e) => Err(Error::incorrect_proposal()),
+            }
+        }
+        None => Err(Error::empty_proposal()),
```
soareschen, line 89: Nested matches can be difficult to understand. Try using helpers like `.map_err()?` and `.ok_or_else()?`.

Inside tests, it is also fine to just return generic errors to indicate test failure. If we find the methods becoming complicated and needing to return fine grained errors, it may also worth considering whether to move the methods to the `ibc-relayer` crate so that they can be also used outside of tests.

tools/integration-test/src/tests/client_upgrade.rs
```diff
@@ -0,0 +1,385 @@
...
+
+        // Wait for the proposal to be processed
+        std::thread::sleep(Duration::from_secs(2));
+
+        let driver = chains.node_a.chain_driver();
+
+        // Retrieve the height which should be used to upgrade the client
+        let upgrade_height = driver
+            .query_upgrade_proposal_height(&Uri::from_str(&driver.0.grpc_address()).unwrap(), 1)
+            .unwrap();
```
soareschen, line 326: Use `?`.

soareschen (changes requested):

tools/integration-test/src/tests/client_upgrade.rs
ljoss17: It's the single implementation for `ClientUpgradeTestOverrides`, I could move it at the bottom if it avoids confusions. What do you think ?

soareschen: It is fine to move it to the config helper even if it is only used once. Functions are used for abstractions and improving readability, not to remove code duplication.

soareschen (approved):

tools/integration-test/src/tests/client_upgrade.rs
ljoss17: I'm sorry, I misunderstood your comment. I had forgotten to remove the duplicate code updating the max_deposit_period. It has been removed in 00d1529dfc2a47d6b06a702ba0795f93691f8eb9

PR #2843, opened 2022-11-08
Add support for dumping Prometheus metrics to JSON

romac:
Closes: #2890

## Description

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/telemetry/src/encoder.rs
+++ b/crates/telemetry/src/encoder.rs
@@ -1,0 +1,234 @@
+// Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
+
+use std::collections::BTreeMap;
+use std::io::{self, Write};
+
+use prometheus::proto::{self, MetricFamily, MetricType};
+use prometheus::{Encoder, Error, Result};
+
+use serde::Serialize;
+
+/// The JSON format of metric family.
+pub const JSON_FORMAT: &str = "application/json";
+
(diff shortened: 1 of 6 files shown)


crates/telemetry/src/encoder.rs
```diff
@@ -0,0 +1,234 @@
+// Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
```
adizere, line 1: A bit of a bummer we can't just import to use this, and instead we'll need to maintain it.

adizere (approved):
In principle looks good. I only did a couple of sanity (manual test) checks of the JSON output.

If there is a use-case on how these would be consumed that is more specific, then might be good to document it in the guide. Thanks Romain!

crates/telemetry/src/encoder.rs
romac: That's actually from the Prometheus text encoder, which I used as a base for the JSON one that I wrote myself. But I had to keep the license since the JSON encoder is therefore technically a derived work.

romac:
> If there is a use-case on how these would be consumed that is more specific, then might be good to document it in the guide. Thanks Romain!

I don't have a specific use case right now, or at least not fleshed out enough that it would be worth documenting, but I'll keep that in mind for when that's the case.

PR #2848, opened 2022-11-09
Remove unused or redundant dependencies, found via `cargo-machete`

romac:
Unfortunately, looking at the `Cargo.lock` diff, it does not seem to remove anything from the transitive dependency graph for the workspace. But it still cuts down on some dependencies when some of the crates are used individually.


adizere (approved):
Seems like this is ready for review? Approving since it looks good, though the impact is unclear.

Issue #1983, opened 2022-03-21
Add packet forwarding middleware tests


adizere:
_Note:_ This is a suggestions that came up in a conversation with @soareschen .

## Context

The IBC packet forward middleware v1.0.1 was added to the Hub features in the Vega upgrade
[https://github.com/cosmos/gaia/blob/main/docs/roadmap/cosmos-hub-roadmap-2.0.md#vega-upgrade-completed-december-14-2021](https://github.com/cosmos/gaia/blob/main/docs/roadmap/cosmos-hub-roadmap-2.0.md#vega-upgrade-completed-december-14-2021)

## Problem statement

There are a couple of problems with this:
- There are no integration tests for this feature.
- It is unclear to the members of our team how the feature works.

Furthermore, it seems that the packet forwarding middleware is a potential solution to address some of the scalability problems in Cosmos, suggesting we should develop the expertise around it and understand it more deeply.

## Acceptance criteria
- Write an integration test and document the functionality.

PR #2849, opened 2022-11-09
Integration test for packet forwarding

ljoss17:
Closes: #1983

## Description

Add integration test for packet forwarding.

Suppose there are chains A, B and C. Chain A can send tokens to C using packet forwarding by transferring the tokens through the channel and port to chain B with the recipient address being:

`{intermediate_refund_address}|{foward_port}/{forward_channel}:{final_destination_address}`

See https://hub.cosmos.network/main/governance/proposals/2021-09-hub-ibc-router/

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


tools/integration-test/src/tests/forward/forward_transfer.rs
```diff
@@ -0,0 +1,418 @@
...
+//!   forwarded.
+//!
+//! - The `InvalidChannelIbcForwardTransferTest` tests the case where the
+//!   sender is refunded if the forward channel specified is invalid.
+//!
+//! - The `InvalidAddressIbcForwardTransferTest` tests the case when the destination
+//!   address is invalid. This results in two scenarios depending on the Gaia
+//!   version.
+//!   For Gaia 6 the intermediate address will receive the refunded tokens.
+//!   For Gaia 8 the sender will receive the refunded tokens.
```
seanchen1991, line 13: I understand that this is the behavior exhibited by Gaia versions 6 and 8, but is this behavior intended for the respective Gaia versions, or are these bugs in the packet forwarding behavior for these versions? 

I'm mostly wondering if this test will need to be modified in the future in the case that this behavior is made uniform across Gaia versions.

ljoss17: I am not sure the behaviour of Gaia 6 is the one which was expected since the atomic forwarding isn't implemented. Gaia 6 uses the v1.0.1 forward middleware module, and atomic forwarding was introduced in v3. But the behaviour of Gaia 8 and I suppose all the next ones shouldn't change.
I will added more information in the comments regarding this.

tools/integration-test/src/tests/forward/forward_transfer.rs
```diff
@@ -0,0 +1,420 @@
...
+//!   This is due to Gaia 6 using the Forward Middleware Module v1.0.1, which
+//!   doesn't have the atomic forwarding which was introduced in v3.
```
seanchen1991, line 15: I think this could be formatted a little more clearly, but otherwise, it looks good! 
```suggestion
//! - The `InvalidAddressIbcForwardTransferTest` tests the case when the destination
//!   address is invalid. This results in two scenarios depending on the Gaia
//!   version:
//!    - For Gaia 6, the intermediate address will receive the refunded tokens. This behavior
//!      deviates from other Gaia versions due to Gaia 6 using the Forward Middleware
//!      Middleware Module v1.0.1, which doesn't make use of atomic forwarding functionality.
//!    - For Gaia 8, the sender will receive the refunded tokens. This is due to Gaia 8 making
//!      use of the Forward Middleware Middleware Module v3 which includes atomic
//!      forwarding functionality.
```

seanchen1991 (approved):

tools/integration-test/src/tests/forward/forward_transfer.rs
ljoss17: This is much better, thanks! I'll apply this change and merge this PR, thank you for the review!

PR #2853, opened 2022-11-10
Represent the `TrustThreshold` as a `num_rational::Ratio`

romac:
Instead of converting to a `Ratio` for equality and ordering comparisons, let's just represent the threshold as a `Ratio` directly.


romac:
Let's see what the CI thinks

seanchen1991:
Does this dis-allow the possibility of `TrustThreshold` having 0 as the denominator?

romac:
> Does this dis-allow the possibility of `TrustThreshold` having 0 as the denominator?

Good catch, fix incoming

PR #2860, opened 2022-11-11
Rename `trust_level` to `trust_threshold` across the codebase

romac:


e2e/e2e/client.py
```diff
@@ -61,7 +61,7 @@ class ClientState:
     frozen_height: Optional[Height]
     latest_height: Height
     max_clock_drift: Duration
-    trust_level: TrustLevel
+    trust_threshold: TrustLevel
```
seanchen1991, line 64: Maybe because this is a Python file, but where is this `TrustLevel` coming from if it's being renamed everywhere else?

romac: Right I forgot to rename the Python type, will do!

seanchen1991 (approved):

PR #2862, opened 2022-11-14
Represent 0/0 trust threshold as `None`

romac:


romac:
I am still not quite sure that this is an improvement over the current version. What do you think?

seanchen1991:
I think an `Option` works better at least from a semantic perspective. Seeing a value of `0/0` is always a little bit alarming. I don't think `None` is amazing either though, since a `None` trust threshold value is not empty in the usual sense; it carries a particular meaning in a particular situation. Perhaps in this case it would make sense to define a bespoke enum like:
```
enum TrustThreshold {
    Ratio(Ratio<u64>),
    ClientStateReset,
}
```
Yeah, it's basically an `Option`, but it's a slight semantic improvement, which, in my opinion is the main thing to capture here. What do you think?

seanchen1991:
Closing this as it doesn't provide much of an improvement over the current logic.

Issue #2851, opened 2022-11-10
Support ed25519 keys in KeyRing


kevinji:
## Summary

Add [ed25519](https://ed25519.cr.yp.to/) support to the relayer's KeyRing.

## Problem Definition

I'm looking to add Solana compatibility to the relayer, which requires adding support for ed25519 keys instead of secp256k1 keys. Ideally, KeyRing is generic over the key type as long as it's defined by a trait.

## Proposal

Modify `KeyRing` to instead be `KeyRing<T: Curve>` and define `impl Curve for Secp256k1` and `impl Curve for Ed25519`. Add support for edd25519 using the [curve25519-dalek crate](https://github.com/dalek-cryptography/curve25519-dalek). This might also require `AddressType` to include a new `Solana` address type, and other interface changes that I'm not 100% sure of yet.

## Acceptance Criteria

KeyRing supports both secp256k1 and ed25519 keys.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

romac:
Hi, thanks for bringing this to our attention! Are you currently working on this?

kevinji:
@romac Yep I'm planning to work on this if there's support for making this change.

romac:
Totally! Would be great if you could post a rough draft of the design you have in mind, and then we can greenlight it :)

kevinji:
Sure, here's a summary of what I think the high-level changes would look like:
- Add a type parameter `T` to [`keyring::KeyEntry`](https://github.com/informalsystems/hermes/blob/master/crates/relayer/src/keyring.rs) where `T: Curve`, `Curve` being a new trait we define that encapsulates public and private keys for both secp256k1 and ed25519.
  - Modify `public_key` and `private_key` in `KeyEntry` to refer to types defined in `T`. (De)serialization should remain the same for secp256k1. Possibly also update `account` although it's not as meaningful as the pubkey is just the account address for Solana.
- Implement [`signature::Signer`](https://docs.rs/signature/latest/signature/trait.Signer.html) and [`signature::Verifier`](https://docs.rs/signature/latest/signature/trait.Verifier.html) for both secp256k1 and ed25519, to allow sharing some code.
[...]

PR #2863, opened 2022-11-15
keyring: Clean up secp256k1 signing with other small changes

kevinji:
These changes prep a larger PR for #2851.

## Description

The KeyRing has a few cleanup changes made. In particular, the signing logic has been updated:

- The `k256` library is no longer used; instead, `secp256k1` is used for both Evmos (aka Ethermint) and Cosmos.
- `.sign_ecdsa` is used for both Evmos and Cosmos as in the current implementation for Evmos, the recovery value `v` is just thrown out via `let (_, sig_bytes)`. I'm not super familiar with Evmos but this seems wrong at a glance, so I left a TODO to figure out how to fix this in the future.

This is the most invasive part of the change, so any help with suggesting how we could test that signing works the same before and after would be helpful.

Separately, there are some miscellaneous changes:

- Use `Secp256k1::signing_only()` instead of `Secp256k1::new()` for clarity.
- Move `sign_message` into `KeyEntry` and remove the duplicated functions in `KeyRing` and at the top level.
- Move `from_seed_file` into `KeyEntry`.
- Simplify the `get_address` implementation.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/relayer/src/keyring.rs
+++ b/crates/relayer/src/keyring.rs
@@ -106,11 +106,13 @@ impl KeyEntry {
         Self::from_key_file(key_file, hd_path)
     }
 
-    // NOTE: Cosmos does not keep `v` (recovery code) in the `(r, s, v)` tuple
-    // for secp256k1, while Ethereum (and therefore Evmos) does.
+    // NOTE: Neither Cosmos nor Ethermint keep `v` (recovery code) in the
+    // `(r, s, v)` tuple for secp256k1.
     //
-    // Cosmos implementation: https://github.com/cosmos/cosmos-sdk/blob/main/crypto/keys/secp256k1/secp256k1_cgo.go
-    // Evmos docs: https://docs.evmos.org/modules/evm/04_transactions.html
+    // Cosmos: https://github.com/cosmos/cosmos-sdk/blob/main/crypto/keys/secp256k1/secp256k1_cgo.go
+    // Ethermint:
+    // - https://github.com/evmos/ethermint/blob/main/crypto/ethsecp256k1/ethsecp256k1.go
+    // - informalsystems/hermes#2863.
     pub fn sign_message(
         &self,
         message: &[u8],
@@ -127,7 +129,6 @@ impl KeyEntry {
         // so `unwrap` is safe.
         let message = Message::from_slice(&hashed_message).unwrap();
 
-        // TODO: Call `.sign_ecdsa_recoverable` for Evmos.
         Ok(Secp256k1::signing_only()
             .sign_ecdsa(&message, &self.private_key.private_key)
             .serialize_compact()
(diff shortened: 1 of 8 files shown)


crates/relayer/src/keyring.rs
```diff
@@ -98,6 +99,40 @@ impl KeyEntry {
...
+                keccak256_hash(message).into()
+            }
+            AddressType::Cosmos | AddressType::Ethermint { .. } => Sha256::digest(message),
+        };
+
+        // SAFETY: hashed_message is 32 bytes, as expected in `Message::from_slice`,
+        // so `unwrap` is safe.
+        let message = Message::from_slice(&hashed_message).unwrap();
+
+        // TODO: Call `.sign_ecdsa_recoverable` for Evmos.
```
romac, line 130: Let's address this TODO before merging

romac: @tomtau We would love to hear what you think we should do here?

tomtau: it should be a recoverable signature, see the previous code:
https://github.com/informalsystems/hermes/pull/2863/files#diff-08bcf6496eddb7428a6a3c7894c2ac3e33be5b3767c0a27a1953eda3dbbabe96L391

you need to enable this feature on secp256k1: https://github.com/rust-bitcoin/rust-secp256k1/blob/master/Cargo.toml#LL28

tomtau: if the pubkey ends with `.ethsecp256k1.PubKey`

kevinji: Right I agree the old code calls `sign_ecdsa_recoverable` but the previous line `let (_, sig_bytes) =` throws away the recovery parameter, which is the first element of the tuple.

I [wrote some test code](https://github.com/kevinji/secp256k1-test/blob/main/src/lib.rs) to validate that the remaining bytes of `sign_ecdsa_recoverable` is the same as the bytes from `sign_ecdsa`.

This is why I was asking what should be done for Evmos as I believe the current behavior is already incorrect.

tomtau: Hm, indeed. Unlike Ethereum, Ethermint doesn't seem to recover pubkeys from signatures (but just looks them up for the account), so it should be ok to use `sign_ecdsa`.

yihuang: yeah, ethermint only do recovery for eth tx that's signed by ethereum compatible wallets, but ibc txs go through normal cosmos-sdk ante handler pipelines.

kevinji: Great, I removed the TODO per the conversation here and updated the comment a bit.

romac (approved):

Issue #2776, opened 2022-10-27
Add a `min_gas_price` verification step in the health check process


seanchen1991:
## Summary

Add a `min_gas_price` verification step in the health check process. 

## Problem Definition

We often ask users to provide us with the `min_gas_price` value as part of their `app.toml` when debugging issues. Now that the SDK exposes this value over gRPC in v0.45.10, we can automatically check for this value as part of Hermes' health check process so that there wouldn't be a need to ask the user for this value directly. 

## Proposal

Add a `min_gas_price` verification step in the health check process that checks that the `gas_price` parameter in the Hermes `config.toml` file is set to be greater than or equal to the `min_gas_price` advertised by the chain. 

This will require generating `0.46.4` IBC proto files in order to access the `cosmos/base/node/v1beta1` gRPC endpoint in order to fetch the chain configuration.

## Acceptance Criteria

- [x] Generate `0.46.4` IBC proto files
- [ ] Added a `min_gas_price` verification step to the health check

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

seanchen1991:
> We probably need to update the protos for that to be able to issue the query and then just gracefully deal with the failure in case the chain does not support it

@romac Could you expand a bit on what you meant by this?

PR #2873, opened 2022-11-17
Add a check for `min_gas_price` in the health check process

seanchen1991:
Closes: #2776

## Description

Adds a step to the health check process that verifies that the `gas_price` parameter in Hermes is >= the on-chain `min_gas_price` parameter.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


crates/relayer/src/chain/cosmos.rs
```diff
@@ -1893,9 +1910,17 @@ fn do_health_check(chain: &CosmosSdkChain) -> Result<(), Error> {
         );
     }
 
+    // Check for `min_gas_price` here
+    // ibc_proto::cosmos::staking::v1beta1::query_client::QueryClient::connect(
```
romac, line 1914: Just a note that ICS consumer chains don't have a staking module and that this query may this fail. I'll check if we can get this information some other way for those chains but in any case we should gracefully handle the case where the query fails because the module is not enabled.

romac: We also need to do the same for the unbonding period.

ci/no-std-check/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
```
romac, line 1: Probably a leftover from the merge, we can remove this.

seanchen1991: [f87e34d](https://github.com/informalsystems/hermes/pull/2873/commits/f87e34de7e4336f47b33696660404609c99f9e98)

crates/relayer/src/chain/cosmos.rs
seanchen1991: > we should gracefully handle the case where the query fails because the module is not enabled

Does this implementation of `query_config_params` handle this? I'm not very clear on that.

crates/relayer/src/config.rs
```diff
@@ -50,6 +53,35 @@ impl Display for GasPrice {
     }
 }
 
+impl TryFrom<String> for GasPrice {
+    type Error = ConfigError;
+
+    fn try_from(price: String) -> Result<Self, Self::Error> {
```
adizere, line 59: This seems like a good first implementation of parsing from String -> GasPrice!

For reference, had a quick look through the SDK and could not find a good example that can help guide us in the future. Closes thing I could find was parsing from strings into decimal coins (eg `10uatoms`). This doesn't work for what we need, which is parsing into f64 coins (`0.03uatoms`).

https://github.com/cosmos/cosmos-sdk/blob/main/types/dec_coin.go#L624-L631

---
That being said, I wasn't very comfortable with this code without a sanity check on it, esp. given the note on "Note: `split_once` does _not_ split inclusively". Wrote a simple test:


```rust
    #[test]
    fn gas_price_try_from() {
        let gp_original = GasPrice::new(10.0, "atom".to_owned());

        let gp_raw: String = gp_original.to_string();
        let gp: GasPrice = gp_raw.try_into().expect("could not parse String into GasPrice");

        assert_eq!(gp, gp_original);
    }
```
seems to fail

```
% cargo test --package ibc-relayer --lib -- config::tests::gas_price_try_from --exact --nocapture
   Compiling ibc-relayer v0.20.0 (/Users/adiseredinschi/Hammers/hermes/crates/relayer)
    Finished test [unoptimized + debuginfo] target(s) in 23.06s
     Running unittests src/lib.rs (target/debug/deps/ibc_relayer-bc3285dcfe99ee96)

running 1 test
thread 'config::tests::gas_price_try_from' panicked at 'assertion failed: `(left == right)`
  left: `GasPrice { price: 10.0, denom: "tom" }`,
 right: `GasPrice { price: 10.0, denom: "atom" }`', crates/relayer/src/config.rs:553:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test config::tests::gas_price_try_from ... FAILED
```


See https://github.com/informalsystems/hermes/commit/3915ad2868fe030c7cc443bf4b7642ef9d821934.

adizere (changes requested):

crates/relayer/src/config.rs
adizere: Update: we can use a different solution: b9966a02

adizere:
One outstanding todo:
- [ ] parsing of denoms should be able to handle the case of multiple min-gas-prices, eg: `minimum-gas-price = 0.25token1;0.0001token2`

seanchen1991:
@adizere Do these changes look good to you?

crates/relayer/src/chain/cosmos.rs
```diff
@@ -318,6 +319,68 @@ impl CosmosSdkChain {
...
+            Ok(response) => {
+                let params = response.into_inner();
+
+                Ok(Some(params))
+            }
+            Err(e) => {
+                if is_unimplemented_node_query(e.clone()) {
+                    Ok(None)
+                } else {
+                    Err(e).map_err(Error::grpc_status)
```
romac, line 366:
```suggestion
                    Err(Error::grpc_status(e))
```

crates/relayer/src/chain/cosmos.rs
```diff
@@ -318,6 +319,68 @@ impl CosmosSdkChain {
...
+    ///     - `Ok(Some(..))` if the query was successful.
+    ///     - `Ok(None) in case the query endpoint is not available.
+    ///     - `Err` for any other error.
+    pub fn query_config_params(&self) -> Result<Option<ConfigResponse>, Error> {
+        crate::time!("query_config_params");
+        crate::telemetry!(query, self.id(), "query_config_params");
+
+        // Helper function to diagnose if the node config query is unimplemented
+        // by matching on the error details.
+        fn is_unimplemented_node_query(err_status: tonic::Status) -> bool {
```
romac, line 336: Let's stake the status by ref here:

```suggestion
        fn is_unimplemented_node_query(err_status: &tonic::Status) -> bool {
```

crates/relayer/src/chain/cosmos.rs
```diff
@@ -318,6 +319,68 @@ impl CosmosSdkChain {
...
+        let request = tonic::Request::new(ibc_proto::cosmos::base::node::v1beta1::ConfigRequest {});
+
+        match self.block_on(client.config(request)) {
+            Ok(response) => {
+                let params = response.into_inner();
+
+                Ok(Some(params))
+            }
+            Err(e) => {
+                if is_unimplemented_node_query(e.clone()) {
```
romac, line 363: so that we don't need to clone it here

```suggestion
                if is_unimplemented_node_query(&e) {
```

crates/relayer/src/config.rs
```diff
@@ -50,6 +53,54 @@ impl Display for GasPrice {
     }
 }
 
+impl TryFrom<String> for GasPrice {
```
romac, line 56: It would more idiomatic to implement `std::str::FromStr` instead of `TryFrom<String>`

crates/relayer/src/config.rs
```diff
@@ -50,6 +53,54 @@ impl Display for GasPrice {
...
+    type Error = ConfigError;
+
+    fn try_from(price_in: String) -> Result<Self, Self::Error> {
+        // TODO: We split by `char::is_alphabetic` delimiter.
+        //      More robust parsing methods might be needed.
```
romac, line 61: Sounds good enough to me for the moment 👍

crates/relayer/src/chain/cosmos.rs
```diff
@@ -1912,9 +1981,31 @@ fn do_health_check(chain: &CosmosSdkChain) -> Result<(), Error> {
         );
     }
 
+    let relayer_gas_price = &chain.config.gas_price;
+    let node_min_gas_prices = chain.min_gas_price()?;
+    let mut found_matching_denom = false;
+
+    for price in node_min_gas_prices {
+        match price.partial_cmp(relayer_gas_price) {
+            Some(Ordering::Greater) => return Err(Error::gas_price_too_low(chain_id.clone())),
```
romac, line 1990: I would invert the condition here to make it slightly more consistent with the euro that's returned, eg.

```suggestion
        match relay_gas_price.partial_cmp(price) {
            Some(Ordering::Lesser) => return Err(Error::gas_price_too_low(chain_id.clone())),
```

crates/relayer/src/config.rs
```diff
@@ -50,6 +53,54 @@ impl Display for GasPrice {
...
+                    denom: denom.to_owned(),
+                })
+            }
+
+            None => Err(Error::invalid_gas_price(price_in)),
+        }
+    }
+}
+
+impl PartialOrd for GasPrice {
```
romac, line 83: Let's perhaps add a comment explaining why we don't implement `Ord` here, ie. because only gas prices for the same denom can be ordered relative to one another.

crates/relayer/src/config.rs
```diff
@@ -509,4 +561,43 @@ mod tests {
...
+        assert_eq!(expected, parsed);
+    }
+
+    #[test]
+    fn parse_empty_gas_price() {
+        let empty_price = "";
+        let parsed = parse_gas_prices(empty_price.to_string());
+
+        assert_eq!(parsed, vec![]);
+    }
```
romac, line 602: Great stuff! Let's add a couple tests for when the `gas_price` string contains malformed/invalid data?

romac (changes requested):
Great work! I have a few nitpicks but looks great otherwise 🙌

crates/relayer/src/config.rs
seanchen1991: [8bc1df9](https://github.com/informalsystems/hermes/pull/2873/commits/8bc1df9b1fedeb490ce6e5576f69e7db3762202f)

romac (approved):

PR #2874, opened 2022-11-18
Add support for ed25519 keys

kevinji:
Closes: #2851

## Description

Add a new trait `SigningKeys` to replace `KeyEntry` that is implemented as `Secp256k1Keys` and `Ed25519Keys`. Parameterize the types in `keyring` (`KeyStore` and `KeyRing`) as well as `chain::ChainEndpoint` on `SigningKeys`.

~~NOTE: This PR is rebased on top of #2863 so it will contain the commits from that PR until it is merged.~~

### Overview

This is a fairly large and invasive change, so here's a broad summary of the changes in the PR:
- Create a new trait `SigningKeys` that encapsulates the logic around key generation, key signing, and address encoding. Delete `KeyEntry` as it is replaced by specific `SigningKeys` implementations.
- Implement `SigningKeys` for secp256k1 as `Secp256k1Keys` and ed25519 as `Ed25519Keys`. Most of the implementation for `Secp256k1Keys` comes from `KeyEntry`; while the type has been updated, it maintains backwards compatibility for Deserializing the old `KeyEntry` type.
- Parameterize all `keyring` types on `SigningKeys`.
- Add a new associated type to `ChainEndpoint` called `SigningKeys`, indicating that a chain can only accept one signature type.
- Add `AnySigningKeys`, which is a enum for the various `SigningKeys` variants that contains some functions using `AsAny` to help cast between `dyn SigningKeys` and the concrete types. This is used in places where the key type is not easily parameterized, like `chain::ChainRuntime`.
- For anything under `chain/cosmos`, replace `KeyEntry` with the concrete `Secp256k1Keys`.
- Under `tools/test-framework`, use the concrete `Secp256k1Keys` to avoid a ton of refactoring needed to parameterize it on `SigningKeys`. This should probably be fixed in a future PR if `Ed25519Keys` are to be tested.

### Major points for review

There are a few parts of this PR that probably require extra scrutiny:
- All of `Secp256k1Keys`: This is mostly taken from the old `KeyEntry` but some changes needed to be made so that it could conform to the `SigningKeys`'s interface. In particular, the underlying data type (and its serialization) has changed, although deserializing the old `KeyEntry` is maintained via the `KeysConversion` enum.
- Dynamic typing via `SigningKeys::as_any()` as well as `AnySigningKeys::downcast()`: this is used in `ChainEndpoint::add_key`.

### Questions to reviewers

- For the serialization of `Secp256k1Keys` (previously `KeyEntry`), do we need to keep the `ExtendedPublicKey` and `ExtendedPrivateKey` as opposed to just the underlying `PublicKey` and `SecretKey`? The extended keys just provide the parameters for the mnemonic-based derivation, but they are never used after they're generated.
- Similarly, do `address` and `account` need to be stored at all when they are uniquely derivable from the `PublicKey`?
- Does `KeyFile` need any parameters besides the mnemonic? Seems like `address` and `pubkey` are uniquely derivable from the private key, which is generated from the mnemonic. Also not quite sure when someone would store a `KeyFile` on disk when the actual `KeyRing` is just persisted on disk after a key is added.
- Is there a cleaner design than the use of `AnySigningKeys`? I couldn't figure out a better way to handle this in places like `ChainRuntime`, as types like `ChainRequest` currently aren't parameterized on any chain-specific types. Similarly, in the test framework, a lot of test code seems like it would need to take an extra type parameter for `SigningKeys`.
- Where should tests for verifying things like signing and (de)serialization go?

______

### PR author checklist:
- [X] Linked to GitHub issue.
- [X] Updated code comments and documentation (e.g., `docs/`).

### Reviewer checklist:


romac:
I just merged #2863, so you can now rebase this PR on top of `master` :)

kevinji:
Should be rebased!

Farhad-Shabani:
Thank you, @kevinji, for your brave Job! Liked how you abstracted the issue. 

**Regarding the points you raised:**

> For the serialization of Secp256k1Keys (previously KeyEntry), do we need to keep the ExtendedPublicKey and ExtendedPrivateKey as opposed to just the underlying PublicKey and SecretKey? The extended keys just provide the parameters for the mnemonic-based derivation, but they are never used after they're generated.
> Similarly, do address and account need to be stored at all when they are uniquely derivable from the PublicKey?

I see the `account` field as a sort of metadata, but about the `address` I feel like you. Might be there for possible future usage. I'd appreciate @romac 's final word on this.

> Does KeyFile need any parameters besides the mnemonic? Seems like address and pubkey are uniquely derivable from the private key, which is generated from the mnemonic. Also not quite sure when someone would store a KeyFile on disk when the actual KeyRing is just persisted on disk after a key is added.

Right. It basically shouldn't be needed to keep it on the disk after loading

> Is there a cleaner design than the use of AnySigningKeys? I couldn't figure out a better way to handle this in places like ChainRuntime, as types like ChainRequest currently aren't parameterized on any chain-specific types. Similarly, in the test framework, a lot of test code seems like it would need to take an extra type parameter for SigningKeys.

Considering Key as a chain-specific property, it makes sense to introduce `anySigningKey` based on `ChainType` once the chains' runtime is spawned. Ideally, not needed to have type `SigningKeys` for `ChainEndpoint` trait and also pass the key as generic `S` to `KeyRing`. However, to flow down key types from that level, as you also mentioned, we should provide a kind of full abstraction for the Solana chain. So, I think your approach is consistent with what we already have.

> Where should tests for verifying things like signing and (de)serialization go?

This could be covered in a future PR

### My Questions
- About naming `*Keys` type as plural. Was there any reason behind it? I guess you did so with this in mind the structs are referring to a keypair (public_key and secret_key). Hw, It looks a bit confusing, I recommend keeping `*keys` singular (e.g. `ed25519_key` instead of `ed25519_keys`)
- Didn't catch why we should have `NewKeyEntry`; couldn't just keep `OldKeyEntry` for the conversion?

crates/relayer/src/keyring/signing_keys.rs
```diff
@@ -0,0 +1,43 @@
+use core::any::Any;
+
+use hdpath::StandardHDPath;
+use serde::{de::DeserializeOwned, Serialize};
+
+use super::{errors::Error, KeyFile};
+use crate::config::AddressType;
+
+// /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\
+// WARNING: Changing this struct in backward incompatible way
```
Farhad-Shabani, line 10: Doesn't comply with the trait

crates/relayer/src/keyring/errors.rs
```diff
@@ -21,9 +21,13 @@ define_error! {
             [ DisplayOnly<anyhow::Error> ]
             |_| { "invalid mnemonic" },
 
-        PrivateKey
+        Secp256k1PrivateKey
```
Farhad-Shabani, line 24: Couldn't have these two as one variant, `InvalidPrivatekey` introduce the type as `{key_type: String}`?

kevinji:
Thanks for leaving the comments; I've pushed some changes. Here are my responses:

> I think that can be replaced by mentioned underlying `PublicKey` and `SecretKey`

Sounds good, I updated the `Secp256k1KeyPair` type to store the underlying keys instead.

> About naming `*Keys` type as plural. Was there any reason behind it? I guess you did so with this in mind the structs are referring to a keypair (public_key and secret_key). Hw, It looks a bit confusing, I recommend keeping `*keys` singular (e.g. `ed25519_key` instead of `ed25519_keys`)

Makes sense, I decided to rename it `key_pair` so it's singular but still more clear than a plural `keys`.

> Didn't catch why we should have `NewKeyEntry`; couldn't just keep `OldKeyEntry` for the conversion?

So the reason I had to duplicate `NewKeyEntry` from `Secp256k1KeyPair` is that I wanted to avoid a circular Serde deserialization. Basically I wanted Old/NewKeyEntry to have standard Serde derives, and then have the real type `Secp256k1KeyPair` rely on `KeyConversion` using Serde's `try_from` feature.

If there's some way to remove the duplication via a macro or something would be happy to switch this out.

crates/relayer-cli/src/commands/keys/add.rs
```diff
@@ -187,44 +195,63 @@ pub fn add_key(
...
-) -> eyre::Result<KeyEntry> {
+) -> eyre::Result<AnySigningKeyPair> {
     let mnemonic_content =
         fs::read_to_string(mnemonic).map_err(|_| eyre!("error reading the mnemonic file"))?;
 
-    let mut keyring = KeyRing::new(Store::Test, &config.account_prefix, &config.id)?;
+    let key_pair = match config.r#type {
+        ChainType::CosmosSdk => {
+            let mut keyring =
+                KeyRing::<Secp256k1KeyPair>::new(Store::Test, &config.account_prefix, &config.id)?;
```
romac, line 232: Let's introduce a constructor here rather have to specify the type? eg.

```suggestion
                KeyRing::new_secp256k1(Store::Test, &config.account_prefix, &config.id)?;
```

crates/relayer-cli/src/commands/keys/delete.rs
```diff
@@ -112,16 +113,26 @@ impl Runnable for KeysDeleteCmd {
 }
 
 pub fn delete_key(config: &ChainConfig, key_name: &str) -> eyre::Result<()> {
-    let mut keyring = KeyRing::new(Store::Test, &config.account_prefix, &config.id)?;
-    keyring.remove_key(key_name)?;
+    match config.r#type {
+        ChainType::CosmosSdk => {
+            let mut keyring =
+                KeyRing::<Secp256k1KeyPair>::new(Store::Test, &config.account_prefix, &config.id)?;
```
romac, line 119: Same

crates/relayer-cli/src/commands/keys/delete.rs
```diff
@@ -112,16 +113,26 @@ impl Runnable for KeysDeleteCmd {
...
 
 pub fn delete_all_keys(config: &ChainConfig) -> eyre::Result<()> {
-    let mut keyring = KeyRing::new(Store::Test, &config.account_prefix, &config.id)?;
-    let keys = keyring.keys()?;
-    for key in keys {
-        keyring.remove_key(&key.0)?;
+    match config.r#type {
+        ChainType::CosmosSdk => {
+            let mut keyring =
+                KeyRing::<Secp256k1KeyPair>::new(Store::Test, &config.account_prefix, &config.id)?;
```
romac, line 130: Same

crates/relayer/src/error.rs
```diff
@@ -545,6 +545,12 @@ define_error! {
             |e| {
                 format_args!("message with length {} is too large for a transaction", e.len)
             },
+
+        InvalidKeyType
+            { key: AnySigningKeyPair }
+            |_| {
+                "Invalid key type for the current chain"
```
romac, line 552: Can we display the invalid type in the error message?

crates/relayer/src/keyring/ed25519_key_pair.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+impl Clone for Ed25519KeyPair {
+    fn clone(&self) -> Self {
+        let Self {
+            keypair: Keypair { secret, public },
+        } = self;
+        Self {
+            keypair: Keypair {
+                secret: SecretKey::from_bytes(secret.as_bytes()).unwrap(),
```
romac, line 57: Let's add a comment explaining why this is needed, ie. because `SecretKey` does not implement `Clone`.

crates/relayer/src/keyring/ed25519_key_pair.rs
```diff
@@ -0,0 +1,120 @@
...
+    fn from_mnemonic(
+        mnemonic: &str,
+        hd_path: &StandardHDPath,
+        _address_type: &AddressType,
+        _account_prefix: &str,
+    ) -> Result<Self, Error> {
+        Self::from_mnemonic_internal(mnemonic, hd_path)
+    }
+
+    // Solana address: base58(pubkey)
```
romac, line 108: What if chains other than Solana use a different encoding? It's not an issue since we don't even support Solana itself at the moment, much less other chains which may use Ed25519 keys, but it might worth thinking about we would solve this eventually. Just some food for thought, no need to address it in this PR.

crates/relayer/src/keyring/errors.rs
```diff
@@ -112,5 +111,29 @@ define_error! {
...
                 format!("invalid HD path: {0}", e.path)
             },
+
+        AddressTypeNotFound
+            {
+                address: Vec<u8>,
+                public_key: Vec<u8>,
+            }
+            |_| {
+                "No address type found that matches the public key"
```
romac, line 121: Let's show the address and the public key in the error message? Otherwise the fields won't be used at all.

crates/relayer/src/keyring/errors.rs
```diff
@@ -112,5 +111,29 @@ define_error! {
...
+            },
+
+        InvalidAddressLength
+            {
+                address: Vec<u8>,
+                expected_length: usize,
+            }
+            |e| {
+                format!("Address length did not match expected length {}", e.expected_length)
```
romac, line 130: Let's show the address or at the very least its length in the error message

crates/relayer/src/keyring/ed25519_key_pair.rs
romac: One way to go about this is similar to what we are doing for Cosmos vs Ethermint addresses, where we have a `Secp256k1AddressType` enum that we use to derive the account. Might be worth adding something similar for Ed25519 keys for future-proofing.

tools/test-framework/src/types/wallet.rs
```diff
@@ -33,9 +33,10 @@ pub struct Wallet {
     /// The address for receiving tokens for this wallet.
     pub address: WalletAddress,
 
-    /// The wallet key information in the form of [`KeyEntry`] that
-    /// is used by the relayer.
-    pub key: KeyEntry,
+    // TODO: Parameterize this type on `SigningKeyPair`
+    /// The wallet key information in the form of [`SigningKeyPair`]
+    /// that is used by the relayer.
+    pub key: Secp256k1KeyPair,
```
romac, line 39: Not that we should eventually do this if we want to test against Solana in the integration tests, but this can be left for a follow-up PR or when full support for Solana is implemented.

romac (changes requested):
Amazing work here, thank you so much!

This all looks great, aside from a few minor nitpicks left inline.

💐

crates/relayer-cli/src/commands/keys/add.rs
romac: And one for the ed25519 keyring

kevinji: Added these functions.

crates/relayer/src/keyring/signing_keys.rs
kevinji: Moved to `Secp256k1KeyPair` per a different comment.

crates/relayer/src/keyring/ed25519_key_pair.rs
kevinji: Added an `Ed25519AddressType`.

tools/test-framework/src/types/wallet.rs
kevinji: Sounds good, going to resolve this for now.

kevinji:
Just pushed some new changes addressing the review comments.

crates/relayer/src/keyring/ed25519_key_pair.rs
```diff
@@ -42,50 +42,82 @@ fn standard_path_to_derivation_path(path: &StandardHDPath) -> DerivationPath {
...
     fn from_key_file(key_file: KeyFile, hd_path: &StandardHDPath) -> Result<Self, Error> {
-        let key_pair = Self::from_mnemonic_internal(&key_file.mnemonic, hd_path)?;
-
-        let public_key_from_file = PublicKey::from_bytes(
-            &bs58::decode(key_file.pubkey)
-                .into_vec()
-                .map_err(Error::bs58_decode)?,
-        )
-        .map_err(Error::invalid_public_key)?;
+        // TODO: Derive this from something in `key_file`
```
romac, line 109: What do you think about adding this to `KeyPairV2`, so that we don't have to introduce a `KeyPairV3` for this change in the future?

crates/relayer/src/keyring/secp256k1_key_pair.rs
```diff
@@ -133,36 +142,39 @@ pub struct Secp256k1KeyPair {
...
 }
 
+// Note: Since this uses Serde's untagged enums, the serialized formats between
+// versions must be incompatible with each other.
 #[derive(Debug, Deserialize)]
 #[serde(untagged)]
-enum KeysConversion {
-    Old(OldKeyEntry),
-    New(NewKeyEntry),
+enum VersionedKeyPair {
```
romac, line 167: 👍

romac:
Awesome, thanks! I left one more question for you, but otherwise it looks great :)

crates/relayer/src/keyring/ed25519_key_pair.rs
kevinji: Hmm this is in the Ed25519 side (Solana) and I think I've already added this to the `KeyPair` type so the only thing that would need to be changed here is updating what we're reading from `KeyFile`. So we should be able to avoid adding a `KeyPairV2` here.

kevinji:
Merged `master` into this branch and fixed the conflicts.

romac (approved):

PR #2912, opened 2022-11-30
Deny Clippy warnings and properly cancel previous runs of same workflow when commits are added

romac:
This setting cancels the previous runs of this workflow when a new commit is added to the PR, branch or tag:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```


.github/workflows/integration.yaml
```diff
@@ -32,6 +32,10 @@ env:
   CARGO_NET_RETRY: 10
   RUSTUP_MAX_RETRIES: 10
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
```
mzabaluev, line 38: There is more going on here than just denying clippy warnings.

Can you describe what this change does?

romac: Updated the code + PR description.

.github/workflows/cargo-doc.yaml
```diff
@@ -6,6 +6,11 @@ on:
       - master
   pull_request: {}
 
+# Cancel previous runs of this workflow when a new commit is added to the PR, branch or tag
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
mzabaluev, line 11: So if not triggered on a `pull_request`, the expression `github.event.pull_request.number || github.ref` should fall back to the ref name and not just crash, right?

mzabaluev (approved):
Looks good to me.

.github/workflows/cargo-doc.yaml
romac: Yes exactly

Issue #2844, opened 2022-11-08
Allow sub-contexts to have different `Error` types with functions to inject errors into `Self::Error`


soareschen:
## Summary

Currently, context traits containing sub-contexts, such as `HasRuntime` and `HasRelayTypes`, require the sub-contexts such as `Self::Runtime`, `Self::SrcChain`, and `Self::DstChain` to have the same `Error` associated type as `Self::Error`. This was done initially for easier understanding of the code.

However, it can be restrictive to require all contexts to share the same concrete error type, especially if the concrete contexts are defined in different crates. As an example, to workaround having a common `Error` type, the concrete runtime context `TokioRuntimeContext` is parameterized with a generic `Error` type, so that it can set its `Error` associated type to be the same as the `Error` type in the concrete chain or relay contexts.

A better design would be to allow each contexts to have their own abstract `Error` types, and require injection functions to inject errors into the parent context. For example, the `HasRuntime` trait can be redefined as follows:

```rust
pub trait HasRuntime: InjectError<<Self::Runtime as HasError>::Error> {
    type Runtime: HasError;

    fn runtime(&self) -> &Self::Runtime;
}
```

When the parent context calls any methods from the sub-context, they can then use `Self::inject_error` to inject error types like `Self::Runtime::Error` into `Self::Error`.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #2816, opened 2022-11-04
Use GAT to define generic MPSC channel context


soareschen:
## Summary

With Rust 1.65 released, it is now possible to use GAT to define generic channel types for our runtime context. This will be very helpful for components that make use of channels, such as the batched message sender component.

We will define a `HasChannelTypes` trait with something as follows:

```rust
trait HasChannelTypes {
  type Sender<T>;
  type Receiver<T>;

  fn send(sender: &Sender<T>, value: T);

  async fn receive(receiver: &Receiver<T>) -> T;
}
```

And then modify `BatchContext` to make use of `HasChannelTypes` to construct the message channels.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

soareschen:
Closed by #2935.

PR #2932, opened 2022-12-08
Allow sub-contexts to have different `Error` types with functions to inject errors into `Self::Error`

soareschen:
Closes: #2844

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


crates/relayer-framework/src/full/one_for_all/impls/batch.rs
```diff
@@ -91,9 +91,11 @@ where
 impl<Relay> HasBatchContext<SourceTarget> for OfaRelayWrapper<Relay>
 where
     Relay: OfaBaseRelay,
-    Relay::SrcChain: OfaFullChain,
+    // TODO: do not require the chain error to be the same as relay error.
+    // this can be fixed in #2816.
+    Relay::SrcChain: OfaFullChain<Error = Relay::Error>,
```
soareschen, line 96: This requires refactoring on the batch context, but the batch context can be removed through #2816. So we will leave it to #2816 to fix this final limitation.

crates/relayer-framework/src/base/runtime/traits/runtime.rs
```diff
@@ -1,7 +1,9 @@
...
+use crate::base::core::traits::error::HasErrorType;
 
-pub trait HasRuntime: HasError {
-    type Runtime: HasError<Error = Self::Error>;
+pub trait HasRuntime: HasErrorType {
+    type Runtime: HasErrorType;
 
     fn runtime(&self) -> &Self::Runtime;
+
+    fn runtime_error(e: <Self::Runtime as HasErrorType>::Error) -> Self::Error;
```
soareschen, line 8: I originally wanted `HasRuntime` to has `InjectError<<Self::Runtime as HasError>::Error>` as the supertrait. But I realized that it would require the implementor to have a generic implementation of `InjectError`, which is not allowed by Rust due to overlapping implementations.

So instead we would have explicit conversion methods for the sub-error types.

crates/relayer-runtime/src/tokio/context.rs
```diff
@@ -11,39 +9,26 @@ use tokio::sync::{mpsc, oneshot};
...
-    fn clone(&self) -> Self {
-        Self::new(self.runtime.clone())
+        Self { runtime }
     }
 }
 
 #[async_trait]
-impl<Error> OfaRuntime for TokioRuntimeContext<Error>
-where
-    Error: From<TokioError> + Debug + Async,
```
ljoss17, line 44: Since the constraint `From<TokioError>` is removed, should the implementation, https://github.com/informalsystems/hermes/blob/soares/explicit-error-injection/crates/relayer-cosmos/src/base/error.rs#L59, be kept or not ?

ljoss17 (approved):
Seems great! Just curious about your opinion on the implementation of `From`

crates/relayer-runtime/src/tokio/context.rs
soareschen: Yup, it can be removed if there is no compile error. Thanks for pointing out!

Issue #2737, opened 2022-10-14
Add preliminary docs for the integration test framework


seanchen1991:
## Summary

Document the integration test framework's API in order to provide context and guidance on how to use it to write tests. 

## Acceptance Criteria

- [x] #2933 
- [x] #3003 
____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

adizere:
Will the integration test framework become superseded due to the new Hermes (context-generic-based) architecture? If so, not sure we should beef up the docs at this point.


Copying @soareschen and @ljoss17.

soareschen:
We could write new tests for the new relayer with more abstract tests, and probably even do model checking using Kani. But the current priority is on the relayer implementation, so new testing infrastructure will likely take at least half year to complete, and existing code will still need to be tested with the current test framework. So it might still be worthwhile to add some documentation and have more dev sessions on it.

adizere:
Let's kick off this? At least a first iteration.

PR #2933, opened 2022-12-09
Clean up documentation for example integration test and add walkthrough of memo.rs

seanchen1991:
Part of #2737.

## Description

As a first stab at improving the documentation of the integration test framework, this PR moves the documentation of the simple example integration test that previously lived at `tools/integration-test/src/tests/example.rs` into `tools/test-framework/src/docs/`. Going forward, we'll centralize the documentation of the framework in this directory. It also adds a walkthrough of the [memo.rs](https://github.com/informalsystems/hermes/blob/master/tools/integration-test/src/tests/memo.rs) test.

Follow up PRs will add more docs walking through more complex integration tests in order to illustrate different use cases and recipes for using the test framework.
______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


adizere:
> Follow up PRs will add more docs walking through more complex integration tests in order to illustrate different use cases and recipes for using the test framework.

Should we open a follow-up issue to capture these needs? Then the present PR can close #2737.

seanchen1991:
> Should we open a follow-up issue to capture these needs

Added #2937 to capture the additional work.

tools/test-framework/src/docs/walkthroughs/simple.rs
```diff
@@ -0,0 +1,65 @@
...
+//!     ) -> Result<(), Error> {
+//!         suspend()
+//!     }
+//! }
+//! ```
+//!
+//! The `example_test` function is the top-most function that drives the test itself. The top-
+//! most test function typically does nothing more than to specify what kind of test we want
+//! to run. In the case of this example, we're running a
+//! [`BinaryChannelTest`](ibc_test_framework::framework::binary::channel::BinaryChannelTest),
```
soareschen, line 34: There are warnings in broken cargo doc links. I think we have to use `crate::` here, since we are referring to the same crate.

tools/test-framework/src/docs/walkthroughs/simple.rs
```diff
@@ -0,0 +1,65 @@
...
+//!
+//! The main logic of the test is implemented in the `run` function of the `BinaryChannelTest`
+//! trait. This trait is implemented for our empty test struct since we're choosing to run a
+//! test between two chains connected via IBC channels. If we had instead opted to run a binary
+//! _chain_ test using the `run_binary_chain_test`, then we would instead implement the
+//! [`BinaryChainTest`](ibc_test_framework::framework::binary::chain::BinaryChannelTest)
+//! trait for our empty test struct.
+//!
+//! The `run` function's parameters are:
+//! 1. `config`: for accessing any relayer configuration values during the course of the test
```
soareschen, line 55: This is the test config, not the relayer config.

soareschen (changes requested):
There are a couple of cargo doc warnings to fix.

soareschen (approved):

Issue #2876, opened 2022-11-21
Default to 2/3 instead of 1/3 for IBC client security parameters


adizere:
## Summary

Replace the 1/3 default with a 2/3 default.

Context: see the recommendation following the work by Tendermint team here: https://github.com/tendermint/tendermint/pull/9420

## Acceptance Criteria

 - [x] replace default trust threshold from 1/3 to 2/3 for IBC client security params
 - [ ] test relaying in a multi-network testate setup and confirm no impact is observed when validator sets continuously overlap
     - we should see no changes in the frequency of client updates or in the cost of client update messages
     - note: we already did some interesting tests in production with 1/1 trust threshold, refer to https://github.com/informalsystems/hermes/pull/1058

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

crodriguezvega:
@adizere Since the trust threshold is [configurable](https://github.com/informalsystems/hermes/blob/master/config.toml#L241), will there be some kind of validation to check that the threshold is >= 2/3? Otherwise an operator might choose to use a lower threshold anyway...

adizere:
> @adizere Since the trust threshold is [configurable](https://github.com/informalsystems/hermes/blob/master/config.toml#L241), will there be some kind of validation to check that the threshold is >= 2/3? Otherwise an operator might choose to use a lower threshold anyway...

The value is already validated to be `> 0 and <=1`, that is, any value chose by an operator with Hermes must be a valid fraction larger than 0 and smaller than 1. But we don't plan to perform validation that it _should always be >= 2/3_. I think it's reasonable to allow operators to select different other values, eg 1/3 if they wish so. Do you see concerns with this?

romac:
We currently [only relay on clients](https://github.com/informalsystems/hermes/blob/2960effefdf3cf0b715215b1e19263d521d8a0f7/crates/relayer/src/supervisor/client_state_filter.rs#L182) with a trust threshold `>= 1/3` and `<= 2/3`. Should we relax the upper bound to `1/1`?

adizere:
> Should we relax the upper bound to `1/1`?

Not sure why we'd need that in the context of this issues. But I don't see problems with this change either!

romac:
[...]

PR #2988, opened 2023-01-06
Switch the default trust threshold for new clients to 2/3

romac:
Closes: #2876

## Description

Note that the default trust threshold used for creating a client with `create client` when no trust threshold is specified on the command line is taken from the chain config's `trust_threshold` setting. We should therefore advise operators to update their configs with the new recommended value as otherwise the changes below won't have any effect outside of tests.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


ancazamfir (approved):
lgtm, thanks @romac !

romac:
I wrote a little tool to fetch the latest 1000 blocks with a step of 1, 5, and 10 and compute the voting power overlap between all adjacent blocks. The results are [here](https://gist.github.com/romac/bb9a14e6d880af82dcfb31c7f5f7755f) (there are multiple files each with a different step).

I did not see an overlap less than 2/3 so I would assume that the impact of this change in terms of client updates will be low if not null.

romac:
Ran some tests on Chihuahua, Evmos, Injective, and Secret and did not find a validator set change greater than 1/3 over any combination of the past 1000 blocks.

Let's merge this :)

Issue #2999, opened 2023-01-09
github actions: rust.yml conflicts with integration.yaml in concurrency


ImJeremyHe:
## Summary of Bug

[rust.yml](https://github.com/informalsystems/hermes/blob/master/.github/workflows/rust.yml#L33) and [integration.yaml](https://github.com/informalsystems/hermes/blob/master/.github/workflows/integration.yaml#L37) share the same name(`Rust`) and they are in the same concurrency group (`${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}`). This makes these 2 actions can not run at the same time.

## Version

main:101de43f24983728b1d7cb5d35b05d0486115b88

## Acceptance Criteria

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #3000, opened 2023-01-10
Fix concurrency conflict between Rust and Integration Tests jobs

romac:
Closes: #2999


seanchen1991 (approved):
Thanks for the quick fix @romac!

Issue #2971, opened 2022-12-20
Enable support for multiple versions of Tendermint and the Cosmos SDK


seanchen1991:
## Background
This issue keeps track of work that contributes to the effort of enabling backwards-compatibility for older Tendermint versions in Hermes. The aim is to ease the burden needing to keep Hermes in lock step with Tendermint and SDK version upgrades.

## Tasks
- [ ] #3017 
- [x] #3022

Issue #2276, opened 2022-06-08
Support Cosmos SDK 0.46 and 0.47 chains side by side


mzabaluev:
## Summary

The relayer should work with chain nodes based on both Cosmos SDK 0.46 / Tendermint 0.34 and Cosmos SDK 0.47 / CometBFT 0.37.

## Problem Definition

While the Cosmos ecosystem transitions to Cosmos SDK 0.47, the relayer should be expected to connect chains working on both old and new versions of the protocols and specifications. There are breaking changes in at least CometBFT 0.37, so the issue is more complex than just updating the libraries to the latest protocol and adding a few backward compatibility adaptations.

## Proposal

- [x] Update `ibc-proto` code generation to CometBFT 0.37, Cosmos SDK 0.47 and a compatible revision of ibc-go definitions TBD. Adapt the code using the protocol structures (#2213).
- [x] Extend tendermint-rs to interoperate with tendermint RPC 0.34 as a static API variant (that is, no need to support both versions with a single client type).
- [x] Implement support for both old and new (i.e. pre- and post-break) version combinations of Tendermint RPC, Cosmos-SDK and IBC with dynamic selection in ibc and ibc-relayer (the same chain endpoint should be able to dynamically discover the versions used by the node and adapt to the protocol).

## Acceptance Criteria

Hermes is able to connect and relay between the following node versions, as confirmed by tests:

- [ ] A chain based on Cosmos SDK 0.46
- [ ] A chain based on Cosmos SDK 0.47

At least one test run should exercise relaying between SDK 0.46 and 0.47 nodes.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
Is this issue still relevant in its current form?

webmaster128:
Right before the release, Cosmos SDK 0.46 switched back from Tendermint 0.35 to 0.34. So at least you only need to worry about one Tendermint version for the two SDK versions.

romac:
Right! I've updated the issue to reflect that, switch from 0.45->0.46 to 0.46->0.47, and update on the progress.

PR #3017, opened 2023-01-18
Update to multi-version tendermint-rs

mzabaluev:
Closes: #2971
Part of: #2276

Requires https://github.com/informalsystems/tendermint-rs/pull/1193

## Description

Update to the changes adding side-by-side support for 0.34 and 0.37 versions of Tendermint into tendermint-rs crates.

_____

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


crates/relayer/src/event/monitor.rs
```diff
@@ -176,14 +176,19 @@ impl EventMonitor {
     pub fn new(
         chain_id: ChainId,
         node_addr: Url,
+        rpc_compat: CompatMode,
```
mzabaluev, line 179: @romac There's also [version discovery code](
https://github.com/informalsystems/hermes/blob/0e81ee6bdaccaf44c407151d58589ab055c95657/crates/relayer/src/chain/cosmos.rs#L718) when instantiating the inner RPC client. That can be removed now and the HTTP client initialized with the previously discovered compatibility mode. On the other hand, it would be still better to reuse that HTTP client's connection to discover the version for the websocket client, rather than connecting twice in a row.

crates/relayer-types/src/applications/ics31_icq/response.rs
```diff
@@ -7,7 +7,7 @@ use prost::Message;
 use std::prelude::v1::*;
 use std::vec;
 use tendermint::merkle::proof::ProofOps as TendermintProofOps;
-use tendermint_proto::crypto::{ProofOp, ProofOps};
+use tendermint_proto::v0_34::crypto::{ProofOp, ProofOps};
```
ancazamfir, line 10: Could you check if this is needed? Were there changes in v0.37?

mzabaluev: No changes as far as I can tell. This change has been reverted now, and the import path is currently pointing at the `v0_37` protobuf types.

romac (approved):

PR #3022, opened 2023-01-20
Test multi-version support against pre-release of IBC-go v7.0.0 which is based on Tendermint 0.37

romac:


romac:
@mzabaluev Let's merge this in your PR so that we can test it on CI.

mzabaluev:
> @mzabaluev Let's merge this in your PR so that we can test it on CI as you make updates to it.

I will pull, thank you.

romac:
Why not just merge it?

PR #3042, opened 2023-01-27
Fix clippy warnings as of Rust 1.67

romac:


ljoss17 (approved):
Should the `clippy.toml` file be updated as well or not necessarily ?

romac:
Oh I forgot we had that, neat. Prevents CI breakage on new Rust versions :) We should only update it when we bump the MSRV, so no need to do that for now.

Issue #2967, opened 2022-12-20
ICS31 Cross-Chain Queries integration tests


ljoss17:
## Summary

The ICS31 implemented in https://github.com/informalsystems/hermes/pull/2915 is currently only tested manually with https://github.com/informalsystems/hermes/wiki/Testing-examples#ics31-cross-chain-queries.

There might also be an compatibility issue due to proto definitions. The suspicion comes from the fact that Stride and Persistence use different proto files, https://github.com/Stride-Labs/stride/blob/main/proto/stride/interchainquery/v1/messages.proto and https://github.com/persistenceOne/persistence-sdk/blob/master/proto/persistence/interchainquery/v1beta1/messages.proto.
The integration tests should be able to easily confirm or discard this suspicion.

## Problem Definition

In order to test ICS31 with Stride, an admin account is required: https://github.com/Stride-Labs/stride/blob/main/utils/admins.go
So the two solutions to my knowledge are:

1. Adding the necessary Admin account and recompile Stride binary
2. Using Stride's dockernet (which is the solution used in the Wiki)

In order to have some running integration tests in Hermes, a solution which allows to trigger Cross-Chain Queries without modifying Stride's binary or using the dockernet is required.

## Proposal

Adding integration tests for ICS31

## Acceptance Criteria

- [ ] Successful integration tests for ICS31

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #3061, opened 2023-02-06
ICS31 integration tests

ljoss17:
Closes: #2967

## Description

This PR adds integration test for ICS31 Cross Chain Queries, using Stride <-> Gaia chains.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -267,9 +267,9 @@ jobs:
       fail-fast: false
       matrix:
         chain:
-          - package: gaia8
-            command: gaiad
-            account_prefix: cosmos
+          - package: .#gaia8 .#stride-no-admin
+            command: gaiad,strided
+            account_prefix: cosmos,stride
     steps:
       - uses: actions/checkout@v2
       - uses: cachix/install-nix-action@v15
@@ -294,10 +294,10 @@ jobs:
           RUST_LOG: info
           RUST_BACKTRACE: 1
           NO_COLOR_LOG: 1
-          CHAIN_COMMAND_PATHS: ${{ matrix.chain.command }},strided
-          ACCOUNT_PREFIXES: ${{ matrix.chain.account_prefix }},stride
+          CHAIN_COMMAND_PATHS: ${{ matrix.chain.command }}
+          ACCOUNT_PREFIXES: ${{ matrix.chain.account_prefix }}
         run: |
-          nix shell .#${{ matrix.chain.package }} .#stride-no-admin -c cargo \
+          nix shell ${{ matrix.chain.package }} -c cargo \
             test -p ibc-integration-test --features ics31 --no-fail-fast -- \
             --nocapture --test-threads=1 ics31::
 
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -261,6 +261,46 @@ jobs:
             test -p ibc-integration-test --features forward-packet --no-fail-fast -- \
             --nocapture --test-threads=1 forward::
 
+  ics31:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        chain:
+          - package: gaia8
+            command: gaiad
+            account_prefix: cosmos
+    steps:
+      - uses: actions/checkout@v2
+      - uses: cachix/install-nix-action@v15
+        with:
+          install_url: https://nixos-nix-install-tests.cachix.org/serve/vij683ly7sl95nnhb67bdjjfabclr85m/install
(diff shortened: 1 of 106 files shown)


.github/workflows/integration.yaml
```diff
@@ -261,6 +261,46 @@ jobs:
             test -p ibc-integration-test --features forward-packet --no-fail-fast -- \
             --nocapture --test-threads=1 forward::
 
+  ics31:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        chain:
```
soareschen, line 269: I think you need to add the new `stride-no-admin` package here.

ljoss17: Testing requires to have one Stride chain and one Gaia chain, if the `stride-no-admin` is added to the `matrix` wouldn't result in having one test running 2 Gaia chains and another test running two Stride chains ?
The idea here was to have the possibility to test multiple Gaia versions if we want/need to, by simply adding them to the `matrix`. Resulting in every test being:

* `Chain from matrix` <-> `Stride chain`

What do you think ?

soareschen: Hmm, I think if you want a CI matrix, then putting stride in it also result in all the combinations with stride executed.

tools/integration-test/src/tests/ics31.rs
```diff
@@ -0,0 +1,159 @@
...
+        chains.node_b.chain_driver().assert_eventual_wallet_amount(
+            &wallet_b.address(),
+            &denom_b.with_amount(a_to_b_amount).as_ref(),
+        )?;
+
+        info!("Registering host-zone");
+
+        simple_exec(
+            "stride",
+            "strided",
```
soareschen, line 117: Does this command always have to be executed with `stride`? Might be better to use the chain driver's command path to execute the command.

tools/test-framework/src/chain/cli/query.rs
```diff
@@ -96,3 +96,60 @@ fn yaml_to_json_value(value: yaml::Value) -> Result<json::Value, Error> {
 
     Ok(parsed)
 }
+
+pub fn query_cross_chain_query_list(
```
soareschen, line 100: This looks like two functionalities mixed into one function. There should be one query function to return the result from `simple_exec`, and another assert function that calls the function, and asserts that the result is as expected.

tools/test-framework/src/chain/driver.rs
```diff
@@ -188,6 +198,15 @@ impl ChainDriver {
         )
     }
 
+    pub fn query_cross_chain_query_list(&self, pending_queries: bool) -> Result<(), Error> {
```
soareschen, line 201: May be better to introduce this as an extension method in `src/chain/ext`.

tools/test-framework/src/chain/driver.rs
```diff
@@ -219,4 +238,31 @@ impl ChainDriver {
 
         Ok(())
     }
+
+    /**
+       Depending on the value of `pending_queries`, assert that a
+       cross chain is pending or that no cross chain queries are
+       pending..
+    */
+    pub fn assert_eventual_cross_chain_query_relayed(
```
soareschen, line 247: May be better to introduce this as an extension method in `src/chain/ext`.

tools/test-framework/src/chain/tagged.rs
```diff
@@ -62,6 +62,9 @@ pub trait TaggedChainDriverExt<Chain> {
         token: &TaggedTokenRef<Chain>,
     ) -> Result<(), Error>;
 
+    fn assert_eventual_cross_chain_query_relayed(&self, pending_queries: bool)
```
soareschen, line 65: May be better to introduce this as an extension method in `src/chain/ext`.

soareschen (changes requested):

tools/integration-test/src/tests/ics31.rs
ljoss17: `gaiad` doesnt have the CLI `<command_path> tx stakeibc register-host-zone`.
I could create move this to a new subdirectory in the `test-framework`:

`test-framework/src/chain/cli/chain-specific/stride/host-zone.rs`

What do you think ?

soareschen: The test itself should not care whether the chain executable supports a given command. If the command is not supported, the test can just fail. We only need to configure the CI matrix such that the supported chain executables are given in the `CHAIN_COMMAND_PATH`.

soareschen (approved):

PR #3076, opened 2023-02-14
Fix `API Docs` CI job

ljoss17:
Closes: #XXX

## Description

The `API Docs` was running into `Error: could not document 'clap'` when using `rustc 1.69.0-nightly (065852def 2023-02-13)`.
~This PR updates the job in order to use the `stable` toolchain.~
`nightly` is required for this job, so instead the version used has been set to a previous `nightly` version, which doesn't have the error with `clap`.
______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


seanchen1991 (approved):

seanchen1991:
Thanks for troubleshooting this Luca. This seems like a temporary fix; we should probably get input from Romain on a more robust solution for this issue when he gets back.

romac:
I think we should strive to use the same Rust version that docs.rs uses to ensure that we can always publish our crates and have their docs build successfully on docs.rs.

In general, I would expect these kind of issues to be fixed promptly in the Rust nightlies, hence we can probably safely ignore those when the docs build breaks because of a dependency.

Also, it looks like docs.rs builds the docs with `cargo doc --no-deps` so we should probably update the job to avoid building docs for dependencies, which would have also resolved the issue at hand here.

Additionally, we could ensure we use the same version as docs.rs by scraping the nightly version used there from https://docs.rs/about/builds and use that in our CI job.

seanchen1991:
Opened #3100 to capture the follow up work on this

Issue #3112, opened 2023-02-21
Hermes cannot get the unbonding period from consumer chains


ancazamfir:
## Summary of Bug
This was reported by a hermes user on a Telegram channel:

<-- begin -->

..We are having some issues with hermes 1.2.0 to set connection between consumer chain and ordinary chain without ICS support (like band protocol)
We found that it is not possible to use Hermes 1.2.0 with consumer chain and other full chain like Band Protocol. It requires hermes to provide
`unbonding_period`
but this parameter is not available because consumer chain doesn't contain
`staking`
module. So we were required to use patched version of hermes (this commit https://github.com/informalsystems/hermes/commit/7defaf067dbe6f60588518ea1619f228d38ac48d)
Maybe there is other way available to connect two chains together and we simple was unable to find it.

Hi there. Yes, Hermes gets
unbounding period from staking params which leads to an error on Neutron chain because it doesn't have staking. Error looks like this
```
ERROR error raised while creating client for chain baryon-1: failed when building client state: gRPC call failed with status: status: Unimplemented, message: "unknown service cosmos.staking.v1beta1.Query", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "content-length": "0", "date": "Thu, 16 Feb 2023 10:42:14 GMT", "server": "Caddy"} }
```
The issue here is about relaying between let's say Neutron and other chain (in my case it was Band protocol). For now it's possible only with this commit https://github.com/informalsystems/hermes/commit/7defaf067dbe6f60588518ea1619f228d38ac48d and setting unbonding_period for baryon-1.

<-- end -->

We should consider the patch mentioned as a short term fix. 
In parallel ask SDK for a gRPC endpoint that works on both sovereign and consumer chains.

## Version
all

## Acceptance Criteria
Consumer chain support

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #3113, opened 2023-02-21
Add `unbonding_period` parameter to the chain config for use with ICS consumer chains

romac:
Closes: #3112 

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/relayer/src/chain/cosmos.rs
+++ b/crates/relayer/src/chain/cosmos.rs
@@ -1957,10 +1957,6 @@ fn do_health_check(chain: &CosmosSdkChain) -> Result<(), Error> {
         )
     })?;
 
-    if chain.historical_entries()? == 0 {
-        return Err(Error::no_historical_entries(chain_id.clone()));
-    }
-
     let status = chain.chain_status()?;
 
     if status.node_info.other.tx_index != TxIndexStatus::On {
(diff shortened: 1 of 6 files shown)


crates/relayer/src/chain/cosmos.rs
```diff
@@ -388,6 +388,10 @@ impl CosmosSdkChain {
     pub fn unbonding_period(&self) -> Result<Duration, Error> {
         crate::time!("unbonding_period");
 
+        if let Some(unbonding_period) = self.config.unbonding_period {
+            return Ok(unbonding_period);
+        }
+
```
ancazamfir, line 394: Should we consider doing this only if the call below shows status `Unimplemented`? And use a non-zero default for unbonding_period.

Also there is a call to `chain.historical_entries()` in `do_health_check()` where we should maybe issue a warning instead of returning error so further checks are done. Or just move this check to be last.

romac: > Should we consider doing this only if the call below shows status Unimplemented? And use a non-zero default for unbonding_period.

I think we should always use the setting's value if it's present, otherwise we may confuse operators when the clients they create etc. don't have the unbonding period they've specified. There is an always chance that they may set it even for a non-ICS consumer chain and that they may set the wrong value, but that's the case for pretty much every setting we have already.

> Also there is a call to chain.historical_entries() in do_health_check() where we should maybe issue a warning instead of returning error so further checks are done. Or just move this check to be last.

Agreed.

ancazamfir: > ...but that's the case for pretty much every setting we have already.

afaik is for parameters that we cannot query. In this case `unbonding_period` could be set to `X` and a query to chain could give `Y`.  Using a bad `X` could result in client errors later during relaying.
Maybe we could have some warning in the health check at least?

ancazamfir (approved):
lgtm!!

Issue #1658, opened 2021-12-08
Consider switching away from rouille


mzabaluev:
## Crate

ibc-relayer-rest, ibc-telemetry

## Summary

The `rouille` crate imposes an unnecessary restriction on its `time` dependency, holding back useful changes.

## Problem Definition

The crates depend on `rouille`, which pins its dependency on `time` to =0.3.2: https://github.com/tomaka/rouille/issues/246

This will come into conflict with changes made in https://github.com/informalsystems/tendermint-rs/pull/1030 and has already caused a rollback in https://github.com/informalsystems/tendermint-rs/issues/1047.

## Proposal

Switch to `warp`, or any other suitable web service framework without dependency issues.

## Acceptance Criteria

The problem is resolved upstream, or this project removes the `rouille` dependency.

____

#### For Admin Use

- [x] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

mzabaluev:
With #1665 in, `rouille` will be the last obstacle to fixing #1478.

PR #3200, opened 2023-03-27
Switch away from `rouille` to `axum` in telemetry and REST servers

romac:
Closes: #1658
Closes: #1478
Closes: #3194

## Description

`rouille` hasn't been updated in a while, and brings with it obsolete dependencies which are either unmaintained or who have CVEs open against them, which is security liability.

This PR switches `rouille` out for `axum`, which cuts down on the number of dependencies of the relayer since we already depend on `axum` via `tonic`.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


Cargo.toml
```diff
@@ -14,8 +14,6 @@ members = [
     "tools/check-guide",
     "tools/query-events",
 ]
-
-[patch.crates-io]
```
ljoss17, line 18: Is this removed for a reason or is it just for aesthetic reasons ?

ljoss17 (approved):
Looks good to me!

Cargo.toml
romac: Not sure how it got removed but there is no real reason to either leave it or keep it.

ljoss17: Ok, with the commented block underneath I have a slight preference if it's kept then. Without the `[patch.crates-io]` the block might be confusing.
Or the commented block could also be removed if it seems better that way

romac: Oh right, didn't realize there was a comment block below, will re-add

PR #3234, opened 2023-04-11
Wait for auto_relay to finish in integration tests

ljoss17:
Closes: #XXX

## Description

Add two traits to the test abstraction:
* `CanWaitForAck`: Queries pending acknowledgments, and waits if there are any.
* `CanShutdown`:
  * The implementation for the current relayer does nothing.
  * The implementation for the relayer-next calls Tokio `JoinHandle::abort` and waits for the `auto_relay` to be finished.

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/relayer-components-extra/src/relay/impls/auto_relayers/parallel_bidirectional.rs
+++ b/crates/relayer-components-extra/src/relay/impls/auto_relayers/parallel_bidirectional.rs
@@ -39,6 +39,9 @@ where
                 .await;
         });
 
+        // Wait for handle1 and handle2 to finish.
+        // Equivalent to Join.
+        // TODO: Confirm with Soares
         handle1.into_future().await;
         handle2.into_future().await;
 
--- a/crates/relayer-cosmos/src/base/impls/transaction.rs
+++ b/crates/relayer-cosmos/src/base/impls/transaction.rs
@@ -219,6 +219,31 @@ fn split_events_by_messages(in_events: Vec<AbciEvent>) -> Vec<Vec<Arc<AbciEvent>
     for event in in_events.into_iter() {
         if event.kind == "message"
             // TODO: What is the purpose of this filter ?
+            // It seems that the event kind "message" from the Tx Response of some chains
+            // only have 1 event attribute:
+            // event kind: message
+            // event attributes: [
+            // EventAttribute {
+            //    key: "sender",
+            //   value: "cosmos1w2jl4lt77j0u3wuvknmrflp9pmwx5zmrr2j8x7",
+            //  index: true,
+            // },
+            // ]
+            // But others have multiple event attributes:
+            // event kind in send_message: message
+            // event.attributes: [
+            // EventAttribute {
+            // key: "action",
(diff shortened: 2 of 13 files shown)


crates/relayer-components-extra/src/relay/impls/auto_relayers/parallel_bidirectional.rs
```diff
@@ -39,6 +39,9 @@ where
                 .await;
         });
 
+        // Wait for handle1 and handle2 to finish.
+        // Equivalent to Join.
+        // TODO: Confirm with Soares
         handle1.into_future().await;
```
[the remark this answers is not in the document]
romac: In any case, it might be slightly cleaner with [`tokio::join!`](https://docs.rs/tokio/latest/tokio/macro.join.html#runtime-characteristics) which will await the two handles in parallel:

```rust
tokio::join!(handle1, handle2);
```

tools/integration-test/Cargo.toml
```diff
@@ -21,6 +21,7 @@ ibc-relayer-components = { version = "0.1.0", path = "../../crates/relayer-compo
 ibc-relayer-runtime    = { path = "../../crates/relayer-runtime" }
 ibc-relayer-cosmos     = { path = "../../crates/relayer-cosmos" }
 ibc-test-framework = { path = "../test-framework" }
+tokio = { version = "1.0", features = ["rt"] }
```
romac, line 24: Let's move this one line below together with non ibc- dependencies?

crates/relayer-components-extra/src/relay/impls/auto_relayers/parallel_bidirectional.rs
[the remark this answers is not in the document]
ljoss17: I think the `into_future` is used in case a runtime other than Tokio is used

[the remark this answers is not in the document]
romac: Oh I didn't know we were aiming to be runtime-agnostic. Worthy goal but a bit overkill imho. Anyway, then perhaps the spawner could expose a `join` method. Although in this context, the current solution is fine as-is.

romac (approved):

crates/relayer-cosmos/src/base/impls/transaction.rs
```diff
@@ -218,7 +218,8 @@ fn split_events_by_messages(in_events: Vec<AbciEvent>) -> Vec<Vec<Arc<AbciEvent>
 
     for event in in_events.into_iter() {
         if event.kind == "message"
-            && event.attributes.len() == 1
+            // TODO: What is the purpose of this filter ?
+            //&& event.attributes.len() == 1
```
romac, line 222: Why this change?

ljoss17: I added more details to the reason behind the change. Let me know if it is clear enough or if I should add more

romac: All good, thanks!

Issue #3254, opened 2023-04-19
Add integration test for `ChanUpgradeTry` step


ljoss17:
## Summary

Add tests for different scenarios for the `ChanUpgradeTry` step.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #3259, opened 2023-04-19
Add integration test for `ChanUpgradeTry` step

ljoss17:
Closes: #3254

## Description

This PR adds a test for the `ChanUpgradeTry` step.

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


tools/integration-test/src/tests/channel_upgrade/manual_channel_upgrade.rs
```diff
@@ -129,3 +140,129 @@ impl BinaryChannelTest for ChannelUpgradeInitHandshake {
...
+    fn modify_test_config(&self, config: &mut TestConfig) {
+        config.bootstrap_with_random_ids = false;
+    }
+
+    fn modify_relayer_config(&self, config: &mut Config) {
+        config.mode.connections.enabled = true;
+
+        config.mode.channels.enabled = false;
+        config.mode.packets.enabled = false;
+        config.mode.clients.enabled = false;
```
romac, line 156: As a sanity check, do we really need this since we are not spawning the supervisor at all at the moment?

ljoss17: Good point, I don't think this is necessary

tools/integration-test/src/tests/channel_upgrade/manual_channel_upgrade.rs
```diff
@@ -129,3 +140,129 @@ impl BinaryChannelTest for ChannelUpgradeInitHandshake {
...
+            channel.clone(),
+            Some(new_version),
+            new_ordering,
+            new_connection_hops,
+            timeout,
+        )?;
+
+        info!("Check that the step ChanUpgradeInit was correctly executed...");
+
+        assert_eventually_channel_upgrade_init(
```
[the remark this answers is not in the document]
ljoss17: I think if the tracing logs are clear enough to know which step failed, we can merge them together when the `ChanUpgradeAck` step is implemented.

tools/integration-test/src/tests/channel_upgrade/manual_channel_upgrade.rs
```diff
@@ -104,24 +97,35 @@ impl BinaryChannelTest for ChannelUpgradeInitHandshake {
         .map_err(|e| eyre!("error creating height for timeout height: {e}"))?;
         let timeout = UpgradeTimeout::Height(timeout_height);
 
-        info!("Initialise channel upgrade process...");
+        info!("Set channel in (INITUPGRADE, OPEN) state...");
 
-        init_channel_upgrade(
-            channel,
+        channel.flipped().build_chan_upgrade_init_and_send(
```
soareschen, line 102: This is not related to the test, but shouldn't channel upgrade init be permissioned and require governance approval? If it is not permissioned, an attacker can potentially DOS a channel by repeatedly trying to upgrade the channel to a version that the counterparty chain do not support.

ljoss17: A chain is able to cancel an upgrade if it doesn't agree on the proposal. Since only the TRY and ACK are "blocking", wouldn't this avoid the DOS scenario you described? https://github.com/cosmos/ibc/blob/main/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md#channelstate

It also seems that it is left to the individual implementation to define the access-control, so I'm guessing when this will be deployed to chains it will be permissioned, https://github.com/cosmos/ibc/blob/main/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md#upgrade-handshake

soareschen (approved):

tools/integration-test/src/tests/channel_upgrade/manual_channel_upgrade.rs
soareschen: Ok makes sense. But that also means the test here probably won't work if tested against chain binary that require governance approval. I guess we will update the tests when official chains start to adopt the protocol.

Issue #3213, opened 2023-04-02
Review channel version code and cleanup


ancazamfir:
## Summary of Bug
We should cleanup the code around channel version according to the changes in https://github.com/cosmos/ibc/issues/628

## Acceptance Criteria

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

seanchen1991:
@ancazamfir could you clarify what changes need to occur in Hermes in order to satisfy the changes proposed in https://github.com/cosmos/ibc/issues/628?

romac:
@seanchen1991 I think it's just a matter of turning this

https://github.com/informalsystems/hermes/blob/a69bf5cf9189fa772921ba0e9a27c25beda8b735/crates/relayer/src/channel.rs#L838-L854

into something like this

```rust
        // If the user supplied a version, use that.
        // Otherwise, use the empty version and let the app decide on it.
        let version = self.dst_version().cloned().unwrap_or_else(Version::empty);
```

And remove `default_by_port`.

romac:
For open try, we already use the version that the app decided on.

seanchen1991:
Closing this as it seems like most chains have not incorporated these changes yet, so they aren't quite relevant for now. We can revisit if/when these changes become necessary.

PR #3267, opened 2023-04-24
Set channel version at the application level 

seanchen1991:
Closes: #3213 

## Description

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


romac:
Perhaps we should actually keep the counterparty version set in the OpenTry message and see if that fixes some tests.

The ordered channel is apparently using an old version of ibc-go (v3) which seems to require the version at the open init step. We should see if we can update it to a newer version.

romac:
Nevermind, looks like pretty much all tests are failing, even on Gaia 8. Maybe we should just drop this PR then, or at least clarify the expected behavior with the ibc-go team.

romac:
Looks like gaia8 (and likely other chains using the same ibc-go version) still expects a version to be set at chan open init, so perhaps we should keep doing that as well.

romac:
Given the resulting change I wonder if we shouldn't just drop the PR altogether? Do we foresee any issues with setting the try version to its init counterpart? If not then let's drop it.

ljoss17:
I agree, the deprecated version in the `Channel` field doesn't seem to bring much to Hermes. I would be keen to drop this PR

Issue #3206, opened 2023-03-29
REST endpoint failing


sryps:
## Summary of Bug

We keep running into a problem where we get an error `Message too long`. It reconnects to the websocket after this error and keeps relaying but every time this happens the REST endpoint dies and wont stand back up until we restart hermes. Since we use the `/state` endpoint for some monitoring, our system alerts on this happening frequently.

```
2023-03-26T17:26:10.764153Z ERROR ThreadId(129) event_monitor{chain=osmosis-1}: failed to collect events: WebSocket driver failed: web socket error: failed to read from WebSocket connection: Space limit exceeded: Message too long: 203625165 > 16777216
2023-03-26T17:26:11.910211Z  INFO ThreadId(129) event_monitor{chain=osmosis-1}:event_monitor.reconnect{chain=osmosis-1}: successfully reconnected to WebSocket endpoint ws://re.dac.te.d:12345/websocket
```

## Version

running: hermes version 1.3.0

## Acceptance Criteria

REST endpoint survives any websocket disconnection, or specifically this event.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

ancazamfir:
> REST endpoint dies and wont stand back up...

What does this mean exactly? For which chains do you see that? And how frequently?

ancazamfir:
Also, could you provide the node config.toml, in particular have you tried with the parameters mentioned [here](https://github.com/tendermint/tendermint/pull/7230#issue-1044218968)?
We had similar [issue](https://github.com/informalsystems/hermes/issues/1299) a while back for which I thought the workaround above worked.

romac:
@sryps I unfortunately haven't been able to reproduce the issue with the REST endpoint even after killing the WebSocket.

If you have the bandwidth, would you will to build and deploy Hermes from this branch: `romac/poll-block-results` and see if Hermes and its REST endpoint survives the Osmosis epoch block?

That bypasses the WebSocket and relies on pulling data from the chain every block.

We've also upgrade the built-in REST and telemetry server to use more modern libraries which may also alleviate the issue, so if you'd rather not run an experimental branch, you can also try building Hermes from `master` and see if you notice any improvements.

sryps:
@romac  so far so good. We can close for now and if it resurfaces I'll let you know! Thanks!

Issue #1299, opened 2021-08-17
Websocket size constraints


greg-szabo:
## Crate
hermes

## Summary of Bug
I need to understand websocket size constraints better. What does this error message mean: (during Osmosis reward distribution)
```
Aug 17 13:17:59 hermes hermes[7621]: Aug 17 13:17:59.046 ERROR [osmosis-1] failed to collect events: WebSocket driver failed: WebSocket Error: failed to read from WebSocket connection: Space limit exceeded: Message length too big: 202887188 > 16777216 (code: 1)
```

## Version
0.6.1

## Acceptance Criteria
no errors :)

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adizere:
@mircea-c wanted to check with you if the problem still appears?

There was a [fix](https://github.com/informalsystems/tendermint-rs/issues/974) from the tm-rs side, which went into 0.22, and Hermes 0.7.2 should have that fix. It's not clear to me, however, if that fixed the problem reported above. Does Hermes still report "Space limit exceeded" ?

romac:
This was actually not released as part of 0.7.2 but only part of a branch and special build of Hermes. I don't know if @mircea-c got a chance to deploy it temporarily and whether it actually fixed the problem?

mircea-c:
Problem still there on the latest osmosis version
```
 2021-12-14T17:23:45.755660Z ERROR ThreadId(61) [osmosis-1] failed to collect events: WebSocket driver failed: web socket error: failed to read from WebSocket connection: Space limit exceeded: Message length too big: 70286072 > 16777216
 2021-12-14T17:23:45.761674Z  INFO ThreadId(61) [osmosis-1] successfully reconnected to WebSocket endpoint ws://10.10.51.208:26690/websocket
```

adizere:
The latest status on websocket improvements is:

- Websocket improvements in tendermint-go (PR [#7230](https://github.com/tendermint/tendermint/issues/7230)) landed into tendermint `0.34.15`
- There is no SDK release to build on tendermint-go `0.34.15`:
[...]

PR #3323, opened 2023-05-09
Add poll-based event source

romac:
Closes: https://github.com/informalsystems/hermes/issues/2850

Fixes: https://github.com/informalsystems/hermes/issues/3190
Fixes: https://github.com/informalsystems/hermes/issues/2809

Might address: #3206 (see also #1299)

## Description

- Fixes https://github.com/informalsystems/hermes/issues/3190 and https://github.com/informalsystems/hermes/issues/2809 by not having to query for events using the non-standard `message` key. 
- Might address #3206 by not having to re-subscribe to the WebSocket when the connection gets closed, which seems to have interfered with the REST endpoints in the past.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


seanchen1991:
As far as how this addresses the REST endpoint failing, Hermes would essentially switch to purely poll-based relaying in the case when the WebSocket connection gets closed?

crates/relayer/src/chain/cosmos.rs
```diff
@@ -286,27 +286,33 @@ impl CosmosSdkChain {
         Ok(())
     }
 
-    fn init_event_monitor(&mut self) -> Result<TxMonitorCmd, Error> {
+    fn init_event_source(&mut self) -> Result<TxEventSourceCmd, Error> {
         crate::time!(
             "init_event_monitor",
```
seanchen1991, line 291:
```suggestion
            "init_event_source",
```

crates/relayer/src/event/source/rpc.rs
```diff
@@ -0,0 +1,385 @@
...
+    util::retry::ConstantGrowth,
+};
+
+use super::{EventBatch, EventSourceCmd, TxEventSourceCmd};
+
+use self::extract::extract_events;
+
+pub type Result<T> = core::result::Result<T, Error>;
+
+pub struct EventSource {
```
seanchen1991, line 37:
```suggestion
/// Captures the mode in which events are fetched from a chain, either via WebSocket
/// or via RPC endpoint.
pub struct EventSource {
```

crates/relayer/src/chain/cosmos.rs
```diff
@@ -286,27 +286,33 @@ impl CosmosSdkChain {
         Ok(())
     }
 
-    fn init_event_monitor(&mut self) -> Result<TxMonitorCmd, Error> {
+    fn init_event_source(&mut self) -> Result<TxEventSourceCmd, Error> {
```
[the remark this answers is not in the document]
romac: Event source is a better name than monitor, especially now that we have multiple different sources.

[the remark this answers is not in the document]
seanchen1991: There are some places where we still use `event_monitor`, which was a bit confusing. So I think we either switch entirely to `event_source` naming, or communicate what the different semantics mean.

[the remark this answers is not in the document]
romac: Yeah et's switch over to `event_source` everywhere

kaisbaccour:
FYI I testnet this branch for wasm between nois-testnet-005 and uni-6 with the event_source.rpc and it seems to work well so far.

romac:
> FYI I testnet this branch for wasm packets between nois-testnet-005 and uni-6 with the event_source.rpc (1s poll_interval) and it seems to work well so far.

Glad to hear, thanks for the heads-up! 🙏

config.toml
```diff
@@ -126,15 +126,34 @@ grpc_addr = 'http://127.0.0.1:9090'
...
+#
+# Default: 500ms
+batch_delay = '500ms'
+
+# The interval at which to poll for blocks when the `event_source` is `rpc`.
+# Note: Only applies when `event_source` is set to `rpc`.
+#
+# Default: 1 second
+poll_interval = '1s'
+
```
ancazamfir, line 150: as discussed, consider grouping the params, e.g.:
```
event_settings = {type = push, websocket_addr = 'ws://127.0.0.1:26557/websocket', interval = '1s'}
event_settings = {type = poll, interval = '1s'}
```

ancazamfir: I am now wondering if we should in fact explore two extra models:
- the current poll one you have in this PR (would be for users who don't have access to websocket server)
- a hybrid one where we only register for `NewBlock` event and act on it by fetching the `/block_results` (or `/tx_results` since we already get begin and end block events)

romac: Settings changes in https://github.com/informalsystems/hermes/pull/3323/commits/829348d7aa16231b159289f7f26d687ff3c763eb

romac: Yeah that sounds like a great idea!

romac: I don't see a `/tx_results` RPC endpoint, do you mean `/tx_search` + `/tx`?

romac: I have a [prototype](https://github.com/informalsystems/hermes/pull/3323) for a hybrid event source, but still using `/block_results` for now.

crates/relayer/src/event/source/rpc/extract.rs
```diff
@@ -0,0 +1,101 @@
...
+        match ibc_event_try_from_abci_event(abci_event) {
+            Ok(event) if should_collect_event(&event) => {
+                if let IbcEvent::DistributeFeePacket(dist) = &event {
+                    // Only record rewarded fees
+                    if let DistributionType::Reward = dist.distribution_type {
+                        telemetry!(fees_amount, chain_id, &dist.receiver, dist.fee.clone());
+                    }
+                }
+
+                events_with_height.push(IbcEventWithHeight { height, event });
```
ancazamfir, line 29:
```suggestion
                } else {
                    events_with_height.push(IbcEventWithHeight { height, event });
               }
```

config.toml
```diff
@@ -128,9 +128,34 @@ rpc_addr = 'http://127.0.0.1:26657'
...
+#      emitted in the absence of NewBlock event. Default: 500ms
+#      Lower values will result in faster event processing, improving the latency of Hermes,
+#      but may split the events into more batches than necessary, requiring more client updates
+#      to be submitted, yielding higher costs. Higher values will result in slower event
+#      processing, increasing the latency of Hermes, but are more likely to batch events together.
+#      The default value provides good latency while minimizing the number of client updates needed.
+
+# b) Poll: for polling for IBC events via the `/block_results` RPC endpoint
+#
+#     `{ mode = 'poll', poll_interval = '1s' }`
```
ancazamfir, line 152: Change to conform with code:
```suggestion
#     `{ mode = 'pull', poll_interval = '1s' }`
```
But wondering if this would cause confusion with `pull` in `mode` and `poll_..` for the interval and should consider using `poll` everywhere.

crates/relayer/src/event/source/rpc.rs
```diff
@@ -0,0 +1,395 @@
...
+                "latest height ({latest_height}) <= latest fetched height ({})",
+                self.last_fetched_height
+            );
+
+            None
+        };
+
+        // Before handling the batch, check if there are any pending shutdown or subscribe commands.
+        if let Some(next) = self.try_process_cmd() {
+            return Ok(next);
```
ancazamfir, line 163: Why do we need to do this again? Maybe clarify in the comment.

ancazamfir (changes requested):
Reviewed the code and did some tests with both modes with ordered and unordered channels and some channel handshake events also. Looks good, left a couple of comments.

crates/relayer/src/event/source/rpc/extract.rs
romac: b9410b21

crates/relayer/src/event/source/rpc.rs
romac: ebee52c7

config.toml
romac: 682efd2b

ancazamfir (approved):
Amazing work @romac !

romac:
Let's merge this after 1.5.1 is released.

seanchen1991 (approved):

Issue #3334, opened 2023-05-15
Allow gRPC calls with query responses larger than `4MiB`


ljoss17:
## Problem Definition

Updating `ibc-proto-rs` to `v0.29.0` updated tonic to `v0.9.0` which introduced a decoding message default limit of `4MiB`.
Some queries need a bigger limit.

## Proposal

* Update the decoding message limit using `max_decoding_message_size()`.
* Check if the `max_decoding_limit` is set, if not default to `32MiB`
* Improve the error logs to include the failed query, https://github.com/informalsystems/hermes/blob/master/crates/relayer/src/error.rs#L98

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #3335, opened 2023-05-15
Allow gRPC calls with larger query responses

ljoss17:
Closes: #3334

## Description

This PR has the following updates:

* Adds a new undocumented configuration `max_grpc_decoding_size` which can be used to specify a new decoding message size upper limit
* Updates the gRPC clients decoding message size limit with `max_decoding_message_size()`. If the the `max_grpc_decoding_size` is not configured, sets the limit to `32 MiB`.
* Adds the name of the failed query to the `GrpcStatus` error output message

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/relayer/src/chain/cosmos.rs
+++ b/crates/relayer/src/chain/cosmos.rs
@@ -1,4 +1,5 @@
 use alloc::sync::Arc;
+use byte_unit::n_mib_bytes;
 use bytes::{Buf, Bytes};
 use core::{
     convert::{TryFrom, TryInto},
@@ -11,7 +12,8 @@ use num_bigint::BigInt;
 use std::{cmp::Ordering, thread};
 
 use tokio::runtime::Runtime as TokioRuntime;
-use tonic::{codegen::http::Uri, metadata::AsciiMetadataValue};
+use tonic::codegen::http::Uri;
+use tonic::metadata::AsciiMetadataValue;
 use tracing::{error, instrument, trace, warn};
 
 use ibc_proto::cosmos::{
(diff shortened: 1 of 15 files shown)


crates/relayer/src/config.rs
```diff
@@ -472,6 +473,8 @@ pub struct ChainConfig {
     pub max_msg_num: MaxMsgNum,
     #[serde(default)]
     pub max_tx_size: MaxTxSize,
+    #[serde(default)]
+    pub max_grpc_decoding_size: Option<Byte>,
```
romac, line 477: Let's set the default value here so that we avoid the if/else at each use site.

romac (changes requested):

crates/relayer/src/chain/cosmos.rs
```diff
@@ -120,6 +122,8 @@ pub mod types;
 pub mod version;
 pub mod wait;
 
+pub const DEFAULT_GRPC_MAX_MESSAGE_LENGTH: u64 = n_mib_bytes(32);
```
romac, line 125: Let's remove this and use a default straight in the config struct to avoid the if/else's below

crates/relayer/src/error.rs
```diff
@@ -94,8 +94,8 @@ define_error! {
             |_| { "gRPC error" },
 
         GrpcStatus
-            { status: GrpcStatus }
-            |e| { format!("gRPC call failed with status: {0}", e.status) },
+            { status: GrpcStatus, query: String }
+            |e| { format!("gRPC call `{}` failed with status: {1}", e.query, e.status) },
```
romac, line 98: 🙌

romac (approved):

PR #3341, opened 2023-05-15
Remove chain runtime

romac:
Closes: #3344

## Description

By removing the chain runtime, we can issue queries concurrently rather than serially, which has the potential to speed up relaying in two cases:

a) when relaying on multiple channels between two or more chains
b) when relaying between a chain A and two or more chains B, C, ... with A<->B, A<->C, ...

Note that it won't let us send multiple txs concurrently because of the monotonic constraint on the account sequence number. To speed up this aspect of the relayer, we would need multiple wallets.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


crates/relayer/src/chain/cosmos.rs
```diff
@@ -144,12 +145,12 @@ pub struct CosmosSdkChain {
     grpc_addr: Uri,
     light_client: TmLightClient,
     rt: Arc<TokioRuntime>,
-    keybase: KeyRing<Secp256k1KeyPair>,
+    keybase: RwLock<KeyRing<Secp256k1KeyPair>>,
 
     /// A cached copy of the account information
-    account: Option<Account>,
+    account: AsyncRwLock<Option<Account>>,
```
romac, line 151: Here's the lock for the account

crates/relayer/src/chain/cosmos.rs
```diff
@@ -707,8 +711,10 @@ impl CosmosSdkChain {
         let key_pair = self.key()?;
         let key_account = key_pair.account();
 
-        let account =
-            get_or_fetch_account(&self.grpc_addr, &key_account, &mut self.account).await?;
+        // Take a write lock on the current account,
+        // excluding anybody from sending txs until we are done.
+        let mut opt_account = self.account.write().await;
```
romac, line 716: We take the lock here for the remaining duration of the function, ie. until the tx has been submitted.

crates/relayer/src/chain/cosmos.rs
```diff
@@ -721,6 +727,33 @@ impl CosmosSdkChain {
...
+        channel_id: &ChannelId,
+        port_id: &PortId,
+        counterparty_payee: &Signer,
+    ) -> Result<(), Error> {
+        let address = self.get_signer()?;
+        let key_pair = self.key()?;
+
+        // Take a write lock on the current account,
+        // excluding anybody from sending txs until we are done.
+        let mut opt_account = self.account.write().await;
```
romac, line 741: We take the lock here for the remaining of the function, ie. until the counterparty payee has been registered.

crates/relayer/src/chain/cosmos.rs
```diff
@@ -144,12 +145,12 @@ pub struct CosmosSdkChain {
     grpc_addr: Uri,
     light_client: TmLightClient,
     rt: Arc<TokioRuntime>,
-    keybase: KeyRing<Secp256k1KeyPair>,
+    keybase: RwLock<KeyRing<Secp256k1KeyPair>>,
```
romac, line 148: I wonder if we are not better off pushing the lock inside the `KeyRing` instead of holding it here, what do you think?

seanchen1991:
Is this PR just waiting on reviews at this point, or does it need more work?

romac:
Reviews and some testing, ideally in prod ;)

PR #3350, opened 2023-05-18
Limit the maximum number of requests which can issued concurrently to the chain

romac:
This PR targets #3341 

---

Controlled via the `max_concurrency` setting in the chain configuration.


romac:
I should probably use a `CondVar` + `Mutex` here instead of pulling in `resevents`. Will get to it later.

crates/relayer/src/chain/handle/base.rs
```diff
@@ -68,7 +72,17 @@ impl Display for BaseChainHandle {
 
 impl BaseChainHandle {
     pub fn new(chain: Arc<ChainImpl>) -> Self {
-        Self { chain }
+        // The semaphore is initialized with the maximum number of concurrent requests.
+        // If that number was specified as 0, then we use the maximum amount of concurrency,
+        // and effectively disable the limit.
```
seanchen1991, line 77: Should this fact be mentioned in the `config.toml` for the `max_concurrency` parameter?

romac: Yes, good catch!

seanchen1991: [920fd69](https://github.com/informalsystems/hermes/pull/3341/commits/920fd697b2d2b16d627ad1cfec1fa31a28c4b10a)

Issue #2852, opened 2022-11-10
Add `--debug` global flag


romac:
## Summary

Add a `--debug` flag which can either be be repeated (`--debug=rpc --debug=profiling`) or take a comma-separated list of sections to enable debugging for (`--debug=rpc,profiling`), or a combination of the two.

The former is easier to use when invoking Hermes from scripts etc, while the latter is shorted to type when invoked manually.

#### `rpc`

Passing `--debug=rpc` to Hermes has the effect of printing out the outgoing and incoming RPC requests, having the same effect as setting `tendermint_rpc=debug` to the tracing filter.

#### `profiling`

Passing `--debug=profiling` to Hermes which enables the profiling output.

## Problem Definition

It's easier to instruct users to add a flag to their Hermes command rather than ask them to set an esoteric variable to a cryptic value.

## Proposal

As above.

## Acceptance Criteria

Hermes can take a global `--debug` flag which can take the `rpc` value and print out the incoming and outgoing RPC requests.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #3352, opened 2023-05-22
Add `--debug=rpc` and `--debug=profiling` flags for enabling debug output

romac:
Closes: #2852 

## Description

This PR implements the proposition in #2852 and removes the `profiling` feature flag.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


ljoss17:
Since we remove the `profiling` feature, won't we see this message for each `time!` call https://github.com/informalsystems/hermes/pull/3352/files#diff-9d723ef7a1f09059ad4272facc9d5ea2b31ecb46aadadfcf8554c3933a5ca52dR38 ?

crates/relayer/src/util/profiling.rs
```diff
@@ -0,0 +1,137 @@
...
+            name,
+            info,
+            start: std::time::Instant::now(),
+        }
+    }
+}
+
+impl Drop for Timer {
+    fn drop(&mut self) {
+        let enabled = enabled();
```
ljoss17, line 60:
```suggestion
        if !enabled() {
            return;
        }
        let enabled = enabled();
```
What about skipping everything if profiling isn't enabled, do you see any downside to this ?

romac: https://github.com/informalsystems/hermes/pull/3352/commits/bc105a818c9a38b9b8c6bdd0f010c85fd1a6b5e2

romac:
> Do we want to enable `profiling-json` for CLIs which are not `hermes start`? If so, we need to set the file when we update the `debug_sections` flag

I don't think it's needed, the console profiling should work fine for one off queries

romac:
Or perhaps only for `clear packets`, but then we may as well add support everywhere

ljoss17:
> Or perhaps only for clear packets, but then we may as well add support everywhere. Which I think can be done in the CliApp directly.

Yes, I was wondering if we should add it here https://github.com/informalsystems/hermes/pull/3352/files#diff-f4ab9a0749ae805544d3bd5157591c8888bd1ae933688a71b63633d5fd791b94R197.
But if there is only one CLI which could potentially use this, it might be better to document in the guide that this debug option is for a `hermes start` instance.
I'll add the changelog and update the guide.

ljoss17 (approved):
Looks good to me. Tested it and seems to work fine.

.changelog/unreleased/features/ibc-relayer-cli/2852-add-debug-flag.md
```diff
@@ -0,0 +1,3 @@
+- Add a global flag `--debug` which allows enabling
+  profiling, rpc debug logs and/or profiling to json file
```
romac, line 2:
```suggestion
- Add a global flag `--debug` which can take one or more of the following values, separated by commas:
  - `rpc`: show RPC debug logs
  - `profiling`: show profiling information in the console
  - `profiling-json`: dump the profiling information to a JSON file in the directory specified in `PROFILING_DIR` env variable if present, or the current directory otherwise.
```

Issue #3359, opened 2023-05-23
Receive packet fails chain proof verification due to duplicate send_packet events 


ancazamfir:
## Summary of Bug
This issue is similar to #2453
Was found when debugging on the `injective - nois` testnet.

For `nois-testnet-005:channel-16 --> injective-888:channel-51` the packet commitment for sequence `3` is present on `nois` and not received by `injective`. When searching for the send_packet event on nois, two Tx-es are returned by the `/tx_search` query which looks like this:
```
http://135.181.252.92:26657/tx_search?query="send_packet.packet_src_channel='channel-16' AND send_packet.packet_sequence='3' AND send_packet.packet_src_port='wasm.nois1xwde9rzqk5u36fke0r9ddmtwvh43n4fv53c5vc462wz8xlnqjhls6d90xc' AND send_packet.packet_dst_channel='channel-51' AND send_packet.packet_dst_port='wasm.inj1492g0dmg4l7dm6nppsy5zhuwh8q350j2wfrxrd'"
```
The `send_packet` event is present in both Tx-es for the same channel and sequence number and can be checked with 
```
hermes query tx events --chain nois-testnet-005 --hash..
```
hermes selects the event from the first Tx. However in this first Tx there are multiple messages, one is the `send_packet` with sequence `3` and the after this there is a `write_acknowledgment` for the packet with sequence `2`. In the events for the write_acknowledgment there is an error:
```
                    EventAttribute {
                        key: "packet_ack",
                        value: "{\"error\":\"ABCI code: 5: error handling packet: see events for details\"}",
                        index: true,
                    },
```
In this case (afaik) all messages are dropped, meaning that the `nextSequenceSend` is not incremented on chain, the commitment is not stored, etc. But it seems that the `send_packet` event is still emitted and stored by tendermint event indexer. So when hermes clears the packet it will find it there.
We should get feedback from the ibc-go team but until then I think we should fix this by getting the event from highest height (for both `/tx_search` and `/block_search` cases)

Note: For #2453 we introduced the `--packet-data-query-height`. I tentatively tried this:
```
hermes --config config_inj.toml tx packet-recv --dst-chain injective-888 --src-chain nois-testnet-005 --src-port wasm.nois1xwde9rzqk5u36fke0r9ddmtwvh43n4fv53c5vc462wz8xlnqjhls6d90xc --src-channel channel-16 --packet-data-query-height 930204
```
[...]

Issue #3358, opened 2023-05-23
Acknowledgement packet not relayed due to failure to retrieve the write_acknowledgement event


ancazamfir:
## Summary of Bug
_[Thanks to @mturkia for the help with debugging]_

This issue is similar to #2867 for `/block_search`.

Was found when debugging on the `injective - nois` testnet.

For `injective-888:channel-51 --> nois-testnet-005:channel-16` direction, the packet commitment for sequence `10` is present on `injective`. When searching for the `write_acknowledgment` event on `nois`, two Tx-es are returned by the `/tx_search` query which looks like this:
```
http://135.181.252.92:26657/tx_search?query="write_acknowledgement.packet_src_channel='channel-51' AND write_acknowledgement.packet_sequence='10' AND write_acknowledgement.packet_src_port='wasm.inj1492g0dmg4l7dm6nppsy5zhuwh8q350j2wfrxrd' AND write_acknowledgement.packet_dst_channel='channel-16' AND write_acknowledgement.packet_dst_port='wasm.nois1xwde9rzqk5u36fke0r9ddmtwvh43n4fv53c5vc462wz8xlnqjhls6d90xc'" 
```

The first Tx that hermes picks is not for the correct packet and the event is filtered by `filter_matching_event` resulting in acknowledgment packet not being relayed:
```
pulled packet data for 0 events out of 1 sequences: 10..=10; events.total=1 events.left=0
```
The fix should be similar to #2870

_Detailed logs and debug steps to be added._

## Acceptance Criteria

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #3361, opened 2023-05-23
Handle duplicate packet events and perform full clearing even in the presence of errors

romac:
Closes: #3359
Closes: #3358

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


crates/relayer/src/chain/cosmos/query/tx.rs
```diff
@@ -121,34 +122,27 @@ pub async fn query_packets_from_txs(
...
-
-        if let Some(event) = event {
-            result.push(event);
+        // Process each tx in descending order
+        'inner: for tx in response.txs {
+            // Check if the tx contains and event which matches the query
+            if let Some(event) = packet_from_tx_search_response(chain_id, request, *seq, tx)? {
+                // We found the event
+                result.push(event);
+                break 'inner;
```
ancazamfir, line 141: Instead of breaking, could we get them all, if more than 1 extracted print a warning with some tx hashes or height info? And then pick the first? In 99.99% of cases there will be only one so there is no perf impact. But it would give very valuable feedback to the user

crates/relayer/src/chain/cosmos/query/tx.rs
```diff
@@ -121,34 +123,49 @@ pub async fn query_packets_from_txs(
...
+        // If more than one event was found for this sequence, log a warning, and use the first one.
+        if tx_events.len() > 1 {
+            warn!("more than one packet event found for sequence {seq}, this should not happen",);
+
+            for (event, hash, height) in &tx_events {
+                warn!("seq: {seq}, tx hash: {hash}, tx height: {height}, event: {event:?}",);
+            }
+
+            let (first_event, _, _) = tx_events.remove(0);
+            result.push(first_event);
```
ancazamfir, line 161: can we pull these out of the `if` as we do it for both branches

ancazamfir (approved):
Looks great! Packets are cleared on the testnet! Thanks @romac !!

romac:
Thank _you_ for diagnosing the issue and coming up with a way to fix it 🙏

Issue #3366, opened 2023-05-24
Adapt buckets for latency metrics to better reflect production numbers


romac:
## Summary of Bug

In production it looks like most of the latency measurement fall in the +inf bucket. We should therefore adapt those to better reflect production latencies.

## Version

Hermes 1.5.0

## Steps to Reproduce

<img width="717" alt="Screenshot 2023-05-24 at 17 58 30" src="https://github.com/informalsystems/hermes/assets/106849/61ab9f76-a1fb-4826-a45d-5d8259d39b14">

<img width="729" alt="Screenshot 2023-05-24 at 17 59 12" src="https://github.com/informalsystems/hermes/assets/106849/adcca3d9-f23c-4218-b7ae-305329862983">

## Acceptance Criteria

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #3382, opened 2023-05-31
Configurable range for histogram buckets for latency metrics

ljoss17:
Closes: #3366

## Description

This PR adds 2 new configurations:

- `submitted_range` used to specify the range of the histogram
    buckets for the `tx_latency_submitted` metric.
- `confirmed_range` used to specify the range of the histogram
    buckets for the `tx_latency_confirmed` metric.

Allowing the buckets to be changed without the need to recompile Hermes

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -110,6 +110,36 @@ host = '127.0.0.1'
 # by the telemetry service. Default: 3001
 port = 3001
 
+# Specify the range of the 10 histogram buckets in ms for the `tx_latency_submitted` metric.
+# Default: { min = 500, max = 10000 } which will result in the following buckets:
+# tx_latency_submitted_bucket{...,le="950"} 0
+# tx_latency_submitted_bucket{...,le="1900"} 1
+# tx_latency_submitted_bucket{...,le="2850"} 1
+# tx_latency_submitted_bucket{...,le="3800"} 1
+# tx_latency_submitted_bucket{...,le="4750"} 1
+# tx_latency_submitted_bucket{...,le="5700"} 1
+# tx_latency_submitted_bucket{...,le="6650"} 1
+# tx_latency_submitted_bucket{...,le="7600"} 1
+# tx_latency_submitted_bucket{...,le="8550"} 1
+# tx_latency_submitted_bucket{...,le="9500"} 1
+# tx_latency_submitted_bucket{...,le="+Inf"} 1
+# submitted_range = { min = 5000, max = 10000 }
+
+# Specify the range of the 10 histogram buckets in ms for the `tx_latency_confirmed` metric.
+# Default: { min = 1000, max = 20000 } which will result in the following buckets:
+# tx_latency_confirmed_bucket{...,le="1900"} 0
+# tx_latency_confirmed_bucket{...,le="3800"} 0
+# tx_latency_confirmed_bucket{...,le="5700"} 0
+# tx_latency_confirmed_bucket{...,le="7600"} 0
+# tx_latency_confirmed_bucket{...,le="9500"} 0
+# tx_latency_confirmed_bucket{...,le="11400"} 0
+# tx_latency_confirmed_bucket{...,le="13300"} 0
+# tx_latency_confirmed_bucket{...,le="15200"} 0
+# tx_latency_confirmed_bucket{...,le="17100"} 0
+# tx_latency_confirmed_bucket{...,le="19000"} 0
+# tx_latency_confirmed_bucket{...,le="+Inf"} 1
+confirmed_range = { min = 1000, max = 20000 }
+
 
 # A chains section includes parameters related to a chain and the full node to which
 # the relayer can send transactions and queries.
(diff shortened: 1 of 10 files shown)


crates/telemetry/src/state.rs
```diff
@@ -880,6 +1049,64 @@ use opentelemetry::sdk::metrics::aggregators::Aggregator;
...
+            "tx_latency_submitted" => Some(Arc::new(histogram(&self.get_submitted_range()))),
+            "tx_latency_confirmed" => Some(Arc::new(histogram(&self.get_confirmed_range()))),
+            "ics29_period_fees" => Some(Arc::new(last_value())),
+            _ => Some(Arc::new(sum())),
+        }
+    }
+}
+
 #[derive(Debug)]
 struct CustomAggregatorSelector;
```
ljoss17, line 1111: Should we remove the previous aggregator selector ?

crates/telemetry/src/state.rs
```diff
@@ -880,6 +1049,64 @@ use opentelemetry::sdk::metrics::aggregators::Aggregator;
...
+        Self {
+            tx_latency_submitted_min,
+            tx_latency_submitted_max,
+            tx_latency_confirmed_min,
+            tx_latency_confirmed_max,
+        }
+    }
+
+    pub fn get_submitted_range(&self) -> Vec<f64> {
+        let step = (self.tx_latency_submitted_max - self.tx_latency_submitted_min) / 10;
```
ljoss17, line 1076: Do we want the number of buckets to be configurable also ?

crates/telemetry/src/state.rs
```diff
@@ -881,7 +1051,60 @@ use opentelemetry::sdk::metrics::aggregators::{histogram, last_value, sum};
...
+            / self.tx_latency_submitted_buckets;
+        (0..=self.tx_latency_submitted_buckets)
+            .scan(self.tx_latency_submitted_range.start, |state, _| {
+                let current_value = *state;
+                if current_value >= self.tx_latency_submitted_range.end {
+                    return Some(self.tx_latency_submitted_range.end);
+                }
+                *state += step;
+                Some(current_value)
+            })
```
romac, line 1087: GitHub Copilot came up with this instead, which I believe has the same semantics:

```suggestion
            (0..=self.tx_latency_submitted_buckets)
                .map(|i| self.tx_latency_submitted_range.start + i * step)
```

crates/telemetry/src/state.rs
```diff
@@ -881,7 +1051,60 @@ use opentelemetry::sdk::metrics::aggregators::{histogram, last_value, sum};
...
+            .map(|e| e as f64)
+            .collect::<Vec<_>>()
+    }
+
+    pub fn get_confirmed_range(&self) -> Vec<f64> {
+        let step = (self.tx_latency_confirmed_range.end - self.tx_latency_confirmed_range.start)
```
romac, line 1093: While we are at it, let's extract that logic into its own helper `fn`?

ljoss17: Good point!

ljoss17: Done 3a11ef7ebd2d06964873ca8654c6437f99f8c274

config.toml
```diff
@@ -110,6 +110,19 @@ host = '127.0.0.1'
 # by the telemetry service. Default: 3001
 port = 3001
 
+[telemetry.buckets]
+# Specify the range of the 10 histogram buckets in ms for the `tx_latency_submitted` metric.
+# Default: { start = 500, end = 10000, buckets = 10 }
+# The default will give the following buckets:
+# [500, 1450, 2400, 3350, 4300, 5250, 6200, 7150, 8100, 9050, 10000]
+# latency_submitted = { start = 500, end = 10000, buckets = 10 }
```
ancazamfir, line 118: Was thinking about giving hints on the strategy for setting these values so I played a bit with this on a local testnet with block time of ~1sec and `batch_delay = 50ms`  trying to figure out the best start and end values. It's not very clear to me what the strategy is to set these values, seemed `end = 5s` worked best. Being also global configs they need to cover all chains that we relay on, so the chain block time cannot be used directly when reasoning about these values. Maybe taking the maximum block time and using a higher number of buckets so smaller block-time chains end up using 2-3 buckets.
Any thoughts on this?

romac: Good idea! So we would not ask for start/end but just the number of buckets and derive start/end from the highest max block time?

ljoss17: The thing I dislike if we derive start/end from max block time is that it would no longer be possible to configure the metrics without changing a non-metric configuration.
I feel like the it should be possible to easily change the buckets in order to have additional information when something unexpected is happening.
I would be more inclined to adding some suggestion in the default `config.toml`, default values if start/end/buckets are not specified, and a more in depth explanation in the guide.

Issue #3223, opened 2023-04-04
Submit the evidence detected by Hermes to all clients of the misbehaving chain


ancazamfir:
## Summary
Submit evidence detected by Hermes to all clients of the misbehaving chain

## Problem Definition
When a chain misbehaves, IBC evidence is not submitted to all clients for that chain, just for the client that was updated with the conflicting header. 
Because of this some clients for the misbehaving chain might be frozen late or not at all.

## Proposal
Currently, with this configuration
```
[mode.clients]
misbehaviour = true
```
hermes monitors headers in `MsgUpdateClient` events for active clients and validates them against the chain's full node headers. In case of misbehaviour, hermes submits `MsgSubmitMisbehaviour` to the respective client and the client is then frozen.

We should attempt to submit the evidence to all clients of the misbehaving chains. Note that since different clients for a chain may have different security parameters, the conflicting header might not verify for other clients. In this case the relayer should skip the evidence submission.

## Acceptance Criteria

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #3219, opened 2023-04-04
For CCV send consumer misbehaviour message (`IcsConsumerMisbehaviour`) to the provider


ancazamfir:
## Summary
When misbehaviour is detected on a chain that is CCV consumer, the evidence should be sent to the provider chain using the new `IcsConsumerMisbehaviour` message.

## Problem Definition
When misbehaviour is detected on a chain that is CCV consumer, the evidence submitted to the provider chain should, in addition to freezing the client, trigger slashing, jailing etc. It has been decided that the best way to achieve this is to send a new message. This is handled by the CCV provider module that freezes the client and then takes actions against the misbehaving validators.

## Acceptance Criteria

____

#### For Admin Use

- [x] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

romac:
Re-opening because we cannot implement this until [this PR](https://github.com/cosmos/interchain-security/pull/826) lands in ICS and on the Hub (likely with Gaia v10.0.0).

Reverted in #3242

PR #3456, opened 2023-06-28
Add `evidence` command for monitoring and reporting misbehavior evidence to counterparty clients

romac:
Closes: #3223 
Closes: #3219

## TODO

- [x] Depends on https://github.com/informalsystems/hermes/pull/3651

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/ci/misbehaviour-ics/double_sign_test.sh
+++ b/ci/misbehaviour-ics/double_sign_test.sh
@@ -522,11 +522,29 @@ $HERMES_BIN keys add --key-name evidence --key-file ${HERMES_PROV_NODE_DIR}/${HE
 $HERMES_BIN keys add --key-name relayer  --key-file ${HERMES_CONS_NODE_DIR}/${HERMES_KEY}.json  --chain consumer
 $HERMES_BIN keys add --key-name evidence --key-file ${HERMES_CONS_NODE_DIR}/${HERMES_KEY2}.json --chain consumer
 
+# CCV connection and channel
 $HERMES_BIN create connection \
-     --a-chain consumer \
+    --a-chain consumer \
     --a-client 07-tendermint-0 \
     --b-client 07-tendermint-0
 
+# Dummy client and connection for the test
+$HERMES_BIN create client \
+    --host-chain provider \
+    --reference-chain consumer \
+    --trusting-period 57600s
+
+$HERMES_BIN create client \
+    --host-chain consumer \
+    --reference-chain provider \
+    --trusting-period 57600s
+
+$HERMES_BIN create connection \
+    --a-chain consumer \
+    --a-client 07-tendermint-1 \
+    --b-client 07-tendermint-1
+
+# CCV channel
 $HERMES_BIN create channel \
     --a-chain consumer \
     --a-port consumer \
--- a/ci/misbehaviour-ics/light_client_attack_freeze_test.sh
+++ b/ci/misbehaviour-ics/light_client_attack_freeze_test.sh
@@ -310,11 +310,29 @@ $HERMES_BIN keys add --key-file  ${PROV_NODE_DIR}/${PROV_KEY}.json --chain provi
 
 waiting 5 "for a block"
 
+# CCV connection 
 $HERMES_BIN create connection \
-     --a-chain consumer \
+    --a-chain consumer \
     --a-client 07-tendermint-0 \
     --b-client 07-tendermint-0
 
+# Dummy client and connection for the test
+$HERMES_BIN create client \
+    --host-chain provider \
+    --reference-chain consumer \
+    --trusting-period 57600s
+
+$HERMES_BIN create client \
+    --host-chain consumer \
+    --reference-chain provider \
+    --trusting-period 57600s
+
+$HERMES_BIN create connection \
+    --a-chain consumer \
+    --a-client 07-tendermint-1 \
+    --b-client 07-tendermint-1
+
+# CCV channel
 $HERMES_BIN create channel \
     --a-chain consumer \
     --a-port consumer \
--- a/ci/misbehaviour-ics/light_client_attack_test.sh
+++ b/ci/misbehaviour-ics/light_client_attack_test.sh
@@ -330,11 +330,29 @@ diag "Consumer Trusted Height: ${CD_TRUSTED_HEIGHT}, Hash: ${CD_TRUSTED_HASH}"
 
 waiting 5 "for a block"
 
+# CCV connection
 $HERMES_BIN create connection \
-     --a-chain consumer \
+    --a-chain consumer \
     --a-client 07-tendermint-0 \
     --b-client 07-tendermint-0
 
+# Dummy client and connection for the test
+$HERMES_BIN create client \
+    --host-chain provider \
+    --reference-chain consumer \
+    --trusting-period 57600s
+
+$HERMES_BIN create client \
+    --host-chain consumer \
+    --reference-chain provider \
+    --trusting-period 57600s
+
+$HERMES_BIN create connection \
+    --a-chain consumer \
+    --a-client 07-tendermint-1 \
+    --b-client 07-tendermint-1
+
+# CCV channel
 $HERMES_BIN create channel \
     --a-chain consumer \
     --a-port consumer \


crates/relayer-cli/src/commands/evidence.rs
```diff
@@ -0,0 +1,362 @@
...
+    Ok(())
+}
+
+fn handle_light_client_attack(
+    rt: Arc<TokioRuntime>,
+    chain: &CosmosSdkChain,
+    lc: LightClientAttackEvidence,
+) -> eyre::Result<()> {
+    // Build the two headers to submit as part of the `MsgSubmitMisbehaviour` message.
+    let (header1, header2) = build_evidence_headers(rt.clone(), chain, lc)?;
```
ancazamfir, line 142: we should also build a client update for the common header and then send it to those clients that don't have the consensus state associated with the common header. Otherwise the evidence cannot be verified on chain.

romac:
Failing test is not due to this PR. Will fix out of band.

romac:
Added a test for [double signing](https://github.com/informalsystems/hermes/actions/runs/6174602564/job/16759566787?pr=3456) and for [light client attacks](https://github.com/informalsystems/hermes/actions/runs/6174602564/job/16759568545?pr=3456).

Not clear yet why the light client attack evidence is not found in the block by the `evidence` command. To be investigated further.

crates/relayer-cli/src/commands/evidence.rs
```diff
@@ -0,0 +1,651 @@
...
+) -> eyre::Result<(TendermintHeader, TendermintHeader)> {
+    let trusted_validator_set = rt
+        .block_on(chain.rpc_client.validators(lc.common_height, Paging::All))?
+        .validators;
+
+    let header1 = {
+        TendermintHeader {
+            signed_header: lc.conflicting_block.signed_header,
+            validator_set: lc.conflicting_block.validator_set,
+            trusted_height: Height::from_tm(lc.common_height, chain.id()),
```
romac, line 596: FIXME: @sainoe is running into an error where the node rejects the evidence because the trusted height is equal to the header height, ie. the height of the conflicting block is the same as the common height. Maybe the light client detector is doing something wrong?

cc @ancazamfir

crates/relayer-cli/src/commands/evidence.rs
```diff
@@ -0,0 +1,651 @@
...
+                .block_on(chain.rpc_client.validators(infraction_height, Paging::All))?
+                .validators;
+
+            let validator_set =
+                validator::Set::with_proposer(validators, signed_header.header.proposer_address)?;
+
+            TendermintHeader {
+                signed_header,
+                validator_set,
+                trusted_height: Height::from_tm(infraction_height, chain.id()),
```
romac, line 299: FIXME: This seems wrong, as the light client will reject a header that has the same trusted height as its header, ie. if both are at `infraction_height` (see line 286 above). Though that hasn't happened in practice so far as far as I can tell, which is weird no?

cc @ancazamfir

romac: Maybe the on-chain misbehavior module does not verify that header with the light client when it gets it as evidence of a duplicate vote?

ancazamfir: yes, we need to get the trusted height in the same way we do for client updates. Retrieve the consensus state at the highest height smaller than the infraction height.

romac: Right, duh. Thank you, will get to that!

crates/relayer-cli/src/commands/evidence.rs
```diff
@@ -0,0 +1,651 @@
...
+        }
+
+        info!("submitting CCV misbehaviour to provider chain {counterparty_chain_id}");
+
+        // Construct the light client block header for the consumer chain at the infraction height
+        let infraction_block_header = {
+            let infraction_height = dv.vote_a.height;
+
+            let trusted_validator_set = rt
+                .block_on(chain.rpc_client.validators(infraction_height, Paging::All))?
```
sainoe, line 283: As in the comment below; this is wrong should use the `trusted_height` instead of the` infraction_height`.

crates/relayer-cli/src/commands/evidence.rs
romac: I am not sure which script exactly @sainoe was using, but I have myself been using this one and could not reproduce the issue he was seeing: https://github.com/informalsystems/hermes/pull/3456/files#diff-a9944185c7e5bf34c51e2a6f1daec2ef8ef0a5cf4ecb3fb81933cf742cb4bb34

crates/relayer-cli/src/commands/evidence.rs
romac: Fixed in [5ff700e](https://github.com/informalsystems/hermes/pull/3456/commits/5ff700e3faaf6cbe42f53ddfc7896000d36de7f2) and cleaned up in [4762131](https://github.com/informalsystems/hermes/pull/3456/commits/47621312d2f075ac589d02eb115487a26cc19b47)

romac:
The failing tests are not due to this PR, and will be fixed out of band.

romac:
Before we can merge this, we need a new release of tendermint-rs which includes https://github.com/informalsystems/tendermint-rs/pull/1353 and https://github.com/informalsystems/tendermint-rs/pull/1351.

crates/relayer-cli/src/commands/evidence.rs
```diff
@@ -0,0 +1,732 @@
...
+
+    if counterparty_client.is_frozen() {
+        warn!(
+            "skipping client `{}` on chain `{}` as it is already frozen",
+            counterparty_client_id,
+            counterparty.id()
+        );
+
+        return Ok(());
+    }
```
ancazamfir, line 499: We should remove from here, still need to submit `MsgSubmitIcsConsumerMisbehaviour` client is frozen. Make the check for `MsgSubmitMisbehaviour` instead.
Also, isn't this checked by the caller in `fetch_all_counterparty_clients`? But will need to remove from there anyway.

crates/relayer-cli/src/commands/evidence.rs
```diff
@@ -0,0 +1,732 @@
...
+    );
+
+    let msg = MsgSubmitMisbehaviour {
+        client_id: counterparty_client_id.clone(),
+        misbehaviour: misbehaviour.to_any(),
+        signer,
+    }
+    .to_any();
+
+    msgs.push(msg);
```
ancazamfir, line 545: only if client is not frozen

crates/relayer-cli/src/commands/evidence.rs
```diff
@@ -0,0 +1,732 @@
...
+                error!("reason: {e}");
+
+                continue;
+            }
+        };
+
+        let counterparty_chain_id = client_state.chain_id();
+        info!("found counterparty client with id {client_id} on counterparty chain {counterparty_chain_id}");
+
+        if client_state.is_frozen() {
```
ancazamfir, line 652: remove this check, we want to submit some evidence event if client is frozen.

ancazamfir (changes requested):

ancazamfir:
Maybe something wrong with my setup but i get:
```
warning: unnecessary parentheses around block return value
   --> crates/relayer-cli/src/commands/tx/client.rs:437:17
    |
437 |                 (self.reference_chain_id != chain.id
    |                 ^
438 |                     && (self.host_chain_id.is_none()
439 |                         || self.host_chain_id == Some(chain.id.clone())))
    |                                                                         ^
    |
    = note: `#[warn(unused_parens)]` on by default
help: remove these parentheses
    |
437 ~                 self.reference_chain_id != chain.id
438 |                     && (self.host_chain_id.is_none()
439 ~                         || self.host_chain_id == Some(chain.id.clone()))
    |
```

ancazamfir:
Few pending issues:
- CI `ics-light-client-attack` fails with `proposer error: nil validator`, see below. I vaguely remember the proposer in validator set being made optional at some point. Not sure if hermes (via tendermint-rs) runs with that version while the ICS runs with the one that is mandatory, or there is another problem. 

```
2023-10-07T23:12:38.142862Z ERROR ThreadId(20) send_messages_and_wait_check_tx{chain=provider tracking_id=light_client_attack_evidence}:send_tx_with_account_sequence_retry{chain=provider account.sequence=10}:estimate_gas: failed to simulate tx. propagating error to caller: gRPC call `send_tx_simulate` failed with status: status: Unknown, message: "failed to execute message; message index: 0: verifying Header1 in Misbehaviour failed: trusted validator set is not tendermint validator set type: fromProto: validatorSet proposer error: nil validator [cosmos/ibc-go/v4@v4.4.2/modules/light-clients/07-tendermint/types/misbehaviour_handle.go:106] With gas wanted: '0' and gas used: '85161' ", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "x-cosmos-block-height": "46"} }
```

- There is one issue with the approach desired by ICS to submit the evidence even if the client is frozen. For LCA we try to build an UpdateClient for the "common height" but we get an error.
```
WARN ThreadId(01) reason: failed to build update client message: client 07-tendermint-0 on chain id provider is frozen: client state reports that client is frozen
```
Then we build the misbehaviour message but this will likely fail if the consensus state for the "common height" is not on chain. 

- There are some unrelated CI tests that are failing (haven't looked at these yet)

romac:
> * CI `ics-light-client-attack` fails with `proposer error: nil validator`, see below. I vaguely remember the proposer in validator set being made optional at some point. Not sure if hermes (via tendermint-rs) runs with that version while the ICS runs with the one that is mandatory, or there is another problem.

I fixed this but now the evidence submission fails with:

> 2023-10-10T09:10:51.961302Z ERROR ThreadId(01) gRPC call `send_tx_simulate` failed with status: status: Unknown, message: "failed to execute message; message index: 1: failed to process misbehaviour for IBC client: cannot process misbehaviour for client (07-tendermint-0) with status Frozen: client is not active [cosmos/ibc-go/v4@v4.4.2/modules/core/02-client/keeper/client.go:201]

romac:
> There are some unrelated CI tests that are failing (haven't looked at these yet)

You can safely ignore those, they're totally unrelated.

ancazamfir:
> > There is one issue with the approach desired by ICS to submit the evidence even if the client is frozen. For LCA we try to build an UpdateClient for the "common height" but we get an error.
> > WARN ThreadId(01) reason: failed to build update client message: client 07-tendermint-0 on chain id provider is frozen: client state reports that client is frozen
> > Then we build the misbehaviour message but this will likely fail (once we fix the proposer nil issue above) if the consensus state for the "common height" is not on chain.
> 
> What do you suggest we do about this?

I client is frozen we can check if the client already has the common height consensus state and in that case just submit the evidence.

romac:
👍 And if it's not present? Then we can't submit anything, right?

ancazamfir:
> > * CI `ics-light-client-attack` fails with `proposer error: nil validator`, see below. I vaguely remember the proposer in validator set being made optional at some point. Not sure if hermes (via tendermint-rs) runs with that version while the ICS runs with the one that is mandatory, or there is another problem.
> 
> I fixed this but now the evidence submission fails with:
> 
> > 2023-10-10T09:10:51.961302Z ERROR ThreadId(01) gRPC call `send_tx_simulate` failed with status: status: Unknown, message: "failed to execute message; message index: 1: failed to process misbehaviour for IBC client: cannot process misbehaviour for client (07-tendermint-0) with status Frozen: client is not active [cosmos/ibc-go/v4@v4.4.2/modules/core/02-client/keeper/client.go:201]

Looks like the script is using the `update client` CLI to submit evidence to the consumer chain. This is later read by the hermes `evidence` command. But the `update client` also freezes the client on the provider chain so the evidence CLI will find it already frozen.
The alternative is to run a light client for the consumer chain, either ibc-go or tendermint-rs but this introduces extra dependencies.

romac:
> If client is alreadyt frozen we can check if the common height consensus state is already on-chain and in that case just submit the ICS evidence.

[9cb4d3e](https://github.com/informalsystems/hermes/pull/3456/commits/9cb4d3ea4c286ac0490761d8e653d22b7cd18a02)

I wonder if we do actually need to submit the regular misbehavior evidence or if we should only submit the ICS one?

romac:
> > If client is alreadyt frozen we can check if the common height consensus state is already on-chain and in that case just submit the ICS evidence.
> 
> [9cb4d3e](https://github.com/informalsystems/hermes/pull/3456/commits/9cb4d3ea4c286ac0490761d8e653d22b7cd18a02)
> 
> I wonder if we do actually need to submit the regular misbehavior evidence or if we should only submit the ICS one?

https://github.com/informalsystems/hermes/pull/3456/commits/2300d6c44ad228618221a247b180a119109afbf3

crates/relayer-cli/src/commands/evidence.rs
```diff
@@ -0,0 +1,779 @@
...
+    counterparty: &BaseChainHandle,
+    misbehaviour: TendermintMisbehaviour,
+) -> Result<(), eyre::Error> {
+    info!(
+        "building light client attack evidence for client `{}` on chain `{}`",
+        counterparty_client_id,
+        counterparty.id(),
+    );
+
+    if !is_counterparty_provider(chain, counterparty) && counterparty_client.is_frozen() {
```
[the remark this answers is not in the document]
romac: https://github.com/informalsystems/hermes/pull/3456/commits/d39546d0d66fbe14f66fd70d6fc8cccc224887dc

crates/relayer-cli/src/commands/evidence.rs
```diff
@@ -0,0 +1,786 @@
...
+        }
+        .to_any();
+
+        msgs.push(msg);
+    };
+
+    // Only send the regular IBC message if the counterparty is not a provider,
+    // or if the counterparty is a provider but its client is not frozen already,
+    // and has a consensus state at the common height (implicity because of the check above).
+    if !counterparty_is_provider || !counterparty_client_is_frozen {
```
ancazamfir, line 555: Would this work?
```suggestion
    if !counterparty_client_is_frozen {
```

crates/relayer-cli/src/commands/evidence.rs
```diff
@@ -0,0 +1,786 @@
...
+            "cannot submit light client attack evidence to client `{}` on provider chain `{}`",
+            counterparty_client_id,
+            counterparty.id()
+        );
+        warn!("reason: client is frozen and does not have a consensus state at height {common_height}");
+
+        return Ok(());
+    }
+
+    let mut msgs = match counterparty_client.wait_and_build_update_client(common_height) {
```
ancazamfir, line 521: Should we update client if `provider_has_common_consensus_state`?

crates/relayer-cli/src/commands/evidence.rs
romac: Oh right, since the client is already frozen we don't need to submit that message, whether it's a provider chain or not.

crates/relayer-cli/src/commands/evidence.rs
romac: I had missed that, good catch! No need indeed, since the provider already has that consensus state.

crates/relayer-cli/src/commands/evidence.rs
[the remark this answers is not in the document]
romac: https://github.com/informalsystems/hermes/pull/3456/commits/78e2ed702820f867ce2e08e076cb72c1545b3ac4

crates/relayer-cli/src/commands/evidence.rs
romac: https://github.com/informalsystems/hermes/pull/3456/commits/3e8b37f58a47dca3c6da72dcc597cf7833bfa192

ancazamfir: Can we do the same for sovereign chains? i.e. rename `provider_has_common_consensus_state` to `counterparty_has_common_consensus_state` and drop the `counterparty_is_provider` check from its definition. Then, if `true`, skip client update for both provider and sovereign chain as well.

crates/relayer-cli/src/commands/evidence.rs
```diff
@@ -0,0 +1,795 @@
...
+            .query_consumer_chains()
+            .unwrap_or_default(); // If the query fails, use an empty list of consumers
+
+        // FIXME: Do do we need to check if the client id also matches?
+        // ie. is it okay to submit a `MsgSubmitIcsConsumerMisbehaviour` to all clients
+        // of the provider chain, or should we only do this for the CCV client, and
+        // use the standard message for other clients?
+        consumer_chains
+            .iter()
+            .any(|(chain_id, _)| chain_id == chain.id())
```
romac, line 641: @ancazamfir Do do we need to check if the client id also matches? ie. is it okay to submit a `MsgSubmitIcsConsumerMisbehaviour` to all clients of the provider chain, or should we only do this for the CCV client, and use the standard message for other clients?

Probably the latter, right?

ancazamfir: Yes, I think so.  @sainoe could you confirm?

ancazamfir (approved):
Yayy! Amazing work @romac!

Issue #2925, opened 2022-12-07
Implement abstract IBC handshake components in relayer framework


soareschen:
## Summary

Currently in the new relayer, we rely on using the v1 relayer to perform IBC handshakes before relaying the packets. However since relayer v1 only works with Cosmos chains, we cannot use it for establishing IBC connections between a Cosmos and non-Cosmos chain.

We would need re-implement the IBC handshakes in a abstract manner in the relayer framework. Once that is done, we would be able to use it to create new IBC connections between Cosmos chains as well as with non-Cosmos chains.

The implementation of the abstract handshake components can follow the same patterns as the `PacketRelayer`. We will be implementing components such as `ConnectionHandshakeRelayer` and `ChannelHandshakeRelayer`.

The initial abstract implementation do not need to be as robust as in the v1 relayer. We will handle logic such as retry in separate issue. The initial implementation also do not need to pick up handshakes initiated by others, or resume from previous handshake.

In this issue, we will focus on a concrete implementation for Cosmos. The implementation of handshake for non-Cosmos chains will be handled in other issues.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #3457, opened 2023-06-30
Implement basic IBC connection handshake in relayer-next

soareschen:
Part of #2925.

## Description

This PR implements the basic IBC connection handshake components in relayer-next. 

The current implementation supports only an end-to-end IBC connection creation, and does not resume connection handshake based on IBC events. It also does not perform additional checks that the created connection on-chain is the same as expected.

Additional features can be defined in future PRs as extra components.

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/relayer-components/src/relay/impls/connection/mod.rs
+++ b/crates/relayer-components/src/relay/impls/connection/mod.rs
@@ -1,1 +1,2 @@
+pub mod open_ack;
 pub mod open_try;
--- a/crates/relayer-components/src/relay/impls/connection/mod.rs
+++ b/crates/relayer-components/src/relay/impls/connection/mod.rs
@@ -1,2 +1,3 @@
 pub mod open_ack;
+pub mod open_confirm;
 pub mod open_try;
--- a/crates/relayer-components/src/relay/impls/connection/mod.rs
+++ b/crates/relayer-components/src/relay/impls/connection/mod.rs
@@ -1,4 +1,5 @@
 pub mod open_ack;
 pub mod open_confirm;
 pub mod open_handshake;
+pub mod open_init;
 pub mod open_try;
--- a/crates/relayer-components/src/relay/impls/connection/mod.rs
+++ b/crates/relayer-components/src/relay/impls/connection/mod.rs
@@ -1,3 +1,4 @@
 pub mod open_ack;
 pub mod open_confirm;
+pub mod open_handshake;
 pub mod open_try;
--- a/crates/relayer-components/src/relay/impls/connection/mod.rs
+++ b/crates/relayer-components/src/relay/impls/connection/mod.rs
@@ -1,3 +1,4 @@
+pub mod bootstrap;
 pub mod open_ack;
 pub mod open_confirm;
 pub mod open_handshake;
--- a/crates/relayer-components/src/relay/impls/connection/mod.rs
+++ b/crates/relayer-components/src/relay/impls/connection/mod.rs
@@ -1,0 +1,1 @@
+pub mod open_try;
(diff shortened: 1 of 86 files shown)


crates/relayer-components/src/relay/impls/connection/open_confirm.rs
```diff
@@ -0,0 +1,76 @@
...
+use crate::chain::traits::message_sender::CanSendMessages;
+use crate::chain::traits::queries::status::CanQueryChainHeight;
+use crate::chain::traits::types::height::CanIncrementHeight;
+use crate::relay::traits::chains::HasRelayChains;
+use crate::relay::traits::connection::open_confirm::ConnectionOpenConfirmRelayer;
+use crate::relay::traits::messages::update_client::CanBuildUpdateClientMessage;
+use crate::relay::traits::target::DestinationTarget;
+use crate::std_prelude::*;
+
+pub struct RelayConnectionOpenConfirm;
```
seanchen1991, line 15:
```suggestion
/// Component responsible for sending a confirmation message from a source chain
/// to a destination chain in response to an `OPEN_ACK` message confirming that 
/// the source chain received the destination chain's acknowledgement, thereby 
/// completing the IBC connection handshake protocol. 
/// This step is known as the `OPEN_CONFIRM` step.
pub struct RelayConnectionOpenConfirm;
```

crates/relayer-components/src/relay/impls/connection/open_handshake.rs
```diff
@@ -0,0 +1,40 @@
...
+
+use crate::chain::traits::types::ibc::HasIbcChainTypes;
+use crate::relay::traits::chains::HasRelayChains;
+use crate::relay::traits::connection::open_ack::CanRelayConnectionOpenAck;
+use crate::relay::traits::connection::open_confirm::CanRelayConnectionOpenConfirm;
+use crate::relay::traits::connection::open_handshake::ConnectionOpenHandshakeRelayer;
+use crate::relay::traits::connection::open_try::CanRelayConnectionOpenTry;
+use crate::std_prelude::*;
+
+pub struct RelayConnectionOpenHandshake;
```
seanchen1991, line 11:
```suggestion
/// Component responsible for performing the IBC connection handshake protocol
/// between a source and a destination chain. Specifically, the `OPEN_TRY`,
/// `OPEN_ACK`, and `OPEN_CONFIRM` steps of the handshake protocol are performed
/// between both chains. Upon successful completion of the handshake protocol,
/// a connection will have been established between both chains.
pub struct RelayConnectionOpenHandshake;
```

crates/relayer-components/src/relay/impls/connection/mod.rs
```diff
@@ -0,0 +1,6 @@
+pub mod bootstrap;
```
seanchen1991, line 1:
```suggestion
//! This module contains the traits and types necessary in order to perform the
//! [IBC connection handshake protocol](https://github.com/cosmos/ibc/tree/main/spec/core/ics-003-connection-semantics) 
//! between two IBC-enabled blockchains. The protocol establishes a connection 
//! between the two participating blockchains. This connection between the two 
//! chains is necessary in order for channels to be established between them.
//!
//! Each of the different `open_*` modules, except for `open_handshake`, defines 
//! a component responsible for a single step in the handshake protocol. The 
//! `open_handshake` module wires together the steps that actually constitute
//! the handshake protocol proper. The `open_init` step is not strictly part
//! of the handshake; it merely signals the start of the handshake process. 
//! The `bootstrap` module is the highest level module that wires together the
//! initialization and handshake steps.

pub mod bootstrap;
```

seanchen1991 (commented):
I took a stab at documenting each of the different steps of the handshake protocol. Would appreciate a pass-over in case any of it isn't accurate.

soareschen:
> I took a stab at documenting each of the different steps of the handshake protocol. Would appreciate a pass-over in case any of it isn't accurate.

Thanks for the work! I have modified your comments a little and add them directly in a commit.

crates/relayer-components/src/relay/impls/connection/mod.rs
soareschen: Technically, the bootstrap function is only used when it is initiated by the relayer user or inside tests. If the connection is initiated by external users, then only the `open_handshake` component is used.

Issue #3211, opened 2023-03-31
Get the `max_expected_time_per_block` using the `/genesis` endpoint


seanchen1991:
## Background

We should take care of this [TODO](https://github.com/informalsystems/hermes/blob/master/crates/relayer/src/link/relay_path.rs#L291-L292) to get the `max_expected_time_per_block` using the `/genesis` endpoint in the `src_max_block_time` and `dst_max_block_time` functions.

## Tasks

- [ ] Refactor `src_max_block_time` and `dst_max_block_time` to use the `max_expected_time_per_block` field fetched from the `/genesis` endpoint

____

#### For Admin Use

- [ ] https://github.com/informalsystems/hermes/labels/tracking or https://github.com/informalsystems/hermes/labels/okr label (as well as any other appropriate labels) applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged

PR #3467, opened 2023-07-10
Use `max_expected_time_per_block` value for the `max_block_time`

ljoss17:
Closes: #3211 

## Description

When Hermes validates the ChainConfig, it queries the /genesis RPC endpoint to retrieve `max_expected_time_per_block` and use its value for the `max_block_time`.

If the query fails, Hermes will use the configured value as a fallback value.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


crates/relayer/src/chain/cosmos/types/app_state.rs
```diff
@@ -0,0 +1,38 @@
...
+#[derive(Debug, Deserialize, Serialize)]
+struct ConnectionGenesisConfig {
+    params: ConnectionGenesisParams,
+}
+
+#[serde_as]
+#[derive(Debug, Deserialize, Serialize)]
+struct ConnectionGenesisParams {
+    #[serde_as(as = "DisplayFromStr")]
+    max_expected_time_per_block: u64,
```
romac, line 37: Can we do this manually rather than pulling in `serde-with` which also brings `darling` with it? We're already quite heavy on dependencies and should try to avoid introducing new ones for a single use.

ljoss17: Good point! I'll remove the `serde-with` dependency

romac (approved):
Awesome!

Issue #3243, opened 2023-04-14
Update the channel worker to perform the upgrade handshake


romac:

ljoss17:
Done by https://github.com/informalsystems/hermes/pull/3228

PR #3569, opened 2023-08-28
Channel upgrade worker handshake

ljoss17:
Closes: #3243

## Description

This PR updates the channel worker in order to handle channel upgrade events and complete an ongoing channel upgrade.

## Blockers

This PR is currently blocked by the following:

__ibc-proto-rs__

- [ ] A `ibc-proto-rs` version with channel upgrade protos from `ibc-go`. Requires `ibc-go` to release a version with the protos. 

__Cosmos Nix__

- [x] Updated Cosmos Nix with a `simapp` which has the channel upgrade capability. Requires `ibc-go` to release a version with channel upgrade capability.

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


crates/relayer-types/src/core/ics04_channel/channel.rs
```diff
@@ -460,20 +459,30 @@ impl State {
...
+        match self {
+            State::Uninitialized => 0,
+            State::Init => 1,
+            State::TryOpen => 2,
+            State::Open(_) => 3,
+            State::Closed => 4,
+            State::Flushing => 5,
+            State::Flushcomplete => 6,
+        }
+    }
```
seanchen1991, line 481: Could we instead implement this conversion using the `From` trait? I.e.
```rust
impl From<State> for i32 {
    fn from(state: State) -> Self {
        match self {
            State::Uninitialized => 0,
            State::Init => 1,
            State::TryOpen => 2,
            State::Open(_) => 3,
            State::Closed => 4,
            State::Flushing => 5,
            State::Flushcomplete => 6,
        }
    }
}
```

romac: I'd rather keep it explicit, From does not really buy us anything here (`.into_i32()` vs `.into()`) since we are not using this State generically. I think we've (and I include myself here ofc) abused a bit of From instances over the years and should from now on try to keep things more explicit unless we have a concrete need for a specific instance.

romac: I would rename the method to `as_i32` for consistency with `as i32` and the general naming conventions in Rust:

- `as_X()`: `&self -> X`
- `into_X()`: `self -> X`

tools/integration-test/src/tests/channel_upgrade/ics29.rs
```diff
@@ -0,0 +1,218 @@
...
+            new_version.clone(),
+            new_version.clone(),
+            old_ordering,
+            old_connection_hops_a,
+            old_connection_hops_b,
+        );
+
+        info!("Will initialise upgrade handshake by sending the ChanUpgradeInit step...");
+
+        // Note: Initialising a channel upgrade this way will eventually be removed.
```
seanchen1991, line 139:
```suggestion
        // Note: Initialising a channel upgrade this way, without requiring a 
        // signature or proof of authority to perform the channel upgrade, will
        // eventually be removed.
```

crates/relayer/src/channel.rs
```diff
@@ -2108,7 +2131,8 @@ fn check_destination_channel_state(
         existing_channel.connection_hops() == expected_channel.connection_hops();
 
     // TODO: Refactor into a method
-    let good_state = *existing_channel.state() as u32 <= *expected_channel.state() as u32;
+    let good_state =
```
romac, line 2134: Let's use `State::less_or_equal_progress` if applicable instead?

crates/relayer-types/src/core/ics04_channel/channel.rs
```diff
@@ -498,28 +489,28 @@ impl State {
```
romac, line 489: This method is becoming unwieldy and complex, and we should either:

a) write good unit tests for it, to ensure that it's correct but also that it is isomorphic to `self.as_i32() <= other.as_i32()`.
b) express it in terms of `as_i32()`

Given that doing (a) enables (b), I think we should do (b) right away.


@ljoss17 What do you think?

ljoss17: I agree, this would be a nice improvement. I will add the unit test/s and update the method

tools/integration-test/src/tests/channel_upgrade/upgrade_handshake_steps.rs
```diff
@@ -0,0 +1,632 @@
...
+        info!("Check that the ChanUpgradeOpen steps were correctly executed...");
+
+        assert_eventually_channel_upgrade_open(
+            &chains.handle_a,
+            &chains.handle_b,
+            &channels.channel_id_a.as_ref(),
+            &channels.port_a.as_ref(),
+            &upgraded_attrs,
+        )?;
+
```
romac, line 208: Shall we check that the version has been properly updated?

tools/integration-test/src/tests/channel_upgrade/upgrade_handshake_steps.rs
```diff
@@ -0,0 +1,632 @@
...
+        relayer.with_supervisor(|| {
+            assert_eventually_channel_upgrade_open(
+                &chains.handle_a,
+                &chains.handle_b,
+                &channels.channel_id_a.as_ref(),
+                &channels.port_a.as_ref(),
+                &upgraded_attrs,
+            )?;
+
```
romac, line 331: Shall we check that the version has been properly updated?

tools/integration-test/src/tests/channel_upgrade/upgrade_handshake_steps.rs
```diff
@@ -0,0 +1,632 @@
...
```
romac, line 469: Shall we check that the version has been properly updated?

tools/integration-test/src/tests/channel_upgrade/upgrade_handshake_steps.rs
```diff
@@ -0,0 +1,632 @@
...
```
romac, line 628: Shall we check that the version has been properly updated?

tools/integration-test/src/tests/channel_upgrade/upgrade_handshake_steps.rs
romac: Oh nice, didn't notice that!

romac (approved):
Amazing work @ljoss17 🚀

Issue #29, opened 2020-03-11
The IBC `ibc` crate should be `no-std`


zmanian:
# Summary

The `relayer-modules` crate should be `no-std`.

## Problem Definition

The `relayer-modules` crate should support `no-std` this would make it more useful for solo nodes running in embedded environments and integrating deployment in wasm environments like Substrate Pallets and Eth2 Execution Environments.

## Proposal

Ideally the full relayer can be in `std` environment and use crates like tokio. 

There is progress in support unclear how this interacts with using core::future in the API. Needs more inverstigation.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

hxrts:
The cdot team raised this same issue as Substrate uses no-std.

adizere:
We'd like to understand the priority for this issue. Suppose we adapt the `modules` crate to no-std in the second trimester of 2021 (after may, before end of august), would that work?

hxrts:
@en @livelybug @toliuyi maybe you can comment on whether this is blocking, or a feature that's nice-to-have?

en:
Sorry, I submitted a duplicate issue( #508 ) today because I missed the notification.

andynog:
the no_std support on `tendermint-rs` repo -> https://github.com/informalsystems/tendermint-rs/issues/824

adizere:
#1158 thoroughly tracks efforts towards no_std. Closing as duplicate of that.

PR #3612, opened 2023-09-14
Fix evidence submission

romac:


crates/relayer/src/light_client/tendermint.rs
```diff
@@ -207,13 +207,18 @@ impl super::LightClient<CosmosSdkChain> for LightClient {
                 challenging_block,
             })) => {
                 warn!("misbehavior detected, reporting evidence to RPC witness node and primary chain");
+                debug!("evidence: {evidence:#?}");
+                debug!("challenging block: {challenging_block:#?}");
+
+                warn!("waiting 5 seconds before reporting evidence to RPC witness node");
```
romac, line 213: This seems to be needed because without it I sometimes get the following error when reporting the evidence:

> failed to add evidence: Invalid evidence: don't have commit at height #29

I wonder if we can do something smarter, though?

cc @ancazamfir

Issue #3636, opened 2023-09-22
Change config format to scope configs by type


hdevalence:
## Summary

Currently, Hermes' `ChainConfig` is one monolithic structure with many options. Although it has a `type` field specifying the chain type, it's not possible to have different config options depending on the chain type.  To fix this, and ease the way into Hermes support for non-SDK chains, it would help to restructure the config format so that configuration options are scoped by chain type.

As an example of the kind of problem this creates, consider the changes in https://github.com/penumbra-zone/hermes/pull/3 . This option is Penumbra-specific, so it should only be present when configuring a Penumbra chain -- it doesn't make sense that all other places where the config is used should have to include a Penumbra-specific option.  And many of the options in the existing `ChainConfig` are actually specific to the Cosmos SDK, and are ignored by our Penumbra `ChainEndpoint` implementation.

Instead, if the config were an enum:

- Each "backend" can define their own configuration structure, with only the options relevant to that backend;
- The top-level config can be an enumeration of all of the backends included in Hermes.

The enum can be nicely represented in a TOML config file using Serde attributes, in particular using Serde's `tag` attribute. To see an example of how this works in practice, look at the authorization policy configurations for Penumbra's `SoftKms` software key-management service, [documented here](https://guide.penumbra.zone/main/pclientd/configure.html#authorization-policy) and implemented [here](https://github.com/penumbra-zone/penumbra/blob/main/crates/custody/src/policy.rs#L23):
```toml
[[kms_config.auth_policy]]
type = 'DestinationAllowList'
allowed_destination_addresses = ['penumbrav2t13vh0fkf3qkqjacpm59g23ufea9n5us45e4p5h6hty8vg73r2t8g5l3kynad87u0n9eragf3hhkgkhqe5vhngq2cw493k48c9qg9ms4epllcmndd6ly4v4dw2jcnxaxzjqnlvnw']

[[kms_config.auth_policy]]
type = 'OnlyIbcRelay'

[[kms_config.auth_policy]]
type = 'PreAuthorization'
method = 'Ed25519'
required_signatures = 1
allowed_signers = ['+Osq5OiWKos57KigDjd3XCG/YLUOSUbuBly4LBBpJTg=']
```
In the config file, the enumeration is collapsed into the `type` or `method` tags, so the user doesn't need to consider nested enums, but the code can have options scoped to the specific backend provider.

Doing this would be a small and incremental step towards multi-chain support for Hermes.

## Acceptance Criteria

[...]

Issue #3641, opened 2023-09-29
ChainEndpoint trait makes restrictive assumptions about signing keys


hdevalence:
## Summary

The current ChainEndpoint trait makes too restrictive assumptions about the shape and role of signing keys. This makes it difficult to properly add support for chains like Penumbra with different authorization behaviors.

## Problem Definition

The [ChainEndpoint trait](https://docs.rs/ibc-relayer/latest/ibc_relayer/chain/endpoint/trait.ChainEndpoint.html) used to create pluggable, chain-specific backends has a SigningKeyPair associated type.

This type must implement [the SigningKeyPair](https://docs.rs/ibc-relayer/latest/ibc_relayer/keyring/trait.SigningKeyPair.html) trait, which demands the following behavior:

* production of signatures on arbitrary byte slices;
* generation of the account corresponding to that key pair;
* a declaration of which supported key type it is (ed25519 or secp256k1), and the assumption that there is a pair of keys (private and public).

This is a layering violation, because it bakes in assumptions from the Cosmos SDK that may not apply to other chains. In fact, none of these properties obtain for Penumbra keys:

* Penumbra signing keys are never used on arbitrary byte slices, because they do not sign the encoding of the transaction data. Instead, they sign over the transaction's _effect hash_, similar to Bitcoin's SIGHASH mechanism, and there is one signature per spend, rather than a single signer for the transaction. And each signature is produced with a fresh randomization of the verification key. There is no API to sign arbitrary byte slices. Instead, a complete TransactionPlan is passed to the custodian, who returns AuthorizationData used to build the transaction.

* Penumbra signing keys control many (2^32) logical accounts simultaneously. Accounts are not disclosed to anyone other than the user who controls them. Each account has multiple addresses, a stable "default address" as well as 2^96 ephemeral addresses. Neither accounts nor addresses are disclosed by shielded transactions. Addresses (representing payment capability) are only disclosed by the user to desired counterparties, but are not published on-chain.

* Penumbra's key heirarchy does not have a pair of keys. Instead it has a tree of keys, with variously attenuated capability: a spending key, representing spending capability, a viewing key, representing viewing capability, and many publicly unlinkable addresses, representing the capability to pay specific accounts controlled by that user's spending key.

However, although these assumptions are all violated by Penumbra's design, there is no real problem with relaying. There's just a different authorization mechanism. The only problem here arises from the layering violation of lifting chain-specific authorization mechanisms into the common ChainEndpoint trait.

## Proposal

[...]

PR #3644, opened 2023-10-01
Change config format to scope configs by type

hdevalence:
Closes: #3636

## Description

This turns out to be a very good place to start to discover requirements for Hermes to operate in a more multi-chain way, since it allows discovering which parts of the code access Cosmos-specific data to do Cosmos-specific logic, and which parts are required for general relaying.

The only change to the existing config format so far is the change in semantics of the `type` field, which now must be `CosmosSdk`. I did not preserve the existing logic and tests that parsing different variations of casing and dashes are supported, because I don't think this is particularly valuable to support.

So far, only the core `ibc-relayer` crate is updated; the cli is not, because it's not necessarily clear how the cli should change to accomodate non-SDK chains.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


crates/relayer/src/chain.rs
```diff
@@ -6,60 +6,3 @@ pub mod handle;
...
-// NOTE(new): When adding a variant to `ChainType`, make sure to update
-//            the `Deserialize` implementation below and the tests.
-//            See the NOTE(new) comments below.
-
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize)]
-/// Types of chains the relayer can relay to and from
-pub enum ChainType {
-    /// Chains based on the Cosmos SDK
-    CosmosSdk,
-}
```
hdevalence, line 21: This is gone entirely. Rather than having the config store an enum _of_ chain types, the config _is_ an enum, so that it can have different subtypes for different configs.

crates/relayer/src/chain/client.rs
```diff
@@ -24,10 +24,18 @@ impl ClientSettings {
...
         // ForeignClient::build_create_client_and_send. Support for
         // heterogeneous chains is left for future revisions.
-        ClientSettings::Tendermint(cosmos::client::Settings::for_create_command(
-            options,
-            src_chain_config,
-            dst_chain_config,
-        ))
+        //
+        // TODO: extract Tendermint-related configs into a separate substructure
+        // that can be used both by CosmosSdkConfig and configs for nonSDK chains.
```
hdevalence, line 29: It seems like the right move here is to split out the Tendermint-related fields on the config into a Tendermint-specific config structure. Then that structure can be a field on both `CosmosSdkConfig` and other Tendermint-using chains (like Penumbra or Namada).

crates/relayer/src/chain/cosmos.rs
```diff
@@ -872,7 +872,17 @@ impl ChainEndpoint for CosmosSdkChain {
     type Time = TmTime;
     type SigningKeyPair = Secp256k1KeyPair;
 
+    fn id(&self) -> &ChainId {
+        &self.config.id
+    }
+
     fn bootstrap(config: ChainConfig, rt: Arc<TokioRuntime>) -> Result<Self, Error> {
+        #[allow(irrefutable_let_patterns)]
```
hdevalence, line 880: (Currently, the pattern is irrefutable, but it won't be in the future, and that should be handled now)

crates/relayer/src/chain/cosmos.rs
```diff
@@ -934,6 +944,16 @@ impl ChainEndpoint for CosmosSdkChain {
...
 
+    fn get_key(&mut self) -> Result<Self::SigningKeyPair, Error> {
+        // Get the key from key seed file
+        let key_pair = self
+            .keybase()
+            .get_key(&self.config.key_name)
+            .map_err(|e| Error::key_not_found(self.config().key_name.clone(), e))?;
+
+        Ok(key_pair)
+    }
```
hdevalence, line 955: cf #3641

crates/relayer/src/config.rs
```diff
@@ -31,7 +31,7 @@ use ibc_relayer_types::core::ics23_commitment::specs::ProofSpecs;
 use ibc_relayer_types::core::ics24_host::identifier::{ChainId, ChannelId, PortId};
 use ibc_relayer_types::timestamp::ZERO_DURATION;
 
-use crate::chain::ChainType;
+//use crate::chain::ChainType;
```
hdevalence, line 34: oops, will clean up after next round of discussion

crates/relayer/src/config.rs
```diff
@@ -150,9 +150,11 @@ impl Display for ExtensionOption {
 pub mod default {
     use super::*;
 
+    /*
     pub fn chain_type() -> ChainType {
         ChainType::CosmosSdk
     }
+     */
```
hdevalence, line 157: ditto above cleanup comment

crates/relayer/src/config.rs
```diff
@@ -556,14 +558,37 @@ pub enum EventSourceMode {
 
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
 #[serde(deny_unknown_fields)]
-pub struct ChainConfig {
+#[serde(tag = "type")]
+pub enum ChainConfig {
+    CosmosSdk(CosmosSdkConfig),
+}
```
hdevalence, line 564: As mentioned in the issue description for #3636, this will produce TOML like
```toml
[[chains]]
type = "CosmosSdk"
id = "chain_id"
rpc_addr = "http://127.0.0.1:26657/"
grpc_addr = "http://127.0.0.1:8080/"
```
The only difference from the current format is that the `type` field is now mandatory, and it must be `CosmosSdk`, not `cosmos-sdk` or `cOsMo-sSdk` or any other variation. I don't think that supporting arbitrary variations is a worthwhile complexity vs ergonomics tradeoff.

crates/relayer/src/config.rs
```diff
@@ -556,14 +558,37 @@ pub enum EventSourceMode {
...
+            Self::CosmosSdk(config) => &config.packet_filter,
+        }
+    }
+
+    pub fn max_block_time(&self) -> Duration {
+        match self {
+            Self::CosmosSdk(config) => config.max_block_time,
+        }
+    }
+}
```
hdevalence, line 584: These are the attributes that are used by Hermes outside of the Cosmos backend.

All of the other attributes are only used inside the Cosmos backend, so this is effectively a way to discover which parts of the configuration data is used by the relayer engine, and which parts are used only by the chain backend.

All other backends' configuration objects will need these fields.

An alternative representation would be to have the `ChainConfig` contain these common fields unconditionally, AND an inner enum with only the non-common configuration data. But I think this is actually worse, because now the backends can't cleanly model their own data: they need access to common fields like the chain id, and those fields would live in an overarching `ChainConfig` object. With the current structure, the common fields would be duplicated across different objects (the `PenumbraConfig` would have an `id` field as well), but this actually makes the code more modular, because each backend can just pass around and work with one object it defines (the cosmos logic can work with a `CosmosSdkConfig`, the penumbra logic can work with a `PenumbraConfig`, and so on).

crates/relayer/src/keyring.rs
```diff
@@ -287,9 +287,10 @@ impl KeyRing<Ed25519KeyPair> {
     }
 }
 
+// Why is this not a method on `ChainConfig`?
```
hdevalence, line 290: Is there any reason this isn't a method on `ChainConfig`, rather than a bare function defined in this other module?

hdevalence:
I was planning to leave self-review comments on the build failures that reveal issues with the CLI interface, but that will need to wait for the CI to run.

crates/relayer/src/config.rs
```diff
@@ -556,14 +558,37 @@ pub enum EventSourceMode {
...
+
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct CosmosSdkConfig {
```
hdevalence, line 588: This should move to the `cosmos/` backend but I left it for now for ease of initial commentary/review.

hdevalence:
I thought it'd be possible to add comments to check failures, but this doesn't seem to be the case. Instead I put them inline.

crates/relayer-cli/src/config.rs
```diff
@@ -87,15 +87,18 @@ pub fn validate_config(config: &Config) -> Result<(), Diagnostic<Error>> {
...
-
-        // Validate gas-related settings
-        validate_gas_settings(&c.id, c)?;
+        // TODO: why are these not methods on the config object?
+        // why are they defined only in the relayer-cli crate?
+        if let ChainConfig::CosmosSdk(cosmos_c) = c {
+            validate_trust_threshold(&cosmos_c.id, cosmos_c.trust_threshold)?;
+            // Validate gas-related settings
+            validate_gas_settings(&cosmos_c.id, cosmos_c)?;
+        }
```
hdevalence, line 101: Shouldn't config validation code be attached to the configs in the core crate, rather than being in the cli? Happy to move it if that sounds good.

crates/relayer-cli/src/commands/clear.rs
```diff
@@ -75,6 +75,11 @@ impl Override<Config> for ClearPacketsCmd {
         })?;
 
         if let Some(ref key_name) = self.key_name {
+            // Q: should the key name be required across chain types, meaning that
+            // key management is common to all chain types, or should key management
+            // be the responsibility of the backend? If key management is common
+            // across backends, how should it be agnostic to the key type? Can it
+            // just be an opaque byte string handled by the backend?
```
hdevalence, line 82: This ties in to https://github.com/informalsystems/hermes/issues/3641 

Perhaps the key manager should only support opaque byte strings, and let the rest be handled by the backend?

romac (commented):
Thank you so much for taking on this whole endeavor. What you've done so far looks really good to me.

Now we need to figure out a good story for backend-specific commands, eg. `keys`, as you've pointed out here and in #3641.

romac:
> Now we need to figure out a good story for backend-specific commands, eg. `keys`, as you've pointed out here and in #3641.

Maybe for now we can restrict these commands to `CosmosSdk` chains, by throwing a runtime error if used against another type of chain?

erwanor:
I have done some work propagating the configuration changes, but I'm leaving some of the refactor items as a last step (as previously agreed above).

I would like to pause and see if CI passes, tie up loose ends, before pushing on the specific TODOs (e.g. extracting the tendermint config) and doing another pass.

erwanor:
This took longer than anticipated, but the branch is finally passing all integration tests. Now, we should be able to move quickly on getting the remaining items done and ready for review by tomorrow:
- [x] small review issues
    - [x] remove dead code
    - [x] move `list_keys` to be a method on `ChainConfig`
    - [x] move config validation out of `relayer-cli`
 
when this is done:
- [ ] extract the tendermint fields into a struct and make the chain config embed it

and finally:
- [ ] move `CosmosSdkConfig` into the `cosmos/` backend
- [ ] move the cosmos specific config validation routines into the sdk backend
- [ ] move `detect_compatibility_mode` into the cosmos sdk impl

erwanor:
@romac I think this is ready for review! We can split out the tendermint config in future work since it has a small surface area, ditto for the websocket methods, for now we can just error if they are not supported.

romac (approved):
Amazing work 🚀 Thanks so much to both of you for seeing this through, much appreciated! 💐

romac:
@hdevalence Before I can merge this, can you please enable edits from maintainers on this PR?

hdevalence:
Hmm, the Github docs only talk about personal-owned forks, I'm not sure if it's possible because we made it from an org fork. But I can certainly add you to our org fork ? https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#

erwanor:
@romac I can't see that button, maybe it's only visible to the PR creator, but I added you as a collaborator to the forked repository: [invitations](https://github.com/penumbra-zone/hermes/invitations)

romac:
Perfect, thank you both! 🙏

seanchen1991:
Will these config changes require updates to the Hermes guide?

romac:
I thought they were all taken care of, but actually found three templates that we missed: https://github.com/informalsystems/hermes/pull/3679

Issue #3655, opened 2023-10-06
Improve documentation related to counterparty payee ICS29


ljoss17:
## Proposal

Version: Hermes 1.6.0

In order to receive ICS29 `recv_fee` the relayer needs to register its address on the counterparty chain. Hermes has a CLI which can be used if some specific address needs to be regisitered

```
hermes fee register-counterparty-payee
```

But if the relayers wallet address needs to be used a configuration exists, https://github.com/informalsystems/hermes/blob/v1.6.0/config.toml#L80,

```
# Auto register the counterparty payee on a destination chain to
# the relayer's address on the source chain. This can be used
# for simple configuration of the relayer to receive fees for
# relaying RecvPacket on fee-enabled channels.
# For more complex configuration, turn this off and use the CLI
# to manually register the payee addresses.
# [Default: false]
auto_register_counterparty_payee = false
```

This configuration is easily missed. In order to avoid this the CLI `fee register-counterparty-payee` could mention this configuration in the help message.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #3654, opened 2023-10-06
Improve `pull`/`poll` mode documentation


ljoss17:
## Proposal

Version: Hermes 1.6.0

Hermes accepts two event source configurations, push and pull mode.
For the pull mode, the mode has an alias and both `pull` and `poll` can be used https://github.com/informalsystems/hermes/blob/v1.6.0/crates/relayer/src/config.rs#L547.

This is not documented but the `poll` is mentioned in the guide, this could lead to some confusion.

This should be documented in:

* The guide: https://github.com/informalsystems/hermes/blob/v1.6.0/guide/src/documentation/configuration/configure-hermes.md#configuring-support-for-wasm-relaying
* The example configuration file: https://github.com/informalsystems/hermes/blob/v1.6.0/config.toml#L165

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

romac:
Let's also mention the WASM issue in the config file.

seanchen1991:
Could you link the issue you have in mind @romac?

romac:
Sorry I mean the note about the pull mode for wasm events that we have in the guide: https://github.com/informalsystems/hermes/blob/v1.6.0/guide/src/documentation/configuration/configure-hermes.md

PR #3661, opened 2023-10-16
Docs improvements

seanchen1991:
Closes: #3655 
Closes: #3654 

## Description

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


guide/src/documentation/configuration/configure-hermes.md
```diff
@@ -162,9 +162,12 @@ event_source = { mode = 'push', url = 'wss://hello:world@mydomain.com:26657/webs
 
 As of version 1.6.0, Hermes supports the relaying of wasm messages natively. This is facilitated by configuring
 Hermes to use pull-based relaying by polling for IBC events via the `/block_results` RPC endpoint. Set
-the `event_source` parameter to pull mode in `config.toml`:
+the `event_source` parameter to pull mode in `config.toml`, which can be done with either an inline table
+with the `mode` set to `'pull'`, along with the desired polling interval, or instead short-handed by just 
+specifying `'poll'`. 
 
 ```toml
+# The `'poll'` option is an alias for `{ mode = 'pull', interval = '1s' }`
```
romac, line 170: That is not correct, specifying just `'poll'` raises the following error:

> The Hermes configuration file at path '/Users/romac/.hermes/config.toml' is invalid, reason: parse error: invalid type: string "poll", expected internally tagged enum EventSourceMode for key `chains.event_source` at line 32 column 16


Not sure why we were mentioning that shorthand. Probably a mistake on my part.

Let's just remove mention of `poll` and the (not-working) shorthand.

romac (changes requested):

guide/src/documentation/configuration/configure-hermes.md
ljoss17: The `poll` is only an alias for `pull`. So the following configurations are equivalent:

* `{ mode = 'pull', interval = '1s' }`
* `{ mode = 'poll', interval = '1s' }`

See: https://github.com/informalsystems/hermes/blob/master/crates/relayer/src/config.rs#L549
As it is mentioned in the guide here https://hermes.informal.systems/documentation/configuration/configure-hermes.html#configuring-support-for-wasm-relaying I think we should either remove the alias and mention from the guide, either specify in the `config.toml` that both the mode `pull` and `poll` are the same.
What do you think?

romac: Agreed. Let's only mention `pull` but we can keep the alias internally in case people mistype it.

ljoss17 (approved):
Looks good to me!

PR #3694, opened 2023-11-13
Add ibc-go v8 to integration tests

ljoss17:
Closes: #3696

## Description

This PR adds ibc-go simapp `v8.0.0` to the chains used to run the integration tests. This requires some changes in the CLIs used to bootstrap the tests:

* The CLIs `add-genesis-account`,  `gentx` and `collect-gentxs` are under a subcommand `genesis`
* `query bank balances` doesn't take a `--denom` flag anymore and won't return the amount if the denom has 0 tokens
* Use `MsgIBCSoftwareUpgrade` instead of `UpgradeProposal` if the ibc-go version used by the chain is `v8.x.x` or higher

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


tools/test-framework/src/chain/cli/query.rs
```diff
@@ -120,3 +194,101 @@ pub fn query_cross_chain_query(
...
+                    "query",
+                    "auth",
+                    "module-accounts",
+                    "--output",
+                    "json",
+                ],
+            )?
+            .stdout;
+            let json_res: HashMap<String, serde_json::Value> =
+                serde_json::from_str(&output).map_err(handle_generic_error)?;
```
romac, line 250: One thing we could do here and in other places where we deal with JSON to avoid peeking manually into the JSON `Value` is to define a struct matching the JSON we expect, derive `Deserialize` for it and then convert the JSON into the struct.

Can be left for later, or skipped completely, will leave you judge of whether it's worth it or not.

tools/test-framework/src/util/proposal_status.rs
```diff
@@ -0,0 +1,79 @@
...
+            4 => Ok(Self::Rejected),
+            5 => Ok(Self::Failed),
+            _ => Err(Error::generic(eyre!(
+                "unknown value for proposal status: `{value}`"
+            ))),
+        }
+    }
+}
+
+impl TryFrom<String> for ProposalStatus {
```
romac, line 33: This is more idiomatic as a `std::str::FromStr` instance.

tools/integration-test/src/tests/client_upgrade.rs
```diff
@@ -241,28 +264,45 @@ impl BinaryChainTest for HeightTooHighClientUpgradeTest {
...
 
-        assert!(wait_for_chain_height(
-            &foreign_clients,
-            target_reference_application_height,
-            MAX_WAIT_FOR_CHAIN_HEIGHT
-        )
-        .is_ok());
+        info!("Assert that the chain upgrade proposal is eventually passed");
+
+        driver.value().assert_proposal_status(
```
romac, line 271: 💯

romac (changes requested):
Amazing work, especially on the tests! And thanks for the doc as well 🚀 

Left some minor comments, but otherwise it looks great!

tools/test-framework/src/chain/cli/query.rs
ljoss17: I agree, it is a good idea as the amount of methods where we extract the JSON output has increased. I think keeping this in a separate PR to clean the test-framework methods would be better. I will open an issue

crates/relayer/src/chain/cosmos.rs
```diff
@@ -1078,9 +1078,9 @@ impl ChainEndpoint for CosmosSdkChain {
         ChainConfig::CosmosSdk(self.config.clone())
     }
 
-    fn ibc_version(&self) -> Result<Option<semver::Version>, Error> {
+    fn ibc_version(&self) -> Result<(Option<semver::Version>, semver::Version), Error> {
```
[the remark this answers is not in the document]
ljoss17: Fixed in 5ab48454d3919abe0f3e2989b419f7f9642a180a

crates/relayer/src/upgrade_chain.rs
```diff
@@ -66,13 +67,54 @@ pub struct UpgradePlanOptions {
...
+            // anymore, use the Cosmos SDK version to determine if the legacy upgrade proposal
+            // has to be used
+            if ibc_version.major < 4 {
+                sdk_version.minor < 50
+            } else {
+                ibc_version.major < 8
+            }
+        }
+        None => sdk_version.minor < 50,
+    };
```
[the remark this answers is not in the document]
ljoss17: Fixed in 3287ddf163a7df1e07cf6a50f73a8d8cdd5bff35

tools/test-framework/src/util/proposal_status.rs
ljoss17: Fixed in a8434f6612acd624fb558609bdeb242dc9da89bb

tools/test-framework/src/chain/tagged.rs
```diff
@@ -80,6 +81,13 @@ pub trait TaggedChainDriverExt<Chain> {
         &self,
         recipient_address: &MonoTagged<Chain, &WalletAddress>,
     ) -> Result<json::Value, Error>;
+
+    /**
+       Tagged version of [`ChainDriver::query_auth_module`].
```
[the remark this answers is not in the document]
ljoss17: Fixed in b7a7af625c2b35700b1a73452a7e58481d03474f

.changelog/unreleased/features/ibc-relayer/3696-new-upgrade-proposal.md
```diff
@@ -0,0 +1,3 @@
+- Use `UpgradeProposal` or `MsgIBCSoftwareUpgrade` when upgrading
+  a chain depending on which ibc-go version of the chain.
+  ([\#3696](https://github.com/informalsystems/hermes/issues/3696))
```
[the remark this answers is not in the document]
ljoss17: Fixed in 056d7787b512e0068321c6008bf7e4b16ea524e0

crates/relayer/src/chain/cosmos.rs
```diff
@@ -1078,9 +1078,9 @@ impl ChainEndpoint for CosmosSdkChain {
         ChainConfig::CosmosSdk(self.config.clone())
     }
 
-    fn ibc_version(&self) -> Result<Option<semver::Version>, Error> {
+    fn version_specs(&self) -> Result<(Option<semver::Version>, semver::Version), Error> {
```
romac, line 1081: Here I meant to suggest both renaming the method and returning the full VersionSpecs struct instead of a tuple.

romac (changes requested):

crates/relayer/src/chain/cosmos.rs
ljoss17: Oh my bad, should I have it return the `Specs` struct https://github.com/informalsystems/hermes/blob/master/crates/relayer/src/chain/cosmos/version.rs#L37. Or would it be better to have a struct with all `Option<>` to avoid issues with non-SDK chains? E.g.:

```
pub struct VersionSpecs {
    pub cosmos_sdk: Option<semver::Version>,
    pub ibc_go: Option<semver::Version>,
    pub tendermint: Option<semver::Version>,
    pub comet: Option<semver::Version>,
}
```

romac: Oh right, yeah that's the way to go!

romac (approved):

PR #3702, opened 2023-11-17
evidence: More reliable evidence reporting + couple of bug fixes

romac:


romac:
It was confirmed that this patch fixes the issues that were seen on the Gaia RS testnet.

PR #3722, opened 2023-12-08
Improve `backlog` metrics

ljoss17:
Closes: #3723

## Description

This PR improves the `backlog` metrics by using the queried pending packet from packet clearing to update the values of the metrics.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/relayer/src/chain/counterparty.rs
+++ b/crates/relayer/src/chain/counterparty.rs
@@ -32,6 +32,7 @@ use crate::channel::ChannelError;
 use crate::client_state::IdentifiedAnyClientState;
 use crate::path::PathIdentifiers;
 use crate::supervisor::Error;
+use crate::telemetry;
 
 pub fn counterparty_chain_from_connection(
     src_chain: &impl ChainHandle,
@@ -501,6 +502,19 @@ pub fn unreceived_packets(
         &path.counterparty_port_id,
         &path.counterparty_channel_id,
     )?;
+
+    telemetry!(
+        update_backlog,
+        commit_sequences
+            .iter()
+            .map(|s| std::convert::Into::<u64>::into(*s))
+            .collect::<Vec<u64>>()
+            .clone(),
+        &counterparty_chain.id(),
+        &path.counterparty_channel_id,
+        &path.counterparty_port_id,
+        &chain.id()
+    );
 
     let packet_seq_nrs =
         unreceived_packets_sequences(chain, &path.port_id, &path.channel_id, commit_sequences)?;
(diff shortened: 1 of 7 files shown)


crates/telemetry/src/state.rs
```diff
@@ -1168,3 +1200,116 @@ impl AggregatorSelector for CustomAggregatorSelector {
         }
     }
 }
+
+#[cfg(feature = "telemetry")]
+mod tests {
```
romac, line 1205: This could probably use a `cfg(test)` attribute as well:

```suggestion
#[cfg(test)]
mod tests {
```

crates/telemetry/src/state.rs
```diff
@@ -968,6 +968,38 @@ impl TelemetryState {
...
+        // as they might have been relayed without the Hermes instance observing it
+        if let Some(path_backlog) = self.backlogs.get(&path_uid) {
+            let backlog = path_backlog.value();
+            let keys_to_remove: Vec<u64> = backlog
+                .iter()
+                .filter(|entry| !sequences.contains(entry.key()))
+                .map(|entry| *entry.key())
+                .collect();
+            for key in keys_to_remove.iter() {
+                self.backlog_remove(*key, chain_id, channel_id, port_id, counterparty_chain_id)
```
romac, line 998: This looks all fine to me, but since we are basically updating the backlog to only contain the given sequence numbers, I wonder if it would be possible to simply override it altogether instead, eg. by clearing it and calling `backlog_insert` for each sequence number?

But perhaps it's not simpler than what we are doing, and since the code is totally fine as is, feel free to ignore.

ljoss17: Looking at it more closely I see advantages and disadvantages in overriding the backlog completly.

Pros:
* If the instance missed a SendPacket the backlog will correctly display the number of pending packets
Cons:
* The timestamp when inserting an entry will be wrong for already observed entries, https://github.com/informalsystems/hermes/blob/8acaf5db9099aad13896011f2fbfe9075b8d338b/crates/telemetry/src/state.rs#L918

I think the best solution would be to override the backlog and update the `backlog_oldest_timestamp` to be `updated_backlog_timestamp` which would contain the timestamp of the last time an entry was inserted or removed.
What do you think about this solution?

crates/relayer/src/chain/counterparty.rs
```diff
@@ -502,6 +503,19 @@ pub fn unreceived_packets(
         &path.counterparty_channel_id,
     )?;
 
+    telemetry!(
+        update_backlog,
+        commit_sequences
+            .iter()
+            .map(|s| std::convert::Into::<u64>::into(*s))
```
romac, line 510:
```suggestion
            .map(|s| u64::from(*s))
```

PR #3756, opened 2024-01-08
Allow specifying sequences to clear as a list of ranges

romac:
See https://github.com/informalsystems/hermes/issues/3672 for the description of the feature and the syntax.

### DIFF
--- a/crates/relayer/src/link/cli.rs
+++ b/crates/relayer/src/link/cli.rs
@@ -1,4 +1,5 @@
 use std::convert::TryInto;
+use std::ops::Range;
 use std::thread;
 use std::time::{Duration, Instant};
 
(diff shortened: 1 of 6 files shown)


crates/relayer/src/link/cli.rs
```diff
@@ -92,25 +93,23 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Link<ChainA, ChainB> {
...
-            )
-            .map_err(LinkError::supervisor)?
-        } else {
-            (
-                maybe_sequences,
-                self.a_to_b.src_chain().query_latest_height().unwrap(),
-            )
-        };
+        // Find the sequence numbers of unreceived packets
+        let (mut sequences, src_response_height) = unreceived_packets(
```
romac, line 97: @ancazamfir I assumed that we want to allow half-open ranges (eg. `2000..` to mean all packets with sequence number 2000 or higher) or even just large ranges, we cannot enumerate all of them as there may be up to 2^64 of them. Therefore we have to query the unreceived packets/acks and then filter by the ranges we are given.

The downside is that we now do have to query for those even if we know exactly which packets we want to clear. The upside is that we won't attempt to query for packets which have already been relayed.

Does that trade-off work for you?

crates/relayer/src/link/cli.rs
```diff
@@ -159,27 +158,26 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Link<ChainA, ChainB> {
...
-            )
-            .map_err(LinkError::supervisor)?
-        } else {
-            Some((
-                maybe_sequences,
-                self.a_to_b.src_chain().query_latest_height().unwrap(),
-            ))
-        }) else {
+        // Find the sequence numbers of unreceived acknowledgements
+        let Some((mut sequences, src_response_height)) = unreceived_acknowledgements(
```
romac, line 162: Same as https://github.com/informalsystems/hermes/pull/3756/files#r1444722668

crates/relayer/src/link/cli.rs
ancazamfir: I think it should be fine. We can think of further optimizations in the future if needed.

ancazamfir (approved):
Very cool! Thanks @romac

PR #3765, opened 2024-01-10
Luca joss/memo and receiver filtering

ljoss17:
Closes: #XXX

## Description

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/relayer-cli/src/commands/clear.rs
+++ b/crates/relayer-cli/src/commands/clear.rs
@@ -135,6 +135,8 @@ impl Runnable for ClearPacketsCmd {
         let opts = LinkParameters {
             src_port_id: self.port_id.clone(),
             src_channel_id: self.channel_id.clone(),
+            max_memo_size: config.mode.packets.max_memo_size.to_usize(),
+            max_receiver_size: config.mode.packets.max_receiver_size.to_usize(),
         };
 
         let fwd_link = match Link::new_from_opts(chains.src.clone(), chains.dst, opts, false, false)
--- a/crates/relayer-cli/src/commands/clear.rs
+++ b/crates/relayer-cli/src/commands/clear.rs
@@ -135,8 +135,8 @@ impl Runnable for ClearPacketsCmd {
         let opts = LinkParameters {
             src_port_id: self.port_id.clone(),
             src_channel_id: self.channel_id.clone(),
-            max_memo_size: config.mode.packets.max_memo_size.to_usize(),
-            max_receiver_size: config.mode.packets.max_receiver_size.to_usize(),
+            max_memo_size: config.mode.packets.ics20_max_memo_size.to_usize(),
+            max_receiver_size: config.mode.packets.ics20_max_receiver_size.to_usize(),
         };
 
         let fwd_link = match Link::new_from_opts(chains.src.clone(), chains.dst, opts, false, false)
--- a/crates/relayer-types/src/core/ics04_channel/packet.rs
+++ b/crates/relayer-types/src/core/ics04_channel/packet.rs
@@ -6,7 +6,8 @@ use ibc_proto::ibc::applications::transfer::v2::FungibleTokenPacketData as RawPa
 use ibc_proto::ibc::core::channel::v1::Packet as RawPacket;
 
 use super::timeout::TimeoutHeight;
-use crate::applications::transfer::packet::PacketData as APacketData;
+use crate::applications::transfer::error::Error as Ics20Error;
+use crate::applications::transfer::packet::PacketData as Ics20PacketData;
 use crate::core::ics04_channel::error::Error;
 use crate::core::ics24_host::identifier::{ChannelId, PortId};
 use crate::timestamp::{Expiry::Expired, Timestamp};
@@ -186,18 +187,11 @@ impl Packet {
         max_memo_size: usize,
         max_receiver_size: usize,
     ) -> Result<(usize, usize), Error> {
-        let any_packet: RawPacketData = match serde_json::from_slice(&self.data) {
-            Ok(v) => v,
-            Err(e) => {
-                return Err(Error::app_module(e.to_string()));
-            }
-        };
-        let packet: APacketData = match any_packet.try_into() {
-            Ok(p) => p,
-            Err(e) => {
-                return Err(Error::app_module(e.to_string()));
-            }
-        };
+        let any_packet: RawPacketData =
+            serde_json::from_slice(&self.data).map_err(Error::serde_json_error)?;
+        let packet: Ics20PacketData = any_packet
+            .try_into()
+            .map_err(|e: Ics20Error| Error::decode_ics20_packet(e.to_string()))?;
         let memo_size = if let Some(memo) = &packet.memo {
             if memo.len() > max_memo_size {
                 memo.len()
--- a/crates/relayer-types/src/core/ics04_channel/packet.rs
+++ b/crates/relayer-types/src/core/ics04_channel/packet.rs
@@ -2,9 +2,11 @@ use std::str::FromStr;
 
 use serde_derive::{Deserialize, Serialize};
 
+use ibc_proto::ibc::applications::transfer::v2::FungibleTokenPacketData as RawPacketData;
 use ibc_proto::ibc::core::channel::v1::Packet as RawPacket;
 
 use super::timeout::TimeoutHeight;
+use crate::applications::transfer::packet::PacketData as APacketData;
 use crate::core::ics04_channel::error::Error;
 use crate::core::ics24_host::identifier::{ChannelId, PortId};
 use crate::timestamp::{Expiry::Expired, Timestamp};
@@ -177,6 +179,39 @@ impl Packet {
             && dst_chain_ts.check_expiry(&self.timeout_timestamp) == Expired;
 
         height_timed_out || timestamp_timed_out
+    }
+
+    pub fn validate_fields(
+        &self,
+        max_memo_size: usize,
+        max_receiver_size: usize,
+    ) -> Result<(usize, usize), Error> {
+        let any_packet: RawPacketData = match serde_json::from_slice(&self.data) {
+            Ok(v) => v,
+            Err(e) => {
+                return Err(Error::app_module(e.to_string()));
+            }
+        };
+        let packet: APacketData = match any_packet.try_into() {
+            Ok(p) => p,
+            Err(e) => {
+                return Err(Error::app_module(e.to_string()));
+            }
+        };
+        let memo_size = if let Some(memo) = &packet.memo {
+            if memo.len() > max_memo_size {
+                memo.len()
+            } else {
+                0
+            }
+        } else {
+            0
+        };
+        if packet.receiver.to_string().len() > max_receiver_size {
+            Ok((memo_size, packet.receiver.to_string().len()))
+        } else {
+            Ok((memo_size, 0))
+        }
     }
 }
 
--- a/crates/relayer/src/config/types.rs
+++ b/crates/relayer/src/config/types.rs
@@ -125,6 +125,10 @@ pub mod max_tx_size {
             Ok(Self(value))
         }
 
+        pub fn unsafe_new(value: usize) -> Self {
+            Self(value)
+        }
+
         pub fn max() -> Self {
             Self(Self::MAX_BOUND)
         }
--- a/crates/relayer/src/link.rs
+++ b/crates/relayer/src/link.rs
@@ -33,6 +33,8 @@ pub use relay_path::{RelayPath, Resubmit};
 pub struct LinkParameters {
     pub src_port_id: PortId,
     pub src_channel_id: ChannelId,
+    pub max_memo_size: usize,
+    pub max_receiver_size: usize,
 }
 
 pub struct Link<ChainA: ChainHandle, ChainB: ChainHandle> {
@@ -43,9 +45,16 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Link<ChainA, ChainB> {
     pub fn new(
         channel: Channel<ChainA, ChainB>,
         with_tx_confirmation: bool,
+        max_memo_size: usize,
+        max_receiver_size: usize,
     ) -> Result<Self, LinkError> {
         Ok(Self {
-            a_to_b: RelayPath::new(channel, with_tx_confirmation)?,
+            a_to_b: RelayPath::new(
+                channel,
+                with_tx_confirmation,
+                max_memo_size,
+                max_receiver_size,
+            )?,
         })
     }
 
@@ -169,7 +178,12 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Link<ChainA, ChainB> {
                 .map_err(LinkError::relayer)?;
         }
 
-        Link::new(channel, with_tx_confirmation)
+        Link::new(
+            channel,
+            with_tx_confirmation,
+            opts.max_memo_size,
+            opts.max_receiver_size,
+        )
     }
 
     /// Constructs a link around the channel that is reverse to the channel
@@ -182,6 +196,8 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Link<ChainA, ChainB> {
         let opts = LinkParameters {
             src_port_id: self.a_to_b.dst_port_id().clone(),
             src_channel_id: self.a_to_b.dst_channel_id().clone(),
+            max_memo_size: self.a_to_b.max_memo_size,
+            max_receiver_size: self.a_to_b.max_receiver_size,
         };
         let chain_b = self.a_to_b.dst_chain().clone();
         let chain_a = self.a_to_b.src_chain().clone();
--- a/crates/relayer/src/link/relay_path.rs
+++ b/crates/relayer/src/link/relay_path.rs
@@ -1389,6 +1389,13 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
         dst_info: &ChainStatus,
         height: Height,
     ) -> Result<(Option<Any>, Option<Any>), LinkError> {
+        crate::time!(
+            "build_recv_or_timeout_from_send_packet_event",
+            {
+                "src_channel": event.packet.source_channel,
+                "dst_channel": event.packet.destination_channel,
+            }
+        );
         let timeout = self.build_timeout_from_send_packet_event(event, dst_info)?;
         if timeout.is_some() {
             Ok((None, timeout))
--- a/crates/relayer/src/link/relay_path.rs
+++ b/crates/relayer/src/link/relay_path.rs
@@ -108,12 +108,17 @@ pub struct RelayPath<ChainA: ChainHandle, ChainB: ChainHandle> {
     // transactions if [`confirm_txes`] is true.
     pending_txs_src: PendingTxs<ChainA>,
     pending_txs_dst: PendingTxs<ChainB>,
+
+    pub max_memo_size: usize,
+    pub max_receiver_size: usize,
 }
 
 impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
     pub fn new(
         channel: Channel<ChainA, ChainB>,
         with_tx_confirmation: bool,
+        max_memo_size: usize,
+        max_receiver_size: usize,
     ) -> Result<Self, LinkError> {
         let src_chain = channel.src_chain().clone();
         let dst_chain = channel.dst_chain().clone();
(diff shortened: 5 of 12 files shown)


crates/relayer/src/config/types.rs
```diff
@@ -175,6 +179,164 @@ pub mod max_tx_size {
...
+
+    use serde::de::Unexpected;
+    use serde::{de::Error as _, Deserialize, Deserializer, Serialize, Serializer};
+
+    impl<'de> Deserialize<'de> for Ics20MaxReceiverSize {
+        fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+        where
+            D: Deserializer<'de>,
+        {
+            let value = usize::deserialize(deserializer)?;
```
romac, line 234: Let's use `byte_unit::Byte` like we do for `max_grpc_decoding_size`.

crates/relayer/src/config/types.rs
```diff
@@ -175,6 +179,164 @@ pub mod max_tx_size {
...
+
+    use serde::de::Unexpected;
+    use serde::{de::Error as _, Deserialize, Deserializer, Serialize, Serializer};
+
+    impl<'de> Deserialize<'de> for Ics20MaxMemoSize {
+        fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+        where
+            D: Deserializer<'de>,
+        {
+            let value = usize::deserialize(deserializer)?;
```
romac, line 313: Let's use `byte_unit::Byte` like we do for `max_grpc_decoding_size`.

crates/relayer/src/link.rs
```diff
@@ -43,9 +45,16 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Link<ChainA, ChainB> {
     pub fn new(
         channel: Channel<ChainA, ChainB>,
         with_tx_confirmation: bool,
+        max_memo_size: usize,
```
romac, line 48: Let's maybe pass and store the whole `LinkParameters` struct here instead of each field? What do you think?

crates/relayer/src/link/relay_path.rs
```diff
@@ -108,12 +108,17 @@ pub struct RelayPath<ChainA: ChainHandle, ChainB: ChainHandle> {
...
+
+    pub max_memo_size: usize,
+    pub max_receiver_size: usize,
 }
 
 impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
     pub fn new(
         channel: Channel<ChainA, ChainB>,
         with_tx_confirmation: bool,
+        max_memo_size: usize,
```
romac, line 120: Same here

crates/relayer-types/src/core/ics04_channel/packet.rs
```diff
@@ -178,6 +181,32 @@ impl Packet {
 
         height_timed_out || timestamp_timed_out
     }
+
+    pub fn validate_fields(
```
romac, line 185: Let's add some doc comments to explain what it does, with a couple examples perhaps?

romac: Also, the semantics of this function are a bit weird.

Instead of returning the size when it's too big and 0 when it's okay, how about returning a bespoke `Error` type like:

```rust
enum ValidationError {
  MemoTooBig { size: usize, max: usize },
  ReceiverTooBig { size: usize, max: usize },
  Other(Error),
}
```

and returning the actual sizes when all is okay?

crates/relayer/src/link/relay_path.rs
```diff
@@ -1381,11 +1389,31 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+                .validate_fields(self.max_memo_size, self.max_receiver_size)
+            {
+                Ok((0, 0)) => Ok((self.build_recv_packet(&event.packet, height)?, None)),
+                Ok((memo_size, receiver_size)) => {
+                    warn!("memo and/or receiver fields were bigger than maximum allowed. Memo field {memo_size}, maximum allowed {}. Receiver field {receiver_size}, maximum allowed {}", self.max_memo_size, self.max_receiver_size);
+                    Ok((None, None))
+                }
+                Err(e) => {
+                    warn!("error validating fields, will simply relay message: {e}");
+                    Ok((self.build_recv_packet(&event.packet, height)?, None))
```
romac, line 1414: Will this warning show for all non-ICS-20 packets? That would be a bit noisy.

We should only show a warning/error if we see that we have an ICS-20 packet and:
- are unable to decode it
- the memo or the receiver fields are too large

What do you think?

crates/relayer-types/src/core/ics04_channel/packet.rs
```diff
@@ -178,6 +181,32 @@ impl Packet {
...
+        &self,
+        max_memo_size: usize,
+        max_receiver_size: usize,
+    ) -> Result<(usize, usize), Error> {
+        let any_packet: RawPacketData =
+            serde_json::from_slice(&self.data).map_err(Error::serde_json_error)?;
+        let packet: Ics20PacketData = any_packet
+            .try_into()
+            .map_err(|e: Ics20Error| Error::decode_ics20_packet(e.to_string()))?;
```
romac, line 194: This is probably overkill, what do you think about just looking at the fields of the RawPacketData instead?

crates/relayer-types/src/core/ics04_channel/packet.rs
```diff
@@ -178,6 +181,32 @@ impl Packet {
...
         height_timed_out || timestamp_timed_out
     }
+
+    pub fn validate_fields(
+        &self,
+        max_memo_size: usize,
+        max_receiver_size: usize,
+    ) -> Result<(usize, usize), Error> {
+        let any_packet: RawPacketData =
+            serde_json::from_slice(&self.data).map_err(Error::serde_json_error)?;
```
romac, line 191: Here we should probably not return an error if the packet data is not an ICS-20 packet, but rather return `Ok((0, 0))` or something like that.

romac (changes requested):

crates/relayer/src/link.rs
ljoss17: Good point, it is cleaner this way

crates/relayer-types/src/core/ics04_channel/packet.rs
ljoss17: Good point I will remove this

crates/relayer/src/config/types.rs
ljoss17: Fixed in 43a79a35c048b1eeef4d50114b2f2e060154fb4e

crates/relayer/src/config/types.rs
ljoss17: Fixed in 43a79a35c048b1eeef4d50114b2f2e060154fb4e

crates/relayer/src/link.rs
ljoss17: Fixed in f2b3a4fc46891f7f047dfb3a6336bf2a31409b59

crates/relayer/src/link/relay_path.rs
ljoss17: Fixed in f2b3a4fc46891f7f047dfb3a6336bf2a31409b59

crates/relayer-types/src/core/ics04_channel/packet.rs
```diff
@@ -178,6 +178,19 @@ impl Packet {
 
         height_timed_out || timestamp_timed_out
     }
+
+    /// This method will return if the memo and receiver fields are valid, e.g. not too big.
+    /// If it fails to decode the ICS20 PacketData then there is no need to validate fields
+    /// so the method will return true.
+    pub fn are_fields_valid(&self, max_memo_size: u64, max_receiver_size: u64) -> bool {
```
romac, line 185:
```suggestion
    /// If this packet data is an ICS-20 packet, check whether or not the memo and receiver fields are under the given max size.
    /// Returns `true` if this packet data is not an ICS-20 or fails to decode as an ICS-20 packet.
    pub fn check_ics20_fields_size(&self, max_memo_size: usize, max_receiver_size: usize) -> bool {
```

crates/relayer-cli/src/commands/clear.rs
```diff
@@ -135,6 +135,8 @@ impl Runnable for ClearPacketsCmd {
         let opts = LinkParameters {
             src_port_id: self.port_id.clone(),
             src_channel_id: self.channel_id.clone(),
+            max_memo_size: config.mode.packets.ics20_max_memo_size.get_bytes(),
+            max_receiver_size: config.mode.packets.ics20_max_receiver_size.get_bytes(),
```
romac, line 139:
```suggestion
            max_memo_size: config.mode.packets.ics20_max_memo_size.get_bytes() as usize,
            max_receiver_size: config.mode.packets.ics20_max_receiver_size.get_bytes() as usize,
```

crates/relayer-types/src/core/ics04_channel/packet.rs
```diff
@@ -178,6 +178,19 @@ impl Packet {
...
+        match serde_json::from_slice::<RawPacketData>(&self.data).map_err(Error::serde_json_error) {
+            Ok(packet) => {
+                packet.memo.len() <= max_memo_size as usize
+                    || packet.receiver.to_string().len() <= max_receiver_size as usize
```
romac, line 189: Shouldn't this be a conjunction?

```suggestion
                    && packet.receiver.to_string().len() <= max_receiver_size as usize
```

romac: Also, let's avoid an allocation here:

```suggestion
                    && packet.receiver.as_ref().len() <= max_receiver_size as usize
```

romac (changes requested):

crates/relayer-types/src/core/ics04_channel/packet.rs
ljoss17: Oh right, since I inverted the assertion it needs to be `&&`, good catch!

crates/relayer/src/link/relay_path.rs
```diff
@@ -1381,11 +1389,27 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+        } else if event
+            .packet
+            .are_fields_valid(self.max_memo_size, self.max_receiver_size)
+        {
             Ok((self.build_recv_packet(&event.packet, height)?, None))
+        } else {
+            debug!(
+                "memo and/or receiver field are too big for packet {:#?}",
+                event.packet
+            );
```
romac, line 1411: Rather than outputting the whole packet here (which may be extremely large and flood the console), let's instead log the memo and receiver sizes and the max allowed sizes.

For this you might have to change the validation function again to return both sizes.

romac: We can perhaps log the whole packet with `trace!`. What do you think @ljoss17  @ancazamfir?

config.toml
```diff
@@ -79,6 +79,18 @@ tx_confirmation = false
 # [Default: false]
 auto_register_counterparty_payee = false
 
+# Set the maximum size for the memo field in ICS20 packets.
+# If the size of the memo field is bigger than the configured
+# one, the packet will not be relayed.
+# [Default: 32768 (32Kb)]
+#ics20_max_memo_size = 32768
```
romac, line 86:
```suggestion
# [Default: "32KiB"]
#ics20_max_memo_size = "32KiB"
```

config.toml
```diff
@@ -79,6 +79,18 @@ tx_confirmation = false
...
+
+# Set the maximum size for the receiver field in ICS20 packets.
+# If the size of the receiver field is bigger than the configured
+# one, the packet will not be relayed.
+# [Default: 32768 (2Kb)]
+#ics20_max_receiver_size = 2048
```
romac, line 92:
```suggestion
# [Default: "2KiB"]
#ics20_max_receiver_size = "2KiB"
```

crates/relayer/src/link/relay_path.rs
ljoss17: I wanted to keep the validation method as simple as possible, and since it's a debug log I would be inclined to have more details in the logs. I could also put it as trace.

Or would you still prefer to have only the field sizes in the log outputs?

crates/relayer-cli/src/commands/clear.rs
ljoss17: Fixed in 0fd448cd8e91d9a825357093c9694cc2e817e6af

crates/relayer/src/link/relay_path.rs
romac: it would be nice to have the sequence number and field sizes in the debug log when we choose to not relay the packet. Will surely help debugging along the line.

Let's dump the whole packet data in `trace!` mode.

ljoss17: I introduced more advanced logs in this commit 9af4e6d0688748f09be91d09b22c7f272e82fb8c.
With debug logs when one or both fields fails and trace logs of the packet when there is any failure.

config.toml
```diff
@@ -79,6 +79,18 @@ tx_confirmation = false
...
+# If the size of the memo field is bigger than the configured
+# one, the packet will not be relayed.
+# [Default: "32KiB"]
+#ics20_max_memo_size = "32KiB"
+
+# Set the maximum size for the receiver field in ICS20 packets.
+# If the size of the receiver field is bigger than the configured
+# one, the packet will not be relayed.
+# [Default: "2KiB"]
+#ics20_max_receiver_size = "2KiB"
```
ancazamfir, line 92: Can you add config on how to achieve old behaviour, i.e. no checks?

crates/relayer/src/link/relay_path.rs
```diff
@@ -1381,11 +1389,55 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+            }
+        );
         let timeout = self.build_timeout_from_send_packet_event(event, dst_info)?;
         if timeout.is_some() {
             Ok((None, timeout))
         } else {
-            Ok((self.build_recv_packet(&event.packet, height)?, None))
+            match event
+                .packet
+                .are_fields_valid(self.max_memo_size, self.max_receiver_size)
```
ancazamfir, line 1405: Why are we not checking this earlier, i.e. for timeouts also? Same question for `build_ack_from_recv_event()`? Maybe even in the supervisor?

config.toml
ljoss17: Good point, I'll slightly change the configuration to have something along the lines of:

`ics20_max_receiver_limit = { enabled = true, size = "2KiB" }`
`ics20_max_memo_limit = { enabled = true, size = "32KiB" }`

ancazamfir:
Left a few comments. In general I am wondering why we don't apply the checks for all msgs that contain a packet, i.e. recv, timeout and ack.

crates/relayer/src/link/relay_path.rs
ancazamfir: Thinking about it more, while it would be nice to have this in the supervisor it would not cover the startup and periodic clearing. So I guess we can check in `relay_path` but let's decide if we want to also apply for the other (than recv) message types.

romac:
Good catch, we should indeed do that! Completely slipped my mind…

crates/relayer-types/src/core/ics04_channel/packet.rs
```diff
@@ -178,6 +199,48 @@ impl Packet {
...
+
+    /// If this packet data is an ICS-20 packet, check whether or not the memo and receiver
+    /// fields are under the given max size.
+    /// Returns `true` if this packet data is not an ICS-20 or fails to decode as an ICS-20 packet.
+    /// Return a ValidationResult::DecodeFailure this packet data is not an ICS-20 or fails to decode
+    /// as an ICS-20 packet.
+    /// Returns ValidationResult::Valid if both the memo and receiver field are valid.
+    /// Returns ValidationResult containing information on the failed fields if one or both fields fail
+    /// the validation.
+    pub fn are_fields_valid(
```
romac, line 211: Let's pull this out of the `Packet` struct as it is specific to ICS-20 and have it instead within the ics20 module as a free-standing function, perhaps named `check_ics20_fields_length`

crates/relayer/src/link/relay_path.rs
```diff
@@ -1381,11 +1389,55 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+            {
+                ValidationResult::Valid => {
+                    Ok((self.build_recv_packet(&event.packet, height)?, None))
+                }
+                ValidationResult::MemoTooBig { size, max } => {
+                    trace!("packet: {:#?}", event.packet);
```
romac, line 1411:
```suggestion
                    trace!("ICS-20 packet failed fields length check: {:#?}", event.packet);
```

crates/relayer/src/link/relay_path.rs
```diff
@@ -1381,11 +1389,55 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+                    debug!("memo field of packet with sequence {} is too big. Memo field size {size}, maximum configured size {max}", event.packet.sequence);
+                    Ok((None, None))
+                }
+                ValidationResult::ReceiverTooBig { size, max } => {
+                    trace!("packet: {:#?}", event.packet);
```
romac, line 1416:
```suggestion
                    trace!("ICS-20 packet failed fields length check: {:#?}", event.packet);
```

crates/relayer/src/link/relay_path.rs
```diff
@@ -1381,11 +1389,55 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+                    debug!("receiver field of packet with sequence {} is too big. Receiver field size {size}, maximum configured size {max}", event.packet.sequence);
+                    Ok((None, None))
+                }
+                ValidationResult::BothTooBig {
+                    memo_size,
+                    memo_max,
+                    receiver_size,
+                    receiver_max,
+                } => {
+                    trace!("packet: {:#?}", event.packet);
```
romac, line 1426:
```suggestion
                    trace!("ICS-20 packet failed fields length check: {:#?}", event.packet);
```

crates/relayer/src/link/relay_path.rs
```diff
@@ -1381,11 +1389,55 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+                    debug!(
+                        "memo and receiver field of packet with sequence {} are too big. \
+                        Memo field size {memo_size}, maximum configured size {memo_max}, \
+                        receiver field size {receiver_size}, maximum configured size {receiver_max}",
+                        event.packet.sequence
+                    );
+                    Ok((None, None))
+                }
+                ValidationResult::DecodeFailure { details } => {
+                    trace!("packet: {:#?}", event.packet);
```
romac, line 1436: I think we can remove this in that case

crates/relayer/src/link/relay_path.rs
```diff
@@ -1381,11 +1389,55 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+                    trace!("failed to decode packet or packet isn't an ICS20 packet, details: {details}");
```
romac, line 1437: Is there a way to determine whether or not the packet was supposed be a valid ICS-20 packet?

ljoss17:
I refactored the memo and receiver field configuration and validation in this commit 451b5781f809fd6bf13b1c66340a67a0e4bc6b27.

I added the option to disable the filter by having the configuration format as following: `{ enabled = true, size = "32KiB" }`

The reason why I put the `Ics20FieldSizeLimit` struct in the `relayer` crate is because I feel like validation linked to the configuration shouldn't be in the `relayer-types` crate.

I only applied the verification https://github.com/informalsystems/hermes/pull/3765/commits/451b5781f809fd6bf13b1c66340a67a0e4bc6b27#diff-072b872f1fbf4ea17384cecde28b35ee339faa9ce969ff9f767647a071da1e36R1405 as I wanted to have your opinion on this solution @ancazamfir @romac. If this seems like a good solution we can think more in detail to where we want to validate these fields

crates/relayer/src/link/relay_path.rs
romac: The check is now performed directly in `generate_operational_data` for all events which pertain to an ICS-04 packet: https://github.com/informalsystems/hermes/pull/3765/commits/3d5687eb534dc72a94c08e95e7afdfb6492d3d22

romac:
The check is now performed directly in `generate_operational_data` for all events which pertain to an ICS-04 packet: https://github.com/informalsystems/hermes/commit/3d5687eb534dc72a94c08e95e7afdfb6492d3d22

romac:
The guide template checker is failing but that's unrelated to this PR. Will fix out of band.

ancazamfir (approved):
Looks good! Thanks @ljoss17!

PR #3768, opened 2024-01-11
Remove assumption that ICS-04 packet data is a valid UTF-8 string

romac:
Closes: #3770

## Description

This PR removes the assumption that the ICS-04 packet data is valid UTF-8, by using the `packet_data_hex` attribute instead of the deprecated `packet_data` attribute.

Relying on the `packet_data` attribute enforces a UTF-8 encoded payload (eg. JSON), disallowing Protobuf-encoded payloads which we are starting to see in the wild.

The `packet_data` atttribute [has been deprecated][0] in favor of `packet_data_hex` since IBC-Go v1.0.0.

[0]: https://github.com/cosmos/ibc-go/blob/fadf8f2b0ab184798d021d220d877e00c7634e26/CHANGELOG.md?plain=1#L1417

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


romac:
Here is how I tested this PR:

1. I [modified ibc-go](https://github.com/informalsystems/ibc-go/commit/817244b0c82ff023f1c8636b60d718d6bc986f25)`transfer`[ module](https://github.com/informalsystems/ibc-go/commit/817244b0c82ff023f1c8636b60d718d6bc986f25) to change the type of the ICS-20 `memo` from `string` to `[]byte`.

2. I used a [modified version](https://github.com/informalsystems/hermes/compare/romac/packet-data-hex...romac/non-utf8-memo?expand=1) of Hermes based on this branch to submit a ICS-20 transfer message with a non-UTF-8 memo, and got the following error:

  ```
  estimate_gas: failed to simulate tx. propagating error to caller: gRPC call `send_tx_simulate` failed with status: status: Internal, message: "failed to decode Protobuf message: EventAttribute.value: Event.attributes: Result.events: SimulateResponse.result: invalid string value: data is not UTF-8 encoded
  ```

  This error comes from somewhere in Comet/SDK, where they assume that an event attribute   value is a valid UTF-8 encoded string, and there is nothing we can do about that.

3. Then I [modified ibc-go again](https://github.com/informalsystems/ibc-go/commit/a2a3c7d61a11dbc85a174e45e11fbae94f5887b5) to not emit an event attribute for the memo and the deprecated `packet_data` attribute, and then error went away, my packet was successfully relayed.

---

So to summarize, while this PR indeed enables Hermes to relay packets with non-UTF-8 packet data, in practice we will never see those because:

1. ibc-go emits the whole packet data as a string in the deprecated `packet_data` event attribute, without ensuring it's actually valid UTF-8.
2. There is an assumption somewhere in the Cosmos SDK or in CometBFT that event attributes must be valid UTF-8 data, leading to the error above.

romac:
> There is an assumption somewhere in the Cosmos SDK or in CometBFT that event attributes must be valid UTF-8 data, leading to the error above

Actually I was wrong. There is indeed [an assumption in Comet v0.37+](https://github.com/tendermint/tendermint/issues/6403) that event attributes are valid UTF-8 strings but that's not where the error above comes from.

In fact this error is yielded not by the chain but by `prost` when it tries to decode the `SimulateResponse` it gets from the chain. This `SimulateResponse` proto has a field `result` which itself contains an [`events`](https://github.com/cosmos/ibc-proto-rs/blob/da8dbee83ed034fffb028a1966cc4c69ae17f053/src/prost/cosmos.base.abci.v1beta1.rs#L150) field of type `Vec<tendermint_proto::abci::Event>`. 

This `tendermint_proto::abci::Event` is an [alias](https://github.com/informalsystems/tendermint-rs/blob/main/proto/src/tendermint.rs#L4) for `tendermint_proto::v0_38::abci::Event` which contains a vector of `EventAttribute`, defined as:

```rust
pub struct EventAttribute {
    #[prost(string, tag = "1")]
    pub key: ::prost::alloc::string::String,
    #[prost(string, tag = "2")]
    pub value: ::prost::alloc::string::String,
    /// nondeterministic
    #[prost(bool, tag = "3")]
    pub index: bool,
}
```

where `key` and `value` are of type `String`, which enforces valid UTF-8.

Hence why decoding the simulation response fails.

I modified `ibc-proto-rs` to instead use `tendermint_proto::v0_34::abci::EventAttribute`, whose `key` and `value` are `Vec<u8>` and was able to successfully relay a packet with non UTF-8 memo, even in the presence of the `packet_data` event attribute in the simulation response.

romac:
Until ibc-go stops emitting the `packet_data` and `packet_ack` attributes, the best we can do is to modify the proto types for `EventAttribute` to use the CometBFT v0.34 definition, which has those specified as `bytes` instead of `string`. WIP PR for ibc-proto-rs [here](https://github.com/cosmos/ibc-proto-rs/pull/180).

maurolacy:
Hi,
I can confirm this solves the issue we were having with using Protobuf in the IBC channel.

Any plans to merge this?

Thanks,

romac:
Glad to hear! We'll include this in the next Hermes release.

PR #3773, opened 2024-01-12
feat: add support for `MsgChannelUpgradeTimeout` and integration test

crodriguezvega:
This PR adds the logic to support the submission of `MsgChannelUpgradeTimeout`. This message can be used to cancel an upgrade when the counterparty is in `FLUSHING` and it hasn't flushed all the packets on its end before the upgrade timeout expired.

The PR also adds an integration test (not fully working yet) where:
- chain A executes upgrade INIT.
- chain B executes upgrade TRY.
- IBC transfer message is submitted on chain A, but the packet is not relayed to chain B (so that we have an in-flight packet and chain A moved to `FLUSHING` on ACK).
- chain A executes ACK.
- Wait for upgrade timeout to pass on chain A.
- chain B executes timeout (cancels upgrade and channel moves back to `OPEN`).

The test is not working yet because the check on successful ACK execution fails (although I see in the logs of chain A that ACK succeeded).

cc: @ljoss17 

Closes: #XXX

## Description

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/relayer-types/src/core/ics04_channel/events.rs
+++ b/crates/relayer-types/src/core/ics04_channel/events.rs
@@ -9,9 +9,11 @@ use crate::core::ics04_channel::channel::Ordering;
 use crate::core::ics04_channel::error::Error;
 use crate::core::ics04_channel::packet::Packet;
 use crate::core::ics04_channel::packet::Sequence;
+use crate::core::ics04_channel::timeout::Timeout;
 use crate::core::ics04_channel::version::Version;
 use crate::core::ics24_host::identifier::{ChannelId, ConnectionId, PortId};
 use crate::events::{Error as EventError, IbcEvent, IbcEventType};
+use crate::timestamp::Timestamp;
 use crate::utils::pretty::PrettySlice;
 
 /// Channel event attribute keys
@@ -145,6 +147,7 @@ pub struct UpgradeAttributes {
     pub upgrade_version: Version,
     pub upgrade_sequence: Sequence,
     pub upgrade_ordering: Ordering,
+    pub upgrade_timeout: Option<Timeout>,
 }
 
 impl UpgradeAttributes {
@@ -609,6 +612,7 @@ impl From<UpgradeInit> for UpgradeAttributes {
             upgrade_version: ev.upgrade_version,
             upgrade_sequence: ev.upgrade_sequence,
             upgrade_ordering: ev.upgrade_ordering,
+            upgrade_timeout: None,
         }
     }
 }
@@ -711,6 +715,7 @@ impl From<UpgradeTry> for UpgradeAttributes {
             upgrade_version: ev.upgrade_version,
             upgrade_sequence: ev.upgrade_sequence,
             upgrade_ordering: ev.upgrade_ordering,
+            upgrade_timeout: None,
         }
     }
 }
@@ -813,6 +818,7 @@ impl From<UpgradeAck> for UpgradeAttributes {
             upgrade_version: ev.upgrade_version,
             upgrade_sequence: ev.upgrade_sequence,
             upgrade_ordering: ev.upgrade_ordering,
+            upgrade_timeout: None,
         }
     }
 }
@@ -915,6 +921,7 @@ impl From<UpgradeConfirm> for UpgradeAttributes {
             upgrade_version: ev.upgrade_version,
             upgrade_sequence: ev.upgrade_sequence,
             upgrade_ordering: ev.upgrade_ordering,
+            upgrade_timeout: None,
         }
     }
 }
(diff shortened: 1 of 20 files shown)


crates/relayer-types/src/core/ics04_channel/events.rs
```diff
@@ -145,6 +147,7 @@ pub struct UpgradeAttributes {
     pub upgrade_version: Version,
     pub upgrade_sequence: Sequence,
     pub upgrade_ordering: Ordering,
+    pub upgrade_timeout: Option<Timeout>,
```
crodriguezvega, line 150: The `upgrade_timeout` attribute is only present in the upgrade timeout event.

crates/relayer-types/src/core/ics04_channel/timeout.rs
```diff
@@ -188,8 +190,8 @@ impl<'de> Deserialize<'de> for TimeoutHeight {
 /// A composite of timeout height and timeout timestamp types, useful for when
 /// performing a channel upgrade handshake, as there are cases when only timeout
 /// height is set, only timeout timestamp is set, or both are set.
-#[derive(Clone, Debug, PartialEq, Eq)]
-pub enum UpgradeTimeout {
```
crodriguezvega, line 192: I renamed this enum to match the naming in ibc-go.

crates/relayer-types/src/core/ics04_channel/timeout.rs
```diff
@@ -204,31 +206,73 @@ pub enum UpgradeTimeout {
...
+        if split.len() != 2 {
+            return Err(TimeoutError::invalid_timeout(value.to_owned()));
         }
+
+        // only timeout timestamp are supported at the moment
+        return split[1]
+            .parse::<Timestamp>()
+            .map(|timestamp| Timeout::Timestamp(timestamp))
+            .map_err(|e| TimeoutError::invalid_timestamp(value.to_owned(), e));
     }
```
crodriguezvega, line 267: I haven't tested this parsing logic yet... If anyone has any recommendations or examples of how to do it, please let me know!

crates/relayer-types/src/core/ics04_channel/timeout.rs
```diff
@@ -204,31 +206,73 @@ pub enum UpgradeTimeout {
...
+            { timeout: String }
+            |e| { format_args!("invalid timeout {0}", e.timeout) },
+    }
+}
+
+impl FromStr for Timeout {
+    type Err = TimeoutError;
+
+    fn from_str(value: &str) -> Result<Self, Self::Err> {
+        let split: Vec<&str> = value.split(' ').collect();
```
crodriguezvega, line 256: I believe the format of the string will be `{<number>-<height>} <timestamp>`, but need to double check, so that's why I split on the empty space.

crates/relayer-types/src/core/ics04_channel/timeout.rs
romac: Do you have examples of the timeout format? If so then we can add a module at the end of the file with a unit test, eg.

```rust
#[cfg(test)]
mod tests {
  use super::*;

  #[test]
  fn parse_timeout_timestamp() {
     let tests = [
       ("timestamp1", Timeout::new(..., ...)),
       ("timestamp2", Timeout::new(..., ...)),
       // etc.
     ];

     for (timestamp, expected) in tests {
       let parsed = Timeout::from_str(timestamp).unwrap();
       assert_eq!(parsed, expected);
     }
  }
}
```

crodriguezvega: Thanks, @romac! I will add something like this!

romac (approved):

crates/relayer/src/channel.rs
```diff
@@ -300,14 +301,13 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
...
         channel: WorkerChannelObject,
-        height: Height,
     ) -> Result<(Channel<ChainA, ChainB>, State), ChannelError> {
         let (a_channel, _) = chain
             .query_channel(
                 QueryChannelRequest {
                     port_id: channel.src_port_id.clone(),
                     channel_id: channel.src_channel_id.clone(),
-                    height: QueryHeight::Specific(height),
+                    height: QueryHeight::Latest,
```
ljoss17, line 310: It seems that when getting a NewBlock we were looking at the state at the height of the NewBlock in order to determine the state of the channel, e.g. which step of the handshake it is at.
This doesn't seem right, I would expect the latest state to be the important one.
Can you confirm this @romac @ancazamfir ?

crates/relayer/src/channel.rs
```diff
@@ -393,7 +393,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
                     QueryChannelRequest {
                         port_id: a_channel.remote.port_id.clone(),
                         channel_id: a_channel.remote.channel_id.clone().unwrap(),
-                        height: QueryHeight::Specific(height),
+                        height: QueryHeight::Latest,
```
ljoss17, line 396: Same as https://github.com/informalsystems/hermes/pull/3773/files#r1457122914

PR #3776, opened 2024-01-13
Updating guide/src/index.md title

Ursulafe:
It currently appears like that : 
Capture d’écran 2024-01-13 à 02 31 22

Seems like the automatic fetch of the version doesn't work. Deleted it, but can manually add the version, like it's done in this file for exemple : https://github.com/informalsystems/hermes/blob/master/guide/README.md

Let me know


romac:
This is working as expected, see https://hermes.informal.systems

PR #3797, opened 2024-01-19
chore: rename upgrade event attributes to match ibc-go

crodriguezvega:
We [renamed a few upgrade event attributes in ibc-go](https://github.com/cosmos/ibc-go/pull/5627) and also [split the upgrade timeout attribute in two (height and timestamp)](https://github.com/cosmos/ibc-go/pull/5612).

Closes: #XXX

## Description

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/relayer-types/src/core/ics02_client/height.rs
+++ b/crates/relayer-types/src/core/ics02_client/height.rs
@@ -11,7 +11,7 @@ use ibc_proto::ibc::core::client::v1::Height as RawHeight;
 use crate::core::ics02_client::error::Error;
 use crate::core::ics24_host::identifier::ChainId;
 
-#[derive(Copy, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[derive(Copy, Clone, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub struct Height {
     /// Previously known as "epoch"
     revision_number: u64,
@@ -30,6 +30,11 @@ impl Height {
             revision_number,
             revision_height,
         })
+    }
+
+    /// Yields the Height as a string.
+    pub fn as_str(&self) -> String {
+        format!("{}-{}", self.revision_number, self.revision_height)
     }
 
     pub fn from_tm(height: tendermint::block::Height, chain_id: &ChainId) -> Self {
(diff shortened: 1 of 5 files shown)


crates/relayer-types/src/core/ics02_client/height.rs
```diff
@@ -32,6 +32,11 @@ impl Height {
         })
     }
 
+    /// Yields the Height as a string.
+    pub fn as_str(&self) -> String {
+        format!("{}-{}", self.revision_number, self.revision_height)
+    }
+
```
romac, line 39: This method is unnecessary, instead you can call `.to_string()` on `Height` to get the same result.

This `to_string(&self) -> String` is method provided by the [`ToString`](https://doc.rust-lang.org/stable/std/string/trait.ToString.html) trait, which is automatically implemented for all types that implement the `Display` trait (which is the case for `Height`, see lines 148-152 below). See the linked doc for more info.

crates/relayer-types/src/timestamp.rs
```diff
@@ -153,6 +153,11 @@ impl Timestamp {
             _ => false,
         }
     }
+
+    /// Yields the Timestamp as a string.
+    pub fn as_str(&self) -> String {
+        format!("{}", self.nanoseconds())
+    }
```
romac, line 160: I would rather we call `timestamp.nanoseconds().to_string()` where needed rather than add a bespoke `to_str` method whose output does not match with the `Display` instance below.

Alternatively we can rename this method to `to_nanoseconds_string` but that seems a bit too ad-hoc.

romac (changes requested):

crates/relayer-types/src/core/ics02_client/height.rs
crodriguezvega: Cool, I removed the method and used `to_string()`.

crates/relayer-types/src/timestamp.rs
crodriguezvega: Also removed the method and used `.nanoseconds().to_string()` instead.

romac (approved):

romac:
Thanks @crodriguezvega, I made one last change while fixing the remaining clips warnings:

Instead of doing `.map_or_else(|| T::default(), |x| x)`, one can do the shorter alternative: `unwrap_or_default()`.

PR #3809, opened 2024-01-23
Publish release automatically when pushing version tag

romac:
Will be useful for releases going forward. I did 1.8.0 manually so we'll see how it goes for v1.8.1.

Releases will be published by GitHub user `ibcbot`, just like the ibc-rs crates.

- [x] Add `ibcbot` as owner of the crates to publish
- [x] Add crates.io token as an action secret for use in the workflow


ljoss17 (approved):
This is great, thanks for doing it!!

Issue #3894, opened 2024-03-14
Hermes can't proceed to next block if it can't parse current block


tkxkd0159:
## Summary
Hermes can't proceed to next block if it can't parse current block

## Problem Definition

I found that Hermes can't proceed to retrieve next block to collect event if it can't interpret current block. Hermes couldn't parse some of my chain blocks which includes `validator_updates` because my chain has different public key type name. So I have to restart my Hermes to initialize current block height by `abci_info`. In conclusion, I hope Hermes can recover itself even if Hermes can't parse some blocks which it includes data Hermes doesn't expect.

## Proposal
I suggest that Hermes try again a certain number of times and then change it to try exploring the next block. Because If it fails after several attempts, it is clear that Hermes cannot parse that block. So it's better to move on to the next block.

## Acceptance Criteria

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

romac:
Thanks for opening this! Do you have any logs exhibiting this problem that we could look at?

tkxkd0159:
This is the log I've got! It queries for same block height even if it's failed many times.
```
2024-03-14T04:48:02.043602Z DEBUG ThreadId(60) event_source.rpc{chain.id=ebony-2}: outgoing request url=https://ebony-rpc-int.line-apps.com/ body={
"jsonrpc": "2.0",
"id": "8c2cad17-b114-421e-8049-2bfa989fe46b",
"method": "abci_info",
"params": null
}
2024-03-14T04:48:02.046327Z DEBUG ThreadId(60) event_source.rpc{chain.id=ebony-2}: incoming response status=200 OK body={"jsonrpc":"2.0","id":"8c2cad17-b114-421e-8049-2bfa989fe46b","result":{"response":{"data":"Finschia","version":"2.0.1","last_block_height":"44774443","last_block_app_hash":"9luNLJzV5chdSX2TKlo8Ag30SvrAW32uT3fpByomvq0="}}}
2024-03-14T04:48:02.046386Z DEBUG ThreadId(60) event_source.rpc{chain.id=ebony-2}: outgoing request url=https://ebony-rpc-int.line-apps.com/ body={
"jsonrpc": "2.0",
"id": "0af222ed-7ea9-4919-b8ce-015bec8d4fac",
"method": "block_results",
"params": {
[...]

PR #3906, opened 2024-03-19
Proceed to next block after a few retries if Hermes can't parse current block during event sourcing

tkxkd0159:
Closes: #3894 

## Description
* In the case of chains that are modified part of cosmos-sdk, Hermes may not be able to interpret a particular block, so in this case, instead of staying in that block, Hermes should move on to the next block.
* It may be a temporary network failure, so retry in `fetch_batches` loop as much as we set it in advance and then move on to the next block.

c.c. @romac 

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


romac (approved):
Excellent 🚀 Thanks so much!

PR #3956, opened 2024-04-18
ICS 031 event extraction perf improvement and refactor

romac:
Let's see if we can re-use the existing infrastructure we use for other events via `append_events`.


ljoss17 (approved):
Thanks for the improvements!! Looks good to me!

Issue #3950, opened 2024-04-16
Add support for multiple connection hops in `tx chan-open-init`


romac:
### [`tx chan-open-init`](https://github.com/informalsystems/hermes/blob/master/crates/relayer-cli/src/commands/tx/channel.rs)

- [x] Add `--connection-hops=conn1,conn2,conn3` parameter to `tx chan-open-init`
- [x] Define `ConnectionHop` struct
- [x] Define `ConnectionHops` struct initially with single field `hops: Vec<ConnectionHop>`
- [x] Fetch alls connections and store them in `ConnectionHops`
  - [x] 1. Fetch destination connection (`--dst-connection`) on the destination
  - [x] 2. Fetch the client state for the client underlying that connection on the destination chain
  - [x] 3. Get the counterparty chain id from the client state
  - [x] 4. Do this for all connection hops, from last to first
- [x] Add `connection_hops: ConnectionHops` field to [`Channel` struct](https://github.com/informalsystems/hermes/blob/master/crates/relayer/src/channel.rs?plain=1#L166)
- [x] Add support for connection hops to [`Channel::build_chan_open_init_and_send`](https://github.com/informalsystems/hermes/blob/master/crates/relayer/src/channel.rs?plain=1#L875)
  - [x] Change `connection_id` field on struct [`Attributes`](https://github.com/informalsystems/hermes/blob/962cbd11e28659383dc997da63129e0d8d700835/crates/relayer-types/src/core/ics04_channel/events.rs#L35) to a `Vec<ConnectionId>`
  - [x] Parse `connection_id`s from `channel_open_init` event into `Attributes.connection_id`
- [x] Write unit tests for `tx chan-open-init` with `--connection-hops`
- [ ] Write integration test for single hop channel

```rust
struct ConnectionHop {
  pub connection: IdentifiedConnectionEnd,
  pub client_state: AnyClientState,
  pub reference_chain_id: ChainId,
}
```

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

joaotav:
Closed by 0ec88bd.

PR #3980, opened 2024-05-03
Add support for multiple connection hops in `tx chan-open-init`

joaotav:
Closes: #3950 

## Description

- [x] https://github.com/informalsystems/hermes/issues/3950
______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


crates/relayer-cli/src/commands/tx/channel.rs
```diff
@@ -108,6 +113,16 @@ pub struct TxChanOpenInitCmd {
...
+
+    #[clap(
+        long = "connection-hops",
+        value_name = "CONNECTION_HOPS",
+        use_delimiter = true,
+        value_delimiter = ',',
+        help = "The comma separated identifiers of the intermediate connections between /
+        a source and destination chain for a multi-hop channel (optional)"
+    )]
+    conn_hop_ids: Option<Vec<ConnectionId>>,
```
romac, line 125: I wonder if we should re-use the `ConnectionIds` type you have introduced here instead of `Vec<ConnectionId>`?

```suggestion
        help = "A list of identifiers of the intermediate connections between /
        a source and destination chain for a multi-hop channel, separated by slashes, eg. connection-1/connection-0 (optional)"
    )]
    conn_hop_ids: Option<ConnectionIds>,
```

crates/relayer-cli/src/commands/tx/channel.rs
```diff
@@ -131,8 +146,84 @@ impl Runnable for TxChanOpenInitCmd {
...
+                        connection_id.clone(),
+                        hop_connection.clone(),
+                    ),
+                    reference_chain_id: hop_conn_client_state.chain_id().clone(),
+                });
+            }
+        }
+
+        // Instantiate ConnectionHops from a vec of ConnectionHop
+        let connection_hops = Some(ConnectionHops::new(assembled_hops));
```
romac, line 224: Should this be `None` if there are no connection hops?

romac (changes requested):
Nice work! Left some minor comments and suggestions inline

crates/relayer-types/src/core/ics24_host/identifier.rs
```diff
@@ -268,6 +268,7 @@ impl FromStr for ConnectionId {
     type Err = ValidationError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let s = s.trim();
```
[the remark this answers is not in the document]
joaotav: You're right, I can't think of a reason why this is necessary. This was probably out of habit.

crates/relayer-cli/src/commands/tx/channel.rs
joaotav: From my understanding, there should always be at least one connection hop. In the case of a single-hop channel, this would be `--dst-connection` pointing to the other end of the channel. Therefore, we always add `--dst-connection` to `connection_hops` and then we process the additional hops if they have been specified. 

From previous manual testing with `gaiad`, the chain requires at least one connection hop to be specified to open the channel. Moreover, in the [ICS33 spec here](https://github.com/cosmos/ibc/blob/main/spec/core/ics-033-multi-hop/README.md?plain=1#L62) it says that a multi-hop packet should be identifiable by the existence of **more than one** connection ID in the `connection_hops` field, implying that single-hop channels have one connection ID in `connection_hops`.

Perhaps this does not need to be an `Option` given that it will always be `Some()` here. At this stage however, having it as an option allows us to pass `None` to `connection_hops` during `Channel` instantiation in other places where we don't have access to the hops yet.

crates/relayer/src/channel.rs
```diff
@@ -857,7 +868,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
             State::Init,
             self.ordering,
             counterparty,
-            vec![self.dst_connection_id().clone()],
+            self.connection_hops.as_ref().unwrap().connection_ids(),
```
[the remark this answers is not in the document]
romac: The call to `.unwrap()` is failing in the test suite for non-multihop channels. I think it's because we need to use `veg![self.dst_connection_id()]` if `self.connection_hops` is `None`.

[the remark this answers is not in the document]
romac:
```suggestion
            self.connection_hops.as_ref().map(|hops| hops.connection_ids()).unwrap_or_else(|| vec![self.dst_connection_id().clone()]),
```

crates/relayer-cli/src/commands/tx/channel.rs
romac: As discussed during the call, let's only pass `Some(_)` if there are multiple hops and `None` if there is a single one.

crates/relayer/src/channel.rs
```diff
@@ -167,6 +167,7 @@ pub struct Channel<ChainA: ChainHandle, ChainB: ChainHandle> {
     pub ordering: Ordering,
     pub a_side: ChannelSide<ChainA>,
     pub b_side: ChannelSide<ChainB>,
+    pub connection_hops: Option<ConnectionHops>,
```
romac, line 170: At a later stage we might want to make this non-optional.

crates/relayer-types/src/core/ics24_host/identifier.rs
```diff
@@ -292,6 +292,47 @@ impl PartialEq<str> for ConnectionId {
     }
 }
 
+// A Vec<ConnectionId> wrapper for the multi-hop channels PoC implementation
+#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+pub struct ConnectionIds(Vec<ConnectionId>);
+
+impl ConnectionIds {
+    pub fn as_vec(&self) -> &[ConnectionId] {
```
romac, line 300:
```suggestion
    pub fn as_slice(&self) -> &[ConnectionId] {
```

crates/relayer-types/src/core/ics24_host/identifier.rs
```diff
@@ -292,6 +292,47 @@ impl PartialEq<str> for ConnectionId {
...
+impl FromStr for ConnectionIds {
+    type Err = ValidationError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let connection_ids = s
+            .split('/')
+            .map(|conn_id| conn_id.parse())
+            .collect::<Result<Vec<ConnectionId>, _>>()?;
+
+        assert!(!connection_ids.is_empty()); // FIXME: Address this in the future to properly ensure that the vector is not empty
```
romac, line 318: Reminder to address to FIXME.

crates/relayer-types/src/core/ics24_host/identifier.rs
```diff
@@ -292,6 +292,50 @@ impl PartialEq<str> for ConnectionId {
...
+    type Err = ValidationError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let connection_ids = s
+            .split('/')
+            .map(|conn_id| conn_id.parse())
+            .collect::<Result<Vec<ConnectionId>, _>>()?;
+
+        if connection_ids.is_empty() {
+            return Err(ValidationError::empty());
```
romac, line 319: Let's define a new error variant for this, perhaps something like `EmptyConnectionHops` with a message like `cannot parse empty list of connection ids into connection hops`

crates/relayer-types/src/core/ics24_host/identifier.rs
```diff
@@ -292,6 +292,50 @@ impl PartialEq<str> for ConnectionId {
...
+impl Display for ConnectionIds {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), FmtError> {
+        let s = self
+            .0
+            .iter()
+            .map(|connection_id| connection_id.as_str())
+            .collect::<Vec<_>>()
+            .join(",");
+
+        write!(f, "{}", s)
```
romac, line 335: We can avoid allocating a new `Vec` before joining the strings by using the [`itertools` crate](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.join)

```suggestion
        use itertools::Itertools;

        write!("{}", self.0.iter().join(","))
```

romac (approved):
Great work! 🚀

crates/relayer-cli/src/commands/tx/channel.rs
```diff
@@ -131,8 +145,96 @@ impl Runnable for TxChanOpenInitCmd {
...
+        }
+
+        // FIXME: For now, pass Some(_) to connection_hops if there are multiple hops and None if there is a single one.
+        // This allows us to keep using existing structs as they are defined (with the single `connection_id` field) while also including
+        // the new `connection_hops` field. When multiple hops are present, pass Some(_) to connection_hops and use that.
+        // When a single hop is present, pass None to connection_hops and use the connection_id stored in `ChannelSide`.
+        let connection_hops = match assembled_hops.len() {
+            0 => Output::error("At least one connection hop is required for opening a channel.")
+                .exit(),
+            1 => None,
```
ancazamfir, line 233: Why is this required? We could still use the same structure but always use the `connection_hops` when we create the handshake message.
(would get rid of `.unwrap_or_else(|| vec![self.dst_connection_id().clone()]),` further down the stack (`crates/relayer/src/channel.rs::build_chan_open_init()`)

romac: Yes we absolutely plan on doing this at a later stage, but for now we wanted to keep as much of the existing code as is to the extent possible.

Issue #3987, opened 2024-05-14
Add support for multiple connection hops in `tx chan-open-try`


romac:
Part of: #3940

### [`tx chan-open-try`](https://github.com/informalsystems/hermes/blob/master/crates/relayer-cli/src/commands/tx/channel.rs#L169)
- [x] Replace call to `tx_chan_cmd!` macro with an inline function on [`TxChanOpenTryCmd`](https://github.com/informalsystems/hermes/blob/25f547595ff71a4910a68889c625ad7b66180de9/crates/relayer-cli/src/commands/tx/channel.rs#L237)
- [x] In the [`run()`](https://github.com/informalsystems/hermes/blob/25f547595ff71a4910a68889c625ad7b66180de9/crates/relayer-cli/src/commands/tx/channel.rs#L236) function for `TxChanOpenTryCmd`, build the reverse channel path, from `--dst-chain` to `--src-chain`, by leveraging information available on the existing `ChannelEnd` that was opened with `tx chan-open-init`
- [x] Add multihop support to [Channel::build_chan_open_try_and_send](https://github.com/informalsystems/hermes/blob/master/crates/relayer/src/channel.rs#L1063)
  - [x] Add `build_multihop_chan_open_try()` method to struct `Channel` 
  - [x] Add `update_channel_path_clients()` method to struct `Channel`
    - [x] Add struct `MultihopProofHeights` to store the heights that are necessary when constructing multihop proofs.
  - [x] Add `build_multihop_proofs()` method to struct `Channel`

### Tests
- [ ] Write integration test for single hop channel
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

joaotav:
Closed by 81123d2.

PR #3997, opened 2024-05-23
Add support for multiple connection hops in `tx chan-open-try`

joaotav:
Closes: #3987 

## Description

- [ ] https://github.com/informalsystems/hermes/issues/3987
______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:


crates/relayer/src/channel.rs
```diff
@@ -273,6 +278,7 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
                 chain,
                 connection.client_id().clone(),
                 connection_id.clone(),
+                None, //FIXME: Unsure what to add here ('None' for now), can we get the hops from the event?
```
joaotav, line 281: Reminder for us to review this in the future.

crates/relayer/src/channel.rs
```diff
@@ -283,11 +289,11 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
                 counterparty_chain,
                 connection.counterparty().client_id().clone(),
                 counterparty_connection_id.clone(),
+                None, //FIXME: Unsure what to add here ('None' for now), can we get the hops from the event?
```
joaotav, line 292: Reminder for us to address this FIXME in the future.

crates/relayer/src/channel.rs
```diff
@@ -1144,16 +1229,17 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> Channel<ChainA, ChainB> {
...
-        self.dst_chain()
-            .query_connection(
-                QueryConnectionRequest {
-                    connection_id: self.dst_connection_id().clone(),
-                    height: QueryHeight::Latest,
-                },
-                IncludeProof::No,
-            )
-            .map_err(|e| ChannelError::query(self.dst_chain().id(), e))?;
+        // Connection must exist on destination -- THIS WAS ALREADY CHECKED ON fn run(), if we are not storing
```
joaotav, line 1232: It just occurred to me that the reason why this check exists in here as well, even though it is also present in `fn run()` for `TxChanOpenTryCmd`, is because the `build_chan_open_try()` method may be called from somewhere other than the CLI functions, in which case this check may not have been performed yet, correct me if I'm wrong. In that case we should keep this.

romac: Yes exactly, so let's keep it as a sanity check.

crates/relayer-types/src/proofs.rs
```diff
@@ -116,3 +116,34 @@ impl ConsensusProof {
...
+        }
+    }
+
+    pub fn query_height(&self) -> Height {
+        self.query_height
+    }
+
+    pub fn consensus_height(&self) -> Option<&Height> {
+        self.consensus_height.as_ref()
+    }
```
romac, line 148: I think we can do away with getters and make the fields `pub` instead, since this is a simple struct which maintains no invariants.

joaotav: Done!

romac (approved):
Nice job! The code looks great!

Left a couple nitpicks, didn't see anything else to comment on :) Feel free to merge after addressing those.

PR #4008, opened 2024-05-30
Add compatibility for ICS20 v2

### LINKED ISSUE #4006: Add ICS20 v2 multiple denom compatibility
## Summary

The new v2 of ICS20 introduces multiple denom transfers, https://github.com/cosmos/ibc-go/issues/5793. This requires some changes in Hermes.

____

#### For Admin Use

- [ ] Not duplicate issue

### LINKED ISSUE #4098: ICS20 v2 Protobuf encoding
## Summary

ICS20 v2 will encode packet data using Protobuf instead of JSON https://github.com/cosmos/ibc-go/issues/6630.

Hermes will need to adapt the verification of memo and receiver fields to decode using Protobuf if the transfer is from ICS20 v2, https://github.com/informalsystems/hermes/blob/v1.10.0/crates/relayer/src/link/relay_path.rs#L1956.

Furthermore if the packet is a forward packet, the `destination_memo`, https://github.com/cosmos/ibc-go/blob/v9.0.0-beta.0/proto/ibc/applications/transfer/v2/packet.proto#L48, should be checked. If not the standard `memo` should be checked, https://github.com/cosmos/ibc-go/blob/v9.0.0-beta.0/proto/ibc/applications/transfer/v2/packet.proto#L38.

____

#### For Admin Use

- [ ] Not duplicate issue

ljoss17:
Closes: #4006
Closes: #4098

## Description

This PR adds the compatibility for ICS20 v2. This includes:

* Conditional deserialization of PacketData if the channel version is `ics20-1` or `ics20-2`
* Update channel Version struct to more easily specify `ics20-1` or `ics20-2`
* Add test to transfer multiple denom in a single message

## Current status of the PR - 30.05.2024

This PR is blocked until a tagged release of ibc-go with ICS20 v2 feature, required for:

* [ ] Creating new ibc-proto-rs release with ICS20 v2 protos
* [ ] Adding a new `simd` to Cosmos Nix to run the Hermes ICS20 v2 test

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


crates/relayer-types/src/applications/transfer/mod.rs
```diff
@@ -23,4 +23,4 @@ pub const MODULE_ID_STR: &str = "transfer";
 pub const PORT_ID_STR: &str = "transfer";
 
 /// ICS20 application current version.
-pub const VERSION: &str = "ics20-1";
+pub const VERSION: &str = "ics20";
```
romac, line 26:
```suggestion
pub const VERSION_PREFIX: &str = "ics20";
```

crates/relayer-types/src/applications/transfer/msgs/transfer.rs
```diff
@@ -41,6 +41,7 @@ pub struct MsgTransfer<C = Coin> {
     pub timeout_timestamp: Timestamp,
     /// optional memo
     pub memo: Option<String>,
+    pub tokens: Vec<C>,
```
romac, line 44: Let's add a doc comment here.

I also wonder if we can't just keep this `tokens` field only instead of having both `token` and `tokens`?

crates/relayer-types/src/applications/transfer/packet.rs
```diff
@@ -45,3 +47,60 @@ impl From<PacketData> for RawPacketData {
         }
     }
 }
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(try_from = "RawPacketDataV2", into = "RawPacketDataV2")]
+pub struct PacketDataV2 {
+    pub token: Vec<PrefixedCoin>,
```
romac, line 54:
```suggestion
    pub tokens: Vec<PrefixedCoin>,
```

crates/relayer-types/src/core/ics04_channel/version.rs
```diff
@@ -52,6 +52,10 @@ impl Version {
             })
             .unwrap_or(false)
     }
+
+    pub fn is_ics20_v2(&self) -> bool {
+        self.0.contains("ics20-2")
```
romac, line 57:
```suggestion
        self.0 == "ics20-2"
```

crates/relayer-types/src/applications/transfer/msgs/transfer.rs
ljoss17: If I'm not mistaken, a channel with version `ics20-1` will require the `token` field since the previous proto didn't have the `tokens` field https://github.com/cosmos/ibc-go/blob/v8.3.1/proto/ibc/applications/transfer/v1/tx.proto#L51.

crates/relayer-types/src/core/ics04_channel/version.rs
ljoss17: This might fail to catch an ICS29 channel with ICS20 v2 since the format would be:

```json
{
  "fee_version": "ics29-1",
  "app_version": "ics20-2"
}
```

crates/relayer-types/src/applications/transfer/packet.rs
ljoss17: done in fc41cdceb1b3993243cbf406faf415bc098f2599

crates/relayer-types/src/applications/transfer/mod.rs
ljoss17: done in fc41cdceb1b3993243cbf406faf415bc098f2599

crates/relayer-types/src/applications/transfer/msgs/transfer.rs
romac: Sure, but this is a domain type, so we can instead opt to only keep a single field here (ie. `tokens`), which would only ever contain a single `Coin` for v1 and potentially multiple for v2. That is, as long as we can still tell somehow which version we are dealing with, maybe from the context?

Or are we using the fact that `token` is `Some` or `None` to figure out which version it is?

ljoss17: Oh this is a good point. I tried to keep the domain type as close to the proto struct as possible since the crate could be used by someone else.

We don't use the fact that `token` is `Some` or `None` to determine the version, but the implementation `From<MsgTransfer> for RawMsgTransfer` might be tricky.
https://github.com/informalsystems/hermes/blob/d05bc5a468904816b669fcc52fe3db64a4315a24/crates/relayer-types/src/applications/transfer/msgs/transfer.rs#L93
If there is only 1 token in the domain type's `tokens`, we would need to find a way to determine if we are building a v1 `RawMsgTransfer` which will have the `token: Some(token)` or the v2 with `tokens: vec!(token)`. But maybe ibc-go has some type of conversion since there is already a conversion for the PacketData and this isn't an issue, https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/internal/convert/convert.go

crates/relayer-types/src/core/ics04_channel/version.rs
romac: Ah right it can contain json, then we should probably try to deserialize the version as a JSON `Value` and look at the `app_version` field, and otherwise fall back to comparing by equality.

crates/relayer-types/src/applications/transfer/msgs/transfer.rs
romac: How about making the `MsgTransfer` domain type an enum, with one variant for v1 and one for v2?

ljoss17: I updated `MsgTransfer` to be an enum in this commit https://github.com/informalsystems/hermes/pull/4008/commits/8ac128e63f7af43e84857c098a49017a69c18f7f

crates/relayer-types/src/core/ics04_channel/version.rs
ljoss17: Used JSON deserialization to verify the ics20 versions in this commit https://github.com/informalsystems/hermes/pull/4008/commits/f68c5e8cadf0d9c4f884b19006ce2d167e95565e

gjermundgaraba:
Given that ICS20 v2 was retracted and will not be released either, I will close the PR (cleaning up a bit).

Issue #4038, opened 2024-06-12
Allow specifying the topology used for Hermes tests


ljoss17:
## Proposal

The current state of Hermes, e4be3e0b0f18eb702e324fb860831db894a4b7a4, creates fully connected chains when running tests with more than 2 chains.

This impacts the run time of some tests, for example `tests::forward::forward_hop_transfer::test_ibc_forward_hop_transfer` can take up to `682.193` seconds to pass. Most of the time is spent in the connection and channel handshake as the bootstrap is fully connecting the 4 chains.
This is not necessary, if the chains are connected linearly, A -> B -> C -> D, the test would still work.

It should be possible to specify the topology used for nary chain tests in order to reduce the run time of some tests.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #4039, opened 2024-06-12
Specify topology for n-ary chain tests

ljoss17:
Closes: #4038 

## Description

This PR refactors the bootstrapping of n-ary chain tests in order to use the specified topology. Only linear and fully connected are currently supported, and if the topology is not specified the linear one is used.

This also facilitates the implementation of new topologies if required.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/tools/test-framework/src/types/topology.rs
+++ b/tools/test-framework/src/types/topology.rs
@@ -35,7 +35,7 @@
 
    Example: Cyclic Topology
 
-   The cylic topology is similar to the linear topology with the addition of the first and
+   The cyclic topology is similar to the linear topology with the addition of the first and
    last chain being connected as well.
    For chains A, B, C and D, the clients are created as follows:
 
--- a/tools/test-framework/src/types/topology.rs
+++ b/tools/test-framework/src/types/topology.rs
@@ -51,7 +51,7 @@ pub enum TopologyType {
     Full,
 }
 
-impl FromStr for TopologyTypes {
+impl FromStr for TopologyType {
     type Err = Error;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
@@ -73,7 +73,7 @@ pub trait Topology<Handle: ChainHandle> {
 pub struct FullyConnectedTopology;
 
 impl<Handle: ChainHandle> Topology<Handle> for FullyConnectedTopology {
-    fn get_topology(
+    fn create_topology(
         &self,
         chain_handles: &Vec<Handle>,
     ) -> Result<HashMap<usize, HashMap<usize, ForeignClient<Handle, Handle>>>, Error> {
@@ -99,7 +99,7 @@ impl<Handle: ChainHandle> Topology<Handle> for FullyConnectedTopology {
 pub struct LinearTopology;
 
 impl<Handle: ChainHandle> Topology<Handle> for LinearTopology {
-    fn get_topology(
+    fn create_topology(
         &self,
         chain_handles: &Vec<Handle>,
     ) -> Result<HashMap<usize, HashMap<usize, ForeignClient<Handle, Handle>>>, Error> {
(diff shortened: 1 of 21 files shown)


tools/test-framework/src/types/topology.rs
```diff
@@ -0,0 +1,145 @@
...
+    ) -> Result<HashMap<usize, HashMap<usize, ForeignClient<Handle, Handle>>>, Error>;
+}
+
+pub struct FullyConnectedTopology;
+
+impl<Handle: ChainHandle> Topology<Handle> for FullyConnectedTopology {
+    fn get_topology(
+        &self,
+        chain_handles: &Vec<Handle>,
+    ) -> Result<HashMap<usize, HashMap<usize, ForeignClient<Handle, Handle>>>, Error> {
```
romac, line 79: Let's either introduce a wrapper type for this with helper methods or at least a type alias?

tools/test-framework/src/types/topology.rs
```diff
@@ -0,0 +1,145 @@
...
+        match s {
+            "linear" => Ok(Self::Linear),
+            "full" => Ok(Self::Full),
+            _ => Err(Error::generic(eyre!("The topology `{s}` does not exist"))),
+        }
+    }
+}
+
+pub trait Topology<Handle: ChainHandle> {
+    fn get_topology(
```
romac, line 67: Since this has side effects make it explicit in the name:

```suggestion
    fn create_topology(
```

```suggestion
    fn bootstrap_topology(
```

tools/test-framework/src/types/topology.rs
```diff
@@ -0,0 +1,145 @@
...
+                clients.insert(i - 1, client);
+            }
+
+            foreign_clients.insert(i, clients);
+        }
+        Ok(foreign_clients)
+    }
+}
+
+pub fn get_topology<Handle: ChainHandle>(value: &str) -> Box<dyn Topology<Handle>> {
```
romac, line 135: How about:

```suggestion
pub fn bootstrap_topology<Handle: ChainHandle>(topology: TopologyType) -> Box<dyn Topology<Handle>> {
```

and leave the parsing of the topology string to the caller?

tools/test-framework/src/types/topology.rs
```diff
@@ -0,0 +1,145 @@
...
+use std::collections::HashMap;
+use std::str::FromStr;
+
+use ibc_relayer::chain::handle::ChainHandle;
+use ibc_relayer::foreign_client::ForeignClient;
+
+use crate::bootstrap::binary::chain::bootstrap_foreign_client;
+use crate::error::Error;
+
+pub enum TopologyTypes {
```
romac, line 49:
```suggestion
pub enum TopologyType {
```

tools/test-framework/src/types/topology.rs
romac: Perhaps a generic `Matrix` type or similarly named?

```rust
pub struct Matrix<T> {
  matrix: HashMap<usize, HashMap<usize, T>
}

impl<T> Matrix<T> {
  // to replace calls like `let outer = map.get(x); let inner = outer.get(y)`
  pub fn get(&self, coords: (usize, usize)) -> Option<&T> {
    // ...
  }

  // to replace nested for loops
  pub fn iter(&self) -> impl Iterator<Item = &T> {
    
  }
}
```

romac (commented):
Amazing! Left some nitpicks inline already, feel free to disregard if you had other plans in mind.

tools/test-framework/src/types/topology.rs
ljoss17: Good point, but I don't know if naming it a `Matrix` might be confusing. For example the linear topology for 3 chains will throw errors when accessing some values, e.g. `[0][2]`, which I wouldn't expect to be an issue in a matrix.
What about something more like `2dHashMap`?

tools/test-framework/src/types/topology.rs
ljoss17: done in b035a6c4f501d03ffc7ae74fe2192371538494d5

tools/test-framework/src/types/topology.rs
romac: Yes sounds better!

romac: 2dHashMap is not a valid identifier in Rust, so perhaps `HashMap2d` or `TwoDimHashMap`?

ljoss17: Oh good catch, I have a small preference for `TwoDimHashMap` as it is slightly more explicit

romac:
Excellent work! Any insights on how much time this buys us on CI?

romac (approved):

ljoss17:
The test which received the most impact is the packet forward test:

* Gaia went from 1h30 to 1h11
* Juno went from 1h31 to 1h06
* Injective went from 1h18 to 58min

The other chains went down a few minutes only as the other main test impacted is the `ternary_transfer`

The remaining very long job is for simapp v8, as it runs the highest number of tests, 86 tests but does not run forward tests. So it went down only from 1h31 to 1h25.

romac:
Very nice, thanks for the report :)

Issue #4047, opened 2024-06-18
Allow ranges for `excluded_sequences` config option


romac:
## Summary

Allow ranges for `excluded_sequences` config option instead of having to specify each sequence number manually.

## Problem Definition

Some operators need to exclude large sequence ranges which are unpractical to specify one by one, eg. we've head of ranges which span 32k sequence numbers.

## Proposal

Change the deserialization code for `excluded_sequences` to allow each element of the list to be either:
- a single sequence number, as an integer: e.g. `123456`
- a range of sequence numbers, as a string: e.g. `"123..456"`

Let's discuss:
- what separator to use (`..`, `..=`, `-`)
- whether or not the range should be inclusive
- whether to allow both inclusive and exclusive ranges

## Acceptance Criteria

The following value for `excluded_sequences` is accepted:

```toml
excluded_sequences = [
  1, 2, 3, 4, 5,
  "6..9",
  10
]
```

and is equivalent to specifying:

```toml
excluded_sequences = [
  1, 2, 3, 4, 5,
  6, 7, 8, 9,
  10
]
```

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate milestone (priority) applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #4051, opened 2024-06-20
Allow ranges for `excluded_sequences` config option

ljoss17:
Closes: #4047

## Description

This PR changes the type of `excluded_sequences` in the configuration to be a `ExcludedSequences` which has custom serialisation and deserialisation. This allows the usage of ranges of sequences in addition to exact values.
It is now possible to specify excluded sequences using:

* Exact values, e.g. `[1, 2, 3]`
* Range with `"..="`, e.g. `["1..=3"]`
* Range with `".."`, e.g. `["1..3"]`
* Range with `"-"`, e.g. `["1-3"]`

These 4 examples will result in the same result.
______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -444,10 +444,10 @@ memo_prefix = ''
 # Specify packet sequences which should not be cleared, per channel.
 #
 # For each channel, specify a list of sequences which should not be cleared. Acceptable value
-# include range of sequences with separator "..=", ".." or "-", eg.
+# include range of sequences with separator "-", eg.
 #
 #   [chains.excluded_sequences]
-#   channel-0 = [1, 2, "3..7", 9, 11],
+#   channel-0 = [1, 2, "3-7", 9, 11],
 #   channel-1 = [4, 5, 6],
 #
 # Default: No filter
--- a/config.toml
+++ b/config.toml
@@ -443,15 +443,15 @@ memo_prefix = ''
 
 # Specify packet sequences which should not be cleared, per channel.
 #
-# For each channel, specify a list of sequences which should not be cleared, eg.
+# For each channel, specify a list of sequences which should not be cleared. Acceptable value
+# include range of sequences with separator "..=", ".." or "-", eg.
 #
-#   excluded_sequences = [
-#     ['channel-0', [1, 2, 3]],
-#     ['channel-1', [4, 5, 6]]
-#  ]
+#   [chains.excluded_sequences]
+#   channel-0 = [1, 2, "3..7", 9, 11],
+#   channel-1 = [4, 5, 6],
 #
 # Default: No filter
-# excluded_sequences = []
+# excluded_sequences = {}
 
 [[chains]]
 id = 'ibc-1'
--- a/.changelog/unreleased/improvements/ibc-relayer/4047-improve-excluded-sequences-config.md
+++ b/.changelog/unreleased/improvements/ibc-relayer/4047-improve-excluded-sequences-config.md
@@ -2,12 +2,10 @@
   ranges of sequence values in addition to exact values.
   Accepted format:
     * Exact sequence, e.g. [1, 2, 3]
-    * "..=" separator, e.g. ["1..=3"]
-    * ".." separator, e.g. ["1..3"]
     * "-" separator, e.g. ["1-3"]
 
   These can be combined making the following configurations equivalent:
-    * `excluded_sequences = { 'channel-0' = [1, "3-5", 7, "9..12", 14, "17..=19"] }`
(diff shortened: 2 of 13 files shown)


romac:
I am a bit wary of support both `..` and `..=` with the same semantics, given that in Rust these mean exclusive and inclusive upper bound, respectively. Granted, it's probably not common knowledge outside of Rust programmers.

Still, I would rather we keep a single syntax for defining ranges. How about keeping only `-`, as it does not collide with Rust ranges, and does not require users to ponder about whether or not it's inclusive or exclusive, provided that we make it explicit in the config file.

What do you think @ljoss17?

config.toml
```diff
@@ -443,15 +443,15 @@ memo_prefix = ''
...
+# For each channel, specify a list of sequences which should not be cleared. Acceptable value
+# include range of sequences with separator "..=", ".." or "-", eg.
 #
-#   excluded_sequences = [
-#     ['channel-0', [1, 2, 3]],
-#     ['channel-1', [4, 5, 6]]
-#  ]
+#   [chains.excluded_sequences]
+#   channel-0 = [1, 2, "3..7", 9, 11],
+#   channel-1 = [4, 5, 6],
```
romac, line 451: While I agree that the new syntax does look better than the former one, I would rather introduce breaking changes to the config at this stage.

Do you think we can support both options? Otherwise let's just revert to the former one.

romac (changes requested):
See comments above. Great job otherwise!

config.toml
ljoss17: Unfortunately I think this was a mistake in the documentation, because if I run the `parse_valid_config()` test, https://github.com/informalsystems/hermes/blob/v1.9.0/crates/relayer/src/config.rs#L842, with the following config, it fails.

config:
<details>
[global]
log_level = 'error'

[mode]

[mode.clients]
enabled = true
refresh = true
misbehaviour = true

[mode.connections]
enabled = false

[mode.channels]
enabled = false

[mode.packets]
enabled = true
clear_interval = 100
clear_on_start = true
tx_confirmation = true
ics20_max_memo_size = { enabled = true, size = "32KiB" }
ics20_max_receiver_size = { enabled = true, size = "2KiB" }

[[chains]]
type = "CosmosSdk"
id = 'chain_A'
rpc_addr = 'http://127.0.0.1:26657'
grpc_addr = 'http://127.0.0.1:9090'
event_source = { mode = 'push', url = 'ws://localhost:26657/websocket', batch_delay = '500ms' }
rpc_timeout = '10s'
account_prefix = 'cosmos'
key_name = 'testkey'
store_prefix = 'ibc'
max_gas = 200000
gas_price = { price = 0.001, denom = 'stake' }
max_msg_num = 4
max_tx_size = 1048576
clock_drift = '5s'
trusting_period = '14days'
trust_threshold = { numerator = '1', denominator = '3' }
excluded_sequences = [
  ['channel-0', [1, 2, 3]],
  ['channel-1', [4, 5, 6]]
]

[chains.packet_filter]
policy = 'allow'
list = [
  ['ica*', '*'],
  ['transfer', 'channel-0'],
]

[[chains]]
type = "CosmosSdk"
id = 'chain_B'
rpc_addr = 'http://127.0.0.1:26557'
grpc_addr = 'http://127.0.0.1:9090'
event_source = { mode = 'push', url = 'ws://localhost:26557/websocket', batch_delay = '500ms' }
rpc_timeout = '10s'
account_prefix = 'cosmos'
key_name = 'testkey'
store_prefix = 'ibc'
gas_price = { price = 0.001, denom = 'stake' }
clock_drift = '5s'
trusting_period = '14days'
trust_threshold = { numerator = '1', denominator = '3' }
address_type = { derivation = 'ethermint', proto_type = { pk_type = '/injective.crypto.v1beta1.ethsecp256k1.PubKey' } }
</details>


error:
```
Caused by:
    TOML parse error at line 25, column 1
       |
    25 | [[chains]]
       | ^^^^^^^^^^
    invalid CosmosSdk config: invalid type: sequence, expected a map
```

ljoss17:
> I am a bit wary of supporting both `..` and `..=` with the same semantics, given that in Rust these mean exclusive and inclusive upper bound, respectively. Granted, it's probably not common knowledge outside of Rust programmers.
> 
> Still, I would rather we keep a single syntax for defining ranges. How about keeping only `-`, as it does not collide with Rust ranges, and does not require users to ponder about whether or not it's inclusive or exclusive, provided that we make it explicit in the config file.
> 
> What do you think @ljoss17?

I agree, having a single separator will make it cleaner. Yes the `-` seems the best option.

config.toml
romac: Oh right, perfect then!

romac (approved):

Issue #4062, opened 2024-06-24
Add support for multiple connection hops in `create channel` command


joaotav:
Part of: #3940

### [`create channel`](https://github.com/informalsystems/hermes/blob/master/crates/relayer-cli/src/commands/create/channel.rs)

- [x] Add a new parameter `--connection-hops` to `create channel`, which should receive the identifiers of the intermediate connections between two chains. This should work similarly to the [`--connection-hops`](https://github.com/informalsystems/hermes/blob/joaotav/multihop-poc/crates/relayer-cli/src/commands/tx/channel.rs#L119) parameter in `tx chan-open-init`.
- [x] Add a new method  `run_multihop_reusing_connection()` to open a multi-hop channel when arguments were passed via `--connection-hops`.
- [x] Add a new associated function `Channel::new_multihop()` to prepare an instance of `Channel` for multi-hop channel handshakes.

### Tests
- [x] Test `create channel` manually for single-hop channels.
- [x] Test `create channel` manually for channels with 2 connection hops.
- [x] Test `create channel` manually for channels with 2+ connection hops.
- [x] Write unit tests for the new `--connection-hops` parameter. 
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

joaotav:
Closed on #4067.

PR #4067, opened 2024-06-25
Add support for multiple connection hops in `create channel` command

joaotav:
Closes: #4062

## Description
- [ ] https://github.com/informalsystems/hermes/issues/4062

______

### PR author checklist:
- [ ] Linked to GitHub issue.

### Reviewer checklist:

### DIFF
--- a/crates/relayer-cli/src/commands/create/channel.rs
+++ b/crates/relayer-cli/src/commands/create/channel.rs
@@ -3,22 +3,25 @@ use abscissa_core::clap::Parser;
 use console::style;
 use dialoguer::Confirm;
 
+use crate::cli_utils::{spawn_chain_runtime, ChainHandlePair};
+use crate::conclude::{exit_with_unrecoverable_error, Output};
+use crate::error::Error;
+use crate::prelude::*;
 use ibc_relayer::chain::handle::ChainHandle;
 use ibc_relayer::chain::requests::{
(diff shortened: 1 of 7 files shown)


crates/relayer-cli/src/commands/create/channel.rs
```diff
@@ -322,6 +323,9 @@ impl CreateChannelCommand {
     ) {
         let config = app_config();
 
+        // Set global registry to get or spawn chain handles
+        set_global_registry(SharedRegistry::new((*app_config()).clone()));
```
joaotav, line 327: Even though the registry is being set here (as it is needed further down in the chain of function calls), the chains are being spawned using `spawn_chain_runtime()` rather than using `get_or_spawn()` within `run_multihop_reusing_connection()`. This was to keep it consistent with `run_reusing_connection()` which was already implemented and uses `spawn_chain_runtime()`. 
As `get_or_spawn()` might give us some performance, if we decide to keep using the registry in the CLIs we should use it here and make it consistent across all methods.

romac (approved):
Great job! 👏

crates/relayer-cli/src/commands/create/channel.rs
romac: Good catch! Yeah let's see by the end of the implementation phase which technique we want to go with.

PR #4080, opened 2024-07-04
Split packet clearing schedule from packet worker

ljoss17:
Closes: #4071 

## Description

This PR removes the packet clearing logic from the `packet_cmd_worker`.
The reason is to avoid blocking the worker if there is a huge amount of pending packets.

The `packet_cmd_worker` will not handle packet clearing anymore, instead an additional background task, `clear_cmd_worker` will schedule pending packets to be cleared. And the packets will be cleared by the when `handle_execute_schedule` is called.
A new configuration `clear_limit` is added in order to only clear up to a maximum of `n` packets every time clearing is triggered. The default value is `50`.

There is an exception for Ordered channels, in which case the clearing is blocking and is done before relaying. This is because relaying without completing the packet clearing before would fail due to an incorrect packet sequence error.

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


romac (approved):
Great work! Left a couple nits

PR #4109, opened 2024-07-30
Fix typos

omahs:
Fix typos


ljoss17 (approved):
Thanks for fixing these typos!

Issue #4143, opened 2024-08-13
Use gRPC query `ConnectionParams` instead of `/genesis` to verify `max_block_time`


ljoss17:
## Summary

In the current version of Hermes, `v1.10.1`, the entire genesis data is queried from `/genesis` in order to verify the `max_block_time` configuration, https://github.com/informalsystems/hermes/blob/v1.10.1/crates/relayer/src/chain/cosmos.rs#L299-L323.

Since genesis data is bigger and bigger, Hermes should use the gRPC query `ConnectionParams` to retrieve only the `maxExpectedTimePerBlock` and use that value to verify the configured value.

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate milestone (priority) applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #4144, opened 2024-08-13
Use gRPC query `ConnectionParams` instead of `/genesis` to verify `max_block_time`

ljoss17:
Closes: #4143

## Description

______

### PR author checklist:
- [x] Linked to GitHub issue.

### Reviewer checklist:


crates/relayer/src/chain/cosmos.rs
```diff
@@ -296,27 +296,27 @@ impl CosmosSdkChain {
...
-
-                    info!(
-                        "Updated `max_block_time` using /genesis endpoint. Old value: `{}s`, new value: `{}s`",
-                        old_max_block_time.as_secs(),
+        // Query Connection Params with gRPC endpoint to retrieve the `max_expected_time_per_block` value and verify the
+        // configured `max_block_time`.
+        // If it is not found, the verification for the configured `max_block_time` is skipped.
+        match self.block_on(query_connection_params(&self.grpc_addr)) {
+            Ok(params) => {
+                warn!("raw max block time: {}", params.max_expected_time_per_block);
```
ancazamfir, line 304:
```suggestion
                info!("chain queried max block time: {}", params.max_expected_time_per_block);
```

crates/relayer/src/chain/cosmos.rs
```diff
@@ -296,27 +296,27 @@ impl CosmosSdkChain {
...
+                let queried_max_expected_time_per_block_as_secs =
+                    params.max_expected_time_per_block / 1_000_000_000;
+                if queried_max_expected_time_per_block_as_secs
+                    != self.config.max_block_time.as_secs()
+                {
+                    warn!(
+                        "Queried `max_block_time` has value: `{}s`, while configured `max_block_time` has value: `{}s`",
```
ancazamfir, line 311:
```suggestion
                        "queried `max_block_time` has value: `{}s`, while configured `max_block_time` has value: `{}s`",
```

crates/relayer/src/chain/cosmos.rs
```diff
@@ -296,27 +296,27 @@ impl CosmosSdkChain {
...
+                        queried_max_expected_time_per_block_as_secs,
                         self.config.max_block_time.as_secs()
                     );
                 }
             }
             Err(e) => {
                 warn!(
-                    "Will use fallback value for max_block_time: `{}s`. Error: {e}",
+                    "Configured value for max_block_time: `{}s` could not be verified. Error: {e}",
```
ancazamfir, line 319:
```suggestion
                    "configured value for max_block_time: `{}s` could not be verified. Error: {e}",
```

crates/relayer/src/chain/cosmos.rs
```diff
@@ -296,27 +296,27 @@ impl CosmosSdkChain {
...
-
-                if old_max_block_time.as_secs() != new_max_block_time.as_secs() {
-                    self.config.max_block_time = new_max_block_time;
-
-                    info!(
-                        "Updated `max_block_time` using /genesis endpoint. Old value: `{}s`, new value: `{}s`",
-                        old_max_block_time.as_secs(),
+        // Query Connection Params with gRPC endpoint to retrieve the `max_expected_time_per_block` value and verify the
+        // configured `max_block_time`.
+        // If it is not found, the verification for the configured `max_block_time` is skipped.
```
ancazamfir, line 301: I think what we are missing is this case:
- user does not configure `max_block_time` in `config.toml`
- the UX should be that if chain can be queried we use  the chain `max_expected_time_per_block`
  - otherwise we use a default (30sec)
  
With this PR, if user does not configure it, the default is automatically loaded in the chain config and  if different we just see a warning, so cannot initialize here.
Before this PR the config was overriden with the genesis result (if avail). Not ideal UX either. 
Maybe to keep the same "bad" UX as before, as we cannot make it better, we override with the query value (with the warning that we do this). And we can figure out how to fix this in a different PR.

In my previous work, we used to have the notion of `configuration` vs `operational` values, kept separate. A tag can show `None` in configuration while showing some default (or other, e.g gRPC output in our case) value in`operational`. This way one can also make a difference between an unconfigured tag vs one configured with the default value (these requiring different UX handling)

ancazamfir (changes requested):

ancazamfir:
soo happy to see the genesis endpoint going away!! thanks @ljoss17 !!

.changelog/unreleased/improvements/ibc-relayer/4143-use-connection-params-in-validate-params.md
```diff
@@ -0,0 +1,3 @@
+- Use the gRPC query `ConnectionParams` to retrieve `maxExpectedTimePerBlock`
+  and verify the configured `max_block_time` instead of using the `/genesis`
+  endpoint. ([\#4143](https://github.com/informalsystems/hermes/issues/4143))
```
romac, line 3:
```suggestion
- Use the `ibc.core.connection.v1.ConnectionParams` gRPC query to retrieve  `maxExpectedTimePerBlock` 
  and check it against the configured `max_block_time` instead of using the `/genesis`
  endpoint. ([\#4143](https://github.com/informalsystems/hermes/issues/4143))
```

crates/relayer/src/chain/cosmos.rs
ljoss17: Oh I forgot to remove this log, I don't need it's necessary to keep it. Or maybe putting it as `debug`, what do you think?

crates/relayer/src/chain/cosmos.rs
ljoss17: I see, I will setup the override using the queried value. This way we would have:

* If the `max_block_time` is not configured (the line is missing), it will use the queried value or the default `30s` if the query fails
* If the `max_block_time` is configured, it will use the queried value or the configured value if the query fails

Doesn't this fit the missing case? The difference is that instead of having an explicit `None`, the `max_block_time` is simply not set.

ljoss17: Updated here ae81050a96c831116d930367b9356e80cb355b48

crates/relayer/src/chain/cosmos.rs
ancazamfir: imo we should have a log, debug is also fine.

crates/relayer/src/chain/cosmos.rs
ancazamfir: > If the max_block_time is configured, it will use the queried value or the configured value if the query fails
> Doesn't this fit the missing case?

imo this last bullet should be:

- If the `max_block_time` is configured, the configured value is used. 

We could still query and issue a warning if the queried value is different but not override. But let's leave it like you have it for now.

ljoss17: Oh I understand. We could do this by using an `Option`. I will create a follow-up issue for this

crates/relayer/src/chain/cosmos.rs
ljoss17: I added a debug log with the unprocessed queried value (before creating a `Duration` using the value) here 7d736c04be9fcadb74482ca7f2674a86be34a6c6

crates/relayer/src/chain/cosmos.rs
ljoss17: https://github.com/informalsystems/hermes/issues/4147

ancazamfir (approved):
thanks @ljoss17 ! 🚀

PR #4424, opened 2026-01-16
perf: opt vector operations in tx query path

songgaoye:
Closes: #XXX

## Description
all 3 remove(0) operations replaced with into_iter().next(). It can avoids the front-shift penalty, O(n) -> O(1)

______

### PR author checklist:
- [ ] Linked to GitHub issue.
  - [ ] If guide has been updated, tag GitHub user `mircea-c`

### Reviewer checklist:

### DIFF
--- a/crates/relayer/src/chain/cosmos/query/tx.rs
+++ b/crates/relayer/src/chain/cosmos/query/tx.rs
@@ -67,8 +67,7 @@ pub async fn query_txs(
                 response.txs.len() <= 1,
                 "packet_from_tx_search_response: unexpected number of txs"
             );
-
-            let tx = response.txs.remove(0);
+            let tx = response.txs.into_iter().next().expect("tx_search was constrained to a single result");
             let event = update_client_from_tx_search_response(chain_id, &request, tx)?;
 
             Ok(event.into_iter().collect())
@@ -95,7 +94,7 @@ pub async fn query_txs(
             if response.txs.is_empty() {
                 Ok(vec![])
             } else {
-                let tx = response.txs.remove(0);
+                let tx = response.txs.into_iter().next().expect("tx_search was constrained to a single result");
                 Ok(all_ibc_events_from_tx_search_response(chain_id, tx))
             }
         }
@@ -165,7 +164,7 @@ pub async fn query_packets_from_txs(
         }
 
         // In either case, use the first (latest) event found for this sequence
-        let (first_event, _, _) = tx_events.remove(0);
+        let (first_event, _, _) = tx_events.into_iter().next().expect("tx_events is known to contain at least one entry");
         result.push(first_event);
     }
 
--- a/crates/relayer/src/chain/cosmos/query/tx.rs
+++ b/crates/relayer/src/chain/cosmos/query/tx.rs
@@ -67,7 +67,11 @@ pub async fn query_txs(
                 response.txs.len() <= 1,
                 "packet_from_tx_search_response: unexpected number of txs"
             );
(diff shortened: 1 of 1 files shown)


crates/relayer/src/chain/cosmos/query/tx.rs
```diff
@@ -67,8 +67,7 @@ pub async fn query_txs(
                 response.txs.len() <= 1,
                 "packet_from_tx_search_response: unexpected number of txs"
             );
-
-            let tx = response.txs.remove(0);
+            let tx = response.txs.into_iter().next().expect("tx_search was constrained to a single result");
```
KirilMihaylov:
```suggestion
            let tx = response.txs.swap_remove(0);
```

crates/relayer/src/chain/cosmos/query/tx.rs
```diff
@@ -95,7 +94,7 @@ pub async fn query_txs(
             if response.txs.is_empty() {
                 Ok(vec![])
             } else {
-                let tx = response.txs.remove(0);
+                let tx = response.txs.into_iter().next().expect("tx_search was constrained to a single result");
```
KirilMihaylov:
```suggestion
                let tx = response.txs.swap_remove(0);
```

crates/relayer/src/chain/cosmos/query/tx.rs
```diff
@@ -165,7 +164,7 @@ pub async fn query_packets_from_txs(
         }
 
         // In either case, use the first (latest) event found for this sequence
-        let (first_event, _, _) = tx_events.remove(0);
+        let (first_event, _, _) = tx_events.into_iter().next().expect("tx_events is known to contain at least one entry");
```
KirilMihaylov:
```suggestion
        let (first_event, _, _) = tx_events.swap_remove(0);
```

KirilMihaylov (commented):
Using [`swap_remove`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.swap_remove) will reduce the readability overhead as it's equivalent to the provided solution.

P.S.:
Leaving this review as a member of the @nolus-protocol dev team.

crates/relayer/src/chain/cosmos/query/tx.rs
KirilMihaylov: On second look, change is unnecessary in the first place because the assertion will fail if there is more than one element.

songgaoye:
> Using [`swap_remove`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.swap_remove) will reduce the readability overhead as it's equivalent to the provided solution.
> 
> P.S.: Leaving this review as a member of the @nolus-protocol dev team.

I don't think so. 
the current solution with into_iter().next() is even better because:

1.Same O(1) performance as swap_remove
2.Consumes the vector - makes it clear the rest of the data is discarded
3.More explicit intent - the code shows we only care about the first element
4.No unnecessary swap - swap_remove still does a swap operation even though we're discarding the vector

KirilMihaylov:
At (1):
> This does not preserve ordering of the remaining elements, but is O(1). If you need to preserve the element order, use [remove](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.remove) instead.

Source: [`std::vec::Vec::swap_remove`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.swap_remove)

At (2):
```rust
{ tx_events }.swap_remove(0)
```

At (3):
In my opinion it's just as explicit, and what's more, `swap_remove` is less verbose.

At (4):
Yes, I admit that it will do one swap, but that's a trade-off justified by being more concise in my eyes.

PR #4426, opened 2026-01-16
perf(relayer): optimize timeout processing with drain and Vec

songgaoye:
Closes: #XXX

## Description

The optimization provides 40-44% performance improvement across all batch sizes by:

-  Replacing HashMap with Vec<Option> for sequential index access

-  Eliminating cloning by using drain() instead of iter() + clone()

-  Pre-allocating vector capacity

Batch Size | Legacy (clone) | Optimized (drain) | Improvement
-- | -- | -- | --
32 | 286 µs | 171 µs | 40%
64 | 548 µs | 321 µs | 41%
128 | 1,115 µs | 640 µs | 43%

______

### PR author checklist:
- [ ] Linked to GitHub issue.
  - [ ] If guide has been updated, tag GitHub user `mircea-c`

### Reviewer checklist:

### DIFF
--- a/crates/relayer/benches/vector_ops.rs
+++ b/crates/relayer/benches/vector_ops.rs
@@ -1,0 +1,210 @@
+use std::str::FromStr;
+use std::time::Duration;
+
+use criterion::{black_box, criterion_group, criterion_main, BatchSize, BenchmarkId, Criterion};
+use ibc_proto::google::protobuf::Any;
+use ibc_relayer::chain::tracking::TrackingId;
+use ibc_relayer::event::IbcEventWithHeight;
+use ibc_relayer::link::operational_data::{OperationalData, OperationalDataTarget, TransitMessage};
+use ibc_relayer_types::core::ics04_channel::events::{SendPacket, WriteAcknowledgement};
+use ibc_relayer_types::core::ics04_channel::packet::Packet;
+use ibc_relayer_types::core::ics04_channel::packet::Sequence;
+use ibc_relayer_types::core::ics04_channel::timeout::TimeoutHeight;
+use ibc_relayer_types::core::ics24_host::identifier::{ChannelId, PortId};
+use ibc_relayer_types::events::IbcEvent;
+use ibc_relayer_types::timestamp::Timestamp;
+use ibc_relayer_types::Height;
+use once_cell::sync::Lazy;
+
+const NUM_BATCHES: usize = 32;
+const TIMEOUT_STRIDE: u64 = 5;
+static PORT_ID: Lazy<PortId> = Lazy::new(|| PortId::from_str("transfer").expect("valid port id"));
+static CHANNEL_ID: Lazy<ChannelId> =
+    Lazy::new(|| ChannelId::from_str("channel-0").expect("valid channel id"));
+
+fn relay_path_benchmark(c: &mut Criterion) {
+    let mut group = c.benchmark_group("relay_path_timeout_processing");
+
+    for &batch_size in &[32usize, 64, 128] {
+        group.bench_with_input(
+            BenchmarkId::new("legacy_clone", batch_size),
+            &batch_size,
+            |b, &size| {
+                b.iter_batched(
+                    || build_operational_batches(size),
+                    |dataset| {
+                        let (_, total) = process_batches_with_clones(dataset);
+                        black_box(total);
+                    },
+                    BatchSize::SmallInput,
+                );
+            },
+        );
+
+        group.bench_with_input(
+            BenchmarkId::new("drain_reuse", batch_size),
+            &batch_size,
+            |b, &size| {
+                b.iter_batched(
+                    || build_operational_batches(size),
+                    |dataset| {
+                        let (_, total) = process_batches_with_drain(dataset);
+                        black_box(total);
+                    },
+                    BatchSize::SmallInput,
+                );
+            },
+        );
+    }
+
+    group.finish();
+}
+
+fn build_operational_batches(batch_size: usize) -> Vec<OperationalData> {
+    (0..NUM_BATCHES)
+        .map(|batch_index| {
+            let mut odata = OperationalData::new(
+                Height::new(0, 10).expect("height"),
+                OperationalDataTarget::Destination,
+                TrackingId::new_static("bench"),
+                Duration::ZERO,
+            );
+
+            for entry in 0..batch_size {
+                let seq = (batch_index * batch_size + entry + 1) as u64;
+                let message = if seq % TIMEOUT_STRIDE == 0 {
+                    make_send_packet(seq)
+                } else {
+                    make_write_ack(seq)
+                };
+
+                odata.push(message);
+            }
+
+            odata
+        })
+        .collect()
+}
+
+fn make_send_packet(sequence: u64) -> TransitMessage {
+    let packet = Packet {
+        sequence: Sequence::from(sequence),
+        source_port: PORT_ID.clone(),
+        source_channel: CHANNEL_ID.clone(),
+        destination_port: PORT_ID.clone(),
+        destination_channel: CHANNEL_ID.clone(),
+        data: vec![0; 32],
+        timeout_height: TimeoutHeight::default(),
+        timeout_timestamp: Timestamp::none(),
+    };
+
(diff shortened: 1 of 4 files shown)


crates/relayer/benches/vector_ops.rs
```diff
@@ -0,0 +1,210 @@
+use std::str::FromStr;
+use std::time::Duration;
+
+use criterion::{black_box, criterion_group, criterion_main, BatchSize, BenchmarkId, Criterion};
```
KirilMihaylov, line 4: [`criterion::black_box`](https://docs.rs/criterion/latest/criterion/fn.black_box.html) is deprecated and says to use [`std::hint::black_box`](https://doc.rust-lang.org/stable/std/hint/fn.black_box.html) instead.

crates/relayer/src/link/relay_path.rs
```diff
@@ -1680,19 +1679,17 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+            let odata_info = odata.info();
+            let odata_tracking_id = odata.tracking_id;
 
-            for gm in odata.batch.iter() {
-                let TransitMessage {
-                    event_with_height, ..
-                } = gm;
-
-                match &event_with_height.event {
+            for gm in odata.batch.drain(..) {
```
KirilMihaylov, line 1691: Capacity is not reused later and is instead replaced (L1739, R1737) with an empty one, which is also pure computationally (`const fn`) and does not allocate.
That can be done via [`std::mem::take`](https://doc.rust-lang.org/stable/std/mem/fn.take.html) and will allow the current or future versions of [`std::vec::Vec`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html) to take advantage of the fact that the whole container is consumed.

```suggestion
            for gm in std::mem::take(&mut odata.batch)
```

crates/relayer/src/link/relay_path.rs
```diff
@@ -1680,19 +1679,17 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
         // to source operational data.
         let mut all_dst_odata = self.dst_operational_data.clone_vec();
 
-        let mut timed_out: HashMap<usize, OperationalData> = HashMap::default();
+        let mut timed_out: Vec<Option<OperationalData>> = vec![None; all_dst_odata.len()];
```
KirilMihaylov: This will only waste computational time and resources as it isn't necessary to allocate it with the same length. On top of that, allocating it with `None` instances also redundant as `all_dst_odata` is traversed sequentially from start to end and elements can just be inserted to the back.
```suggestion
        let mut timed_out = vec![];
```


If proved necessary, it could be allocated
```suggestion

```

KirilMihaylov, line 1682: \[timed_out 1/4\]

This will only waste computational time and resources as it isn't necessary to allocate it with the same length. On top of that, allocating it with `None` instances also redundant as `all_dst_odata` is traversed sequentially from start to end and elements can just be inserted to the back.
```suggestion
        let mut timed_out = vec![];
```


If proven necessary, it could be allocated with `max(1, half-capacity)`, as I believe that it's highly unlikely during normal circumstances to get past half-capacity.
```rust
let mut timed_out = Vec::with_capacity(1.max(all_dst_odata.len() >> 1));
```

crates/relayer/src/link/relay_path.rs
```diff
@@ -1680,19 +1679,17 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
 
         // For each operational data targeting the destination chain...
         for (odata_pos, odata) in all_dst_odata.iter_mut().enumerate() {
```
KirilMihaylov, line 1685: \[timed_out 2/4\]

```suggestion
        for odata in &mut all_dst_odata {
```

crates/relayer/src/link/relay_path.rs
```diff
@@ -1702,36 +1699,37 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
+                                    dst_current_height,
+                                    OperationalDataTarget::Source,
+                                    odata_tracking_id,
+                                    self.channel.connection_delay,
+                                )
+                            })
+                            .push(TransitMessage {
+                                event_with_height: gm.event_with_height.clone(),
+                                msg: new_msg,
+                            });
```
KirilMihaylov, line 1719: \[timed_out 3/4\]

```suggestion
                            let mut odata = OperationalData::new(
                                dst_current_height,
                                OperationalDataTarget::Source,
                                odata_tracking_id,
                                self.channel.connection_delay,
                            );

                            odata.push(TransitMessage { msg: new_msg, ..gm });

                            timed_out.push(odata);
```

crates/relayer/src/link/relay_path.rs
```diff
@@ -1747,13 +1745,13 @@ impl<ChainA: ChainHandle, ChainB: ChainHandle> RelayPath<ChainA, ChainB> {
...
         // Handle timed-out events
-        if timed_out.is_empty() {
+        if timed_out.iter().all(|entry| entry.is_none()) {
             // Nothing timed out in the meantime
             return Ok(());
         }
 
         // Schedule new operational data targeting the source chain
-        for (_, new_od) in timed_out.into_iter() {
+        for new_od in timed_out.into_iter().flatten() {
```
KirilMihaylov, line 1754: \[timed_out 4/4\]

```suggestion
        for new_od in timed_out {
```

crates/relayer/benches/vector_ops.rs
KirilMihaylov, line 1: This micro-benchmark tests the difference between [`std::vec::Vec`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html) and [`std::collections::HashMap`](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html) as it does _**NOT**_ test against the used implementation but rather a dummy out-of-sync version of the algorithm.

That makes this micro-benchmark wholly redundant as the algorithm I have provided doesn't allocate without necessity _**AND**_ does not allocate the whole index range.

KirilMihaylov (commented):
Heavy refactoring necessary.

P.S.:
Leaving this review as a member of the @nolus-protocol dev team.

crates/relayer/src/link/relay_path.rs
songgaoye: Using `Vec::with_capacity` there is intentional: we need to take ownership of every TransitMessage so we can move “timed-out” ones into timed_out and keep only the survivors. 
The current flow (for gm in odata.batch.drain(..) { … } followed by odata.batch = retain_batch;) gives us owned gms and preserves ordering; the price is allocating a fresh buffer for retain_batch. 
Replacing this with `std::mem::take` doesn’t actually remove that allocation: once you take the batch you still need a second vector to store the retained items (because retain_mut/retain only hand out &mut references, which we can’t move), so you’d end up allocating again for that second vector or cloning messages. 
In short, there’s no practical win here—`mem::take` can’t simultaneously give us ownership of the drained messages and let us reuse the original buffer without deeper restructuring, so the current approach at `relay_path.rs` (lines 1688-1739) is fine as-is.

KirilMihaylov: I'm sorry but I'll have to strongly disagree. Please look at the signature and provided documentation of `std::mem::take`. It temporarily needs the given mutable reference, replacing it with `Default::default()`, and giving the old result. It's a convenience wrapper around `std::mem::replace`.

The code base is already in unacceptable state and we at @nolus-protocol are currently rewriting parts of it, with possibility for upstreaming, so I personally would rather not add more unnecessary complexity.

I have already removed the need for `retain_batch` as it's very error-prone and directly add back to `odata.batch`.

KirilMihaylov: Leaving our implementation that can be used as a reference and proof of concept that this approach works.

https://github.com/nolus-protocol/hermes/blob/hermes-lite-stable/crates/relayer/src/link/relay_path.rs#L1723

Link to the commit: https://github.com/nolus-protocol/hermes/commit/52479c319aafdab1a32c7ec41f26efda74eb890d

PR #4429, opened 2026-01-30
feat(cardano): add Cardano chain support (Gateway) + Mithril-based 08-cardano light client

floor-licker:
### Summary 

cc: @fabianbormann , @nemo83 , @cardano-foundation

This PR adds a Cardano chain backend to Hermes enabling Hermes to relay IBC between Cosmos chains and Cardano.

The Cardano backend integrates through a gRPC API that exposes the Cosmos-style IBC query surfaces and transaction submission, so Hermes can implement the standard `ChainEndpoint` responsibilities (queries, proofs, message building, event parsing) without requiring Cardano full-node semantics inside Hermes itself. On the light-client side, the Cosmos chain tracks Cardano using client type 08-cardano with Mithril headers (`/ibc.lightclients.mithril.v1.MithrilHeader`).

I will avoid doing a deep dive on Mithril and Cardano consensus dynamics in this PR, but the Mithril header carries the evidence needed to authenticate a HostState update transaction, from which the verifier can extract the committed `ibc_state_root` used for IBC membership and non-membership. Once that root is known, membership and non-membership checks follow the standard ICS-23 model by verifying `protobuf ibc.core.commitment.v1.MerkleProof` bytes against the committed root. The Gateway is treated as a data availability layer for queries and proofs, while the verification on the counterparty chain is intended to be purely cryptographic. That is, this setup for Cardano maintains core IBC trust principles, and lends itself to a fully trustless bridge implementation. 

This integration requires a Cardano notion of “height”. In this design `Height.revision_height` is treated as a Cardano block number as surfaced by db-sync and by Mithril cardano-transactions snapshots, **not a Cardano slot**. Because Mithril certificates are checkpoint-based rather than per-block, Hermes includes configurable waiting behavior after Cardano transaction submission: Hermes waits until the latest certified Mithril snapshot height is greater than or equal to the inclusion height of the submitted transaction, preventing the relayer from building proofs at a height that the Cosmos-side Mithril light client can't yet verify.

 The PR also adds Cardano key handling using the standard Hermes keyring pattern (serialized keys under `~/.hermes/keys/...`), and implements Cardano event polling and parsing needed for handshake and packet relaying flows.
Existing Cosmos chain behaviour stays the same. The new functionality is isolated behind the new chain type and client types, and standard proof decoding and query flows are reused wherever possible.
 
Testing was performed with the existing Hermes unit test suite and additional Cardano-specific tests, and the Cardano config includes explicit parameters for Mithril waiting and logging so operational behaviour is observable rather than relying on fixed sleeps. 

Attesting canonical on-chain state on Cardano is quite tricky, so the Mithril based solution (which is using STM signatures from stakepool operators to attest to transaction inclusion rather than explicit on-chain state per se) is currently the best solution available for an IBC light client. The underlying dynamic is essentially to prove transaction inclusion and then derive transaction outputs which dictate the "trusted" IBC host-state modification which that transaction made. If there's any interest in understanding more about the struggle for a Cardano canonical UTXO-set, or a Cardano notion of something like an `app_hash`, or a Mithril-like service for on-chain state, this discussion is a good starting point: CIP-0165, https://github.com/cardano-foundation/CIPs/pull/1083

## cardano-ibc-incubator
The cardano-ibc-incubator project brings IBC to Cardano. This PR contains the Hermes integration required to complete the bridge, which we are otherwise maintaining in a separate fork. Cardano integration presents a lot of unique challenges for IBC integration, to name a few: UTXO model means transaction building differs fundamentally from acccount-based chains, Cardano has no native websocket event model or RPC interface, Cardano has no analogous KV store and in fact no contract state whatsoever, finally Cardano has no consensus-signed app state root which complicates things substantially. See the repo here: https://github.com/cardano-foundation/cardano-ibc-incubator

  ## Compatibility 

  - Introduced a new chain type Cardano in Hermes config and wiring it into runtime spawning.
  - Implemented a `CardanoChainEndpoint` backed by a Cardano Gateway gRPC service for queries, tx
    building, event polling, and tx submission.
  - Added the Cosmos-side Cardano light client domain types under 08-cardano, implemented via
    Mithril headers (`/ibc.lightclients.mithril.v1.MithrilHeader`) in ibc-relayer-types.
  - Implemented the IBC happy path for Cardano <-> Cosmos relaying: client/connection/channel
    handshakes, packet relay, channel close, and timeout-on-close.
  - Added/adjusted tests + docs

  This PR adds new variants/types in public enums:

  - `ChainConfig::Cardano`
  - `ClientType::Cardano`
  - `AnyClientState::Mithril`, `AnyConsensusState::Mithril`,  and`AnyHeader` Mithril support

## Design overview
Hermes drives the normal relaying (scheduling, retries, path processing, and eventdriven progression), asks the Gateway to construct unsigned Cardano transactions for the relevant IBC messages, signs those transactions locally using the configured Cardano key, and then submits the signed transaction back through the Gateway. The Gateway is treated as a data plane for Cardano queries, transaction construction, submission, and event retrieval, and Hermes translates Gateway-returned events into Hermes internal `IbcEvent` model so the rest of the relayer logic is idiomatic.
 
Cardano transactions are signed as Ed25519 witnesses over the CBOR (which is like JSON) transaction body and witness set, using Pallas primitives/codecs to ensure Conway-era compatibility. The integration supports two key input modes: a mnemonic path that derives the signing key using CIP‑1852, and a bech32-encoded private key (ed25519_sk...) path for environments where mnemonic derivation is incompatible or where keys originate outside Hermes tooling. Internally, the CIP‑1852 derivation uses `m/1852'/1815'/account'/2'/0'`, and the only user controlled component is the account index (wired to the Hermes HD path “account” value); the rest of the derivation path is fixed to match Cardano conventions.

 An important Cardano-specific nuance is the meaning of “height.” Hermes heights are the standard (`revision_number`, `revision_height`), but for this integration `revision_height` is treated as a Cardano block number as surfaced by db-sync and by Mithril cardano-transactions snapshots, not a Cardano slot number. This matters because the Cosmos-side Cardano-tracking client verifies at a snapshot/block-number notion of height, and it affects proof construction, update-client flows, and how the relayer reasons about “what is committed.” **Because Mithril is checkpoint-based a Cardano transaction can be included in a block before it is covered by a Mithril-certified transaction snapshot.** If Hermes immediately proceeds to the next IBC step after inclusion, it can attempt to build or use proofs at a height that the Cosmos-side Mithril client cannot yet verify. To prevent this class of false-negative failures, we treat “included” and “certified” as distinct, and wait after submission until the Gateway reports a Mithril snapshot whose certified `block_number` is greater than or equal to the inclusion block number of the submitted transaction. This wait is configurable through Cardano chain config knobs (timeout, polling interval, and periodic progress logging), and is intended to be observable rather than relying on fixed sleeps.
 
On the light client side, the Cosmos chain tracks Cardano using a single client type `08-cardano`, where headers are Mithril headers (`/ibc.lightclients.mithril.v1.MithrilHeader`). In the current model, consensus state carries the committed `ibc_state_root` extracted from certified HostState transaction evidence contained in the Mithril header, and membership/nonmembership verification follows the standard IBC approach by verifying ICS‑23 proofs against that committed root. The implementation avoids trying to split “consensus tracking” and “state proof verification” across multiple client IDs, because the main connection/channel logic assumes a single `client_id` is the reference point for proof verification.
 
From a relayer plumbing perspective, the Cardano endpoint implements the query, proof, and event surfaces Hermes relies on for core relaying workflows. Proofs are handled in the canonical format Hermes expects, namely protobuf `ibc.core.commitment.v1.MerkleProof` bytes, and the Cardano endpoint decodes those bytes into Hermes `MerkleProof` type so the standard proof-handling flow is reused.

## Cardano signing

 A quick note on dependencies and Cargo.lock: the Cardano integration introduces a small set of new crates for transaction signing and encoding (pallas-* for Cardano CBOR/witness handling, blake2 for Cardano-style hashing, and slip10 for Ed25519 key derivation from a BIP‑39 seed). You may notice Cargo.lock grows more than the raw code diff would suggest, this is because slip10 currently pulls an older ed25519-dalek/hmac stack alongside the newer versions already used elsewhere in Hermes, so the lockfile contains both. This is a known consequence of the current choice of derivation crate and can be revisited to converge on a single crypto stack.
 
On key input, Hermes supports two ways to provide a Cardano signing key. You can supply a BIP‑39 mnemonic, in which case Hermes derives a key using a hardened-only SLIP‑0010 (Ed25519) path (m/1852'/1815'/account'/2'/0'). This is deterministic and convenient for local testing and relayer-managed keys, but it is not guaranteed to match the exact derivation used by all Cardano wallet stacks (many CIP‑1852 implementations use non-hardened steps for the final segments). For production and maximum interoperability, prefer importing a Cardano private key directly via the ed25519_sk... bech32 format, which avoids derivation ambiguity. The CLI still uses `--mnemonic-file`, but that file may contain either a mnemonic phrase or an ed25519_sk... key,  the implementation just branches depending on what it sees in the file, see that implementation at `relayer/crates/relayer/src/chain/ cardano/signing_key_pair.rs`.

  ## Follow-ups
For clarity, the current feature set is prioritized to enable round-trip ICS-20 token transfers, so some non-implicated features are currently stubbed, but return clear error messages about not yet being implemented. 

 - [] Cardano/Mithril misbehaviour detection
 - [] Gateway-backed balance/UTXO query
 - [] `hermes listen` cli path wiring
 - [] balance queries (it might seem surprising that this is not needed but the transaction building is handled by the gateway. Cardano transactions do have some odd dynamics.This is not required for ICS‑20 relaying because the Gateway builds Cardano transactions (including UTxO selection and fee calculation) and Hermes only signs and submits them. Balance queries are mainly an operator UX feature. This is likely to be a follow-up PR soon as it will be useful for general telemetry.)

### PR author checklist:
- [ ] Added changelog entry, using [`unclog`](https://github.com/informalsystems/unclog). **will wait for feedback**
- [N/A] Linked to GitHub issue.
- [ ] If guide has been updated, tag GitHub user `mircea-c` **will wait for feedback**
- [ ] Tagged *one* reviewer who will be the one responsible for shepherding this PR. **will reach out outside of GitHub**

### Reviewer checklist:


floor-licker:
Hey @gjermundgaraba, just wondering if you might be able to take a look at this

floor-licker:
Migrated to https://github.com/cardano-foundation/hermes-relayer to be maintained by Cardano Foundation for the time being.

